PR #71, opened 2020-10-24
[RFC] Query builder

piodul:
This is an implementation of query builders.
- There is a main `Builder` object which can be configured with default parameters:
  ```rust
  let mut builder = Builder::new();
  builder.set_no_paging_by_default();
  builder.set_default_consistency(Consistency::All);
  ```
- You can use it to create a new statement and adjust query parameters:
  ```rust
  builder
      .build()
      .with_page_size(1024)
      .query("SELECT (a, b, c) FROM ks.t");
  ```
- Prepared statements are supported as well:
  ```rust
  builder
      .build()
      .with_consistency(Consistency::One)
      .prepare("INSERT INTO ks.t (a, b, c) VALUES (?, ?, ?)");
  ```

I'm not sure this is the right direction, so I'd really like to hear your opinion on how to improve this API.

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use scylla::statement::builder::Builder;
 use scylla::transport::session::Session;
 use std::env;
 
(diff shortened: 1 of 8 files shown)


piodul:
Superseded by https://github.com/scylladb/scylla-rust-driver/pull/156

PR #102, opened 2020-11-09
Remove cargo clippy warnings

cvybhu:
Currently running `cargo clippy` produces 12 warnings which makes it hard to notice tips for newly written code.
This PR applies clippy's suggestions to remove the warnings.

Clippy is buggy and might show no output when ran on already build project
Running `touch $(find ./scylla) && cargo clippy` allows to refresh the build and clippy works ok

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,6 +18,8 @@ jobs:
     - uses: actions/checkout@v2
     - name: Format check
       run: cargo fmt --verbose --all -- --check
+    - name: Clippy check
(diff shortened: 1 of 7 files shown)


piodul:
Can we make a CI check based on `cargo clippy`? We already have one for `cargo fmt` ([here](https://github.com/scylladb/scylla-rust-driver/blob/main/.github/workflows/rust.yml#L20)).

cvybhu:
Sounds good, added command for clippy in `rust.yml`

piodul (approved):

PR #131, opened 2020-12-10
Removed duplicate code in proc macro

Jasperav:
This PR removes duplicate code from the proc macro crate


scylla-macros/Cargo.toml
```diff
@@ -13,3 +13,4 @@ proc-macro = true
 [dependencies]
 syn = "1.0"
 quote = "1.0" 
+proc-macro2 = "1.0"
```
Jasperav, line 16: This is needed to return the Ident type in parser.rs

scylla-macros/src/parser.rs
```diff
@@ -0,0 +1,21 @@
+use proc_macro::TokenStream;
+use syn::{DeriveInput, parse2, Data, Fields, FieldsNamed};
+use proc_macro2::Ident;
+
+/// Parsers the tokens_input to a DeriveInput and returns the struct name from which it derives and
```
piodul, line 5: Nit: s/Parsers/Parses

scylla-macros/src/parser.rs
```diff
@@ -0,0 +1,21 @@
...
+/// the named fields
+pub(crate) fn parse(tokens_input: TokenStream, current_derive: &str) -> (Ident, FieldsNamed) {
```
piodul, line 7: This function would benefit from a more descriptive name. How about `parse_struct_with_named_fields`?

scylla-macros/Cargo.toml
piodul: Nit: add a newline at the end of file

scylla-macros/src/parser.rs
```diff
@@ -1,21 +1,22 @@
...
+            ),
         },
-        _ => panic!("derive({}) works only on structs!", current_derive)
+        _ => panic!("derive({}) works only on structs!", current_derive),
     };
 
     (struct_name, struct_fields)
-}
\ No newline at end of file
+}
```
piodul, line 22: I think we can squash both commits together. The second only fixes formatting in the first one, I think it's better to have correct formatting from the beginning.

piodul (commented):
Sorry for the long delay. Left some minor comments.

piodul:
@cvybhu would you like to take a look? This PR changes the code you wrote.

scylla-macros/src/parser.rs
```diff
@@ -0,0 +1,22 @@
+use proc_macro::TokenStream;
+use proc_macro2::Ident;
+use syn::{parse2, Data, DeriveInput, Fields, FieldsNamed};
+
+/// Parsers the tokens_input to a DeriveInput and returns the struct name from which it derives and
+/// the named fields
+pub(crate) fn parse(tokens_input: TokenStream, current_derive: &str) -> (Ident, FieldsNamed) {
+    let input = parse2::<DeriveInput>(tokens_input.into()).expect("No DeriveInput");
```
cvybhu, line 8: Is there a reason why `parse_macro_input!(tokens_input as DeriveInput);` was changed to `parse2::...`?
It looks like `parse_macro_input!` provides some error information on failure which I'm not sure if expect does:
https://github.com/dtolnay/syn/blob/fc37ee397164e2c0cc884b1f4949c7ebe5355246/src/parse_macro_input.rs#L109

Additionally `parse2` documentation states that it should be used with `TokenStream` from `proc_macro2` and when  using `TokenStream` from `proc_macro` `parse` should be used instead:
https://docs.rs/syn/1.0.57/syn/fn.parse2.html

cvybhu (changes requested):
Found two things I think we should change

scylla-macros/src/parser.rs
Jasperav: You mean a new module called Parsers?

scylla-macros/src/parser.rs
```diff
@@ -0,0 +1,22 @@
+use proc_macro::TokenStream;
+use proc_macro2::Ident;
```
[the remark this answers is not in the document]
Jasperav: Ah, CLion didn't suggest syn::Ident at all, that's why I though I needed to use proc_macro2

scylla-macros/src/parser.rs
Jasperav: See: https://docs.rs/syn/1.0.57/syn/macro.parse_macro_input.html:

'This macro must be called from a function that returns proc_macro::TokenStream'

It triggers a compile error if that condition isn't met

scylla-macros/src/parser.rs
Jasperav: I tried adding all my commits into one, but it went wrong and now the commits are duplicated :(

scylla-macros/src/parser.rs
cvybhu: Oh okay that makes sense :+1:

We could also probably remove `.into()` since we are no longer converting from `proc_macro2::TokenStream` to `proc_macro::TokenStream`

Jasperav: I added this change in a new PR: https://github.com/scylladb/scylla-rust-driver/pull/139. This branch had a lot of commits that I didn't want to have and I was unable to remove them... I'm not a git expert

scylla-macros/src/parser.rs
Jasperav: I created a new PR: https://github.com/scylladb/scylla-rust-driver/pull/139

scylla-macros/src/parser.rs
Jasperav: @piodul  This is the only comment I haven't processed in the new PR: https://github.com/scylladb/scylla-rust-driver/pull/139. What do you exactly mean?

piodul: I just meant to say that there is a typo, and `Parsers` should be changed to `Parses` (without the "r").

`s/Parsers/Parses` is the `sed` syntax for changing the first occurrence of `Parsers` into `Parses`. Sorry, I shouldn't have assumed that everybody knows this tool.

PR #137, opened 2021-01-03
Add SCYLLA_URI parameter to session tests

cvybhu:
Currently all tests in `session_test.rs` connect to `127.0.0.1:9042`

This PR adds an option to specify a different uri using an environment variable, like in the examples.
Tests can then be run like this:
```bash
SCYLLA_URI=172.17.0.3:9042 cargo test -- --ignored
```

### DIFF
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -6,7 +6,8 @@ use crate::transport::session::Session;
 #[tokio::test]
 #[ignore]
 async fn test_unprepared_statement() {
-    let session = Session::connect("127.0.0.1:9042", None).await.unwrap();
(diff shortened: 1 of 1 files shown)


piodul (approved):

piodul:
@cvybhu please rebase onto the newest master. I re-ran the CI and I'm not entirely sure why it failed.

Logs from the CI shows errors which were caused by switching to a newer version of clippy. Those errors are already fixed on `main`, but here they happen again - perhaps GitHub checks against a result of a merge of this branch with an older `main` branch. Rebasing may help.

cvybhu:
@piodul - done, I rebased and the CI passes now

piodul:
@cvybhu thanks

PR #138, opened 2021-01-03
Add more checks to CI

cvybhu:
* Add `--examples` and `--tests` flags to `cargo clippy` to make it check examples and tests
* Add `--examples` to `cargo build` to also build examples

Before these changes it was possible to submit an example that doesn't compile and CI would pass

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,8 +19,8 @@ jobs:
     - name: Format check
       run: cargo fmt --verbose --all -- --check
     - name: Clippy check
-      run: cargo clippy --verbose -- -D warnings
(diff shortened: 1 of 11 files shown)


Jasperav:
Maybe add -- --test-threads=1 to cargo test, since database tests can conflict when running parallel

cvybhu:
I didn't notice that the tests could conflict, good catch
`test_unprepared_statement` and `test_batch`  used the same table `ks.t`

I fixed them so that now all tests are  independent of each other so we can probably leave multiple threads until there will be some tests that require a single thread

cvybhu:
Clippy got an update with new warnings, I guess I can just fix this in this PR since it updates CI anyway

piodul (approved):

PR #140, opened 2021-01-14
Reorganize connection management

### LINKED ISSUE #107: Add retrieving replication strategy information from the cluster

### LINKED ISSUE #118: Establish connections to all available nodes on startup

cvybhu:
## Description
This PR reorganizes the way connections and topology are managed in `Session`.  

Now connections to a node are created as soon as we get the information that this node exists.
Each connection is now managed by a `ConnectionKeeper` which spawns a worker in the background that opens the connection and handles reopening on failure (TODO - it doesn't reopen yet but it's not a priority for now)

The way connections are kept has also changed - `struct Node` now contains all information about this node and all node's connections. In the background each node spawns a worker which detects if the Node has resharded and updates the connections to match new `ShardInfo`

```rust
/// Node represents a cluster node along with it's data and connections
pub struct Node {
    pub address: SocketAddr,
    pub datacenter: Option<String>,
    pub rack: Option<String>,

    pub connections: Arc<AsyncRwLock<NodeConnections>>,

    _worker_handle: RemoteHandle<()>,
}

pub enum NodeConnections {
    /// Non shard-aware ex. a Cassandra node connection
    Single(ConnectionKeeper),
    /// Shard aware Scylla node connections
    Sharded {
        shard_info: ShardInfo,
        shard_conns: HashMap<Shard, ConnectionKeeper>,
    },
}
```
Updating topology information and creating new `Nodes` was separated to a new struct - `Cluster`:
```rust
pub struct Cluster {
    data: Arc<RwLock<Arc<ClusterData>>>,

    refresh_channel: tokio::sync::mpsc::Sender<RefreshRequest>,
    _worker_handle: RemoteHandle<()>,
}

pub struct ClusterData {
    pub known_peers: HashMap<SocketAddr, Arc<Node>>, // Invariant: nonempty after Cluster::new()
    pub ring: BTreeMap<Token, Arc<Node>>,            // Invariant: nonempty after Cluster::new()
    pub keyspaces: HashMap<String, Keyspace>,
}
```

The connections were moved inside the `Node` to make load balancing implementation easier - currently there are two maps behind `RwLock` for ring structure and node connections. This new representation forces an invariant that for each `Node` that we find in the ring there are active `Connections` that we can use.

Additionaly `ClusterData` is kept in an `Arc` which plays well with load balancing where we need to keep a reference to the ring for a long time, which was hard to do correctly when this reference was behind `RwLock`. Now we can just quickly clone the `Arc` containing up to date `ClusterData` and not worry about keeping any locks locked for a long time.

## Architecture
Here is a short description about each component and what it does:

### `ConnectionKeeper`
Manages a `Connection` - tries to keep it open. Spawns in the background `ConnectionKeeperWorker::work` which handles all this work. If given `ShardInfo` in the constructor will connect to `shard_info.shard`. Has the option to use a channel where after opening each new `Connection` this connection's `ShardInfo` will be sent.
```rust
/// ConnectionKeeper keeps a Connection to some address and works to keep it open
pub struct ConnectionKeeper {
    conn_state_receiver: tokio::sync::watch::Receiver<ConnectionState>,
    _worker_handle: RemoteHandle<()>,
}

#[derive(Clone)]
pub enum ConnectionState {
    Initializing, // First connect attempt ongoing
    Connected(Arc<Connection>),
    Broken(QueryError),
}

pub type ShardInfoSender = Arc<std::sync::Mutex<tokio::sync::watch::Sender<Option<ShardInfo>>>>;

impl ConnectionKeeper {
    /// Creates new ConnectionKeeper that starts a connection in the background
    /// # Arguments
    ///
    /// * `address` - IP address to connect to
    /// * `compression` - preferred compression method to use
    /// * `shard_info` - ShardInfo to use, will connect to shard number `shard_info.shard`
    /// * `shard_info_sender` - channel to send new ShardInfo after each connection creation
    pub fn new(
        address: SocketAddr,
        compression: Option<Compression>,
        shard_info: Option<ShardInfo>,
        shard_info_sender: Option<ShardInfoSender>,
    ) -> Self
```
### `Node`
Contains all information and connections to a `Node`. On creation creates a single `ConnectionKeeper` to connect to the node in the background. Then spawns `NodeWorker::work` in the background which receives `Connection`'s `ShardInfo` using a channel. If `NodeWorker` detects that the node has resharded it will create new `NodeConnections` to match new sharding.
```rust
/// Node represents a cluster node along with it's data and connections
pub struct Node {
    pub address: SocketAddr,
    pub datacenter: Option<String>,
    pub rack: Option<String>,

    pub connections: Arc<AsyncRwLock<NodeConnections>>,

    _worker_handle: RemoteHandle<()>,
}

pub enum NodeConnections {
    /// Non shard-aware ex. a Cassandra node connection
    Single(ConnectionKeeper),
    /// Shard aware Scylla node connections
    Sharded {
        shard_info: ShardInfo,
        shard_conns: HashMap<Shard, ConnectionKeeper>,
    },
}
```
### `Cluster`
Manages up to date information about cluster topology and creates new `Nodes` which immediately open connections. Spawns a `ClusterWorker` in the background which periodically fetches new topology information from system tables and rebuilds `ClusterInfo`. Topology is read using `TopologyReader` which manages its own set of control connections to not break metrics.
```rust
pub struct Cluster {
    data: Arc<RwLock<Arc<ClusterData>>>,

    refresh_channel: tokio::sync::mpsc::Sender<RefreshRequest>,
    _worker_handle: RemoteHandle<()>,
}

pub struct ClusterData {
    pub known_peers: HashMap<SocketAddr, Arc<Node>>, // Invariant: nonempty after Cluster::new()
    pub ring: BTreeMap<Token, Arc<Node>>,            // Invariant: nonempty after Cluster::new()
    pub keyspaces: HashMap<String, Keyspace>,
}
```
## Errors
Doing changes this big also caused errors in `transport_errors.rs` to become obsolete, I had to create new ones.
I chose to separate 4 diferent kinds of errors that can happen during a Query:
```rust
/// Error that occured during query execution
#[derive(Error, Debug, Clone)]
pub enum QueryError {
    /// Database sent a response containing some error
    #[error(transparent)]
    DBError(#[from] DBError),

    /// Caller passed an invalid query
    #[error(transparent)]
    BadQuery(#[from] BadQuery),

    /// Input/Output error has occured, connection broken etc.
    #[error("IO Error: {0}")]
    IOError(Arc<std::io::Error>),

    /// Unexpected or invalid message received
    #[error("Protocol Error: {0}")]
    ProtocolError(&'static str),
}
```

## Fixed issues
Fixes #118 
Fixes #107 (Makes #135 obsolete)


cvybhu:
@piodul @psarna could you review

scylla/src/transport/cluster.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        Ok(result)
+    }
+}
+
+impl ClusterWorker {
+    pub async fn work(mut self) {
+        use tokio::time::{Duration, Instant};
+
+        let refresh_duration = Duration::from_secs(60); // Refresh topology every 60 seconds
```
psarna, line 141: I suppose it would be nice to make this parameter configurable eventually. And to allow triggering the topology sync on demand when users want it, e.g. because they are aware that topology just changed and they don't want to wait 60 seconds for the changes to be applied. But it's good as is for now.

piodul:
I appreciate the detailed description of the PR.

However, please try splitting this change into multiple, logically separated commits. It looks like there are multiple things being changed at once, but everything is put into a single commit (followed by a few commits with fixes). Smaller commits makes reviewing and maintaining easier.

For example, you could have separate commits for `ConnectionKeeper`, `Node`, `Cluster`, and errors rework. Each commit should compile and pass tests.

scylla/src/transport/node.rs
```diff
@@ -0,0 +1,173 @@
...
+
+                    NodeConnections::Sharded {
+                        shard_info: shard_info.clone(),
+                        shard_conns: connections,
+                    }
+                }
+            };
+
+            // Update node.connections
+            // Use std::mem:;swap to minimalize time spent holding write lock
```
psarna, line 166: typo: :; -> ::

scylla/src/transport/node.rs
```diff
@@ -0,0 +1,173 @@
...
+                }
+                (None, None) => continue, // Nothing chaged, go back to waiting for a change
+                _ => {}
+            }
+
+            cur_shard_info = new_shard_info;
+
+            // We received updated node ShardInfo
+            // Create new node connections. It will happen rarely so we can probably afford it
+            // TODO: Maybe save some connections instead of recreating?
```
psarna, line 134: I don't know if I follow this comment correctly, but if we already received ShardInfo from a node, we would never receive another one unless the node was restarted first, which also implies that all of its connections are broken. Changing the number of shards in Scylla is only available upon restart.

psarna (commented):
Looks good.

scylla/src/transport/cluster.rs
cvybhu: There is already an option to refresh on demand - there is `Session::refresh_topology` which calls `Cluster::refresh_topology`.

About configuration I think it can be easily added later after we add multiple thinks to configure (TLS, password, load balancing policy etc..) which would require some `SessionBuilder` or something like that.

psarna: Oh, right, it used to be manual only at first.

scylla/src/transport/node.rs
cvybhu: The idea was that after Node has resharded its connection will die and it's `ConnectionKeeper` will create a new `Connection` which will send new `ShardInfo` to `NodeWorker`. Then `NodeWorker` detects that there was a change and creates new connections to correct shards.

psarna: Makes sense. This TODO could be removed though - since all connections are reset, you won't be able to save any, so it only makes sense to create all of them from scratch.

cvybhu:
> I appreciate the detailed description of the PR.
> 
> However, please try splitting this change into multiple, logically separated commits. It looks like there are multiple things being changed at once, but everything is put into a single commit (followed by a few commits with fixes). Smaller commits makes reviewing and maintaining easier.
> 
> For example, you could have separate commits for `ConnectionKeeper`, `Node`, `Cluster`, and errors rework. Each commit should compile and pass tests.

This PR was developed all at once - there was some empty skeleton with `Cluster`, `Node` and `ConnectionKeeper` with no functionality and gradually new functionally was added and things refactored. It is all interconnected with each other. I will try splitting it up, but it feels a bit artificial to do and will take time which could be spent eg. implementing the remaining CQL types.

scylla/src/transport/node.rs
cvybhu: They are reset but their `ConnectionKeeper` will immediately reconnect so it's a bit wasteful to discard it's work altogether.

piodul (commented):
LGTM, left some style comments.

scylla/src/transport/cluster.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    pub async fn refresh_topology(&self) -> Result<(), QueryError> {
+        let (response_sender, response_receiver) = tokio::sync::oneshot::channel();
+
+        self.refresh_channel
+            .send(RefreshRequest {
+                response_chan: response_sender,
+            })
+            .await
+            .unwrap_or_else(|_| panic!("Bug in Cluster::refresh_topology sending"));
```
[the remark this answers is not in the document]
cvybhu: :+1:, It was a panic because because `SendError` didn't implement `Debug` so `expect` didn't compile, but it was because `RefreshRequest` didn't have `Debug` so I fixed that and now `expect` works.

scylla/src/routing.rs
```diff
@@ -81,12 +64,21 @@ impl ShardInfo {
 
     /// Randomly choose a source port `p` such that `shard_of(t) == shard_of_source_port(p)`.
     pub fn draw_source_port_for_token(&self, t: Token) -> u16 {
```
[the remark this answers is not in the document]
cvybhu: :+1:, removed

cvybhu:
@piodul I applied your suggestions and squashed everything into two clean commits.

I tried to split `Reorganize connection management` but it is not that easy - If we want to fake small changes that pass CI eg. first only add `ConnectionKeeper` then I need to write a lot of glue code which never existed before to fit `ConnectionKeeper` into existing `Session`. 

If we really want to do this I can finish it before Tuesday, but maybe we could just merge it as-is?

scylla/src/transport/topology.rs
```diff
@@ -1,364 +1,337 @@
...
+    },
+    NetworkTopologyStrategy {
+        // Replication factors of datacenters with given names
+        datacenter_repfactors: HashMap<String, usize>,
+    },
+    LocalStrategy, // replication_factor == 1
+    Other {
+        name: String,
+        data: HashMap<String, String>,
+    },
```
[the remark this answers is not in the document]
cvybhu: Supporting only these strategies was previously approved by psarna in #135, so I'm gonna wait for his opinion before adding `EverywhereStrategy`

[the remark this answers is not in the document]
piodul: I guess that it's perfectly fine to do it in a follow-up, if we decide to do it at all.

piodul:
> If we really want to do this I can finish it before Tuesday, but maybe we could just merge it as-is?

@cvybhu No need to reorganize it further, it's OK as it is now.

piodul (approved):

PR #154, opened 2021-02-06
Local control connection should be using rpc_address instead of listen address

dtzxporter:
DataStax (nodejs driver) claims that you shouldn't use system.local's rpc_address unless the client specifies SNI in the options. For now I have omitted that and will use rpc_address which is meant for clients to receive. DataStax defaults to using the initial connections address for the control connection (Which we could expose from conn) but I've left a comment explaining for now.

Also cleaned up parsing of ipv6 in the connection with port function.

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -7,7 +7,7 @@ use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::io::ErrorKind;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::net::{SocketAddr};
 use std::sync::Arc;
 use std::sync::Mutex as StdMutex;
 
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -490,11 +490,11 @@ async fn connect_with_source_port(
 ) -> Result<TcpStream, std::io::Error> {
     if addr.is_ipv4() {
         let socket = TcpSocket::new_v4().unwrap();
-        socket.bind(format!("{}:{}", addr, source_port).parse().unwrap())?;
+        socket.bind(format!("{}:{}", addr.ip(), source_port).parse().unwrap())?;
         Ok(socket.connect(addr).await?)
     } else if addr.is_ipv6() {
         let socket = TcpSocket::new_v6().unwrap();
-        socket.bind(format!("[{}]:{}", addr, source_port).parse().unwrap())?;
+        socket.bind(format!("[{}]:{}", addr.ip(), source_port).parse().unwrap())?;
         Ok(socket.connect(addr).await?)
     } else {
         Err(std::io::Error::new(ErrorKind::Other, "invalid ip address").into())
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -488,16 +488,17 @@ async fn connect_with_source_port(
     addr: SocketAddr,
     source_port: u16,
 ) -> Result<TcpStream, std::io::Error> {
-    if addr.is_ipv4() {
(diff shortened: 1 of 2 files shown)


scylla/src/transport/topology.rs
```diff
@@ -177,6 +177,10 @@ async fn query_peers(conn: &Connection, connect_port: u16) -> Result<Vec<Peer>,
     let typed_local_rows =
         local_rows.into_typed::<(IpAddr, Option<String>, Option<String>, Option<Vec<String>>)>();
 
+    // Here we are considering the local node to be just a regular peer, but datastax claims that unless
+    // you have SNI enabled, you shouldn't trust rpc_address (So they just use the initial connection's address)
+    // we could expose this from conn & use it directly instead of system.local.rpc_address
+
```
piodul, line 183: While usage of `listen_address` was clearly wrong and `rpc_address` is better, I don't understand the issue related to disabled SNI. Could you explain it, or at least point to some resources with an explanation?

scylla/src/transport/connection.rs
```diff
@@ -488,11 +488,17 @@ async fn connect_with_source_port(
...
+    if addr.is_ipv4() {
+        let socket = TcpSocket::new_v4().unwrap();
+        socket.bind(format!("{}:{}", addr.ip(), source_port).parse().unwrap())?;
+        Ok(socket.connect(addr).await?)
+    } else if addr.is_ipv6() {
+        let socket = TcpSocket::new_v6().unwrap();
+        socket.bind(format!("[{}]:{}", addr.ip(), source_port).parse().unwrap())?;
+        Ok(socket.connect(addr).await?)
+    } else {
+        Err(std::io::Error::new(ErrorKind::Other, "invalid ip address"))
```
piodul, line 500: This is dead code. A `SocketAddr` can be either IPv4 or IPv6, nothing else.

You can avoid the `else` clause by replacing the `if`/`else if`/`else` chain with pattern matching:

```rust
match addr {
    SocketAddr::V4(_) => {
        // ...
    }
    SocketAddr::V6(_) => {
        // ...
    }
}
```

piodul (changes requested):

scylla/src/transport/topology.rs
dtzxporter: https://github.com/datastax/nodejs-driver/blob/master/lib/control-connection.js#L716

I'm just referencing what is done in data stack's driver. They are ignoring the result entirely unless the client specifies sni here. (And using the initial connection's address)

scylla/src/transport/connection.rs
```diff
@@ -488,11 +488,18 @@ async fn connect_with_source_port(
...
 ) -> Result<TcpStream, std::io::Error> {
-    // TODO: handle ipv6?
-    let socket = TcpSocket::new_v4()?;
-    let source_addr = IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0));
-    socket.bind(SocketAddr::new(source_addr, source_port))?;
-    Ok(socket.connect(addr).await?)
+    match addr {
+        SocketAddr::V4(_) => {
+            let socket = TcpSocket::new_v4()?;
+            socket.bind(format!("0.0.0.0:{}", source_port).parse().unwrap())?;
```
piodul, line 494: This line:

- Allocates a new string on the heap,
- Writes `0.0.0.0:` to it and appends to it the `source_port` converted to decimal,
- Parses the complete address.

While I don't think this function is on a critical path, there is no need to do all of that in order to initialize a `SocketAddr` structure. Using `SocketAddr::new` is much cheaper:

```rust
let source_addr = IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0));
socket.bind(SocketAddr::new(source_addr, source_port))?;
```

Similar change should be done to the IPv6 case, too.

piodul (approved):

PR #178, opened 2021-02-26
Checking if query starts with 'use ' is now more efficient

Jasperav:
## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -280,18 +280,17 @@ impl Session {
         query: impl Into<Query>,
         values: impl ValueList,
     ) -> Result<Option<Vec<result::Row>>, QueryError> {
-        let query: Query = query.into();
-        let query_text: &str = query.get_contents();
+        let query = query.into();
+        let query_text = query.get_contents();
 
         // In case the user tried doing session.query("use keyspace ks") run session::use_keyspace
-        if query_text.to_lowercase().starts_with("use ") {
+        if query_is_setting_keyspace(query_text) {
+            // TODO: replace with log library in https://github.com/scylladb/scylla-rust-driver/issues/158
             eprintln!("Warning: Raw USE KEYSPACE queries are experimental, please use session::use_keyspace instead");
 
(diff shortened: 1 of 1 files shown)


scylla/src/transport/session.rs
```diff
@@ -280,18 +280,17 @@ impl Session {
         query: impl Into<Query>,
         values: impl ValueList,
     ) -> Result<Option<Vec<result::Row>>, QueryError> {
-        let query: Query = query.into();
```
Jasperav, line 283: The explicit types here are not needed

scylla/src/transport/session.rs
```diff
@@ -280,18 +280,17 @@ impl Session {
...
-        let query_text: &str = query.get_contents();
+        let query = query.into();
+        let query_text = query.get_contents();
 
         // In case the user tried doing session.query("use keyspace ks") run session::use_keyspace
-        if query_text.to_lowercase().starts_with("use ") {
+        if query_is_setting_keyspace(query_text) {
```
Jasperav, line 287: I think it's cleaner to do this in a separate method (testing is easier that way)

scylla/src/transport/session.rs
```diff
@@ -280,18 +280,17 @@ impl Session {
...
             eprintln!("Warning: Raw USE KEYSPACE queries are experimental, please use session::use_keyspace instead");
 
-            let mut keyspace_name: &str = &query_text["use ".len()..].trim_end_matches(';').trim();
-
-            let case_sensitive: bool = keyspace_name.starts_with('"');
-
-            keyspace_name = keyspace_name.trim_matches('"');
+            let keyspace_name = &query_text["use ".len()..].trim_end_matches(';').trim();
+            let case_sensitive = keyspace_name.starts_with('"');
+            let keyspace_name = keyspace_name.trim_matches('"');
```
Jasperav, line 293: Introduced a new variable, else I get a compile error when I remove the explicit &str type

Jasperav:
Fixed: https://github.com/scylladb/scylla-rust-driver/issues/176

PR #192, opened 2021-03-08
Run CI database tests in parallel

cvybhu:
@Jasperav added a flag to run tests sequentially one after another in CI in #170, but I think it's not really needed.
All tests are written in a way that allows them to be run in parallel without any conflicts, I don't see any reason why something would break.

Running them in prallel could improve CI times - currently a single CI run takes 5 minutes out of which 3 minutes is spent running tests.

## Pre-review checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,8 +23,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Run tests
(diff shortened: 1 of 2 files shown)


cvybhu:
Okay nevermind, during `Run tests` most of the time is taken by compiling, tests take an insignificant amount of time compared to this.

PR #220, opened 2021-04-14
Preallocate space for SerializedValues

Jasperav:

### DIFF
--- a/scylla/src/frame/value.rs
+++ b/scylla/src/frame/value.rs
@@ -579,28 +579,28 @@ macro_rules! impl_value_list_for_tuple {
     }
 }
 
-impl_value_list_for_tuple!(T0, T1; 0, 1; 1);
-impl_value_list_for_tuple!(T0, T1, T2; 0, 1, 2; 2);
-impl_value_list_for_tuple!(T0, T1, T2, T3; 0, 1, 2, 3; 3);
-impl_value_list_for_tuple!(T0, T1, T2, T3, T4; 0, 1, 2, 3, 4; 4);
-impl_value_list_for_tuple!(T0, T1, T2, T3, T4, T5; 0, 1, 2, 3, 4, 5; 5);
-impl_value_list_for_tuple!(T0, T1, T2, T3, T4, T5, T6; 0, 1, 2, 3, 4, 5, 6; 6);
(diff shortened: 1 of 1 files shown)


piodul (commented):
Thanks for the patch! I have only two comments:

- Could you make the commit title/description more descriptive? E.g. "Preallocate space for SerializedValues`?
- Could you extend this optimization to include tuples of other sizes than 0 and 1, too? You could e.g. pass the length of the tuple as an additional argument of the `impl_value_list_for_tuple` macro and use it as an argument for `SerializedValues::with_capacity` inside the macro.

scylla/src/frame/value.rs
```diff
@@ -579,28 +579,28 @@ macro_rules! impl_value_list_for_tuple {
...
+                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12; 12);
 impl_value_list_for_tuple!(T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13;
-                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13);
+                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13; 13);
 impl_value_list_for_tuple!(T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14;
-                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14);
+                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14; 14);
 impl_value_list_for_tuple!(T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15;
-                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15; 15);
```
piodul, line 603: The tuple sizes are off by one (they should be larger by one). For example, the first `impl_value_list_for_tuple` corresponds to tuple of size 2.

piodul (approved):

PR #224, opened 2021-04-22
Add schema_change event, schema_change response

asledz:
# Description 

Add deserialization for schema change server event and for schema change response for schema altering query.
Important note: the way to make sure, if the schema is same for each node will be schema agreement query.

## Pre-review checklist

### DIFF
--- a/scylla/src/frame/response/event.rs
+++ b/scylla/src/frame/response/event.rs
@@ -7,7 +7,7 @@ use std::net::SocketAddr;
 pub enum Event {
     TopologyChange(TopologyChangeEvent),
     StatusChange(StatusChangeEvent),
-    SchemaChange, // TODO specify
+    SchemaChange(SchemaChangeEvent),
 }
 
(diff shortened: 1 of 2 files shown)


asledz:
I feel like I couldn't find any way to make it useful. I thought about adding schema information to nodes, but apart from informative value for user I can’t see what else to do with the information about changing the schema (maybe automatic schema agreement if any of the nodes differ option?)

I implemented adding the info about schema, but I am not sure about adding this. Would appreciate some feedback 0:)

psarna:
Schema change events can be useful once we start tracking the current schemas on the driver side (which we don't now, but e.g. cqlsh does for autocompletion).  I think that the code is good as is for now, but please test it manually and post some logs indicating that the events were successfully received.

asledz:
Every even has been tested, also: on every example and test we can see the correct behavior!

Zrzut ekranu 2021-05-14 o 11 44 03

PR #284, opened 2021-08-21
Stop parsing UDT fields when buffer is empty.

merlleu:
When you create an User Defined Type, then insert a frozen udt into a table, then add a field to this UDT (on both Scylla schema and Rust code) the driver will fail to parse the UDT with the error: "Invalid message: Error parsing message: failed to fill whole buffer".
This is because the driver tries to read the new field from the buffer but it reached the end of it.
This fix simply stop parsing UDT fields when buffer is empty.

Fixes: Stop parsing UDT fields when buffer is empty.

## Pre-review checklist

### DIFF
--- a/scylla/src/frame/response/result.rs
+++ b/scylla/src/frame/response/result.rs
@@ -706,6 +706,13 @@ fn deser_cql_value(typ: &ColumnType, buf: &mut &[u8]) -> StdResult<CqlValue, Par
             let mut fields: BTreeMap<String, Option<CqlValue>> = BTreeMap::new();
 
             for (field_name, field_type) in field_types {
+                // If a field is added to a UDT and we read an old (frozen ?) version of it, 
+                // the driver will fail to parse the whole UDT.
+                // This is why we break the parsing after we reach the end of the serialized UDT.
+                if buf.is_empty() {
+                    break
+                }
+
                 let mut field_value: Option<CqlValue> = None;
                 if let Some(mut field_val_bytes) = types::read_bytes_opt(buf)? {
                     field_value = Some(deser_cql_value(field_type, &mut field_val_bytes)?);
--- a/scylla/src/frame/response/result.rs
+++ b/scylla/src/frame/response/result.rs
@@ -706,11 +706,11 @@ fn deser_cql_value(typ: &ColumnType, buf: &mut &[u8]) -> StdResult<CqlValue, Par
             let mut fields: BTreeMap<String, Option<CqlValue>> = BTreeMap::new();
 
             for (field_name, field_type) in field_types {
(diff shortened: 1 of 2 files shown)


scylla/src/frame/response/result.rs
```diff
@@ -706,6 +706,13 @@ fn deser_cql_value(typ: &ColumnType, buf: &mut &[u8]) -> StdResult<CqlValue, Par
             let mut fields: BTreeMap<String, Option<CqlValue>> = BTreeMap::new();
 
             for (field_name, field_type) in field_types {
+                // If a field is added to a UDT and we read an old (frozen ?) version of it,
```
piodul, line 709: There is a trailing space at the end of the line, please remove.

scylla/src/frame/response/result.rs
```diff
@@ -706,6 +706,13 @@ fn deser_cql_value(typ: &ColumnType, buf: &mut &[u8]) -> StdResult<CqlValue, Par
             let mut fields: BTreeMap<String, Option<CqlValue>> = BTreeMap::new();
 
             for (field_name, field_type) in field_types {
+                // If a field is added to a UDT and we read an old (frozen ?) version of it, 
+                // the driver will fail to parse the whole UDT.
+                // This is why we break the parsing after we reach the end of the serialized UDT.
+                if buf.is_empty() {
+                    break
```
piodul, line 713: Missing semicolon at the end of the line.

piodul:
Thanks for the contribution! The fix seems reasonable.

Could you add a unit test?

Additionally, please improve the commit message. If you copy what you wrote in the PR's description to the commit description, it will be fine.

merlleu:
> Could you add a unit test?

Done.

> Additionally, please improve the commit message. If you copy what you wrote in the PR's description to the commit description, it will be fine.
I've renamed the pull request but I'm not sure that's what you meant

piodul:
> > Additionally, please improve the commit message. If you copy what you wrote in the PR's description to the commit description, it will be fine.
> 
> I've renamed the pull request but I'm not sure that's what you meant

The name of the pull request is fine. There used to be a single commit on the branch, so I meant changing its description (`"added backtrace for debug\n\ndebug UDT parsing..."`) into something more descriptive. However, I think it's fine - I can probably do it myself just before merging.

merlleu:
> The name of the pull request is fine. There used to be a single commit on the branch, so I meant changing its description ("added backtrace for debug\n\ndebug UDT parsing...") into something more descriptive. However, I think it's fine - I can probably do it myself just before merging.

Oh ok, it's because I squashed all my test commits into this one (each line is a commit name) and forgot to change the description.

piodul (approved):

PR #302, opened 2021-09-28
Client-side timestamps

### LINKED ISSUE #262: Client-side timestamps
The driver lacks the ability to use client-side timestamps.

While, of course, it is possible to set a custom timestamp for a statement via the CQL `USING TIMESTAMP` clause, the Datastax Java driver also allows to:

- Programmatically set the timestamp for the query - [`Statement.setQueryTimestamp()`](https://docs.datastax.com/en/drivers/java/4.10/com/datastax/oss/driver/api/core/cql/Statement.html#setQueryTimestamp-long-),
- Use a client-side timestamp generator which automatically generates a new timestamp just before sending the statement.

Source: https://docs.datastax.com/en/developer/java-driver/4.10/manual/core/query_timestamps/

havaker:

This pull request adds support for specifying client-side timestamps for a given query.
```rust
let mut regular_query = Query::new(query_str.to_string());
regular_query.set_timestamp(Some(42));
session.query(regular_query, (,)).await?;
```
Refs: #262 (not saying fixes, because issue is resolved partially - there are no generators implemented, only programmatically setting the timestamp for queries)

### DIFF
--- a/scylla/src/statement/query.rs
+++ b/scylla/src/statement/query.rs
@@ -108,6 +108,19 @@ impl Query {
     pub fn get_tracing(&self) -> bool {
         self.config.tracing
     }
+
+    /// Sets the default timestamp for this statement in microseconds.
+    /// If not None, it will replace the server side assigned timestamp as default timestamp
+    /// If a statement contains a `USING TIMESTAMP` clause, calling this method won't change
+    /// anything
+    pub fn set_timestamp(&mut self, timestamp: Option<i64>) {
+        self.config.timestamp = timestamp
+    }
+
+    /// Gets the default timestamp for this statement in microseconds.
+    pub fn get_timestamp(&self) -> Option<i64> {
+        self.config.timestamp
+    }
 }
 
 impl From<String> for Query {
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -913,3 +913,111 @@ async fn test_await_timed_schema_agreement() {
         .await
         .unwrap();
 }
+
+#[tokio::test]
+async fn test_timestamp() {
+    let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+    let session = SessionBuilder::new().known_node(uri).build().await.unwrap();
+
+    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1}", &[]).await.unwrap();
+    session
+        .query("DROP TABLE IF EXISTS ks.t_timestamp;", &[])
+        .await
+        .unwrap();
+    session
+        .query(
+            "CREATE TABLE IF NOT EXISTS ks.t_timestamp (a text, b text, primary key (a))",
+            &[],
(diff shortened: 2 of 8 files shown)


scylla/src/statement/query.rs
```diff
@@ -108,6 +108,14 @@ impl Query {
     pub fn get_tracing(&self) -> bool {
         self.config.tracing
     }
+
+    pub fn set_timestamp(&mut self, timestamp: Option<i64>) {
```
piodul, line 112: Please put a docstring similarly to what you did for `PreparedStatement::set_timestamp`.

scylla/src/statement/query.rs
```diff
@@ -108,6 +108,14 @@ impl Query {
...
+        self.config.timestamp = timestamp
+    }
+
+    pub fn get_timestamp(&self) -> Option<i64> {
```
piodul, line 116: Ditto

scylla/src/transport/session_test.rs
```diff
@@ -913,3 +913,83 @@ async fn test_await_timed_schema_agreement() {
...
+        })
+        .collect();
+    results.sort();
+    assert_eq!(
+        results,
+        vec![
+            (1, 2, &String::from("higher timestamp")),
+            (2, 2, &String::from("higher timestamp"))
+        ]
+    );
```
piodul, line 994: Could you test batches here, too?

scylla/src/transport/session_test.rs
```diff
@@ -913,3 +913,83 @@ async fn test_await_timed_schema_agreement() {
...
+        .unwrap();
+
+    let mut results: Vec<(i32, i32, &String)> = rs
+        .iter()
+        .map(|r| {
+            let a = r.columns[0].as_ref().unwrap().as_int().unwrap();
+            let b = r.columns[1].as_ref().unwrap().as_int().unwrap();
+            let c = r.columns[2].as_ref().unwrap().as_text().unwrap();
+            (a, b, c)
+        })
```
piodul, line 985: Couldn't `into_typed` be used here instead of casting the types manually?

scylla/src/transport/session_test.rs
piodul: Additionally, you could also check that the timestamp is equal to the (higher) default timestamp by using the `writetime` CQL function: https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useWritetime.html

piodul (changes requested):
Left some small comments, otherwise LGTM

scylla/src/statement/query.rs
havaker: Done

scylla/src/statement/query.rs
havaker: Done

scylla/src/transport/session_test.rs
havaker: Done

scylla/src/transport/session_test.rs
havaker: It could, fixed

havaker:
v2: applied suggestions from @piodul's review, refactored `set_timestamp` test to be more readable

PR #310, opened 2021-10-03
Remove batch fields and added new initializer

Jasperav:
## Pre-review checklist

- [ ] I added relevant tests for new features and bug fixes. -> Not needed

I have some sort of wrapper around batch in which I have the batch statement vec already there, it would be a waste to call `append_statement` in a loop for that case. 

With the new initalizer, I can add them directly to the batch. 

I removed the getter/setter methods, I think those methods are cluttering up the code and users can access the config fields themself fairly easy.


Jasperav:
Maybe the get/setters are useful to hide certain fields for batching

Issue #283, opened 2021-08-20
Implement Value for CqlValue


pkolaczk:
I want to be able to construct a list of parameters to a query dynamically - they are of different types and their number is known only at runtime.

`Vec` doesn't cut it, because it requires all values to be the same concrete type.
A tuple doesn't cut it either, because it requires the number and types of items to be known at compile time.

It would be nice if I could pass a `Vec<CqlType>` to a query - looks like the proper class is already there, but unfortunately it doesn't implement `Value`.

piodul:
As a workaround, for now you can use [`SerializedValues`](https://docs.rs/scylla/0.2.1/scylla/frame/value/struct.SerializedValues.html). An example:

```rust
use anyhow::Result;
use futures::StreamExt;
use scylla::frame::value::SerializedValues;
use scylla::transport::session::Session;
use scylla::SessionBuilder;
use std::env;

#[tokio::main]
async fn main() -> Result<()> {
    let uri = env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
    let session: Session = SessionBuilder::new().known_node(uri).build().await?;

    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1}", &[]).await?;
    session
        .query(
            "CREATE TABLE IF NOT EXISTS ks.t (a int, b int, c text, primary key (a, b))",
            &[],
        )
        .await?;

    let mut vals = SerializedValues::new();
    vals.add_value(&1i32)?;
    vals.add_value(&2i32)?;

    // &str doesn't work yet, unfortunately
    vals.add_value(&"The quick brown fox jumps over the lazy dog".to_string())?;

    session
        .query("INSERT INTO ks.t (a, b, c) VALUES (?, ?, ?)", vals)
        .await?;

    let mut iter = session
        .query_iter("SELECT a, b, c FROM ks.t", ())
        .await?
        .into_typed::<(i32, i32, String)>();

    while let Some(r) = iter.next().await {
        let (a, b, c) = r?;
        println!("{} {} {}", a, b, c);
    }

    Ok(())
}

```

pkolaczk:
Thanks! That's neat and even simpler than what I did by providing my own `CqlValue` wrapper that implements `Value`.

PR #314, opened 2021-10-10
Implement `Value` for `CqlValue`

havaker:

Fixes: #283


havaker:
Marking as a draft until more tests are added

scylla/src/frame/response/result.rs
```diff
@@ -96,7 +95,7 @@ pub enum CqlValue {
     UserDefinedType {
         keyspace: String,
         type_name: String,
-        fields: BTreeMap<String, Option<CqlValue>>,
+        fields: Vec<(String, Option<CqlValue>)>,
```
piodul, line 98: Here, please add a remark that the order of fields in the vector must match the order of fields as defined in the UDT and that the driver does not check this by itself, so incorrect data will be written if the order is wrong.

scylla/src/frame/cql_collections_test.rs
```diff
@@ -121,49 +119,40 @@ async fn test_cql_collections() {
...
-    }
+        .unwrap()
+        .into_typed::<(Vec<i32>,)>()
+        .map(Result::unwrap)
+        .map(|(v,)| v.into_iter())
+        .flatten()
+        .collect();
+    received_elements.sort_unstable();
+
+    assert_eq!(received_elements, vec![2, 3, 4, 4, 9, 17]);
```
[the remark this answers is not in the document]
havaker: Removed flattening, tests are verifying list's elements order now

scylla/src/frame/response/result.rs
havaker: Done

PR #327, opened 2021-11-03
scylla/transport/topology: accept shorter topology names

QuentinPerez:
this patch handles shorter topology names, otherwise the
strategy is considered as `Other` and the default behavior in this case with
a TokenAwarePolicy is to override RF to 1.

see https://github.com/scylladb/scylla-rust-driver/blob/main/scylla/src/transport/load_balancing/token_aware.rs#L128-L132 and https://docs.scylladb.com/getting-started/ddl/#create-keyspace-statement

## Pre-review checklist


psarna (approved):

QuentinPerez:
so fast 😅

PR #331, opened 2021-11-08
transport/session: do not serialize values twice

QuentinPerez:
👋  just a small PR.

- to not serialize the values twice, `values.serialized()` was called in `query_paged` then `query`
- to use try_join_all instead of `collect::<Vec<_>> + collect::<Result<Vec<_>>>`

**EDIT**: just saw that there was an impl of ValueList for Result<SerializedValues> means that the first item is wrong.
If it's ok for you I keep the PR open for the "correctness"

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -401,6 +401,7 @@ impl Session {
         paging_state: Option<Bytes>,
     ) -> Result<QueryResult, QueryError> {
         let query: Query = query.into();
+        let serialized_values = values.serialized()?;
 
         let response = self
             .run_query(
@@ -410,7 +411,7 @@ impl Session {
                 |connection: Arc<Connection>| {
                     // Needed to avoid moving query and values into async move block
                     let query_ref = &query;
-                    let values_ref = &values;
+                    let values_ref = &serialized_values;
                     let paging_state_ref = &paging_state;
 
                     async move {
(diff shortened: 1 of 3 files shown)


scylla/src/transport/session.rs
```diff
@@ -401,22 +401,23 @@ impl Session {
         paging_state: Option<Bytes>,
     ) -> Result<QueryResult, QueryError> {
         let query: Query = query.into();
-        let serialized_values = values.serialized();
```
piodul, line 403: The `ValueList` trait was designed so that `serialized()` is idempotent, i.e. calling `serialized()` on the result of previous `serialized()` is very cheap. The implementation of `ValueList` for `SerializedValues` just returns an object which wraps a reference to that `SerializedValues` - see [here](https://github.com/scylladb/scylla-rust-driver/blob/b6553c6d2ea05578f3648fd58f7f18266a57a8fe/scylla/src/frame/value.rs#L679-L683). Therefore serializing once at the beginning is a good thing because if we pass the serialized values to functions which take a `ValueList` as a parameter and call `serialized()`, those functions won't perform any actual serialization.

After your change, the heavy serialization will be done in `connection.query(...)` on line 418. This lambda may be called multiple times if the driver decides to retry the query, so this commit will actually increase the amount of work done in such situations.

I noticed that you removed `ValueList` impl for `SerializedResult`, which I think is a good thing, but it causes `query_paged` to no longer compile. However, this can be fixed in much simpler way than what this commit does - just add a question mark to the above line, like this:

```rust
let serialized_values = values.serialized()?;

```

piodul (changes requested):

piodul (approved):

piodul:
Merged manually to resolve conflicts, additionally removed one unused import in tests to get rid of a compiler warning.

Issue #336, opened 2021-11-12
Expose column spec information for the `_iter` interface


piodul:
Related to #271.

In #276, information about column names became available for the results of `Session::query` and `Session::execute` through the `QueryResult::col_specs` field.

We should also expose this information through the iterator object returned by `Session::query_iter` and `Session::execute_iter` methods.

cvybhu:
I implemented it when adding `query_all` to `Session`, but eventually these changes were dropped.
There is a commit on this branch: https://github.com/cvybhu/scylla-rust-driver/commits/queryall-backup:

https://github.com/cvybhu/scylla-rust-driver/commit/6bf71829821a1969a8c19f05b43eadba17242040

piodul:
Fixed by b46f913

PR #337, opened 2021-11-12
Add getting column spec for RowIterator

macher259:
Fixes: #336 
We add a getter for `ColumnSpec` for `*_iter` query methods. 
Enables users to learn about column names when they are using `*_iter` methods.
I took implementation from [cvybhu@6bf7182](https://github.com/cvybhu/scylla-rust-driver/commit/6bf71829821a1969a8c19f05b43eadba17242040) and made tests for it.
## Pre-review checklist

### DIFF
--- a/scylla/src/transport/iterator.rs
+++ b/scylla/src/transport/iterator.rs
@@ -18,7 +18,7 @@ use crate::cql_to_rust::{FromRow, FromRowError};
 use crate::frame::{
     response::{
         result,
-        result::{Row, Rows},
+        result::{ColumnSpec, Row, Rows},
         Response,
     },
     value::SerializedValues,
@@ -211,6 +211,11 @@ impl RowIterator {
         &self.tracing_ids
     }
 
+    /// Returns specification of row columns
+    pub fn get_column_specs(&self) -> &[ColumnSpec] {
+        &self.current_page.metadata.col_specs
+    }
+
     fn is_current_page_exhausted(&self) -> bool {
         self.current_row_idx >= self.current_page.rows.len()
     }
@@ -363,6 +368,11 @@ impl<RowT> TypedRowIterator<RowT> {
     /// If tracing was enabled returns tracing ids of all finished page queries
     pub fn get_tracing_ids(&self) -> &[Uuid] {
         self.row_iterator.get_tracing_ids()
+    }
+
+    /// Returns specification of row columns
+    pub fn get_column_specs(&self) -> &[ColumnSpec] {
+        self.row_iterator.get_column_specs()
     }
 }
 
--- a/scylla/src/transport/iterator.rs
+++ b/scylla/src/transport/iterator.rs
@@ -106,7 +106,7 @@ impl RowIterator {
         }
     }
 
-    pub(crate) fn new_for_query(
+    pub(crate) async fn new_for_query(
         query: Query,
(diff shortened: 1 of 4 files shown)


scylla/src/transport/session_test.rs
```diff
@@ -75,6 +75,14 @@ async fn test_unprepared_statement() {
...
         ]
     );
+    let query_result = session
+        .query_iter("SELECT a, b, c FROM ks.t", &[])
+        .await
+        .unwrap();
+    for (spec, name) in query_result.get_column_specs().iter().zip(["a", "b", "c"]) {
+        assert_eq!(spec.name, name); // Check column name.
+        assert_eq!(spec.table_spec.ks_name, "ks");
+    }
```
piodul, line 85: This test doesn't detect the case when the `get_column_specs` is empty. The `zip` method takes two iterators and produces a new one which stops after the shorter of them finishes ([doc link](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip)) - so if the columns specs slice is empty, the loop will immediately stop.

Besides, it's precisely what happens now. The way that the `RowIterator` is implemented now is that it caches a single page of rows and lazily updates it after the page is exhausted and the `next` method is called - but the iterator starts with a dummy, empty page (see e.g. [here](https://github.com/scylladb/scylla-rust-driver/blob/83199faac35accfd18910d9b5fa7cf5f08384248/scylla/src/transport/iterator.rs#L151)) which has empty metadata. So, until you call `next()` on the iterator for the first time, you will get empty metadata.

To solve this problem, let's change the `Session::execute_iter` and `Session::query_iter` so that they return an iterator already initialized with the first page. Most of the changes needed to implement this should happen in `RowIterator::new_for_query` and `RowIterator::new_for_prepared_statement`.

scylla/src/transport/iterator.rs
```diff
@@ -146,16 +146,24 @@ impl RowIterator {
 
         tokio::task::spawn(worker_task);
 
-        RowIterator {
+        let pages_received = receiver.recv().await.unwrap()?;
```
piodul, line 149: It's not really obvious if the worker always pushes at least one value before the channel is closed. If not, the `unwrap()` here will trigger a panic which will crash the application - and that is something we would like to avoid.

Let's be defensive and use `unwrap_with_default()` instead - it will return a `Default::default()` page if the channel does not produce any page.

Additionally, the variable holds only one page, so the page in the variable name should be singular (`page_received`).

The same comments apply to the logic in `new_for_prepared_statement`.

piodul: After our conversation in private I was convinced that unwrapping here is OK. It's a bug if there are no pages returned.

piodul:
Merged. The commits needed some cleaning up, so in order not to delay merging this PR further I reorganized them a little before merging. Some tips for the future:

- The third commit looks like a squash of three commits with unedited commit message after the squash. Remember that the first line in the message serves as a very short summary of what the commit does. That commit introduced a change which cause the RowIterator to have the first page already loaded after it was returning, but the commit message started with `result: removed deriving Default trait for Rows struct.` - which wasn't very informative. I rearranged the commit message a bit to make it more clear.
- You had a commit which introduced tests and then another which fixes them. It's better to have one commit which introduces correct tests, so I squashed them.

Issue #346, opened 2021-11-19
Expose schema metadata


piodul:
Other drivers provide information about the current schema of the cluster (e.g. gocql - see [here](https://pkg.go.dev/github.com/gocql/gocql#KeyspaceMetadata)). This driver currently does not. This kind of information is useful for generic applications which do not depend on a particular schema.

The driver already collects some basic information about keyspace replication strategy (see the `ClusterData` struct) so that the load balancing strategies work correctly, although it doesn't expose it.

This metadata should be periodically updated by the driver and made available to users. The `TopologyReader` could be extended to do that on cluster topology info refresh (although its name should be changed e.g. to MetadataReader in such case).

As a first step, we could expose some basic information about tables' schema:

- Column names and types (the type should include information if it is frozen or not),
- Column kinds (if they belong to partition key, clustering key, are they static or regular),
- Ordering of columns in partition key and clustering key,
- (Optional, but will be useful for https://github.com/scylladb/scylla-rust-driver/issues/194) The custom partitioner for the table (scylla-specific).

PR #349, opened 2021-11-23
Expose schema metadata

Ponewor:
This pull request introduces an insight into the schema of the cluster. From now on it is possible to access keyspace elements such as tables, their columns, types, kinds (regular, static, clustering, partition_key), ordering of partition key and clustering columns, UDTs, partitioner which are stored in `Metadata` structure (former `TopologyInfo`). The client has a way to disable this feature using a proper flag in `SessionConfig`.

This pull request introduces new external dependency namely strum.

## Pre-review checklist

Fixes: #346

### DIFF
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -81,6 +81,7 @@ impl Cluster {
     pub async fn new(
         initial_peers: &[SocketAddr],
         pool_config: PoolConfig,
+        fetch_schema_metadata: bool,
     ) -> Result<Cluster, QueryError> {
         let cluster_data = Arc::new(ArcSwap::from(Arc::new(ClusterData {
             known_peers: HashMap::new(),
@@ -101,6 +102,7 @@ impl Cluster {
                 initial_peers,
                 pool_config.connection_config.clone(),
                 server_events_sender,
+                fetch_schema_metadata,
             ),
             pool_config,
 
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -37,11 +37,11 @@ pub struct Datacenter {
 
 #[derive(Clone)]
 pub struct ClusterData {
-    pub known_peers: HashMap<SocketAddr, Arc<Node>>, // Invariant: nonempty after Cluster::new()
-    pub ring: BTreeMap<Token, Arc<Node>>,            // Invariant: nonempty after Cluster::new()
-    pub keyspaces: HashMap<String, Keyspace>,
-    pub all_nodes: Vec<Arc<Node>>,
-    pub datacenters: HashMap<String, Datacenter>,
+    pub(crate) known_peers: HashMap<SocketAddr, Arc<Node>>, // Invariant: nonempty after Cluster::new()
+    pub(crate) ring: BTreeMap<Token, Arc<Node>>, // Invariant: nonempty after Cluster::new()
+    pub(crate) keyspaces: HashMap<String, Keyspace>,
+    pub(crate) all_nodes: Vec<Arc<Node>>,
+    pub(crate) datacenters: HashMap<String, Datacenter>,
 }
 
 // Works in the background to keep the cluster updated
@@ -194,7 +194,7 @@ impl Cluster {
 impl ClusterData {
     /// Returns an iterator to the sequence of ends of vnodes, starting at the vnode in which t
     /// lies and going clockwise. Returned sequence has the same length as ring.
-    pub fn ring_range<'a>(&'a self, t: &Token) -> impl Iterator<Item = Arc<Node>> + 'a {
+    pub(crate) fn ring_range<'a>(&'a self, t: &Token) -> impl Iterator<Item = Arc<Node>> + 'a {
         let before_wrap = self.ring.range(t..).map(|(_token, node)| node.clone());
         let after_wrap = self.ring.values().cloned();
 
@@ -213,7 +213,7 @@ impl ClusterData {
         }
     }
 
-    pub async fn wait_until_all_pools_are_initialized(&self) {
+    pub(crate) async fn wait_until_all_pools_are_initialized(&self) {
         for node in self.all_nodes.iter() {
             node.wait_until_pool_initialized().await;
         }
@@ -221,7 +221,7 @@ impl ClusterData {
 
     /// Creates new ClusterData using information about topology held in `metadata`.
     /// Uses provided `known_peers` hashmap to recycle nodes if possible.
-    pub fn new(
+    pub(crate) fn new(
         metadata: Metadata,
         pool_config: &PoolConfig,
         known_peers: &HashMap<SocketAddr, Arc<Node>>,
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -5,7 +5,7 @@ use crate::transport::connection::{Connection, VerifiedKeyspaceName};
 use crate::transport::connection_pool::PoolConfig;
 use crate::transport::errors::QueryError;
 use crate::transport::node::Node;
-use crate::transport::topology::{Keyspace, TopologyInfo, TopologyReader};
+use crate::transport::topology::{Keyspace, Metadata, MetadataReader};
 
 use arc_swap::ArcSwap;
 use futures::future::join_all;
@@ -50,7 +50,7 @@ struct ClusterWorker {
     cluster_data: Arc<ArcSwap<ClusterData>>,
 
     // Cluster connections
-    topology_reader: TopologyReader,
+    metadata_reader: MetadataReader,
     pool_config: PoolConfig,
 
     // To listen for refresh requests
@@ -97,7 +97,7 @@ impl Cluster {
         let worker = ClusterWorker {
             cluster_data: cluster_data.clone(),
 
-            topology_reader: TopologyReader::new(
+            metadata_reader: MetadataReader::new(
                 initial_peers,
                 pool_config.connection_config.clone(),
                 server_events_sender,
@@ -121,7 +121,7 @@ impl Cluster {
             _worker_handle: worker_handle,
         };
 
-        result.refresh_topology().await?;
+        result.refresh_metadata().await?;
 
         Ok(result)
     }
@@ -130,7 +130,7 @@ impl Cluster {
         self.data.load_full()
     }
 
-    pub async fn refresh_topology(&self) -> Result<(), QueryError> {
+    pub async fn refresh_metadata(&self) -> Result<(), QueryError> {
         let (response_sender, response_receiver) = tokio::sync::oneshot::channel();
 
         self.refresh_channel
@@ -138,12 +138,12 @@ impl Cluster {
                 response_chan: response_sender,
             })
             .await
-            .expect("Bug in Cluster::refresh_topology sending");
+            .expect("Bug in Cluster::refresh_metadata sending");
         // Other end of this channel is in ClusterWorker, can't be dropped while we have &self to Cluster with _worker_handle
 
         response_receiver
             .await
-            .expect("Bug in Cluster::refresh_topology receiving")
+            .expect("Bug in Cluster::refresh_metadata receiving")
         // ClusterWorker always responds
     }
 
@@ -217,22 +217,22 @@ impl ClusterData {
         }
     }
 
-    /// Creates new ClusterData using information about topology held in `info`.
+    /// Creates new ClusterData using information about topology held in `metadata`.
     /// Uses provided `known_peers` hashmap to recycle nodes if possible.
     pub fn new(
-        info: TopologyInfo,
+        metadata: Metadata,
         pool_config: &PoolConfig,
         known_peers: &HashMap<SocketAddr, Arc<Node>>,
         used_keyspace: &Option<VerifiedKeyspaceName>,
     ) -> Self {
         // Create new updated known_peers and ring
         let mut new_known_peers: HashMap<SocketAddr, Arc<Node>> =
-            HashMap::with_capacity(info.peers.len());
+            HashMap::with_capacity(metadata.peers.len());
         let mut ring: BTreeMap<Token, Arc<Node>> = BTreeMap::new();
         let mut datacenters: HashMap<String, Datacenter> = HashMap::new();
-        let mut all_nodes: Vec<Arc<Node>> = Vec::with_capacity(info.peers.len());
+        let mut all_nodes: Vec<Arc<Node>> = Vec::with_capacity(metadata.peers.len());
 
-        for peer in info.peers {
+        for peer in metadata.peers {
             // Take existing Arc<Node> if possible, otherwise create new one
             // Changing rack/datacenter but not ip address seems improbable
             // so we can just create new node and connections then
@@ -276,7 +276,7 @@ impl ClusterData {
         ClusterData {
             known_peers: new_known_peers,
             ring,
-            keyspaces: info.keyspaces,
+            keyspaces: metadata.keyspaces,
             all_nodes,
             datacenters,
         }
@@ -427,12 +427,12 @@ impl ClusterWorker {
     }
 
     async fn perform_refresh(&mut self) -> Result<(), QueryError> {
-        // Read latest TopologyInfo
-        let topo_info = self.topology_reader.read_topology_info().await?;
+        // Read latest Metadata
+        let metadata = self.metadata_reader.read_metadata().await?;
         let cluster_data: Arc<ClusterData> = self.cluster_data.load_full();
 
         let new_cluster_data = Arc::new(ClusterData::new(
-            topo_info,
+            metadata,
             &self.pool_config,
             &cluster_data.known_peers,
             &self.used_keyspace,
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -281,6 +281,13 @@ impl ClusterData {
             datacenters,
         }
     }
+
+    /// Access keyspaces details collected by the driver
+    /// Driver collects various schema details like tables, partitioners, columns, types.
+    /// They can be read using this method
+    pub fn get_keyspace_info(&self) -> &HashMap<String, Keyspace> {
+        &self.keyspaces
+    }
 }
 
 impl ClusterWorker {
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -15,6 +15,7 @@ use uuid::Uuid;
 
 use super::connection::QueryResponse;
 use super::errors::{BadQuery, NewSessionError, QueryError};
+use crate::cql_to_rust::FromRow;
 use crate::frame::response::cql_to_rust::FromRowError;
 use crate::frame::response::result;
 use crate::frame::value::{BatchValues, SerializedValues, ValueList};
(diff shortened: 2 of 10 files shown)


scylla/src/transport/session.rs
```diff
@@ -46,7 +46,7 @@ use openssl::ssl::SslContext;
 
 /// `Session` manages connections to the cluster and allows to perform queries
 pub struct Session {
-    cluster: Cluster,
+    pub cluster: Cluster,
```
havaker, line 49: Exposing `cluster` as public will cause it to become part of public API, which is undesirable. If your goal was to provide a view to `keyspaces` field of `ClusterData`, It would be better to make a getter in form of public `Session` member function.

scylla/src/transport/session.rs
```diff
@@ -860,6 +862,13 @@ impl Session {
         self.metrics.clone()
     }
 
+    /// Access keyspaces details collected by the driver
+    /// Driver collects various schema details like tables, partitioners, columns, types.
+    /// They can be read using this method
+    pub fn get_keyspaces(&self) -> HashMap<String, Keyspace> {
```
havaker, line 868: From what I know, `HashMap<String, Keyspace>` can be quite heavy, as it contains description of all tables and types. Cloning it on each `get_keyspace` call is expensive, so it's worth getting rid of it. Returning a raw reference is not an option because of lifetime issues, but it could be worked around by returning an `Arc<Hashmap<_,_>>`, or a struct implementing `Deref` (similar to `MutexGuard`):
```rust
pub fn get_keyspaces(&self) -> KeyspacesRef {
    KeyspacesRef { cluster_data: self.cluster.get_data() }
}
```
```rust
#[derive(Clone)]
pub struct KeyspacesRef {
    cluster_data: Arc<ClusterData>
}

impl Deref for KeyspacesRef {
    type Target = HashMap<String, Keyspace>;

    fn deref(&self) -> &Self::Target {
        &self.cluster_data.keyspaces
    }
}
```

It could be used in a normal way:

```rust
let keyspaces = session.get_keyspaces();
println!("keyspace system = {:?}", keyspaces.get("system").unwrap());
```

The advantage of returning `KeyspaceRef` over `Arc<Hashmap<_,_>>` is that you do not have to create an `Arc` in the middle of `ClusterData`.

Ponewor: Isn't it weird that `KeyspacesRef` actually points to `ClustedData` not `Keyspace`?

havaker: Weirdness is subjective ;p
It has to point to `ClusterData`, because `ClusterData` owns `Keyspace`, and wee need to keep the owner alive while having a reference to its field - `Keyspace`. Besides, `KeyspaceRef` fields are invisible to the library user, so it won't be confusing (I hope).

@piodul any thoughts about this?

scylla/src/transport/topology.rs
```diff
@@ -41,6 +43,52 @@ pub struct Peer {
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Column {
+    pub type_: ColumnType,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ColumnType {
+    pub cql_type: CqlType,
+    pub frozen: bool,
```
piodul, line 67: Being frozen or not is relevant only for compound types: lists, sets, maps, tuples and user-defined types. Let's change the `CqlType` as follows:

```rust
pub enum CqlType {
    Native(NativeType),
    Compound {
        frozen: bool,
        type_: CompoundType,
    },
}
```

scylla/src/transport/topology.rs
```diff
@@ -290,12 +334,213 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+                    frozen = true;
+                    map_string_to_column(conn, keyspace_name, inner_type.to_string())
+                        .await?
+                        .cql_type
+                }
+                "map" | "tuple" => {
+                    let types = inner_type.split(", ").collect::<Vec<_>>();
+                    if outer_type == "map" {
+                        if types.len() != 2 {
+                            return Err(QueryError::InvalidMessage(format!(
```
piodul, line 544: I think that `ProtocolError` would be more suitable instead of your uses of `InvalidMessage`. AFAIK `InvalidMessage` means that an unparsable frame was received from the node, while `ProtocolError` means that the frame was parsed correctly, but something about its contents ws unexpected.

scylla/src/transport/session_test.rs
```diff
@@ -1345,4 +1345,49 @@ async fn test_schema_types_in_metadata() {
...
+                    d int STATIC,
+                    e int,
+                    f int,
+                    PRIMARY KEY ((c, e), b, a)
+                  )",
+            &[],
+        )
+        .await
+        .unwrap();
+
```
piodul, line 1382: To be 100% sure that the test works with clusters with size larger than one node, you should wait for schema agreement here.

scylla/src/transport/session_test.rs
```diff
@@ -1391,3 +1391,47 @@ async fn test_column_kinds_in_metadata() {
...
+                    g int,
+                    h int,
+                    i int STATIC,
+                    PRIMARY KEY ((c, e), b, a)
+                  )",
+            &[],
+        )
+        .await
+        .unwrap();
+
```
piodul, line 1430: Wait for schema agreement here.

scylla/src/transport/session_test.rs
```diff
@@ -1073,3 +1075,274 @@ async fn test_prepared_config() {
...
+        .query(
+            "CREATE TABLE IF NOT EXISTS table_b (
+                        a text PRIMARY KEY,
+                        b frozen<map<int, int>>
+                     )",
+            &[],
+        )
+        .await
+        .unwrap();
+
```
piodul, line 1144: Wait for schema agreement here.

scylla/src/transport/session_test.rs
```diff
@@ -1435,3 +1435,38 @@ async fn test_primary_key_ordering_in_metadata() {
...
+    session.query("USE test_metadata_ks", &[]).await.unwrap();
+
+    session
+        .query(
+            "CREATE TABLE t (pk int, ck int, v int, PRIMARY KEY (pk, ck, v))WITH cdc = {'enabled':true}",
+            &[],
+        )
+        .await
+        .unwrap();
+
```
piodul, line 1463: Wait for schema agreement here.

scylla/src/transport/session_test.rs
```diff
@@ -1435,3 +1435,38 @@ async fn test_primary_key_ordering_in_metadata() {
     assert_eq!(table.partition_key, vec!["c", "e"]);
     assert_eq!(table.clustering_key, vec!["b", "a"]);
 }
+
+#[tokio::test]
+async fn test_table_partition_in_metadata() {
```
piodul, line 1440: This test assumes that the cluster it connects to is Scylla. We running all tests with both Scylla and Cassandra in CI. The test should detect if it connected to Scylla and should skip itself if it is not.

scylla/src/transport/topology.rs
```diff
@@ -449,13 +450,52 @@ async fn query_table_schema(
         clustering_key[position as usize] = column_name;
     }
 
+    let partitioner = query_table_partitioner(conn, keyspace_name, table_name)
+        .await
+        .unwrap_or("".to_string());
```
piodul, line 455: I know that the custom partitioner info might not be present (Cassandra won't have this table and, but we shouldn't just ignore all errors. What if the query times out? In such case the user will get incorrect information from the metadata; it's better to fail the whole refresh in such situation instead.

If the `system_schema.scylla_tables` table is not present or there is no `partitioner` column, you will get `DbError::Invalid` - you can detect this and e.g. return `None` in such case.

piodul (changes requested):

scylla/src/transport/session.rs
piodul: @havaker @Ponewor Sometimes such a pattern is occasionally useful (e.g. [`MappedMutexGuard`](https://docs.rs/tokio/1.14.0/tokio/sync/struct.MappedMutexGuard.html) in tokio) but I don't think it is necessary here - we can just return `Arc<ClusterData>` directly. Additionally, please make its fields `pub(crate)` instead of `pub`, and provide a `pub` getter which returns a reference to the hashmap. We will probably want to expose more of its fields in the future, but for now let's restrict ourselves to the per-keyspace metadata.

scylla/src/transport/topology.rs
```diff
@@ -41,6 +43,52 @@ pub struct Peer {
...
+    Float,
+    Int,
+    BigInt,
+    Text,
+    Timestamp,
+    Inet,
+    List(Box<ColumnType>),
+    Map(Box<ColumnType>, Box<ColumnType>),
+    Set(Box<ColumnType>),
+    UserDefinedType(String, Vec<(String, ColumnType)>),
```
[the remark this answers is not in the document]
Ponewor: UDTs were not in the scope of this issue and that is why I have decided to do it this way but clearly expanding this issue in the way you suggest yields better results.

scylla/src/transport/topology.rs
Ponewor: The thing with `ProtocolError` is that it only accepts `&'static str`...

scylla/src/transport/topology.rs
Ponewor: Aren't tuples always frozen?

piodul: Indeed they are, my mistake. How about using the following variants: `Native, Collection, Tuple, UserDefinedType` (Collection and UserDefinedType would have the frozen flag)?

scylla/src/transport/topology.rs
piodul: I see... Changing all ProtocolErrors to accept String would be a lot of effort, so let's keep using InvalidMessage there. I doubt that users will ever want to differentiate between those variants, anyway.

Please add FIXME comments to the places where you return InvalidMessage telling that ProtocolError should be returned there but cannot due to this issue.

piodul:
@Ponewor One more thing - some users have an enormous amount of keyspaces and tables in their cluster but they might be not interested in accessing information about schema. We should provide an opt-out switch in the session configuration to turn off fetching schema information (fetching topology information should be always on).

scylla/src/transport/session_test.rs
Ponewor: Is there a way to get this information from the `session` itself or should I somehow modify workflows to pass this information to the runtime (e.g. with environment variable)?

piodul: I think that an environment variable set by the workflow is a better choice.

scylla/src/transport/topology.rs
Ponewor: `None`? So do you prefer to store the partitioner as an `Option<String>` rather than `String`?

piodul: Yes. The lack of a custom partitioner is represented by `null` in the `partitioner` column, so I think it makes sense to use `Option`.

scylla/src/transport/session.rs
Ponewor: I don't get it? What should be returning `Arc<ClusterData>`?

piodul: The idea is to write two getters:

- `Session::get_cluster_data(&self) -> Arc<ClusterData>`
- `ClusterData::get_keyspace_info(&self) -> &HashMap<String, Keyspace>`

Ponewor:
> @Ponewor One more thing - some users have an enormous amount of keyspaces and tables in their cluster but they might be not interested in accessing information about schema. We should provide an opt-out switch in the session configuration to turn off fetching schema information (fetching topology information should be always on).

I have tried to do this the simplest way and here is what I have achieved:
The struct
```rust
pub struct Keyspace {
    pub strategy: Strategy,
    pub tables: HashMap<String, Table>,
    pub user_defined_types: HashMap<String, Vec<(String, CqlType)>>,
}
```
has turned into

```rust
pub struct Keyspace {
    pub strategy: Strategy,
    pub tables: Option<HashMap<String, Table>>,
    pub user_defined_types: Option<HashMap<String, Vec<(String, CqlType)>>>,
}
```
as `strategy` is required and `tables` and `user_defined_types are optional`. I do not like it. It makes using the API cumbersome as it may be observed in tests. In the future `Keyspace` will gain additional fields and all of them would be `Option`s if we continue this way. Maybe it would be better to turn it into
```rust
pub struct Keyspace {
    pub strategy: Strategy,
    pub schema: Option<Schema>,
}

pub struct Schema {
    pub tables: HashMap<String, Table>,
    pub user_defined_types: HashMap<String, Vec<(String, CqlType)>>,
}
```
? What do you think? Maybe I am being overly dramatic over few `unwrap`s? If I am not, is schema the right name for this additional struct? I wonder if schema is a more general term containing all of the keyspaces.

piodul:
> ? What do you think? Maybe I am being overly dramatic over few `unwrap`s? If I am not, is schema the right name for this additional struct? I wonder if schema is a more general term containing all of the keyspaces.

@Ponewor I agree that the `Option`s make it cumbersome. I think we can get rid of them and just leave the hashmaps empty if fetching schema information is disabled - if the user disabled it then they won't be interested in looking at those fields. A docstring comment for those fields mentioning this behavior would be nice to have.

scylla/src/transport/topology.rs
```diff
@@ -301,12 +371,274 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+    Ok(rows
+        .into_typed::<(Option<String>,)>()
+        .next()
+        .ok_or(QueryError::DbError(
+            DbError::Invalid,
+            format!(
+                "no entry for table {}.{} in system_schema.scylla_tables",
+                keyspace_name, table_name
+            ),
+        ))?
```
Ponewor, line 637: So it turns out that not all tables have a corresponding entry in system_schema.scylla_tables. Is it the expected behaviour?

scylla/src/transport/topology.rs
```diff
@@ -301,12 +371,274 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+    let mut clustering_key_map: HashMap<i32, String> = HashMap::new();
+
+    for row in rows.into_typed::<(String, String, i32, String)>() {
+        let (column_name, kind, position, type_) = row.map_err(|_| {
+            QueryError::ProtocolError("system_schema.columns has invalid column type")
+        })?;
+
+        if type_ == INVALID_TYPE {
+            continue;
+        }
```
Ponewor, line 495: So there's this column `system.IndexInfo.value` with type `empty`? What is it?

scylla/src/transport/cluster.rs
```diff
@@ -37,11 +37,11 @@ pub struct Datacenter {
 
 #[derive(Clone)]
 pub struct ClusterData {
```
piodul, line 39: Although `ClusterData` is `pub` in this module, the module `cluster` itself is not declared public, therefore the `ClusterData` struct is not accessible from outside of the `scylla` crate. Please make the `cluster` module public.

Additionally:
- Please go over all public methods of `ClusterData` which were not added in this PR and make them `pub(crate)`.
- Make the `Cluster` struct `pub(crate)`.

scylla/src/transport/topology.rs
```diff
@@ -21,7 +24,7 @@ pub struct MetadataReader {
     control_connection_address: SocketAddr,
     control_connection: NodeConnectionPool,
 
-    // when control connection fails, TopologyReader tries to connect to one of known_peers
+    // when control connection fails, MetadataReader tries to connect to one of known_peers
```
piodul, line 27: Nit: there seem to be some leftover renames in this commit which should be included in the first commit.

scylla/src/transport/topology.rs
```diff
@@ -549,6 +557,37 @@ fn map_string_to_cql_type(type_: &str) -> Result<CqlType, InvalidCqlType> {
     })
 }
 
+async fn query_table_partitioner(
```
piodul, line 621: As with `query_columns`, please fetch all the contents of `scylla_tables` into a hashmap in one go.

scylla/src/transport/topology.rs
```diff
@@ -380,9 +381,54 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+
+    for row in rows.into_typed::<(String, Vec<String>, Vec<String>)>() {
+        let (type_name, field_names, field_types) = row.map_err(|_| {
+            QueryError::ProtocolError("system_schema.types has invalid column type")
+        })?;
+
+        let mut fields = Vec::with_capacity(field_names.len());
+
+        for (field_name, field_type) in field_names.iter().zip(field_types.iter()) {
+            fields.push((field_name.clone(), map_string_to_cql_type(field_type)?));
```
piodul, line 443: If you use `field_names.into_iter()`, you can avoid cloning the `field_name`.

scylla/src/transport/session.rs
```diff
@@ -102,6 +102,9 @@ pub struct SessionConfig {
     pub disallow_shard_aware_port: bool,
 
     pub default_consistency: Consistency,
+
+    /// If true, full schema is fetched with every metadata refresh.
+    pub fetch_schema_metadata: bool,
```
piodul, line 107: We need a corresponding method in `SessionBuilder` for this field.

piodul (changes requested):

scylla/src/transport/cluster.rs
havaker: Wouldn't it be less invasive to reexport `ClusterData` with `pub use` in `transport/mod.rs` and keep `cluster` module private? Marking many things as `pub(crate)` is a bit tedious.

```rust
pub use cluster::ClusterData;
```

piodul: I agree, that's a better idea.

Ponewor: Is it ok now?

scylla/src/transport/topology.rs
```diff
@@ -301,12 +400,276 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+    Ok(match NativeType::from_str(type_) {
+        Ok(type_) => CqlType::Native(type_),
+        _ => CqlType::UserDefinedType {
+            frozen,
+            name: type_.to_string(),
+        },
+    })
+}
+
+async fn query_table_partitioners(
```
piodul, line 646: The CI check which runs the tests with a Cassandra instance fails because the `scylla_tables` table does not exist. This function already used to handle such a situation in one of the previous versions of this PR.

scylla/src/transport/cluster.rs
piodul: The `ring_range`, `wait_until_all_pools_are_initialized` and `new` methods are still visible. The documentation generated by `cargo doc` shows this:

image

Those methods weren't exposed to the users before, they are unrelated to the functionality introduced by this PR so we shouldn't expose them in this PR. Changing their visibility to `pub(crate)` should hide them.

To be clear,`get_keyspace_info` is visible and that's OK.

scylla/src/transport/topology.rs
```diff
@@ -301,12 +400,276 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+        for (position, column_name) in partition_key_columns {
+            partition_key[position as usize] = column_name;
+        }
+
+        let mut clustering_key = vec!["".to_string(); clustering_key_columns.len()];
+        for (position, column_name) in clustering_key_columns {
+            clustering_key[position as usize] = column_name;
+        }
+
+        result.insert(
```
piodul, line 576: Tip: you can avoid cloning keyspace and table names. You can assign the partitioner to a variable first, then call 

```rust
result.insert((keyspace_name, table_name), Table { columns, parition_key, clustering_key, partition })
```

scylla/src/transport/topology.rs
```diff
@@ -301,12 +400,276 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
 
         let strategy: Strategy = strategy_from_string_map(strategy_map)?;
 
-        result.insert(keyspace_name, Keyspace { strategy });
+        result.insert(
+            keyspace_name.clone(),
```
piodul, line 404: Tip: You can avoid cloning here, too, see my other comment about `(keyspace_name, table_name)`.

scylla/src/transport/topology.rs
```diff
@@ -301,12 +400,276 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+    for row in rows.into_typed::<(String, String)>() {
+        let (keyspace_name, table_name) = row.map_err(|_| {
+            QueryError::ProtocolError("system_schema.tables has invalid column type")
+        })?;
+
+        result
+            .entry(keyspace_name.clone())
+            .or_insert(HashMap::new())
+            .insert(
+                table_name.clone(),
```
piodul, line 482: Same here.

piodul (commented):
Looks good overall. Please fix the issues found by the CI + address my comments and I think the PR will be ready for merging.

scylla/src/transport/topology.rs
Ponewor: It does not work this way, Rust still complains about the value being moved even after variable assignment :(

piodul: There is a way out:

```rust
let keyspace_and_table_name = (keyspace_name, table_name);

let partitioner = partitioners
    .remove(&keyspace_and_table_name)
    .unwrap_or_default();

result.insert(
    keyspace_and_table_name,
    Table {
        columns,
        partition_key,
        clustering_key,
        partitioner,
    },
);
```

Ponewor: Huh, much simpler than I have expected. I have tried something among those lines but with no success. Thanks :)

scylla/src/transport/topology.rs
```diff
@@ -293,20 +385,297 @@ async fn query_keyspaces(conn: &Connection) -> Result<HashMap<String, Keyspace>,
...
+    conn: &Connection,
+) -> Result<HashMap<(String, String), Option<String>>, QueryError> {
+    let mut partitioner_query = Query::new(
+        "select keyspace_name, table_name, partitioner from system_schema.scylla_tables",
+    );
+    partitioner_query.set_page_size(1024);
+
+    let rows = match conn.query_all(&partitioner_query, &[]).await {
+        Err(QueryError::DbError(DbError::Invalid, message))
+            if message == "table scylla_tables does not exist" =>
```
piodul, line 659: I'd rather not rely on the error string. Different Cassandra versions might have some differences in the error message (e.g. spelling, capitalization, quotes, etc.). I understand that you would like this to match on the "table scylla_tables does not exist" error only, but currently I think checking the error code is the best thing we can do without making this check fragile. Please remove the string comparison.

In the future, we can improve the situation by adding detection of the database vendor and version and we would check this information before deciding to query or not. I suggest leaving a FIXME here.

piodul (approved):

PR #401, opened 2022-02-17
result: add possibility to borrow map values

procrastinationfighter:
This PR adds method to `CqlValue` that allows to borrow map values represented as `CqlValue::Map`.
I decided to use the naming convention analogous to sets and lists: the borrowing methods for set and list are called `as_set` and `as_list` and the method that takes ownership of the value is called `into_vec`, therefore for maps it is `as_map` and `into_pair_vec` respectively.
## Pre-review checklist


piodul (approved):
LGTM. We fixed a bug in the CI configuration but it won't affect your PR unless you rebase, so please do.

Issue #459, opened 2022-06-17
Wait for schema agreement automatically


piodul:
Currently, after sending a schema-altering query, the node handling the query does not wait until other nodes in the cluster learn about the new schema element, so it's the users' responsibility to make sure to wait for schema agreement before using the new schema element. This includes cases such as creating a table in keyspace, sending queries to new table, using a newly-introduced UDT etc. It is easy to forget about this when writing an application, and such problems happen much more rarely when developing locally with a single node.

I propose to add automatic wait for schema agreement after a schema-altering query in order to help users avoid such bugs. This should be easy to implement because Scylla/Cassandra responds with a different message to schema-altering queries (see [here](https://github.com/scylladb/scylla-tools-java/blob/d4133b54c9c2753891b072c3464ef6df1cfaa716/doc/native_protocol_v4.spec#L742-L751)).

Waiting for schema agreement more than necessary is never a bug, but might slow down applications which do a lot of schema changes (e.g. a migration). In case where somebody wishes to create a keyspace and then a lot of tables in it, it makes sense only to wait after creating a keyspace and after creating all the tables rather than after every query. To support such use case, it should be possible to opt-out of this feature (probably makes sense to put this into an execution profile after they are implemented).

Automatic wait for schema agreement should probably use a configurable timeout (what should be the default?).

PR #460, opened 2022-06-17
session: automatically wait for schema agreement

piodul:
Now, the session automatically waits for schema agreement after
executing a statement which modifies the schema. The rationale for this
change is that, previously, it was the user's responsibility to wait for
schema agreement and it was easy to forget about this check.

The automatic wait for schema agreement is never a bug, although it can
slow down some programs which do a lot of schema changes in a short
succession. Those applications can opt-out of this feature by disabling
the feature in the configuration and can still wait for schema agreement
manually.

The session will wait for schema agreement for a configured amount of
time, and then - if the agreement is not reached - it will return
a TimeoutError.

Fixes: #459

## Pre-review checklist

- [ ] ~I have split my patch into logically separate commits.~

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -158,6 +158,13 @@ impl QueryResponse {
         }
     }
 
+    pub fn as_schema_change(&self) -> Option<&result::SchemaChange> {
+        match &self.response {
+            Response::Result(result::Result::SchemaChange(sc)) => Some(sc),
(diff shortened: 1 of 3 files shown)


piodul:
Fun fact: 6 of our current tests fail when ran against a two-node cluster because they don't wait for schema agreement. This PR fixes those tests.

psarna (approved):

Issue #469, opened 2022-07-09
Expose a `batch` method on CachingSession


colin-grapl:
Currently there doesn't seem to be any support for batch operations on CachingSession.

As `Session` exposes a `.batch`, which takes a Batch, which itself takes PreparedStatement's, it'd be nice to have that natively supported.

colin-grapl:
https://github.com/colin-grapl/scylla-rust-driver/blob/main/scylla/src/transport/caching_session.rs#L97

Wrote this up pretty quickly. I think it gets the idea across. I may have botched the error handling.

cvybhu:
Thank you for the suggestion and code.

It turns out that `CachingSession` doesn't need to handle repreparing statements.
Once #473 is merged it should be easy to add `CachingSession::batch` as well.

One improvement I would do is to do the `add_prepared_statement`s in parallel using something like `join_all`.

colin-grapl:
Ah, that's awesome. That was by far the most confusing part of this. I agree entirely re: using join_all to do this concurrently.

I'll update the PR to:
1. Remove the reprep code (yay)
2. Optimize this a bit

colin-grapl:
^ Done, feel free to grab that code and do whatever you'd like. I don't really want to go through a PR process myself since I have a workaround in my own code.

PR #473, opened 2022-07-12
Remove reprepare code from CachingSession, add reprepare tests

cvybhu:
Sometimes the prepared statement cache gets cleared and we need to prepare a `PreparedStatement` again.
This happens e.g. after `ALTER TABLE` where a column might change its name.

In such cases `Session` automatically reprepares the `PreparedStatement` under the hood. This happens in `Connection::execute` and `Connection::batch`.

`CachingSession` should also handle such cases automatically. It has some code that is supposed to handle them, but it isn't actually needed. Even if it weren't there `Session` would handle repreparing everything,

This PR first adds tests which ensure that repreparation works correctly in `Session` and `CachingSession`, and then removes repreparation code from `CachingSession`. After removal the tests still pass.

This affects #469, implementing `CachingSession::batch` should be much simpler once we don't have to care about repreparing.

I added `tracing-subscriber` to dev-dependencies. It makes it easier to start the test with logs enabled and see that `Session` actually reprepares the statements as it should.

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -41,6 +41,7 @@ smallvec = "1.8.0"
 
 [dev-dependencies]
 criterion = "0.3"
+tracing-subscriber = "0.3.14"
 
 [[bench]]
 name = "benchmark"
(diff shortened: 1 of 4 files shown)


psarna (approved):

cvybhu:
Could we merge this?
Pinging @piodul in case his review is needed as well

piodul (approved):

Issue #476, opened 2022-07-15
Instructions/scripts for running tests on non-Linux systems


piodul:
Many of our unit tests require a Scylla instance to be available. Currently, our recommendation is to start a Scylla Docker container and point to its IP via an environment variable when running `cargo test` on the host. The problem with this approach is that it does not work on MacOS and Windows - unlike on Linux, on those systems the internal Docker IPs are not reachable from host.

A workaround for that is publishing the container's ports and connecting to `127.0.0.1:9042`, which works due to a workaround introduced by https://github.com/scylladb/scylla-rust-driver/pull/173: when querying system tables, instead of fetching rpc_address for the current node, the control connection's address is used instead. This works OK for current tests, but doesn't work well with multi-node clusters (system.peers contains docker containers' IPs) and definitely won't work if we decide to add tests which require a multi-node cluster. Additionally, this is a bit problematic for https://github.com/scylladb/scylla-rust-driver/issues/470 and we would like to remove it.

Because we would like to remove this workaround, instructions are needed for non-Linux contributors on how to run tests. We could introduce a script which runs the tests in a Docker container on non-Linux systems (scylla-go-driver already does that, [see here](https://github.com/scylladb/scylla-go-driver/blob/72ce015d584c1bc0530688af01e26729bbf8bfa2/Makefile#L48-L58)).

PR #479, opened 2022-07-20
Added scripts for running tests in docker.

wprzytula:
This way developers on platforms other than Linux will be able to test
their changes without hassle, and we can remove the workaround #173
that has served for this purpose. Fixes: #476

## Pre-review checklist

### DIFF
--- a/scripts/test_in_docker/test_in_docker.ps1
+++ b/scripts/test_in_docker/test_in_docker.ps1
@@ -1,0 +1,28 @@
+﻿if ( -not (docker container ls | Select-String scylla_rust_driver)) {
+        echo "Scylla container must be running to execute tests (use 'make scylla-up')."
+        exit 1
+}
+
+docker image ls -q scylla_rust_driver_testing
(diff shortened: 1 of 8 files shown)


scripts/test_in_docker/test_in_docker.ps1
```diff
@@ -0,0 +1,28 @@
+﻿if ( -not (docker container ls | Select-String scylla_rust_driver)) {
+        echo "Scylla container must be running to execute tests (use 'make scylla-up')."
+        # exit 1
```
piodul, line 3: It looks like it is commented out. Why?

piodul (commented):
Please modify the `Testing` section in `CONTRIBUTING.md` so that it mentions the new scripts that you added.

scripts/test_in_docker/test_in_docker.ps1
wprzytula: Accidentally left commented after tinkering with Windows. Will fix this.

PR #489, opened 2022-08-02
Enhanced CI cluster: now 3-node instead of 1-node

wprzytula:
As it enables testing more sophisticated scenarios such as retries
and speculative executions, we made the cluster in CI consist
of 3 nodes. The existing tests pass seamlessly on such cluster.

## Pre-review checklist

### DIFF
--- a/test/docker-compose.yml
+++ b/test/docker-compose.yml
@@ -1,0 +1,35 @@
+version: "3.7"
+
+networks:
+  public:
+    name: scylla_rust_driver_public
+    driver: bridge
+    ipam:
+      driver: default
+      config:
+        - subnet: 172.42.0.0/16
+services:
+  scylla1:
+    image: scylladb/scylla
+    volumes:
+      - ${GITHUB_WORKSPACE}:/workspace
+    networks:
+      public:
+        ipv4_address: 172.42.0.2
+    command: --rpc-address 172.42.0.2 --listen-address 172.42.0.2 --seeds 172.42.0.2 --skip-wait-for-gossip-to-settle 0
+  scylla2:
+    image: scylladb/scylla
+    volumes:
+      - ${GITHUB_WORKSPACE}:/workspace
+    networks:
+      public:
+        ipv4_address: 172.42.0.3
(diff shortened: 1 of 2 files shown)


wprzytula:
The changes are ready for review.

test/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      - 127.0.0.1:9042:9042
+      - 127.0.0.1:9142:9142
+      - 127.0.0.1:19042:19042
+      - 127.0.0.1:19142:19142
+    volumes:
+      - ${GITHUB_WORKSPACE}:/workspace
+    networks:
+      public:
+        ipv4_address: 172.42.0.2
+    command: --rpc-address 172.42.0.2 --listen-address 172.42.0.2 --seeds 172.42.0.2 --skip-wait-for-gossip-to-settle 0 --thread-affinity false
```
piodul, line 24: `--thread-affinity false` is redundant here, by default Scylla Docker instances use `--overprovisioned` which already disables thread affinity and does even more. See here: https://github.com/scylladb/scylladb/blob/master/dist/docker/commandlineparser.py

test/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      config:
+        - subnet: 172.42.0.0/16
+services:
+  scylla1:
+    image: scylladb/scylla
+    ports:
+      - 127.0.0.1:9042:9042
+      - 127.0.0.1:9142:9142
+      - 127.0.0.1:19042:19042
+      - 127.0.0.1:19142:19142
```
piodul, line 18: Are those published ports used? The nodes are using `172.x.x.x` addresses as `rpc-address`, so the driver will attempt to connect to that address directly rather than on localhost. I assume that everything is running on Linux, so the `172.x.x.x` addresses are reachable.

Perhaps we could get rid of the published ports to reduce confusion, and set SCYLLA_URI appropriately when running tests.

piodul (approved):

Jasperav:
@wprzytula I am using this docker-compose file for my own ci, it is great to have multiple nodes running. I am unable to make it work on macOS, I asked it in the Slack group but no easy solution. Do you have a solution to make it work on macOS? Slack question: https://scylladb-users.slack.com/archives/C2NLNBXLN/p1663944339848569

PR #495, opened 2022-08-08
Fixed missing tokio feature in scylla-cql

wprzytula:
Contrary to previous assumptions, tokio feature "io-util" is required
for `scylla-cql` crate to build. Moreover, I added the new crate
to Cargo workspace.

## Pre-review checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,6 +26,8 @@ jobs:
       run: cargo clippy --verbose --examples --tests
     - name: Cargo check without features
       run: cargo check --manifest-path "scylla/Cargo.toml" --features ""
+    - name: Build scylla-cql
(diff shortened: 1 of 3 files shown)


piodul (commented):
Good catch. The CI probably didn't check this because we are building everything together. AFAIK, If a crate requires a feature from a particular package, then it is enabled globally - hence, when running `cargo check`, `scylla-cql` is built against `tokio` with "io-util" enabled because the `scylla` crate requires it.

To prevent such issues in the future, please also add a CI step which builds the `scylla-cql` crate only. Remember to put it both in the `build` (newest rust) and `min_rust` (minimum supported rust version) jobs in the `rust.yml` workflow.

piodul (approved):

PR #497, opened 2022-08-10
Remove unnecessary `pub`

Ten0:
(non backwards-compatible)

Alternately we could make this return an explicit type so that it's expressable, but I doubt that would actually be necessary for anyone.
https://github.com/scylladb/scylla-rust-driver/blob/db8d52893bdef56977872a09f51fda2fa423168f/scylla-cql/src/frame/value.rs#L189


wprzytula:
Closing as obsolete.

PR #525, opened 2022-08-19
Remove lines starting with # when generating Sphinx docs

cvybhu:
Fixes: #521

I tried to do the same thing that we do with mdbook source to get rid of sphinx ` ```eval_rst ` bits.
First I moved all Sphinx related stuff to a subdirectory called `sphinx`,
Then I made it so that on each sphinx build source markdown is generated on the fly by taking the original files and modifying them to remove lines starting with `#`.

Sadly there are some problems with this approach.

To generate the docs we use `sphinx-multiversion`, which supposedly is able to generate the docs for many different versions by looking at the git history or something like that. Right now the only version on https://rust-driver.docs.scylladb.com is `stable`, but maybe we would like to make it work in the future.

Will it work if documentation source files are not tracked by git but generated each time they're needed?
`sphinx-multiversion` already doesn't work after this change, it says `No matching refs found!`
I found some issue that says this is becaus the files are not commited to git: https://github.com/Holzhaus/sphinx-multiversion/issues/75, but I tried commiting them and that didn't help.

Changing make target from `sphinx-multiversion` to `dirhtml` allows to generate html for the latest version, but that isn't an ideal solution :/.

I'm not sure what to do about it, I would have to learn sphinx better or ask dgarcia for help here.


cvybhu:
@dgarcia360 any tips on how to approach this?

dgarcia360:
@cvybhu In the [Cpp driver](https://github.com/scylladb/cpp-driver/tree/master/docs ) project we are also postprocessing the docs before generating the HTML.

To do so, we edited the [Makefile](https://github.com/scylladb/cpp-driver/blob/master/docs/Makefile#L7) ``setup`` function to copy the contents to a folder named ``_source`` before every other command runs:

```
SOURCEDIR     = _source

.PHONY: setup
setup:
	$(POETRY) install
	$(POETRY) update
	cp -TLr source $(SOURCEDIR)
```

Then, we edited the ``multiversion`` function from the same file to build the docs from a bash script:

```
.PHONY: multiversion
multiversion: setup
	$(POETRY) run ./_utils/multiversion.sh
	@echo
	@echo "Build finished. The HTML pages are in $(BUILDDIR)/dirhtml."
```

The [bash script](https://github.com/scylladb/cpp-driver/blob/master/docs/_utils/multiversion.sh) runs ``sphinx-multiversion`` with the option ``--pre-build``. This option allows you to pass a command that will be executed just before the docs for a specific version are built.

```
#! /bin/bash	
	 
	cd .. && sphinx-multiversion docs/source docs/_build/dirhtml \
	 --pre-build "python <PATH_TO_YOUR_SCRIPT>'"
````

When applying the changes, please try to stick to the same organization and file / directory names the cpp-driver uses . This will make things easier if we need to update the theme or workflows in the future.

cvybhu:
> @cvybhu In the [Cpp driver](https://github.com/scylladb/cpp-driver/tree/master/docs) project we are also postprocessing the docs before generating the HTML.
> 
> To do so, we edited the [Makefile](https://github.com/scylladb/cpp-driver/blob/master/docs/Makefile#L7) `setup` function to copy the contents to a folder named `_source` before every other command runs:
> 
> ```
> SOURCEDIR     = _source
> 
> .PHONY: setup
> setup:
> 	$(POETRY) install
> 	$(POETRY) update
> 	cp -TLr source $(SOURCEDIR)
> ```
> 
> Then, we edited the `multiversion` function from the same file to build the docs from a bash script:
> 
> ```
> .PHONY: multiversion
> multiversion: setup
> 	$(POETRY) run ./_utils/multiversion.sh
> 	@echo
> 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/dirhtml."
> ```
> 
> The [bash script](https://github.com/scylladb/cpp-driver/blob/master/docs/_utils/multiversion.sh) runs `sphinx-multiversion` with the option `--pre-build`. This option allows you to pass a command that will be executed just before the docs for a specific version are built.
> 
> ```
> #! /bin/bash	
> 	 
> 	cd .. && sphinx-multiversion docs/source docs/_build/dirhtml \
> 	 --pre-build "python <PATH_TO_YOUR_SCRIPT>'"
> ```
> 
> When applying the changes, please try to stick to the same organization and file / directory names the cpp-driver uses . This will make things easier if we need to update the theme or workflows in the future.

Thank you, closing in favour of #537

PR #526, opened 2022-08-22
tracing: Get list of nodes from  TracingInfo

cvybhu:
Adds a simple getter which allows to get the list of nodes involved in a traced query.

## Pre-review checklist

### DIFF
--- a/scylla/src/tracing.rs
+++ b/scylla/src/tracing.rs
@@ -1,4 +1,5 @@
 use crate::statement::Consistency;
+use itertools::Itertools;
 use std::collections::HashMap;
 use std::net::IpAddr;
 use std::num::NonZeroU32;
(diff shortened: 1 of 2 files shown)


cvybhu:
> For some reason cargo clippy retrurns 0 on warnings :O

Ah nevermind I didn't notice the `RUSTFLAGS=-Dwarnings` in CI script

psarna (approved):
Beautiful in its simplicity. As we already discussed, it's rather unlikely for the tracing info to have more than a few entries, so a linear getter sounds perfectly acceptable as opposed to trying to track the replicas somewhere else. Note that this getter would print all nodes taking part in the query, regardless if they act as coordinators or replicas, but if we assume token awareness, it would effectively return replica addresses.

piodul (approved):

PR #537, opened 2022-08-25
Remove lines starting with # from Sphinx docs 1/2

cvybhu:
This PR adds preprocessing for Sphinx markdown so that code lines starting with # are removed.
This is a better version of #525, using the instructions provided by @dgarcia360.
I tried to keep it similar to `cpp-driver` code.

The `setup` target now copies the `source` directory to `_source` and takes care of removing #s in all files withing `_source` using a python script.

The `multiversion` target isn't modified yet. I tried to add the `--pre-build` flag, but it fails with `file not found` error.
I think this is because `sphinx-multiversion` checks out to the `main` branch and tries to run the script there, but it isn't merged yet so it doesn't find it. Because of that preprocessing for `sphinx-multiversion` will be done in a follow-up PR(#538) once this PR is merged and the script ends up on the `main` branch.

## Pre-review checklist


cvybhu:
dgarcia seems busy, but we can try merging it and seeing what happens. I'm 90% sure that it's gonna work.
If something goes wrong it can be reverted quickly.

dgarcia360 (commented):
The docs build without errors and drops the hashes locally as expected. 

However, it will not drop the hashes in production since the command used is [make multiversion](https://github.com/scylladb/scylla-rust-driver/pull/537/files#diff-cdec797dddf8fc387304501d6bd4f318e1e9067c6fae012dfa8b69fd85fd3248L69) and not `make preview`.

To solve this, we should call `sphinx-multiversion` with the ``--pre-build`` option as explained in this comment: https://github.com/scylladb/scylla-rust-driver/pull/525#issuecomment-1223826414

I was thinking we could merge this PR, and then I submit another one adding the ``pre-build`` option to the multiversion command. What do you think?

docs/Makefile
```diff
@@ -5,7 +5,7 @@ SPHINXOPTS    =
 SPHINXBUILD   = $(POETRY) run sphinx-build
 PAPER         =
 BUILDDIR      = _build
-SOURCEDIR     = source
+SOURCEDIR     = _source
```
dgarcia360, line 8: We should add ``docs/_source`` to the .gitignore file.

cvybhu:
v2:
* Added `_docs/source` to .`.gitignore`
* Print removed lines in docs workfflow to make sure thay are correct

PR #541, opened 2022-08-31
Add separate struct for view metadata to store more info

Gor027:
Added MaterlizedView struct to store base table name for a view.
This will allow to find all materialized views for a given table name, or find the base table metadata for a certain view.

## Pre-review checklist


cvybhu (changes requested):
Looks good.

Could you extend `test_views_in_schema_info` in `scylla/src/transport/session_test.rs` to test that base table names are correct.

cvybhu (approved):

Issue #544, opened 2022-09-06
Immediately fetch new topology data after schema changes


cvybhu:
When there is a schema altering query (`CREATE KEYSPACE`, `DROP TABLE`, ...) the driver automatically awaits schema agreement to make sure that the created entity can be used on any node.

Queries like `CREATE KEYSPACE` affect load balancing, but new topology information is not fetched immediately. The queries will be sent to the wrong nodes until the new information is fetched by periodic topology updates (60s (!)).

We could call `refresh_metadata` inside `handle_auto_await_schema_agreement` to fix this.

Pros:
* Load balancing chooses the right nodes as soon as possible, important in benchmarks
* Tests that handle metadata information don't require calling `refresh_metadata` immediately

Cons:
* Performance problems?

piodul:
@Gor027 is probably working on this right now, it's needed for https://github.com/scylladb/cpp-rust-driver/pull/72 .

Schema altering queries are rather rare so fetching the full metadata shouldn't be a big problem. The response to a schema altering query contains information what was changed by the schema, so perhaps we could optimize this later and only fetch information on the control connection about the changed items.

wyfo:
> We could call refresh_metadata inside handle_auto_await_schema_agreement to fix this.

Does it means that when several schema-related queries are made successively, `refresh_metadata` would be called for each of them successively?

Gor027:
I have tested the approach described in the above comment in the `cpp-rust` bindings with `CassandraTypesTests.UDT` test which creates a new user-defined type and then tries to get information about it from the keyspace metadata. The schema metadata is being refreshed with a little delay because of `tokio::select` in `ClusterWorker::work()` which picks a branch randomly.

PR #545, opened 2022-09-06
Session: refresh metadata on schema change 

Gor027:
Fixes: #544 

Added refreshing schema metadata after reaching schema agreement after a
schema-changing query.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -122,6 +122,7 @@ pub struct Session {
     default_consistency: Consistency,
     auto_await_schema_agreement_timeout: Option<Duration>,
     request_timeout: Option<Duration>,
+    refresh_metadata_on_auto_schema_agreement: bool,
 }
 
 /// This implementation deliberately omits some details from Cluster in order
@@ -206,6 +207,10 @@ pub struct SessionConfig {
     pub request_timeout: Option<Duration>,
 
     pub address_translator: Option<Arc<dyn AddressTranslator>>,
+
+    /// If true, full schema metadata is fetched after successfully reaching a schema agreement.
+    /// It is true by default but can be disabled if successive schema-altering statements should be performed.
+    pub refresh_metadata_on_auto_schema_agreement: bool,
 }
 
 /// Describes database server known on Session startup.
@@ -250,6 +255,7 @@ impl SessionConfig {
             auto_await_schema_agreement_timeout: Some(std::time::Duration::from_secs(60)),
             request_timeout: Some(Duration::from_secs(30)),
             address_translator: None,
+            refresh_metadata_on_auto_schema_agreement: true,
         }
     }
 
@@ -448,6 +454,8 @@ impl Session {
             default_consistency: config.default_consistency,
             auto_await_schema_agreement_timeout: config.auto_await_schema_agreement_timeout,
             request_timeout: config.request_timeout,
+            refresh_metadata_on_auto_schema_agreement: config
+                .refresh_metadata_on_auto_schema_agreement,
         };
 
         if let Some(keyspace_name) = config.used_keyspace {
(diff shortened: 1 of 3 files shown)


piodul (commented):
The change you made makes sense - when a client performs a schema change, Scylla/Cassandra will send schema change events to all clients so that other clients than the one issuing the query learns about the new changes relatively quickly. However, I don't think it fixes https://github.com/scylladb/scylla-rust-driver/issues/544 . The events are not synchronized with the the end of the schema-altering query in any way and there is no guarantee that, after the query finishes, the metadata returned by `session.get_data()` will reflect the changes.

piodul (changes requested):
Looks good, but needs a unit test.

scylla/src/transport/session.rs
```diff
@@ -606,6 +606,10 @@ impl Session {
             }
         }
 
+        if response.as_schema_change().is_some() {
+            self.refresh_metadata().await?;
+        }
```
piodul, line 611: Additionally, I'd like to see a flag added to configuration which allows to switch this feature off (it would be turned on by default) so that the users who don't like this feature can switch this off (e.g. they want to perform many schema changes in succession).

scylla/src/transport/session.rs
```diff
@@ -606,6 +613,10 @@ impl Session {
             }
         }
 
+        if self.refresh_metadata_on_schema_agreement && response.as_schema_change().is_some() {
```
piodul, line 616: The name of the flag is `refresh_metadata_on_schema_agreement`. However, I'm not sure its name aligns with what it really does:

- If automatic wait for schema agreement is disabled and `refresh_metadata_on_schema_agreement` is true, the driver will still refresh metadata, even though it didn't wait for schema agreement.
- If the user explicitly waits for schema agreement (`session.await_schema_agreement()`), metadata won't be refreshed after it finishes.

Perhaps we should move the wait into `await_schema_agreement` in order to align its behavior with the name. @cvybhu what do you think?

scylla/src/transport/session_test.rs
```diff
@@ -2206,3 +2206,38 @@ async fn test_prepare_batch() {
...
+    let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+    let session = SessionBuilder::new().known_node(uri).build().await.unwrap();
+
+    let ks = unique_keyspace_name();
+    session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'SimpleStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
+    session.use_keyspace(ks.clone(), false).await.unwrap();
+
+    session
+        .query(
+            "CREATE TYPE IF NOT EXISTS udt (field1 int, field2 uuid, field3 text)",
```
piodul, line 2221: Nit: it would perhaps be slightly better no to use `IF NOT EXISTS` here, it would express the intention of the test better. The UDT is not supposed to be in the initial metadata, it should appear only after the statement is executed and the metadata is refreshed.

scylla/src/transport/session.rs
Gor027: Maybe it would be a better idea to name it `refresh_metadata_on_auto_schema_agreement` and make it to refresh metadata only if the automatic wait for schema agreement is enabled. I think it would be better that the user explicitly calls `refresh_metadata` after manually waiting for schema agreement.

piodul: Yes, this is also a valid option.

piodul: The problem from the first bullet point is still there.

Gor027: Fixed it, tnx.

piodul (approved):

Issue #554, opened 2022-09-20
Remove BatchResult in favor of QueryResult


devrandom:
`BatchResult` is really the same as ~~`QueryResponse`~~ `QueryResult`, except that the `response` field is dropped.  This makes it impossible to get at the `[applied]` bools, and therefore the caller can't tell if the batch wasn't applied because it had any failing conditions.  `Session::batch` should just return `QueryResult` instead.

(changed description to refer to `QueryResult`, because a query error is not possible for a batch return value)

cvybhu:
`Session::batch` returns a result, which indicates whether the batch succeeded.

```rust
    pub async fn batch(
        &self,
        batch: &Batch,
        values: impl BatchValues,
    ) -> Result<BatchResult, QueryError> {
```
If the batch was applied successfully it will return `Ok(BatchResult)`, otherwise it's gonna be `Err(QueryError)`.

> [applied] bools

Could you clarify which bools you would like to access?

devrandom:
When a batch has failing conditions, it doesn't error, it returns a `Response::Result`, as it should.  The result's first column is `[applied]`, and the value in that column is set to false (i.e. "batch not applied") if the condition is false anywhere in the batch.

demonstrating this with a manual cqlsh example:
```
cqlsh:ks> BEGIN BATCH
UPDATE ks.data set version = 1 WHERE key = 'x1a' and client = 0x01 IF version = 0;
UPDATE ks.data set version = 1 WHERE key = 'x1b' and client = 0x01 IF version = 0;
APPLY BATCH;

 [applied] | client | key | version
-----------+--------+-----+---------
     False |   0x01 | x1a |       1
     False |   0x01 | x1b |       0
```

When running this with a breakpoint inside `Session::batch`, I do see the right values in the `QueryResponse`, but it's thrown away when constructing the `BatchResult`.

devrandom:
BTW, the schema for the above is `CREATE TABLE IF NOT EXISTS ks.data (client blob, key varchar, version bigint, value blob, primary key (client, key))` if that helps.

devrandom:
I updated the issue description above to clarify this is about batches with failing conditions, not about other errors.

cvybhu:
Ah this looks like it's because of #100.
This should be finally fixed, I will take a look at it, thank you for reporting.

[...]

PR #555, opened 2022-09-21
batch: Replace BatchResult with QueryResult

cvybhu:
`BatchResult` is a struct used to represent the result of a batch query.
The difference between BatchResult and QueryResult is that BatchResult doesn't contain any rows. Batches only support INSERT and UPDATE statements, so it made sense to create a type with stricter constraints that could be used to represent this rowless result.

It turns out that in certain cases batches do return rows.
For batches with LWT queries, like this one:
```
BEGIN BATCH
UPDATE ks.data set version = 1 WHERE key = 'x1a' and client = 0x01 IF version = 0;
UPDATE ks.data set version = 1 WHERE key = 'x1b' and client = 0x01 IF version = 0;
APPLY BATCH;
```

The database returns rows with information about the changes performed:
```
    [applied] | client | key | version
-----------+--------+-----+---------
        False |   0x01 | x1a |       1
        False |   0x01 | x1b |       0
```

It wasn't possible to access these rows because BatchResult had no rows, as described in #554.

The solution is to remove BatchResult and use QueryResult instead. Batch is now like any other query which might return rows.

There's one field of QueryResult that doesn't make sense for batches: `paging_state`, but we can ignore it.
QueryResult has a ton of convenience methods that can be used to access the received rows, we would have to duplicate them all for BatchResult.
I think we can have one useless field, this much duplication isn't worth it.

At first I thought that maybe we could just expose a list of `[applied]` bools, not rows, but this isn't enough.
There are other columns that tell us the values of columns involved in the LWT. This might be useful to some users, we can't just discard it.

This change brings us a step closer to implementing #100

Fixes: #554 

## Pre-review checklist


scylla/src/transport/session_test.rs
```diff
@@ -2309,3 +2309,117 @@ async fn test_rate_limit_exceeded_exception() {
...
+        .unwrap();
+
+    let mut batch: Batch = Batch::default();
+    batch.append_statement("UPDATE tab SET r2 = 1 WHERE p1 = 0 AND c1 = 0 IF r1 = 0");
+    batch.append_statement("INSERT INTO tab (p1, c1, r1, r2) VALUES (0, 123, 321, 312)");
+    batch.append_statement("UPDATE tab SET r1 = 1 WHERE p1 = 0 AND c1 = 0 IF r2 = 0");
+
+    let batch_res: QueryResult = session.batch(&batch, ((), (), ())).await.unwrap();
+
+    // Scylla returns 5 rows, but Cassandra returns only 1
```
piodul, line 2343: Rows? Did you mean "columns" instead?

cvybhu: right, fixed

piodul (approved):

Issue #558, opened 2022-09-28
Make it possible to configure DcAwareRoundRobinPolicy _not_ to return remote replicas


piodul:
Currently, the load balancing plans returned `DcAwareRoundRobinPolicy` will first return nodes from the local DC and then those from the remote DCs. However, sometimes it is useful to make sure that the nodes from other DCs are not returned at all to make sure that they do not serve any user queries.

Some other drivers already provide this possibility, for example:

- CPP: https://github.com/scylladb/cpp-driver/blob/739b87c15428d1d26ec0160444b7b693543772c2/include/cassandra.h#L1043-L1045
- Python: https://github.com/scylladb/python-driver/blob/e4c508405215496fabadd91e175f2f9f713c785a/cassandra/policies.py#L220-L225

PR #560, opened 2022-09-28
dc_aware_round_robin: Nodes in remote DCs optional

wprzytula:
The feature was already available in other drivers: we may want or may not want that the produces plan contain nodes from remote DCs at the end, as a fallback. Now configurable as constructor flag.
Fixes: #558 

## Pre-review checklist


wprzytula:
@piodul ready.

scylla/src/transport/load_balancing/dc_aware_round_robin.rs
```diff
@@ -80,8 +82,11 @@ impl LoadBalancingPolicy for DcAwareRoundRobinPolicy {
             "DC Aware"
         );
 
-        let plan = rotated_local_nodes.chain(rotated_remote_nodes);
-        Box::new(plan)
+        if self.include_remote_nodes {
+            Box::new(rotated_local_nodes.chain(rotated_remote_nodes))
```
piodul, line 86: Let's put the calculation of remote nodes inside this branch so that they do not need to be calculated when not used. The `trace!` command will have to be moved into this branch and the sibling branch, and then modified appropriately.

scylla/src/transport/load_balancing/dc_aware_round_robin.rs
```diff
@@ -106,12 +111,15 @@ impl ChildLoadBalancingPolicy for DcAwareRoundRobinPolicy {
         let remote_nodes_rotation = super::compute_rotation(index, remote_nodes.len());
         let rotated_remote_nodes = super::slice_rotated_left(&remote_nodes, remote_nodes_rotation);
 
-        let plan = rotated_local_nodes
-            .chain(rotated_remote_nodes)
-            .cloned()
-            .collect::<Vec<_>>()
-            .into_iter();
-        Box::new(plan)
+        let plan = if self.include_remote_nodes {
```
piodul, line 114: Ditto

wprzytula:
v2: fixes after @piodul review.

scylla/src/transport/load_balancing/dc_aware_round_robin.rs
```diff
@@ -11,13 +11,15 @@ use tracing::trace;
 pub struct DcAwareRoundRobinPolicy {
     index: AtomicUsize,
     local_dc: String,
+    include_remote_nodes: bool,
 }
 
 impl DcAwareRoundRobinPolicy {
-    pub fn new(local_dc: String) -> Self {
+    pub fn new(local_dc: String, include_remote_nodes: bool) -> Self {
```
piodul, line 18: I think that it would be better to have a setter method for the include_remote_nodes flag. The name of the method (e.g. set_include_remote_nodes) will give more meaning to the boolean flag when looking at the code that initializes the policy. Moreover, this will reduce the amount of code broken by this change.

docs/source/load-balancing/dc-robin.md
```diff
@@ -25,9 +29,9 @@ let local_dc_name: String = "us_east".to_string();
...
     .known_node("127.0.0.1:9042")
-    .load_balancing(Arc::new(DcAwareRoundRobinPolicy::new(local_dc_name)))
+    .load_balancing(Arc::new(DcAwareRoundRobinPolicy::new(local_dc_name, true)))
     .build()
     .await?;
 # Ok(())
 # }
-```
\ No newline at end of file
+```
```
piodul, line 37: Do you know why GitHub marks this line as a change?

piodul (commented):
Looks good, just one comment about the API

docs/source/load-balancing/dc-robin.md
wprzytula: It's a matter of whitespace: my editor automatically added a newline, which had been missing at the end of file.

piodul (approved):

PR #565, opened 2022-10-01
Fix derives so that they work with generics

Ten0:
Derives were written without using the dedicated [`item.generics.split_for_impl()`](https://docs.rs/syn/latest/syn/struct.Generics.html#method.split_for_impl) which resulted in them not working as soon as the struct contained generics. Notably, it was impossible to derive `ValueList` on a struct that contained an `&str`, forcing the user to reallocate.

## Pre-review checklist

### DIFF
--- a/examples/value_list.rs
+++ b/examples/value_list.rs
@@ -35,6 +35,23 @@ async fn main() {
         .await
         .unwrap();
 
+    // You can also use type generics:
+    #[derive(scylla::ValueList)]
+    struct MyTypeWithGenerics<S: scylla::frame::value::Value> {
+        k: i32,
+        my: Option<S>,
+    }
+
+    let to_insert_2 = MyTypeWithGenerics {
+        k: 18,
+        my: Some("Some string".to_owned()),
+    };
+
+    session
+        .query("INSERT INTO ks.my_type (k, my) VALUES (?, ?)", to_insert_2)
+        .await
+        .unwrap();
+
     let q = session
         .query("SELECT * FROM ks.my_type", &[])
         .await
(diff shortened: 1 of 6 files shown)


scylla-macros/src/from_row.rs
```diff
@@ -1,12 +1,15 @@
 use proc_macro::TokenStream;
 use quote::{quote, quote_spanned};
-use syn::spanned::Spanned;
+use syn::{spanned::Spanned, DeriveInput};
 
 /// #[derive(FromRow)] derives FromRow for struct
 /// Works only on simple structs without generics etc
```
Ten0, line 6: To-do fix comment

examples/value_list.rs
```diff
@@ -20,14 +20,14 @@ async fn main() {
         .unwrap();
 
     #[derive(scylla::ValueList)]
-    struct MyType {
+    struct MyType<'a> {
```
piodul, line 23: Does your PR make it possible to use type parameters as well (not only lifetime parameters)? If so, could you extend the example and add another structure which is generic over a type parameter?

piodul (commented):
Looks good (only one suggestion), but before merging I would like somebody more familiar with how procedural macros work in our project to take a look as well. @cvybhu @havaker?

examples/value_list.rs
Ten0: Yes it does, and given the way split_for_impl works if would be hard for someone to break one without breaking the other.

Ten0: 23695bbc104959f183e12a6ab3f74f75ad8baefd

piodul (approved):

PR #566, opened 2022-10-04
session_builder: fixed doc example

wprzytula:
In `address_translator()`, fixed bad formatting and added port numbers to supplied addresses.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session_builder.rs
+++ b/scylla/src/transport/session_builder.rs
@@ -595,7 +595,7 @@ impl SessionBuilder {
     /// # Ok(())
     /// # }
     /// ```
-    /// /// # Example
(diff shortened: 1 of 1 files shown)


scylla/src/transport/session_builder.rs
```diff
@@ -605,15 +605,15 @@ impl SessionBuilder {
     /// # use scylla::transport::session::{AddressTranslator, TranslationError};
     ///
```
piodul, line 606: It looks like there is an unnecessary space at the beginning of the box with the example code (above `let mut translation_rules = ...`):

image

piodul (approved):

Issue #561, opened 2022-09-29
Make it possible to provide a host filter


piodul:
Sometimes, the user may not be interested in connecting to every node in the cluster. For example, one may want to connect to the local DC only in order to avoid the large cross-DC latency when the driver establishes its first connections. Another example would be connecting to a single, specific node for testing purposes.

Gocql already offers such functionality: https://github.com/gocql/gocql/blob/0da64451f8dedea8e6c90030b9c4be3566317138/filters.go#L6-L10. A similar interface sounds like would be a good fit our driver, too.

PR #567, opened 2022-10-04
Host filter

piodul:
This PR adds host filters. With host filters, it is possible to prevent some nodes from establishing connections to the cluster at all. For example, the set of nodes for which connections are allowed can be restricted to the local DC.

Tested manually with cql-stress and a 3 node cluster - applied an allowlist of nodes and observed in metrics that forbidden nodes do not have additional connections and do not receive any requests.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/561

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/topology.rs
+++ b/scylla/src/transport/topology.rs
@@ -4,6 +4,7 @@ use crate::statement::query::Query;
 use crate::transport::connection::{Connection, ConnectionConfig};
 use crate::transport::connection_pool::{NodeConnectionPool, PoolConfig, PoolSize};
 use crate::transport::errors::{DbError, QueryError};
+use crate::transport::host_filter::HostFilter;
 use crate::transport::session::{AddressTranslator, IntoTypedRows};
 use crate::utils::parse::{ParseErrorCause, ParseResult, ParserState};
 
@@ -36,6 +37,7 @@ pub(crate) struct MetadataReader {
     fetch_schema: bool,
 
     address_translator: Option<Arc<dyn AddressTranslator>>,
+    host_filter: Option<Arc<dyn HostFilter>>,
 }
 
 /// Describes all metadata retrieved from the cluster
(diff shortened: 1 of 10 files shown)


scylla/src/transport/topology.rs
```diff
@@ -251,6 +255,9 @@ impl MetadataReader {
         let mut result = self.fetch_metadata(initial).await;
         if let Ok(metadata) = result {
             self.update_known_peers(&metadata);
+            if initial {
+                self.check_that_control_connection_is_an_accepted_host(&metadata);
```
cvybhu, line 259: Would it be possible to do the same filtering that we do in `update_knoiwn_peers` in the constructor?
We would be sure that the control connection is connected to an accepted peer.

Forcing the user to specify only known_peers that are accepted by the host filter seems overly restrictive.
I can imagine that someone might want to write a single app with predefined list of peers and then deploy it in multiple DCs, something like:
```rust
let session = SessionBuilder::new()
    .known_peers(["eu-1", "eu-2", "eu-3", "us-1", "us-2", "us-3"])
    .host_filter(make_host_filter_for_dc(get_dc_from_env_variable()));
    .build()
```

cvybhu (changes requested):

cvybhu:
Let's also make an issue that documentation for `HostFilter` is missing in the book.

scylla/src/transport/topology.rs
piodul: It would be nice if it were possible, but unfortunately not all kinds of filters can be implemented this way. For example, we don't really know which nodes are in which DC until we fetch topology data from system.local/system.peers.

piodul:
v2:
- Re-exported the `host_filter` module in library root,
- `AllowListHostFilter::new()` now accepts types which are convertible to an iterator over items that implement `ToSocketAddrs`. For example, addresses provided in textual form via `Vec<&str>` will work now,
- In addition to warning the user about the control connection being established to a node outside the filtered set, the driver will now re-establish the control connection to another node from the filtered set.

piodul:
> Let's also make an issue that documentation for `HostFilter` is missing in the book.

https://github.com/scylladb/scylla-rust-driver/issues/570

Issue #568, opened 2022-10-04
Passing batch values by reference doesn't work


cvybhu:
On current main (acf1dd949dd9a7ada240f08e4c29abf0b5a8d38a), trying to pass references to batch values doesn't work.

Passing them by value like this works:
```rust
let values = ((1_i32, 2_i32, "abc"), (), (1_i32, 4_i32, "hello"));
session.batch(&batch, values).await.unwrap();
```

But trying to use a reference fails:
```rust
let val1: i32 = 1;
let values = ((&val1, 2_i32, "abc"), (), (1_i32, 4_i32, "hello"));
session.batch(&batch, values).await.unwrap();
```
```
let values = ((&val1, 2_i32, "abc"), (), (1_i32, 4_i32, "hello"));
    |                    ^^^^^ borrowed value does not live long enough
393 |
394 |     session.batch(&batch, values).await.unwrap();
    |     ----------------------------- argument requires that `val1` is borrowed for `'static`
...
457 | }
    | - `val1` dropped here while still borrowed
```
This is unacceptable. For `i32` there is no difference whether we copy the value or use a reference, but anything else like collections, tuples, UDTs, shouldn't be cloned frivolously.

The regression is caused by #508, which introduced a lot of fun lifetimes to `BatchValues`.

In particular I think it's caused by the GAT workaround.
```rust
pub trait BatchValues: for<'r> BatchValuesGatWorkaround<'r> {}
impl<T: for<'r> BatchValuesGatWorkaround<'r> + ?Sized> BatchValues for T {}
```
We require that every thihg implementing `BatchValues` also implements `BatchValuesGatWorkaround` for ANY lifetime `'r`.
When T contains a short lived reference, it cannot match the requirements for longer lifetimes, the reference would have to be ``static`.

We might revert #508 and rethink the approach.

Caused by: #508
Refs: #448

cc @Ten0

Ten0:
Aha I see it looks like something using the implicit bounds trick described at https://sabrinajewson.org/blog/the-better-alternative-to-lifetime-gats might solve this issue. I'll try to have a look at this :)

PR #569, opened 2022-10-04
Revert "Pick connections based on batch first statement's shard"

cvybhu:
The new code, and in particualr the fancy GAT workaroound, cause problems when trying to pass batch values by reference, as described in #568.

Fixes: #568
Unfixes: #448

This reverts commit 24ee9545774291607f591e2a4ad2ba1bbe7ab96e.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -388,7 +388,8 @@ async fn test_batch() {
     batch.append_statement(&format!("INSERT INTO {}.t_batch (a, b, c) VALUES (7, 11, '')", ks)[..]);
     batch.append_statement(prepared_statement.clone());
 
-    let values = ((1_i32, 2_i32, "abc"), (), (1_i32, 4_i32, "hello"));
+    let four_value: i32 = 4;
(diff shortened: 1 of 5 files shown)


scylla/src/transport/session_test.rs
```diff
@@ -388,7 +388,8 @@ async fn test_batch() {
     batch.append_statement(&format!("INSERT INTO {}.t_batch (a, b, c) VALUES (7, 11, '')", ks)[..]);
     batch.append_statement(prepared_statement.clone());
 
-    let values = ((1_i32, 2_i32, "abc"), (), (1_i32, 4_i32, "hello"));
+    let four_value: i32 = 4;
+    let values = ((1_i32, 2_i32, "abc"), (), (1_i32, &four_value, "hello"));
```
piodul, line 392: Just a tiny nitpick/clarification: `"abc"` and `"hello"` already are references (`&str`), but their lifetime is `'static`. The problem with the code that is being reverted is that it doesn't work for references with non-static lifetimes.

No need to do anything here, I think it is obvious that the commit description talks about non-static references.

piodul (approved):

PR #572, opened 2022-10-05
Add support for custom authentication

### LINKED ISSUE #562: Add support for custom authentication
#### Motivation:

Custom authentication is useful for integrating with more complex authentication systems like Kerberos, instead of authenticating with a username and password. This will also allow us to easily add custom authentication to the [Cpp-Rust](https://github.com/scylladb/cpp-rust-driver) driver.

#### Suggestion:

My suggestion is inspired by the implementation of custom authentication in the C++ driver and the current authentication implementation in the Rust driver:

1. Introduce an `Authenticator` trait

```Rust
use bytes::BufMut;

pub trait Authenticator: Send + Sync {
    async fn initial_response(&self, response: &mut impl BufMut) -> bool;

    async fn evaluate_challenge(&self, token: Option<Vec<u8>>, response: &mut impl BufMut) -> bool;

    async fn success(&self, token: Option<Vec<u8>>) -> bool;

    fn error() -> &str;
}
```

2. Add authenticator to `ConnectionConfig`:

```Rust
pub struct ConnectionConfig {
    /* ... */
    pub authenticator: Arc<dyn Authenticator>;
}
```
The `authenticator` should be either initialized based on the `SessionConfig`, like `ssl_context` is being set currently, or there should be a default authenticator, in my opinion, the current implementation should be the default authentication:

```Rust
struct PlainTextAuthenticator {
    username: String,
    password: String,
}

impl Authenticator for PlainTextAuthenticator {
  async fn initial_response(&self, response: &mut impl BufMut) -> bool {
      let username_as_bytes = self.username.as_ref().unwrap().as_bytes();
      let password_as_bytes = self.password.as_ref().unwrap().as_bytes();

      // The body of AuthResponse is a single [bytes] value (i32 length and then contents)
      let buf_size = 2 + username_as_bytes.len() + password_as_bytes.len();
      response.put_i32(buf_size.try_into()?);

      response.put_u8(0);
      response.put_slice(username_as_bytes);
      response.put_u8(0);
      response.put_slice(password_as_bytes);
      return true;
  }

  /* The other functions should just return true */
}
```
 
The `response: &mut impl BufMut` later can be sent by calling `Connection::send_request`.

3. The `evaluate_challenge`  and `success` should be called respectively for the `AuthChallenge` and `AuthSuccess` authentication responses. In case the functions return `false`, the `error()` function can be used to retrieve the error string.

Gor027:
Added AuthenticatorProvider and AuthenticatorSession traits. AuthenticatorProvider serves as a factory to create a new AuthenticatorSession instance for each session.

AuthenticatorSession trait has 3 methods to be implemented by a user:
 * initial_response: to initiate a request to begin an authentication exchange. Resources required for this specific exchange can be stored in the struct implementing AuthenticatorSession trait.
 * evaluate_challenge: to handle an authentication challenge initiated by the server. The information contained in the token parameter is authentication protocol specific. It may be NULL or empty.
 * success: to handle the success phase of exchange. The token parameters contain information that may be used to finalize the request.

Fixes #562 

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/request/auth_response.rs
+++ b/scylla-cql/src/frame/request/auth_response.rs
@@ -1,37 +1,18 @@
 use crate::frame::frame_errors::ParseError;
 use bytes::BufMut;
-use std::convert::TryInto;
 
 use crate::frame::request::{Request, RequestOpcode};
-use crate::frame::Authenticator;
+use crate::frame::types::write_bytes_opt;
 
 // Implements Authenticate Response
 pub struct AuthResponse {
-    pub username: Option<String>,
-    pub password: Option<String>,
-    pub authenticator: Authenticator,
+    pub response: Option<Vec<u8>>,
 }
 
 impl Request for AuthResponse {
     const OPCODE: RequestOpcode = RequestOpcode::AuthResponse;
 
     fn serialize(&self, buf: &mut impl BufMut) -> Result<(), ParseError> {
-        if self.username.is_none() || self.password.is_none() {
-            return Err(ParseError::BadDataToSerialize(
-                "Bad credentials: username or password missing. You can use SessionBuilder::user(\"user\", \"pass\") to provide credentials.".to_string(),
-            ));
-        }
-        let username_as_bytes = self.username.as_ref().unwrap().as_bytes();
-        let password_as_bytes = self.password.as_ref().unwrap().as_bytes();
-
-        // The body of AuthResponse is a single [bytes] value (i32 length and then contents)
-        let buf_size = 2 + username_as_bytes.len() + password_as_bytes.len();
-        buf.put_i32(buf_size.try_into()?);
-
-        buf.put_u8(0);
-        buf.put_slice(username_as_bytes);
-        buf.put_u8(0);
-        buf.put_slice(password_as_bytes);
-        Ok(())
+        write_bytes_opt(self.response.as_ref(), buf)
     }
 }
--- a/scylla/src/authentication/mod.rs
+++ b/scylla/src/authentication/mod.rs
@@ -1,0 +1,89 @@
+use async_trait::async_trait;
+use bytes::{BufMut, BytesMut};
+
+/// Type to represent an authentication error message.
+pub type AuthError = String;
+
+/// Trait used to represent a user-defined custom authentication.
+#[async_trait]
+pub trait AuthenticatorSession: Send + Sync {
+    /// To handle an authentication challenge initiated by the server.
+    /// The information contained in the token parameter is authentication protocol specific.
+    /// It may be NULL or empty.
+    async fn evaluate_challenge(
+        &mut self,
+        token: Option<&[u8]>,
+    ) -> Result<Option<Vec<u8>>, AuthError>;
+
+    /// To handle the success phase of exchange.
+    /// The token parameters contain information that may be used to finalize the request.
+    async fn success(&mut self, token: Option<&[u8]>) -> Result<(), AuthError>;
+}
+
+/// Trait used to represent a factory of [`AuthenticatorSession`] instances.
+/// A new [`AuthenticatorSession`] instance will be created for each session.
+///
+/// The custom authenticator can be set using SessionBuilder::authenticator_provider method.  
+///
+/// Default: [`PlainTextAuthenticator`] is the default authenticator which requires username and
+/// password. It can be set by using SessionBuilder::user(\"user\", \"pass\") method.
+#[async_trait]
+pub trait AuthenticatorProvider: Sync + Send {
+    /// A pair of initial response and boxed [`AuthenticatorSession`]
+    /// should be returned if authentication is required by the server.
+    async fn start_authentication_session(
+        &self,
+        authenticator_name: &str,
+    ) -> Result<(Option<Vec<u8>>, Box<dyn AuthenticatorSession>), AuthError>;
+}
+
+struct PlainTextAuthenticatorSession;
+
+#[async_trait]
+impl AuthenticatorSession for PlainTextAuthenticatorSession {
+    async fn evaluate_challenge(
+        &mut self,
+        _token: Option<&[u8]>,
+    ) -> Result<Option<Vec<u8>>, AuthError> {
+        Err("Challenges are not expected during PlainTextAuthentication".to_string())
+    }
+
+    async fn success(&mut self, _token: Option<&[u8]>) -> Result<(), AuthError> {
+        Ok(())
+    }
+}
+
+/// Default authenticator provider that requires username and password if authentication is required.
+pub struct PlainTextAuthenticator {
+    username: String,
+    password: String,
+}
+
+impl PlainTextAuthenticator {
+    /// Creates new [`PlainTextAuthenticator`] instance with provided username and password.
+    pub fn new(username: String, password: String) -> Self {
+        PlainTextAuthenticator { username, password }
+    }
+}
+
+#[async_trait]
+impl AuthenticatorProvider for PlainTextAuthenticator {
+    async fn start_authentication_session(
+        &self,
+        _authenticator_name: &str,
+    ) -> Result<(Option<Vec<u8>>, Box<dyn AuthenticatorSession>), AuthError> {
+        let mut response = BytesMut::new();
+        let username_as_bytes = self.username.as_bytes();
+        let password_as_bytes = self.password.as_bytes();
+
+        response.put_u8(0);
+        response.put_slice(username_as_bytes);
+        response.put_u8(0);
+        response.put_slice(password_as_bytes);
+
+        Ok((
+            Some(response.to_vec()),
+            Box::new(PlainTextAuthenticatorSession),
+        ))
+    }
+}
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -96,6 +96,7 @@
 pub use scylla_cql::frame;
 pub use scylla_cql::macros::{self, *};
 
+pub mod authentication;
 pub mod history;
 pub mod routing;
 pub mod statement;
--- a/scylla/src/transport/authenticate_test.rs
+++ b/scylla/src/transport/authenticate_test.rs
@@ -1,4 +1,8 @@
+use crate::authentication::{AuthError, AuthenticatorProvider, AuthenticatorSession};
 use crate::utils::test_utils::unique_keyspace_name;
+use async_trait::async_trait;
+use bytes::{BufMut, BytesMut};
+use std::sync::Arc;
 
 #[tokio::test]
 #[ignore]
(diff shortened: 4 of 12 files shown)


piodul (changes requested):

scylla-cql/src/frame/response/authenticate.rs
```diff
@@ -42,3 +44,96 @@ impl AuthChallenge {
         })
     }
 }
+
```
piodul, line 47: The `frame::response::authenticate` module is supposed to be related to the `AUTHENTICATE` frame in the CQL protocol. It's not a good place for the new authentication-related types that you introduced.

Let's add an `authentication` module in the root of the `scylla` crate and put those types there.

scylla-cql/src/frame/response/authenticate.rs
```diff
@@ -42,3 +44,96 @@ impl AuthChallenge {
...
+pub type AuthError = String;
+
+/// Trait used to represent a user-defined custom authentication.
+/// evaluate_challenge: to handle an authentication challenge initiated by the server.
+/// The information contained in the token parameter is authentication protocol specific.
+/// It may be NULL or empty.
+///
+/// success: to handle the success phase of exchange.
+/// The token parameters contain information that may be used to finalize the request.
```
piodul, line 56: Instead of putting descriptions of methods to the main docstring, you should place them beside `evaluate_challenge` and `success`. Please look at e.g. `prepared_statement.rs` for examples.

scylla/src/transport/connection.rs
```diff
@@ -1246,31 +1243,48 @@ pub async fn open_named_connection(
...
+                                    "Unexpected response to Authenticate Response message",
+                                ))
+                            }
+                        }
+                    }
+                },
+                None => return Err(QueryError::InvalidMessage(
+                    "Authentication is required. You can use SessionBuilder::user(\"user\", \"pass\") to provide credentials \
+                    or SessionBuilder::authenticator_provider to provide custom authenticator".to_string(),
+                )),
```
piodul, line 1287: The `open_named_connection` function is already very long and is becoming even longer, how about moving the authentication logic to a separate function?

scylla-cql/src/frame/response/authenticate.rs
piodul: Please make sure to apply this advice to other public types that you introduced.

scylla/src/transport/connection.rs
```diff
@@ -1292,6 +1253,69 @@ pub async fn open_named_connection(
...
+        "AllowAllAuthenticator" => AllowAllAuthenticator,
+        "PasswordAuthenticator" => PasswordAuthenticator,
+        "org.apache.cassandra.auth.PasswordAuthenticator" => CassandraPasswordAuthenticator,
+        "org.apache.cassandra.auth.AllowAllAuthenticator" => CassandraAllowAllAuthenticator,
+        "com.scylladb.auth.TransitionalAuthenticator" => ScyllaTransitionalAuthenticator,
+        _ => unimplemented!(
+            "Authenticator not supported, {}",
+            authenticate.authenticator_name
+        ),
+    };
```
piodul, line 1270: Let's get rid of this code and switch to using just `String`/`&str` instead of an enum to represent the authenticator name. The `Authenticator` itself should verify in `start_authentication_session` that it recognizes and supports the authentication kind.

scylla/src/lib.rs
```diff
@@ -96,6 +96,7 @@
 pub use scylla_cql::frame;
 pub use scylla_cql::macros::{self, *};
 
+pub mod authenticate;
```
piodul, line 99: IMO `authentication` would be a better name for the module. Module names are usually nouns, not verbs. The `frame::response::authenticate` module is called like that most likely because it contains a definition of the `AUTHENTICATE` frame (although it contains definitions of other frames, too, so maybe at some point we should change it).

scylla/src/authenticate/mod.rs
```diff
@@ -0,0 +1,97 @@
+use crate::frame::Authenticator;
+use async_trait::async_trait;
+use bytes::{BufMut, BytesMut};
+
+/// Type to represent an authentication error message.
+pub type AuthError = String;
+
+/// Trait used to represent a user-defined custom authenticate.
```
piodul, line 8: "authentication"

scylla/src/authenticate/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+#[async_trait]
+pub trait AuthenticatorSession: Send + Sync {
+    /// To handle an authenticate challenge initiated by the server.
+    /// The information contained in the token parameter is authenticate protocol specific.
```
piodul, line 12: "authentication"

scylla/src/authenticate/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+/// A new [`AuthenticatorSession`] instance will be created for each session.
+///
+/// The custom authenticator can be set using SessionBuilder::authenticator_provider method.  
+///
+/// Default: [`PlainTextAuthenticator`] is the default authenticator which requires username and
+/// password. It can be set by using SessionBuilder::user(\"user\", \"pass\") method.
+#[async_trait]
+pub trait AuthenticatorProvider: Sync + Send {
+    /// A pair of initial response and boxed [`AuthenticatorSession`]
+    /// should be returned if authenticate is required by the server.
```
piodul, line 34: "authentication"

test/tls/scylla.yaml
```diff
@@ -226,7 +226,7 @@ batch_size_fail_threshold_in_kb: 50
 # Out of the box, Scylla provides org.apache.cassandra.auth.{AllowAllAuthenticator,
 # PasswordAuthenticator}.
 #
-# - AllowAllAuthenticator performs no checks - set it to disable authentication.
+# - AllowAllAuthenticator performs no checks - set it to disable authenticate.
```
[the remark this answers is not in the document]
Gor027: Yep. Unfortunately, I did not notice the auto-refactored lines. Will fix it.

scylla-cql/src/frame/request/auth_response.rs
```diff
@@ -1,37 +1,22 @@
...
 use crate::frame::request::{Request, RequestOpcode};
-use crate::frame::Authenticator;
 
 // Implements Authenticate Response
 pub struct AuthResponse {
-    pub username: Option<String>,
-    pub password: Option<String>,
-    pub authenticator: Authenticator,
+    pub response: Option<Vec<u8>>,
+    pub authenticator: String,
```
piodul, line 9: The `AUTH_RESPONSE` frame contains only one field. The `authenticator` does not belong here.

scylla-cql/src/frame/request/auth_response.rs
```diff
@@ -1,37 +1,22 @@
...
 impl Request for AuthResponse {
     const OPCODE: RequestOpcode = RequestOpcode::AuthResponse;
 
     fn serialize(&self, buf: &mut impl BufMut) -> Result<(), ParseError> {
-        if self.username.is_none() || self.password.is_none() {
-            return Err(ParseError::BadDataToSerialize(
-                "Bad credentials: username or password missing. You can use SessionBuilder::user(\"user\", \"pass\") to provide credentials.".to_string(),
-            ));
+        if let Some(bytes) = self.response.as_ref() {
+            buf.put_slice(bytes.as_slice());
```
piodul, line 17: The `AUTH_RESPONSE` field has a single field which is serialized as `[bytes]`. The `[bytes]` is a concept from the CQL protocol description an is semantically equivalent to `Option<Vec<u8>>`.

What you did above is not how you serialize `[bytes]`. Please refer to the CQL protocol spec on how it should be serialized.

https://github.com/scylladb/scylla-tools-java/blob/5f2b91d774843c343e9fb07c84fb6298bd07a4e9/doc/native_protocol_v4.spec

scylla/src/authentication/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+        let mut response = BytesMut::new();
+        let username_as_bytes = self.username.as_bytes();
+        let password_as_bytes = self.password.as_bytes();
+        let buf_size = 2 + username_as_bytes.len() + password_as_bytes.len();
+
+        // The body of response is a single [bytes] value (i32 length and then contents)
+        match buf_size.try_into() {
+            Ok(size) => response.put_i32(size),
+            Err(err) => return Err(err.to_string()),
+        }
```
piodul, line 84: Hint: the length of the message is a part of the `[bytes]`. The AuthenticatorProvider should not be concerned with it.

scylla/src/transport/authenticate_test.rs
```diff
@@ -21,3 +25,58 @@ async fn authenticate_superuser() {
...
+struct CustomAuthenticatorProvider;
+
+#[async_trait]
+impl AuthenticatorProvider for CustomAuthenticatorProvider {
+    async fn start_authentication_session(
+        &self,
+        _name: &str,
+    ) -> Result<(Option<Vec<u8>>, Box<dyn AuthenticatorSession>), AuthError> {
+        let mut response = BytesMut::new();
+        let cred = "20\0cassandra\0cassandra";
```
piodul, line 54: Why does this test pass at all? You have prepended the string with its length, but expressed as text ("20"), not as an i32 length.

test/tls/scylla.yaml
```diff
@@ -227,7 +227,7 @@ batch_size_fail_threshold_in_kb: 50
 # PasswordAuthenticator}.
 #
 # - AllowAllAuthenticator performs no checks - set it to disable authentication.
-# - PasswordAuthenticator relies on username/password pairs to authenticate
+# - PasswordAuthenticator relies on username/password pairs to authentication
```
piodul, line 230: This one was actually legit - "relies on username/password pairs to _authenticate_ users", here it makes sense to use the verb.

Please do not perform blind search and replace like this in comments. If you do, you need to manually verify that the changes make sense.

If you need to change the name of a variable, either use refactoring tools offered by your IDE or change the name at the definition and use the compiler to guide you to places where the names should actually be changed.

scylla/src/authentication/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+///
+/// Default: [`PlainTextAuthenticator`] is the default authenticator which requires username and
+/// password. It can be set by using SessionBuilder::user(\"user\", \"pass\") method.
+#[async_trait]
+pub trait AuthenticatorProvider: Sync + Send {
+    /// A pair of initial response and boxed [`AuthenticatorSession`]
+    /// should be returned if authentication is required by the server.
+    async fn start_authentication_session(
+        &self,
+        name: &str,
```
piodul, line 36: The name `name` is not too descriptive, it's not obvious what the `name` refers to. How about `authenticator_name`?

scylla/src/transport/connection.rs
```diff
@@ -1292,6 +1248,59 @@ pub async fn open_named_connection(
     Ok((connection, error_receiver))
 }
 
+async fn start_authenticate(
```
piodul, line 1251: The name `start_authenticate` implies that the function starts the authentication process but does not finish it. How about `perform_authentication` instead?

scylla-cql/src/frame/request/auth_response.rs
```diff
@@ -1,37 +1,25 @@
...
 
     fn serialize(&self, buf: &mut impl BufMut) -> Result<(), ParseError> {
-        if self.username.is_none() || self.password.is_none() {
-            return Err(ParseError::BadDataToSerialize(
-                "Bad credentials: username or password missing. You can use SessionBuilder::user(\"user\", \"pass\") to provide credentials.".to_string(),
-            ));
-        }
-        let username_as_bytes = self.username.as_ref().unwrap().as_bytes();
-        let password_as_bytes = self.password.as_ref().unwrap().as_bytes();
+        if let Some(bytes) = self.response.clone() {
```
piodul, line 15: Please be mindful of clones and avoid them if you can. Here, you probably can do `self.response.as_ref()` instead of `self.response.clone()`.

scylla-cql/src/frame/request/auth_response.rs
```diff
@@ -1,37 +1,25 @@
...
+            buf.put_i32(bytes.len().try_into()?);
```
piodul, line 16: `self.response` is, in the terminology of the CQL protocol spec, a `[bytes]` object. `[bytes]` is serialized as follows:

- if null, then serialize `-1` as a little-endian `i32` and that's it,
- if not null, then serialize its length as a little-endian `i32` and then just serialize the contents.

We have a module `scylla-cql/src/frame/types.rs` which provides functions for serialization and deserialization of CQL protocol spec objects. You need the `types::write_bytes_opt` function - which does not exist, but I'm sure you can add it yourself after looking at how other functions are implemented in this file.

scylla/src/authentication/mod.rs
piodul: This code shouldn't be here. The `start_authentication_session` function should just return the [bytes] value (called `token` in the protocol specification). For password-based authentication, the token should not contain its length. The layer higher above the code is responsible for serializing the AuthResponse message, which requires serializing the token length and then its contents.

Even though the message is incorrect, Scylla accepts it. This is Scylla's fault as it is too permissive. It parses the token from authentication response from the back, first looking for password and then for the username and ignores the rest - so, if there is garbage at the beginning of the token then it will be ignored. Try running the authentication tests with Cassandra and you will see that it rejects it.

piodul: Link to Scylla's token parsing code: https://github.com/scylladb/scylladb/blob/ff5527deb139275b9742e14bc62984caadd754f1/auth/sasl_challenge.cc#L30-L61

scylla/src/transport/authenticate_test.rs
```diff
@@ -21,3 +25,60 @@ async fn authenticate_superuser() {
...
+
+#[async_trait]
+impl AuthenticatorProvider for CustomAuthenticatorProvider {
+    async fn start_authentication_session(
+        &self,
+        _authenticator_name: &str,
+    ) -> Result<(Option<Vec<u8>>, Box<dyn AuthenticatorSession>), AuthError> {
+        let mut response = BytesMut::new();
+        let cred = "\0cassandra\0cassandra";
+        let cred_length = 20;
```
piodul, line 55: There is absolutely no reason for hardcoding the length of the string literal from the previous line. You should just do `let cred_length = cred.len();` and that's it.

Perhaps you thought that Rust's `&str` is like `const char*` from C? Unlike C, Rust's strings are not null-terminated. The `&str` type is a pair `(pointer to the beginning of the string, length of the string)`, so the length of the string is immediately known and calling `.len()` is super cheap here and works even if there are null bytes inside it.

scylla/src/transport/authenticate_test.rs
```diff
@@ -21,3 +25,60 @@ async fn authenticate_superuser() {
...
+
+        response.put_i32(cred_length);
```
piodul, line 57: I assume that the custom authenticator here works the same as the plain text authenticator. Like in the case of the `PlainTextAuthenticator`, you shouldn't serialize the length here - it's not a part of the response token, and `AuthResponse::serialize` takes care of serializing the length of the token.

piodul (changes requested):
I think the PR looks good overall, only some small issues left before it can be merged. Please also remember to fix the issue I mentioned in this comment: https://github.com/scylladb/scylla-rust-driver/pull/572#discussion_r1022553836

piodul (approved):

PR #573, opened 2022-10-05
Pick connections based on batch first statement's shard - v2

### LINKED ISSUE #448: Token aware load balancing for batches

### LINKED ISSUE #568: Passing batch values by reference doesn't work

Ten0:
Fixes #448 

Follow-up on #508 being reverted in #569 because of #568 - first commit reintroduces #508, while second commit fixes #568.

-> This re-fixes #448, without reintroducing #568.

#568 is effectively fixed by using default generic argument on the `BatchValuesGatWorkaround` trait which will generate an implicit  `where Self: 'r` bound in `for<'r> BatchValuesGatWorkaround<'r>` (becoming in effect `for<'r where Self: 'r> BatchValuesGatWorkaround<'r>`). This trick is explained in more detail at https://sabrinajewson.org/blog/the-better-alternative-to-lifetime-gats.

We also get rid of lending iteration in `BatchValuesIterator`, giving more flexibility for future shard-pick heuristic improvements (this is also unlocked by that same implicit bound).

--

Update: c7f7640 now also moves to the recently-stabilized GATs.  This removes the need for the previously described GAT workaround. The implicit bound is replaced by an explicit `where Self: 'r` bound.
Since we don't need `for<'r> X::AssociatedType<'r>: Something` bounds anywhere, we don't hit the limitations described [here](https://sabrinajewson.org/blog/the-better-alternative-to-lifetime-gats), and the simplification is probably worth losing those.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/request/batch.rs
+++ b/scylla-cql/src/frame/request/batch.rs
@@ -1,4 +1,4 @@
-use crate::frame::frame_errors::ParseError;
+use crate::frame::{frame_errors::ParseError, value::BatchValuesIterator};
 use bytes::{BufMut, Bytes};
 use std::convert::TryInto;
 
@@ -54,9 +54,21 @@ where
         // Serializing queries
         types::write_short(self.statements_count.try_into()?, buf);
 
-        for (statement_num, statement) in self.statements.clone().enumerate() {
+        let mut n_serialized_statements = 0usize;
+        let mut value_lists = self.values.batch_values_iter();
+        for statement in self.statements.clone() {
             statement.serialize(buf)?;
-            self.values.write_nth_to_request(statement_num, buf)?;
+            value_lists.write_next_to_request(buf).ok_or_else(|| {
+                ParseError::BadDataToSerialize(
+                    "Mismatch between statement counts for batch query".to_owned(),
+                )
+            })??;
+            n_serialized_statements += 1;
+        }
+        if n_serialized_statements != self.statements_count {
+            return Err(ParseError::BadDataToSerialize(
+                "Mismatch between statement counts for batch query".to_owned(),
+            ));
         }
 
         // Serializing consistency
--- a/scylla-cql/src/frame/request/batch.rs
+++ b/scylla-cql/src/frame/request/batch.rs
@@ -54,22 +54,35 @@ where
         // Serializing queries
         types::write_short(self.statements_count.try_into()?, buf);
 
-        let counts_mismatch_err = || {
-            ParseError::BadDataToSerialize(
-                "Length of provided values must be equal to number of batch statements".to_owned(),
-            )
+        let counts_mismatch_err = |n_values: usize, n_statements: usize| {
+            ParseError::BadDataToSerialize(format!(
+                "Length of provided values must be equal to number of batch statements \
+                    (got {n_values} values, {n_statements} statements)"
+            ))
         };
         let mut n_serialized_statements = 0usize;
         let mut value_lists = self.values.batch_values_iter();
-        for statement in self.statements.clone() {
+        for (idx, statement) in self.statements.clone().enumerate() {
             statement.serialize(buf)?;
             value_lists
                 .write_next_to_request(buf)
-                .ok_or_else(counts_mismatch_err)??;
+                .ok_or_else(|| counts_mismatch_err(idx, self.statements_count))??;
             n_serialized_statements += 1;
         }
-        if n_serialized_statements != self.statements_count || value_lists.skip_next().is_some() {
-            return Err(counts_mismatch_err());
+        if value_lists.skip_next().is_some() {
+            return Err(counts_mismatch_err(
+                std::iter::from_fn(|| value_lists.skip_next()).count() + 1,
+                n_serialized_statements,
+            ));
+        }
+        if n_serialized_statements != self.statements_count {
+            // We want to check this to avoid propagating an invalid construction of self.statements_count as a
+            // hard-to-debug silent fail
+            return Err(ParseError::BadDataToSerialize(format!(
+                "Invalid Batch constructed: not as many statements serialized as announced \
+                    (batch.statement_count: {announced_statement_count}, {n_serialized_statements}",
+                announced_statement_count = self.statements_count
+            )));
         }
 
         // Serializing consistency
--- a/scylla-cql/src/frame/request/batch.rs
+++ b/scylla-cql/src/frame/request/batch.rs
@@ -54,21 +54,22 @@ where
         // Serializing queries
         types::write_short(self.statements_count.try_into()?, buf);
 
+        let counts_mismatch_err = || {
+            ParseError::BadDataToSerialize(
+                "Length of provided values must be equal to number of batch statements".to_owned(),
+            )
+        };
         let mut n_serialized_statements = 0usize;
         let mut value_lists = self.values.batch_values_iter();
         for statement in self.statements.clone() {
             statement.serialize(buf)?;
-            value_lists.write_next_to_request(buf).ok_or_else(|| {
-                ParseError::BadDataToSerialize(
-                    "Mismatch between statement counts for batch query".to_owned(),
-                )
-            })??;
+            value_lists
+                .write_next_to_request(buf)
+                .ok_or_else(counts_mismatch_err)??;
             n_serialized_statements += 1;
         }
-        if n_serialized_statements != self.statements_count {
-            return Err(ParseError::BadDataToSerialize(
-                "Mismatch between statement counts for batch query".to_owned(),
-            ));
+        if n_serialized_statements != self.statements_count || value_lists.skip_next().is_some() {
+            return Err(counts_mismatch_err());
         }
 
         // Serializing consistency
--- a/scylla-cql/src/frame/request/batch.rs
+++ b/scylla-cql/src/frame/request/batch.rs
@@ -66,7 +66,7 @@ where
             statement.serialize(buf)?;
             value_lists
                 .write_next_to_request(buf)
-                .ok_or_else(|| counts_mismatch_err(idx, self.statements_count))??;
+                .ok_or_else(|| counts_mismatch_err(idx, self.statements.clone().count()))??;
             n_serialized_statements += 1;
         }
         if value_lists.skip_next().is_some() {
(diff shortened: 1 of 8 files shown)


cvybhu:
Thank you for the updated PR and sorry that this is taking so long.

I'm hesitating to merge this because the code is so complex and hard to reason about.
I'm not 100% sure that we won't hit another unexpected lifetime issue.

Perhaps it would be possible to implement this in a simpler way.
I'll sit down and try to explore other possibilities.

Another factor is that support for GATs has just become stable, so maybe we could avoid having strange GAT workarounds, although that would require raising the minimum supported rustc version.

cvybhu:
@piodul do you think we could raise the MSVR?

Ten0:
I'm not sure whether that would solve it as we may still hit (now or later) the limitations of GATs related to HKTs described in [the article](https://sabrinajewson.org/blog/the-better-alternative-to-lifetime-gats), but otherwise I agree it would improve readability. I would also be in favor of eventually rewriting this using GATs if it works. 

> I'm not 100% sure that we won't hit another unexpected lifetime issue.

I have to admit I might have rushed the implementation a bit last time, relying too much on that the relevant tests would already exist. :/
That being said, with the new tests and with `BatchValuesIterator` now being generic on a lifetime to explicitly specify the (possibly longer) lifetimes it may output, I'm reasonably confident this shouldn't happen again.

> I'm hesitating to merge this because the code is so complex and hard to reason about.

Note that the difference between this and the previously merged one is just [this small commit](https://github.com/scylladb/scylla-rust-driver/pull/573/commits/9154660dbf90aee91d60c5c19416bf6065fd7307) which doesn't feel like it changes much to [the general idea](https://github.com/scylladb/scylla-rust-driver/pull/508#issuecomment-1218490696).

piodul (commented):
I share some of the worries about the complexity with @cvybhu, mainly because this shows up in the public interface. However, after educating myself a bit about GATs, the trick you used and reading the PR thoroughly I understand (I think) why the "GAT workaround trick" is used and I would be fine with merging it. I suppose users don't usually implement `BatchValues` themselves but rather use provided ones, so it should be fine for now.

I agree about using the newly stabilized GATs if possible, however this can be done as a followup. Bumping up MSRV should be fine, we currently allow ourselves to do it during major releases (I consider going from 0.X to 0.(X+1) to be a major release).

scylla-cql/src/frame/value.rs
```diff
@@ -230,6 +215,78 @@ impl<'a> Iterator for SerializedValuesIterator<'a> {
...
+///
+/// This trait is not implemented directly, but rather implemented through `BatchValuesGatWorkaround`
+/// (until GATs are made available in Rust)
+pub trait BatchValues: for<'r> BatchValuesGatWorkaround<'r> {}
+impl<T: for<'r> BatchValuesGatWorkaround<'r> + ?Sized> BatchValues for T {}
+
+pub trait BatchValuesGatWorkaround<'r, ImplicitBounds = &'r Self> {
+    type BatchValuesIter: BatchValuesIterator<'r>;
+    fn batch_values_iter(&'r self) -> Self::BatchValuesIter;
+    fn len(&'r self) -> usize;
```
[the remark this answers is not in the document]
Ten0: It is used there: https://github.com/scylladb/scylla-rust-driver/blob/a818ae55c3c08d12083c67ebae66d4ab425c0462/scylla/src/transport/connection.rs#L645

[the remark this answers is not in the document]
piodul: OK. I think this check can be removed. It is redundant because you added a later check that compares statement and value counts match during serialization (in `Batch::serialize`). This error should only happen if the API is misused, so I think it is fine to detect it later, even at a point when we have already spent some time serializing the request.

scylla-cql/src/frame/request/batch.rs
```diff
@@ -54,9 +54,21 @@ where
...
-        for (statement_num, statement) in self.statements.clone().enumerate() {
+        let mut n_serialized_statements = 0usize;
+        let mut value_lists = self.values.batch_values_iter();
+        for statement in self.statements.clone() {
             statement.serialize(buf)?;
-            self.values.write_nth_to_request(statement_num, buf)?;
+            value_lists.write_next_to_request(buf).ok_or_else(|| {
+                ParseError::BadDataToSerialize(
+                    "Mismatch between statement counts for batch query".to_owned(),
+                )
```
piodul, line 64: I think the message should contain information about how many values were expected and how many there actually were.

You should also check whether the iterator has more elements than there are statements and return an error in such case.

scylla-cql/src/frame/request/batch.rs
```diff
@@ -54,9 +54,21 @@ where
...
+            })??;
+            n_serialized_statements += 1;
+        }
+        if n_serialized_statements != self.statements_count {
+            return Err(ParseError::BadDataToSerialize(
+                "Mismatch between statement counts for batch query".to_owned(),
+            ));
         }
```
piodul, line 72: I'm not sure if I understand the purpose of this check. AFAIK `n_serialized_statements == self.statements_count` is always true because we only construct `Batch` internally and `self.statements_count` is just the length of the iterator (see `Connection::batch_with_consistency`). If that was the purpose then I don't think this check is needed.

scylla-cql/src/frame/value.rs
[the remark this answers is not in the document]
Ten0: IIRC there also the type that differs at that point so we can't return the user-friendly error without more significant changes. Since the user has to know the count in values so that it matches the count in statements I feel like keeping the len requirement is unlikely to be a significant constraint.
It should also be possible to update the serialization error so that it may propagate that error back in a typed manner, though I'd have to look into it a bit to be sure of the extent of the interface update that this would require.

scylla-cql/src/frame/request/batch.rs
Ten0: As Batch is constructed in a separate file and seems to be generic to allow for evolutions in the way statements are managed, it seems to me error-prone for the future to rely on something non-local to this function to guarantee that the serialization respects the scylla protocol (that is, we have serialized as many values as we have announced).
If that were to happen, it looks like it would lead to an extremely hard to debug error.

scylla-cql/src/frame/request/batch.rs
Ten0: done

scylla-cql/src/frame/request/batch.rs
Ten0: added rationale comment

scylla-cql/src/frame/value.rs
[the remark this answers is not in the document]
Ten0: Updated the interface this way: bc80d1a3503ba1cc90d15063f39e112c856c51d9

Ten0:
I've added a commit to move to actual GATs.
Only the min_rust (that explicitly specifies Rust 1.65 as MSRV) builds, as other builds use the default image's toolchain, [whose Rust version is outdated](https://github.com/actions/runner-images/blob/3a8564acb18ccc15ca5c4f7accf9de3b3b0bc96a/images/linux/Ubuntu2004-Readme.md#rust-tools).

piodul:
> I guess we could merge the previous one then open a separate PR for moving to actual GATs.

Ok, let's do this. Please drop the last commit so that the CI checks pass and we can merge it.

Ten0:
done

piodul:
@Ten0 it looks like the clippy version got updated in the CI environment and complains about some issues in your PR. Could you fix them?

scylla-cql/src/frame/value_tests.rs
```diff
@@ -650,12 +650,46 @@ fn tuple_batch_values() {
 fn ref_batch_values() {
     let batch_values: &[&[i8]] = &[&[1, 2], &[2, 3, 4, 5], &[6]];
 
-    assert_eq!(<&&&&[&[i8]] as BatchValues>::len(&&&&batch_values), 3);
+    return check_ref_bv::<&&&&&[&[i8]]>(&&&&batch_values);
```
piodul, line 653: One ampersand too many in `&&&&&[&[i8]]`?

Ten0: batch_values already has one reference level so I think this is correct

piodul (approved):

PR #576, opened 2022-10-10
session: add keyspaces_to_fetch configuration

### LINKED ISSUE #574: Do not fetch all keyspaces
I have dozens of keyspaces with hundreds of tables in a cluster. I'm only using one keyspace for my application, but the driver still fetch all the tables/views/udt of each keyspaces, so dozens of thousands of rows. That's a lot of memory/CPU/network waste.

Moreover, I don't even use any of the metadata fetched in the `ClusterData.keyspaces`, except the keyspace strategy that the driver use itself automatically.

The driver should make metadata fetching configurable, to prevent this waste of resource.

wyfo:
This attribute allows filtering fetched keyspaces in metadata queries

Fixes #574 

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -127,6 +127,7 @@ impl Cluster {
     pub async fn new(
         initial_peers: &[SocketAddr],
         pool_config: PoolConfig,
+        keyspaces_to_fetch: Vec<String>,
         fetch_schema_metadata: bool,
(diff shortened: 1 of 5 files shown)


wyfo:
As explained in https://github.com/scylladb/scylla-rust-driver/issues/574#issuecomment-1273856773, I don't believe that's this is a good solution.
But that's a temporary one.

piodul (approved):

PR #578, opened 2022-10-14
LWT routing optimisation

wprzytula:
This enables optimisation of LWT queries routing. If the cluster supports the LWT mark protocol extension, then it will be negotiated and henceforth prepared statements which contain an LWT will be routed to the replicas only.
Fixes: #100 
## Pre-review checklist

### DIFF
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -104,6 +104,11 @@ pub mod transport;
 
 pub(crate) mod utils;
 
+/// This module is NOT part of the public API (it is `pub` only for internal use of integration tests).
+/// Future minor releases are free to introduce breaking API changes inside it.
+#[doc(hidden)]
+pub use utils::test_utils;
+
 pub use statement::batch;
 pub use statement::prepared_statement;
 pub use statement::query;
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -25,6 +25,7 @@ pub struct PreparedStatement {
     statement: String,
     page_size: Option<i32>,
     partitioner_name: PartitionerName,
+    is_confirmed_lwt: bool,
 }
 
 impl Clone for PreparedStatement {
@@ -37,6 +38,7 @@ impl Clone for PreparedStatement {
             statement: self.statement.clone(),
             page_size: self.page_size,
             partitioner_name: self.partitioner_name.clone(),
+            is_confirmed_lwt: self.is_confirmed_lwt,
         }
     }
 }
@@ -44,6 +46,7 @@ impl Clone for PreparedStatement {
 impl PreparedStatement {
     pub(crate) fn new(
         id: Bytes,
+        is_lwt: bool,
         metadata: PreparedMetadata,
         statement: String,
         page_size: Option<i32>,
@@ -57,6 +60,7 @@ impl PreparedStatement {
             page_size,
             config,
             partitioner_name: Default::default(),
+            is_confirmed_lwt: is_lwt,
         }
     }
 
@@ -94,6 +98,17 @@ impl PreparedStatement {
     /// will always be sent to a random node/shard.
     pub fn is_token_aware(&self) -> bool {
         !self.metadata.pk_indexes.is_empty()
+    }
+
+    /// Returns true if it is known that the prepared statement contains
+    /// a Lightweight Transaction. If so, the optimisation can be performed:
+    /// the query should be routed to the replicas in a predefined order
+    /// (i. e. always try first to contact replica A, then B if it fails,
+    /// then C, etc.). If false, the query should be routed normally.
+    /// Note: this a Scylla-specific optimisation. Therefore, the result
+    /// will be always false for Cassandra.
+    pub fn is_confirmed_lwt(&self) -> bool {
+        self.is_confirmed_lwt
     }
 
     /// Computes the partition key of the target table from given values —
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -334,6 +334,9 @@ impl Connection {
             Response::Error(err) => return Err(err.into()),
             Response::Result(result::Result::Prepared(p)) => PreparedStatement::new(
                 p.id,
+                self.features
+                    .protocol_features
+                    .prepared_flags_contain_lwt_mark(p.prepared_metadata.flags as u32),
                 p.prepared_metadata,
                 query.contents.clone(),
                 query.get_page_size(),
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -1531,13 +1531,28 @@ impl VerifiedKeyspaceName {
 #[cfg(test)]
 mod tests {
     use scylla_cql::errors::BadQuery;
+    use scylla_cql::frame::protocol_features::{
+        LWT_OPTIMIZATION_META_BIT_MASK_KEY, SCYLLA_LWT_ADD_METADATA_MARK_EXTENSION,
+    };
+    use scylla_cql::frame::types;
+    use scylla_proxy::{
+        Condition, Node, Proxy, Reaction, RequestFrame, RequestOpcode, RequestReaction,
+        RequestRule, ResponseFrame,
+    };
+
+    use tokio::select;
+    use tokio::sync::mpsc;
 
     use super::super::errors::QueryError;
     use super::ConnectionConfig;
     use crate::query::Query;
+    use crate::transport::connection::open_connection;
     use crate::utils::test_utils::unique_keyspace_name;
     use crate::{IntoTypedRows, SessionBuilder};
+    use std::collections::HashMap;
     use std::net::SocketAddr;
+    use std::str::FromStr;
+    use std::sync::Arc;
 
     // Just like resolve_hostname in session.rs
     async fn resolve_hostname(hostname: &str) -> SocketAddr {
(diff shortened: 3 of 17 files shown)


wprzytula:
Tests are yet to be done, but first I'd love a review to make sure the approach is right. @piodul @cvybhu

scylla/src/transport/load_balancing/mod.rs
```diff
@@ -92,6 +92,62 @@ fn slice_rotated_left<T>(slice: &[T], mid: usize) -> impl Iterator<Item = &T> +
     begin.iter().chain(end.iter())
 }
 
+/// This allows the optimisation in case of statement being a LWT: if so,
+/// the statements' queries are routed to the replicas only
+/// (as it speeds LWT execution up significantly).
+pub(crate) fn make_plan_considering_lwt<'a>(
```
piodul, line 98: I think this is an overkill. You are forcing token awareness upon users which don't want it. Moreover, you aren't taking DC awareness into consideration.

The implementation in gocql just made sure that the token aware host policy always returns replicas in the specified order. Would there be some problems if we implemented it that way? The only one I can see is that a token aware policy could be composed with another policy in such a way that the parent policy breaks this ordering in some bad way.

scylla/src/statement/prepared_statement.rs
```diff
@@ -96,6 +100,14 @@ impl PreparedStatement {
         !self.metadata.pk_indexes.is_empty()
     }
 
+    /// Returns true if it is known that the prepared statement contains
+    /// a Lightweight Transaction. If so, the optimisation will be performed:
+    /// the query will be routed first to the replicas. If false, the query
+    ///  will not necessarily be sent to the replicas.
+    pub fn is_lwt(&self) -> bool {
+        self.is_lwt
+    }
```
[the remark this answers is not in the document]
wprzytula: What about making it `pub(crate)` ?

scylla/src/transport/load_balancing/mod.rs
wprzytula: I believe I've followed Java driver implementation: https://github.com/scylladb/java-driver/blob/ed54dad73c96c724d8a4e28da9dc26893c031313/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java#L135
If this implementation is bad, perhaps we should change it in Java driver as well?

scylla/src/transport/load_balancing/mod.rs
```diff
@@ -21,13 +21,15 @@ pub use token_aware::TokenAwarePolicy;
 pub struct Statement<'a> {
     pub token: Option<Token>,
     pub keyspace: Option<&'a str>,
+    pub is_lwt: bool,
```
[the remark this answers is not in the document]
wprzytula: WDYT about `is_confirmed_lwt`?

scylla/src/transport/load_balancing/mod.rs
piodul: I'm not entirely sure. I read the discussion about the LWT optimization in the gocql repo and it turns out that the Java behavior might be better when using SERIAL consistency whereas the implementation in gocql should work better with LOCAL_SERIAL.

Ref: https://github.com/scylladb/gocql/issues/40#issuecomment-616740156

I'm not sure whether the issue was considered during development of the support in the Java driver, however I think the gocql behavior a bit more sense as the user retains more control over how replicas are chosen.

@avelanarius @Lorak-mmk could you comment on the current behavior of the Java driver?

wprzytula: > The implementation in gocql just made sure that the token aware host policy always returns replicas in the specified order. Would there be some problems if we implemented it that way?

What you say is about covering the case of LWTs in use cases when a token aware policy is used. What if it is not used, and instead there is a round robin policy, which essentially contradicts the idea of preserving the specified order?

piodul: I'd leave the round robin policy as it is. It isn't very useful by itself and is usually used as a child policy (e.g. along with token aware policy).

wprzytula:
v2:
- moved towards gocql solution (optimisation only present in TokenAwarePolicy, in other cases leaving default behaviour),
- addressed naming and docstrings comments
@piodul Could you take a look now?

scylla/src/transport/load_balancing/mod.rs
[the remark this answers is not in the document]
piodul: The new name is OK, however the docstring should state that this is a Scylla-specific flag and it will be always false for Cassandra.

scylla/src/statement/prepared_statement.rs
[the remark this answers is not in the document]
piodul: Now it's dead code... Let's not keep dead code in the repo. After the name change and stating in the docstring that this flag only works with Scylla and is always false for Cassandra, I think it should be OK for it to be kept as `pub`.

piodul:
Apart from small comments, looks good.

@wprzytula what about automatic tests? Could it be done using the proxy? If not, did you do any manual tests? If you are not going to include automatic tests then please describe your manual tests and put them in the PR's description.

scylla/src/transport/connection.rs
```diff
@@ -1664,4 +1680,89 @@ mod tests {
...
+                    .proxy_address(proxy_addr)
+                    .request_rules(make_rules(options_without_lwt_optimisation_support))
+                    .build_dry_mode(),
+            )
+            .build()
+            .run()
+            .await
+            .unwrap();
+
+        // We must interrupt the driver's full connection opening, because our proxy does not interact further after Startup.
```
piodul, line 1734: Why do we need to interrupt it? What happens if the connection opening process proceeds too far before the timeout? What guarantee do we have that 10ms isn't too short or too long?

Couldn't we intercept the `Options` and `Startup` messages and modify the LWT bit mask there, then let the real Scylla handle the rest? The LWT mask doesn't introduce large changes to the protocol - unless you send LWT queries explicitly everything should be fine.

scylla/src/lib.rs
```diff
@@ -104,6 +104,8 @@ pub mod transport;
 
 pub(crate) mod utils;
 
+pub use utils::test_utils;
```
piodul, line 107: This should not be a part of the main crate's public interface. Perhaps we can introduce yet another crate (ugh) just for test utilities, use this crate as a `dev-dependency` and never publish it.

Ref: https://stackoverflow.com/questions/44539729/what-is-an-idiomatic-way-to-have-shared-utility-functions-for-integration-tests

scylla/src/transport/connection.rs
```diff
@@ -1664,4 +1680,89 @@ mod tests {
...
+        }
+
+        let proxy_addr = SocketAddr::from_str("127.0.0.54:9042").unwrap();
+
+        let config = ConnectionConfig::default();
+
+        let (startup_tx, mut startup_rx) = mpsc::unbounded_channel();
+
+        let options_without_lwt_optimisation_support =
+            [].into_iter().collect::<HashMap<String, Vec<String>>>();
```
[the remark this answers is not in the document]
wprzytula: Definitely.

scylla/src/statement/batch.rs
```diff
@@ -17,6 +17,8 @@ pub struct Batch {
 
     pub statements: Vec<BatchStatement>,
     batch_type: BatchType,
+
+    contains_confirmed_lwt: bool,
```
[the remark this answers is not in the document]
wprzytula: The comment is no longer relevant, as we've decided to withdraw from applying the LWT routing optimisation in batch case.

scylla/src/transport/connection.rs
wprzytula: > Why do we need to interrupt it? What happens if the connection opening process proceeds too far before the timeout? What guarantee do we have that 10ms isn't too short or too long?

You are right here, we have no such guarantee. I'll fix this with a `select!` over `open_connection` and `startup_rx`.

> Couldn't we intercept the `Options` and `Startup` messages and modify the LWT bit mask there, then let the real Scylla handle the rest? The LWT mask doesn't introduce large changes to the protocol - unless you send LWT queries explicitly everything should be fine.

I believe the purpose of such unit tests is to minimize amount of actions performed and isolate the specific part; in this case, further messages between driver and cluster are needless for this test's aim.

scylla/src/lib.rs
wprzytula: IIUC, you suggest such an arrangement: (arrows signify dependency)
scylla (lib) -> scylla-test-utils -> scylla (integration tests) 
Is the above correct?

piodul: If "A -> B" means "B depends on A", then yes.

piodul: Ugh, but if we do this we won't be able to use the utilities in `scylla`'s unit tests, which is sad.

Perhaps it would work if we marked the module as `#[cfg(test)]`? If it shows up in documentation, we can also mark if with `#[doc(hidden)]` - this is also what one of the SO answers suggests.

scylla/src/transport/connection.rs
piodul: > You are right here, we have no such guarantee. I'll fix this with a `select!` over `open_connection` and `startup_rx`.

You should also make sure that the proxy doesn't respond to any requests after the `Startup` message. I'm not sure how the proxy reacts to requests when it is in dry mode and none of the rules match, but just ignoring them (not returning any response) should work here.

wprzytula: That's exactly the default behaviour of a proxy in dry mode - ignoring requests.

scylla/src/lib.rs
wprzytula: If we mark the module as `#[cfg(test)]`, then it will work only in that crate's unit tests. I've already checked that.

wprzytula: So far no good solution for that, so for now I'm leaving the present solution.

wprzytula: Addressed the problem by use of `#[doc(hidden)]` and a proper comment.

wprzytula:
I've rebased on current main and adjusted for changed `CachingSession` (by extending `RawPreparedStatementData` with `is_confirmed_lwt` value).

scylla/src/transport/session.rs
```diff
@@ -1040,6 +1041,7 @@ impl Session {
                 Statement {
                     token: self.calculate_token(ps, first_serialized_value)?,
                     keyspace: ps.get_keyspace_name(),
+                    is_confirmed_lwt: false,
```
piodul, line 1044: What if `ps` is a confirmed LWT statement? Perhaps it would make sense to forward this flag here.

scylla/tests/retries.rs
```diff
@@ -11,14 +11,17 @@ use tracing::info;
...
+    WorkerError,
 };
 
 #[tokio::test]
 #[ntest::timeout(5000)]
 async fn speculative_execution_is_fired() {
     const TIMEOUT_PER_REQUEST: Duration = Duration::from_millis(1000);
-    let res = test_with_3_node_cluster(217, |proxy_uris, translation_map, mut running_proxy| async move {
+
+    // We choose Unaware to support testing Cassandra
```
piodul, line 23: Why does the proxy panic with Cassandra in query mode? Wouldn't it make more sense in this case to switch to the "unaware" mode?

scylla/tests/lwt_optimisation.rs
```diff
@@ -0,0 +1,149 @@
...
+    ResponseOpcode, ResponseReaction, ResponseRule, ShardAwareness, WorkerError,
+};
+
+#[tokio::test]
+#[ntest::timeout(20000)]
+async fn if_lwt_optimisation_mark_offered_then_negotiatied_and_lwt_routed_optimally() {
+    // This is just to increase the likelyhood that only intended prepared statements (which contain this mark) are captures by the proxy.
+    const MAGIC_MARK: i32 = 123;
+
+    // We choose Unaware to support testing Cassandra
```
piodul, line 23: Our "cassandra" check in CI sets up only one Cassamdra node as far as I know.

wprzytula: That's true, but one can run this test on a multi-node Cassandra cluster and verify that everything works.

scylla/tests/retries.rs
wprzytula: IIUC, you suggest that once a proxy detects a non-shard-aware cluster (effectively a cluster without shards, most probably Cassandra), it should switch from `QueryNode` to `Unaware` ?
I didn't do it so in order to have explicit errors. I can change that, though, but only if I log a warning about that.

scylla/src/transport/session.rs
piodul: Please correct me if I'm wrong, but I think that we decided to postpone it until a PR which adds support for routing batches is merged. That PR was merged recently (https://github.com/scylladb/scylla-rust-driver/pull/573), so changing the highlighted line slightly should just work.

On the other hand, you should probably add tests for it, so it's not a matter of changing one line after all. If that's a large amount of work then it's OK to do in a followup.

scylla/tests/retries.rs
piodul: I think the semantic "just do what the underlying node does" makes more sense. As you found out, the current semantic of `QueryNode` is not useful with Cassandra clusters, and I'm not sure if I see a use for the current semantic - maybe you do, and if so then maybe those two semantics could live in parallel as separate modes?

scylla/tests/lwt_optimisation.rs
piodul: In that case we need to extend the CI to set up three nodes instead of one. You seem to already have done some improvements towards Cassandra support both in the proxy and in the "cassandra" CI check, so why not finish this?

It would probably be the best to extract all the improvements related to Cassandra + proxy support to a separate PR.

scylla/src/transport/session.rs
wprzytula: Let's do it in a followup.

piodul:
```
---- speculative_execution_is_fired stdout ----
Unique name: test_rust_1669812108_0
thread 'speculative_execution_is_fired' panicked at 'timeout: the function call took 5008 ms. Max time 5000 ms', scylla/tests/retries.rs:19:1
```

Looks like the `speculative_execution_is_fired` test took too long and failed :/ This is unrelated to the PR, but it shows that our timeouts in tests are definitely too short (this one got 5 seconds). I saw during local tests that Scylla spends a lot of time processing schema updates - here, it was Cassandra that failed, but I would suspect that it was too slow due to similar reasons.

wprzytula:
> ```
> ---- speculative_execution_is_fired stdout ----
> Unique name: test_rust_1669812108_0
> thread 'speculative_execution_is_fired' panicked at 'timeout: the function call took 5008 ms. Max time 5000 ms', scylla/tests/retries.rs:19:1
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> Looks like the `speculative_execution_is_fired` test took too long and failed :/ This is unrelated to the PR, but it shows that our timeouts in tests are definitely too short (this one got 5 seconds). I saw during local tests that Scylla spends a lot of time processing schema updates - here, it was Cassandra that failed, but I would suspect that it was too slow due to similar reasons.

As a quick mitigation, I'll just increase timeout in the integration tests to 30 seconds each. It will additionally trigger CI check again, so that should suffice to pass CI and merge this.

piodul (approved):

PR #583, opened 2022-10-26
proxy: Enabled dry mode (run with simulated nodes)

wprzytula:
It has appeared to be useful to have a feature in the proxy to interact with the driver without having any Scylla nodes up. This is especially useful for unit testing.

## Pre-review checklist

### DIFF
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -485,6 +485,15 @@ impl Doorkeeper {
         &mut self,
         connection_no: usize,
     ) -> Result<(SocketAddr, TcpStream, TcpStream), DoorkeeperError> {
+        let (driver_stream, driver_addr) = self.make_driver_stream(connection_no).await?;
+        let cluster_stream = self.make_cluster_stream(driver_addr).await?;
+        Ok((driver_addr, driver_stream, cluster_stream))
+    }
+
+    async fn make_driver_stream(
+        &mut self,
+        connection_no: usize,
+    ) -> Result<(TcpStream, SocketAddr), DoorkeeperError> {
         let (driver_stream, driver_addr) =
             self.listener.accept().await.map_err(|err| {
                 DoorkeeperError::DriverConnectionAttempt(self.node.proxy_addr, err)
@@ -493,7 +502,13 @@ impl Doorkeeper {
             "Connected driver from {} to {}, connection no={}.",
             driver_addr, self.node.proxy_addr, connection_no
         );
+        Ok((driver_stream, driver_addr))
+    }
 
+    async fn make_cluster_stream(
+        &mut self,
+        driver_addr: SocketAddr,
+    ) -> Result<TcpStream, DoorkeeperError> {
         let cluster_stream = if self.node.shard_awareness.is_aware() {
             let shards = match self.shards_number {
                 None => {
@@ -550,7 +565,7 @@ impl Doorkeeper {
         }
         .map_err(|err| DoorkeeperError::NodeConnectionAttempt(self.node.real_addr, err))?;
 
-        Ok((driver_addr, driver_stream, cluster_stream))
+        Ok(cluster_stream)
     }
 
     fn next_port_to_same_shard(&self, port: u16) -> u16 {
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -468,7 +468,7 @@ impl Doorkeeper {
     ) -> Result<(SocketAddr, TcpStream, TcpStream), DoorkeeperError> {
         let (driver_stream, driver_addr) =
             self.listener.accept().await.map_err(|err| {
-                DoorkeeperError::DriverConnectionAttempt(self.node.real_addr, err)
+                DoorkeeperError::DriverConnectionAttempt(self.node.proxy_addr, err)
             })?;
         info!(
             "Connected driver from {} to {}, connection no={}.",
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -1933,4 +1933,132 @@ mod tests {
                 .expect("Request processing was not concurrent");
         running_proxy.finish().await.unwrap();
     }
+
+    #[tokio::test]
+    #[ntest::timeout(1000)]
+    async fn dry_mode_proxy_drops_incoming_frames() {
+        let node1_proxy_addr = next_local_address_with_port(9876);
+        let proxy = Proxy::new([Node::new_dry_mode(node1_proxy_addr, None)]);
+        let running_proxy = proxy.run().await.unwrap();
+
+        let params = FrameParams {
+            flags: 0,
+            version: 0x04,
+            stream: 0,
+        };
+        let opcode = FrameOpcode::Request(RequestOpcode::Options);
+
+        let body = random_body();
+
+        let mut conn = TcpStream::connect(node1_proxy_addr).await.unwrap();
+
+        write_frame(params, opcode, &body, &mut conn).await.unwrap();
+        // We assert that after sufficiently long time, no error happens inside proxy.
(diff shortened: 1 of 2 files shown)


piodul:
Could you provide some examples where dry mode would be useful? Currently, we require a Scylla cluster to be available when running `cargo test`, so I don't understand why we need the dry mode.

wprzytula:
> Could you provide some examples where dry mode would be useful? Currently, we require a Scylla cluster to be available when running `cargo test`, so I don't understand why we need the dry mode.

Consider a situation that you want to test particular behaviour in case of Scylla or Cassandra in versions that are different than those available in the test cluster. Then you can simply mock their behaviour with the proxy running in dry mode.

piodul:
> Consider a situation that you want to test particular behaviour in case of Scylla or Cassandra in versions that are different than those available in the test cluster. Then you can simply mock their behaviour with the proxy running in dry mode.

Could you give some examples of the version-specific behaviors that you would like to test in this manner?

If we want to support multiple versions of Scylla/Cassandra, we should just run all the tests multiple times, against all supported versions.

wprzytula:
> > Consider a situation that you want to test particular behaviour in case of Scylla or Cassandra in versions that are different than those available in the test cluster. Then you can simply mock their behaviour with the proxy running in dry mode.
> 
> Could you give some examples of the version-specific behaviors that you would like to test in this manner?
> 
> If we want to support multiple versions of Scylla/Cassandra, we should just run all the tests multiple times, against all supported versions.

In this PR, I'm testing various server-side behaviour regarding LWT_OPTIMIZATION_MARK. In this exact case, I could do with running one test on Cassandra workflow and another one on Scylla, but in general it is often much more precise to test some feature in isolation, when the environment (in this case the server side) is strictly controlled. That's what the proxy's dry mode is for.

piodul (changes requested):
Please split the PR into smaller steps. There is a lot of things going on at once and it's very hard to track all of the changes if they are all introduced in one commit.

wprzytula:
> Please split the PR into smaller steps. There is a lot of things going on at once and it's very hard to track all of the changes if they are all introduced in one commit.

I've done my best and extracted two more commits: removal of a small piece of dead code and a unit test. However, the vast majority of this PR must be given in a single commit, as it's an encapsulated unity - the code would not compile if I split it in any way. Will you accept it after this explanation?

piodul:
> I've done my best and extracted two more commits: removal of a small piece of dead code and a unit test. However, the vast majority of this PR must be given in a single commit, as it's an encapsulated unity - the code would not compile if I split it in any way. Will you accept it after this explanation?

I still think the commit `proxy: Enabled dry mode (run with simulated nodes)` is much too complicated. A good, readable PR is not just about taking the final diff and splitting it into multiple commits. Ideally, every commit should do a single, logical step. For example, refactors and adding functionality should be separate.

I don't think I agree that the majority of the PR must be done in a single commit. I can see that the problematic commit does multiple things and you could try to introduce them step by step - here are some examples:

- You are moving some fields from `Node` to `NodeType::Real`. You could introduce `NodeType` with a single variant `NodeType::Real` and move the fields there in one commit.
- The same with conversion of `InternalNode` to an enum: `struct InternalNode` looks like `InternalNode::Real`, so you could move the fields in one commit.
- `response_rules` have been wrapped in an `Option` - maybe could be moved to its own commit?
- It looks like you moved a lot of code to `spawn_workers` - could this movement be a separate commit?
- Adding `NodeType::Simulated` and `InternalNode::Simulated` - could probably be done in separate or a single commit, whatever makes more sense.

I don't see more of them at the moment (but most likely there are more), and I may be wrong about some of them because I didn't dig too deeply into the commit.

Mixing multiple refactors with adding features results in commits which are hard to read and whose changes are hard to follow. As a reviewer, I would like to verify that those things were done correctly, but it's hard to do when they are interleaved.

Please remember to describe the reason for each commit in their messages. This includes refactors as well.

wprzytula:
> I still think the commit `proxy: Enabled dry mode (run with simulated nodes)` is much too complicated. A good, readable PR is not just about taking the final diff and splitting it into multiple commits. Ideally, every commit should do a single, logical step. For example, refactors and adding functionality should be separate.

Thank you for that detailed advice. I followed it and the refactored commits are ready for review. I hope you will grasp it easily now.

scylla-proxy/src/proxy.rs
```diff
@@ -61,12 +61,17 @@ impl ShardAwareness {
     }
 }
 
+/// Node can be either Real (truly backed by a Scylla node) or Simulated
```
piodul, line 64: Nit: I think it would be actually fine to merge `proxy: Added Node::Simulated for dry mode` commit with the one that introduces the `NodeType::Simulated` enum variant - they are logically related to each other and one wouldn't obscure the other. Besides, you could avoid having to use `#[allow(unused)]`.

scylla-proxy/src/proxy.rs
```diff
@@ -419,24 +456,27 @@ impl Doorkeeper {
...
             .map_err(|err| DoorkeeperError::DriverConnectionAttempt(node.proxy_addr(), err))?;
 
-        #[allow(irrefutable_let_patterns)]
         if let InternalNode::Real {
             shard_awareness,
             real_addr,
             ..
         } = node
         {
-            let doorkeeper_type = if shard_awareness.is_aware() {
```
piodul: This is an unrelated refactor. Please move it to a separate commit.

piodul (commented):
Thanks, it looks much better now. I think some commits could still be split a bit further and I left some suggestions, but I also tried to leave some comments regarding implementation itself.

One more thing that I would appreciate is a short, high-level description of the dry mode, i.e. what it is, what for, how it works and how it affects the operation of the proxy. For example, I remember that during review of the LWT PR it wasn't obvious for me that the proxy just ignores the requests that aren't handled by proxy's rules. This description should be placed in the PR, and - even better - as a docstring in the code, so that other people that will work on the proxy in the future will have some high-level overview.

scylla-proxy/src/proxy.rs
```diff
@@ -147,6 +147,25 @@ struct InternalNode {
     request_rules: Arc<Mutex<Vec<RequestRule>>>,
     response_rules: Arc<Mutex<Vec<ResponseRule>>>,
 }
+
+impl From<Node> for InternalNode {
```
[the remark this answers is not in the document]
piodul: I think it is too terse. I would rephrase it using simpler grammar and maybe elaborate on why you think this refactor enhances clarity.

For example: "The logic in Proxy::new responsible for converting Node to InternalNode is abstracted away into `impl From<Node> for InternalNode`. This shortens and simplifies the Proxy::new function."

wprzytula:
WDYT now, @piodul?

scylla-proxy/src/proxy.rs
```diff
@@ -1933,4 +1933,132 @@ mod tests {
...
+        };
+        let opcode = FrameOpcode::Request(RequestOpcode::Options);
+
+        let body = random_body();
+
+        let mut conn = TcpStream::connect(node1_proxy_addr).await.unwrap();
+
+        write_frame(params, opcode, &body, &mut conn).await.unwrap();
+        // We assert that after sufficiently long time, no error happens inside proxy.
+        tokio::time::sleep(Duration::from_millis(3)).await;
```
piodul, line 1957: Note for the future: we should have a way to gracefully close the proxy, i.e. prevent it from accepting more connections and wait until the current connections are closed by the client. This would allow us to avoid such sleeps as above.

scylla-proxy/src/proxy.rs
piodul: Not really resolved, but I'll let it slide this time ;)

piodul (approved):
LGTM

Thanks for taking the time to split everything up into parts, it made the review considerably easier.

PR #584, opened 2022-10-27
Makefile

havaker:
## Description

This pull request tidies up our methods to run tests and makes setting up test cluster easier.

The main thing that was added is a Makefile. Thanks to its presence it is now possible to run a cargo test suite against a 3-node cluster using only `make test` command - no need to invoke `docker-compose` manually. Doing a format check, cargo check, clippy, test and build can now be done with `make ci`.

Tests in Docker (`scripts/test_in_docker/`) were modified to use the same `docker-compose.yml` as the CI and Makefile. They were also moved to `test/dockerized/` directory, and made to be run by a `dockerized-test` `make` target.
I chose to drop the `test_in_docker.ps1` script that was written to make it easier to run tests on Windows.
Running tests in a Docker container on Windows changes the build target to the Linux one - Windows-specific code paths are not tested in that way. I don't see any benefit of maintaining this script, especially given that Windows users can use WSL to run tests.
As for now, I think that `test/dockerized` makes sense only on macOS.

This pull request is a draft because I have to add commit messages that explain motivations behind some changes.

## Pre-review checklist


Makefile
```diff
@@ -0,0 +1,78 @@
+COMPOSE := docker-compose -f test/cluster/docker-compose.yml
+
+.PHONY: all
+all: test
+
+.PHONY: ci
+ci: fmt-check check check-without-features clippy test build
```
piodul, line 7: Apart from running static checks, this command also runs tests. Considering users that will have to run tests in a docker container, it would be nice if there were two versions of this target - one which runs tests in a docker container, and the other which doesn't.

havaker:
v2:

- added `dockerized-ci` make target
- refactored `test/cluster/wait.sh` to use `cqlsh` in docker network
- make clippy fail on warnings

piodul (commented):
A note about how commits should be organized in a PR: I noticed that you introduce some functionality in one commit and then fix some issues with it in following commits (e.g.: add waiting for all nodes in the cluster to become available, add missing targets etc.). This makes review awkward when reviewing commit by commit as sometimes I want to point out issues with the code only to realize a bit later that the issue is fixed. To be fair, it was much easier for me to review the PR by looking at the "Files changed" tab.

The also applies to addressing review remarks by adding more commits to the PR. Although I understand that it technically shows the reviewer more clearly what has been fixed since the last review round, it is still awkward for new reviewers and - more importantly - for people that will look at the git log after the PR is merged.

At Scylla, we usually ask people to provide a short log of changes done since the previous version, and GitHub allows to compare changes with the branch before force-push.

I would recommend squashing some commits together for the sake of cleaner commit history.

havaker:
I agree that [make: fail on clippy warnings](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/bc705563bb0dd2ea89564dd19bf884401fdf6450) is a fix and it shouldn't been a separate commit. However,

> A note about how commits should be organized in a PR: I noticed that you introduce some functionality in one commit and then fix some issues with it in following commits (e.g.: add waiting for all nodes in the cluster to become available, add missing targets etc.). 

I did not think about it as fixing issues. For me it was more like adding new functionality in small, atomic steps - e.g. waiting for cluster wasn't introduced in the [root: introduce a top-level makefile](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/74451de79eab00c730686a7dd47309c6a4f52415) because I felt that splitting it into separate commits would make them easier to manage.
As for _add missing targets_, I had a similar reasoning - I've treated `docs`, `clean` as if they were new features, so I've added them in a separate commits. `dockerized-ci` target, which I've added after your review as a separate commit, might have been better to add to the one that introduced `dockerized-test` target.

The difficulties with reviewing may have been caused by rearranging files, I did not have ideas how to make commits that move files less painful to review.

> I would recommend squashing some commits together for the sake of cleaner commit history.

I'll squash [make: fail on clippy warnings](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/bc705563bb0dd2ea89564dd19bf884401fdf6450) to  [root: introduce a top-level makefile](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/74451de79eab00c730686a7dd47309c6a4f52415), move changes introduced by [make: add dockerized-ci target](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/221fd3c9da4b06a91a7f207ba9ce954b255dfc37) to the commit that created `dockerized-test`, and experiment with things that might make git history look cleaner and easier to reason about.

havaker:
I agree that [make: fail on clippy warnings](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/bc705563bb0dd2ea89564dd19bf884401fdf6450) is a fix and it shouldn't been a separate commit. However,

> A note about how commits should be organized in a PR: I noticed that you introduce some functionality in one commit and then fix some issues with it in following commits (e.g.: add waiting for all nodes in the cluster to become available, add missing targets etc.). 

I did not think about it as fixing issues. For me it was more like adding new functionality in small, atomic steps - e.g. waiting for cluster wasn't introduced in the [root: introduce a top-level makefile](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/74451de79eab00c730686a7dd47309c6a4f52415) because I felt that splitting it into separate commits would make them easier to manage.
As for _add missing targets_, I had a similar reasoning - I've treated `docs`, `clean` as if they were new features, so I've added them in a separate commits. `dockerized-ci` target, which I've added after your review as a separate commit, might have been better to add to the one that introduced `dockerized-test` target.

The difficulties with reviewing may have been caused by rearranging files, I did not have ideas how to make commits that move files less painful to review.

> I would recommend squashing some commits together for the sake of cleaner commit history.

I'll squash [make: fail on clippy warnings](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/bc705563bb0dd2ea89564dd19bf884401fdf6450) to  [root: introduce a top-level makefile](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/74451de79eab00c730686a7dd47309c6a4f52415), move changes introduced by [make: add dockerized-ci target](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/221fd3c9da4b06a91a7f207ba9ce954b255dfc37) to the commit that created `dockerized-test`, and experiment with things that might make git history look cleaner and easier to reason about.

havaker:
Oops, I rebased too deep and changed the commit author of some of colin's work

havaker:
v3:

- cleaner git history

test/dockerized/run.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [ -z "$CARGO_HOME" ]; then
+    CARGO_REGISTRY=$HOME/.cargo/registry
+else
+    CARGO_REGISTRY=$CARGO_HOME/registry
+fi
+
+docker run --name "scylla-rust-driver-testing" \
+    --network scylla_rust_driver_public \
+    -v "$(pwd)/../..:/scylla_driver:Z" \
```
piodul, line 22: This has a not-nice consequence of creating a `.cargo_home` directory in the main directory which is owned by the root user. Can we avoid it somehow?

test/dockerized/run.sh
```diff
@@ -0,0 +1,31 @@
...
+    -v "$CARGO_REGISTRY:/usr/local/cargo/registry:z" \
```
piodul, line 23: This might actually cause a similar problem to the mount above. I understand that this is done to save time (so that the container doesn't have to download everything every time).

Perhaps we should create a `Dockerfile` which doesn't mount anything but rather copies files inside. There is a nice trick described in this article about caching dependencies when using a Dockerfile: http://whitfin.io/speeding-up-rust-docker-builds/ .

havaker: Won't this method require building a new image for each `make dockerized-test` invocation?

I agree that mounting is a problem, but polluting local image store with one-time test images can be a problem too:(

havaker: Also, I don't see an easy way to translate this http://whitfin.io/speeding-up-rust-docker-builds/ trick to our multi-workspace cargo project.

havaker:
v4:

- avoid mounting to dockerized test containers

Note on the organization of commits:
I chose to create [test: dockerized: do not mount anything to containers](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/120e3a9b120f07c0de3dbc3c569192d521b0e22a) as a separate commit, even though it is fixing some test behavior. I did so because the affected behavior was introduced in a separate pull request ([test: make dockerized tests use the same docker-compose.yml as normal…](https://github.com/scylladb/scylla-rust-driver/pull/584/commits/4b3ceeaa493d4e3abb157ddb81caaaf82327c332) only moved it around)

test/dockerized/run.sh
piodul: > Won't this method require building a new image for each `make dockerized-test` invocation?
> 
> I agree that mounting is a problem, but polluting local image store with one-time test images can be a problem too:(

Maybe there is a way to cache only some of the steps, but not the last one which only builds and runs the tests?

> Also, I don't see any easy way to translate this http://whitfin.io/speeding-up-rust-docker-builds/ trick to our multi-workspace cargo project.

Won't copying the `Cargo.lock` and all `Cargo.toml`, while preserving the directory structure, work?

I was thinking about something like this:

```Dockerfile
# ...

RUN mkdir ./scylla/ ./scylla-cql/ ./scylla-proxy/
COPY ./Cargo.lock ./Cargo.lock
COPY ./Cargo.toml ./Cargo.toml
COPY ./scylla/Cargo.toml ./scylla/Cargo.coml
COPY ./scylla-cql/Cargo.toml ./scylla-cql/Cargo.toml
COPY ./scylla-proxy/Cargo.toml ./scylla-proxy/Cargo.toml

# Build the dependencies
RUN cargo build

# ...
```

Maybe it is possible to do in a smarter way with some globs so that we don't have to update the Dockerfile every time we add a new crate.

havaker: The snippet above fails due to missing `examples` workspace member.

This is a minimal example that I came up with for `RUN cargo build` to pass.
```dockerfile
WORKDIR /driver
RUN mkdir -p ./scylla/src ./scylla-cql/src ./scylla-proxy/src ./scylla-macros/src ./examples/src
COPY ./Cargo.lock ./Cargo.lock
COPY ./Cargo.toml ./Cargo.toml

COPY ./scylla/Cargo.toml ./scylla/Cargo.toml
COPY ./scylla-cql/Cargo.toml ./scylla-cql/Cargo.toml
COPY ./scylla-proxy/Cargo.toml ./scylla-proxy/Cargo.toml
COPY ./scylla-macros/Cargo.toml ./scylla-macros/Cargo.toml
COPY ./examples/Cargo.toml ./examples/Cargo.toml

RUN touch scylla/src/lib.rs
RUN touch scylla-cql/src/lib.rs
RUN touch scylla-proxy/src/lib.rs
RUN touch scylla-macros/src/lib.rs
RUN touch examples/src/lib.rs

COPY ./scylla/benches/benchmark.rs ./scylla/benches/benchmark.rs
COPY ./scylla-cql/benches/benchmark.rs ./scylla-cql/benches/benchmark.rs

RUN cargo build

COPY  . .
```

Despite the complexity, it still does not work. Running `cargo test` in the built container fails due to mysterious missing imports, however `cargo clean && cargo test` works as it should.

I think that going further with this solution will increase the maintenance costs significantly - adding new crates would make us update this `Dockerfile`, which we don't test in CI.

havaker: Also, the demand for dockerized tests is not known. If we find a lot of people dissatisfied with `make dockerized-test` speed, we can improve it then.
As for now, I'm in favor of simple solutions that don't add a lot of maintenance costs (just like `Dockerfile` from 4th version;)) .

piodul: I agree, the current solution (as in the current state of the PR) seems fine and we can refine it later if somebody needs it.

Issue #527, opened 2022-08-23
Extend RetryPolicy and RetryDecision to support SerialConsistency


wprzytula:
The current PR with `DowngradingConsistencyRetryPolicy` only supports specifying `RegularConsistency`. Moreover, the mentioned policy does not retry on LWTs. This should be supported.

wprzytula:
`RetryPolicy` receives a `QueryInfo` struct, which contains a `LegacyConsistency`. Should `RetryDecision` contain a `LegacyConsistency` as well, or fields with both consistencies?
1. If we use `LegacyConsistency`, how should we behave if a faulty `RetryPolicy` outputs `RetryDecision` with wrong type of consistency (i.e. the other `LegacyConsistency` variant than one received in `QueryInfo`)? Should we panic, or should we warn and... what? Return a ProtocolError? (or, perhaps this is a situation when adding a UserFaultError or similar is finally justified?).
2. If we use two fields (with `SerialConsistency` and `[Regular]Consistency`), then we probably should put them in `Option` (because why require from a `RetryPolicy` implementation to output two consistencies when only one is adequate). But again, what should we do when a suitable consistency field is `None`?
@cvybhu @piodul

piodul:
The Java and CPP driver seem to output only one consistency. How do their downgrading consistency retry policies work with LWT? I think we should just follow what they do, if possible.

Regarding using two consistencies: a single query can have both regular and sequential consistency set and they will both affect the query. While I'm not familiar with LWT at all, according to C*'s documentation this is true: https://docs.datastax.com/en/cassandra-oss/2.2/cassandra/dml/dmlConfigSerialConsistency.html

PR #585, opened 2022-10-31
retry_policy: Added support for SerialConsistency

wprzytula:
RetryDecision contains Option<Consistency> now instead of Consistency. None conveys meaning of "retry with the same CL as before", which is especially suitable when the previous consistency was Serial (even though such a case seems to be impossible in the code ATM). Moreover, covered one special case of LWT Unavailable error in DowngradingConsistencyPolicy, mirroring Java driver's behaviour.
Fixes: #527 
## Pre-review checklist

### DIFF
--- a/scylla/src/history.rs
+++ b/scylla/src/history.rs
@@ -672,7 +672,7 @@ mod tests {
         history_collector.log_attempt_error(
             attempt_id,
             &QueryError::TimeoutError,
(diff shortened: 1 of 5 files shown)


piodul (commented):
One small nitpick, otherwise looks good.

wprzytula:
v2: corrected the nitpick.

piodul (approved):

PR #588, opened 2022-11-07
connection: clearly mark db-side warnings as such

piodul:
The CQL protocol allows the database to return some warnings along with the response and the driver prints them using the tracing::warn! macro. Those warnings, when reported by tracing, may be confused with a driver-side error, so this commit attempts to make it more clear that the warning was generated on the db-side.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -839,7 +839,10 @@ impl Connection {
         )?;
 
         for warn_description in &body_with_ext.warnings {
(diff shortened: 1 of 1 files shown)


cvybhu (approved):
LGTM

Just for reference, this is how an example warning looks like:
`
2022-11-07T10:21:09.564554Z  WARN scylla::transport::connection: Response from the database contains a warning warning="This query should use ALLOW FILTERING and will be rejected in future versions."
`

PR #590, opened 2022-11-08
Implement Value for Vec<&dyn Value> and Vec<Box<dyn Value>>

rukai:
## Pre-review checklist

The new test demonstrates the new types that are now possible to provide.
Supporting `Vec<&dyn Value>` and `Vec<Box<dyn Value>>` here is very rarely useful but I ran into a case where I needed it for abstracting across multiple cassandra drivers in order to easily test my cassandra proxy implementation across multiple drivers.

### DIFF
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -473,10 +473,17 @@ impl<V: Value> Value for MaybeUnset<V> {
     }
 }
 
-// Every &impl Value should also implement Value
-impl<T: Value> Value for &T {
+// Every &impl Value and &dyn Value should also implement Value
+impl<T: Value + ?Sized> Value for &T {
     fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
         <T as Value>::serialize(*self, buf)
+    }
+}
+
+// Every Boxed Value should also implement Value
+impl<T: Value + ?Sized> Value for Box<T> {
+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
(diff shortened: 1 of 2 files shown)


piodul (changes requested):
I think we can add support for those types differently, in a more generic way, without much effort.

- `Vec<&dyn Value>`

	We already have `impl<T: Value> Value for Vec<T>` and `impl<T: Value> Value for &T`, so at the first glance `Vec<&dyn Value>` should be supported, too. However, Rust adds an implicit bound `T: Sized` and `dyn Value` does not automatically implement `Sized`, so `&dyn Value` currently doesn't implement `Value`.

	However, if you explicitly drop the assumption about `Sized` from the impl for reference, like this:

	```diff
	-impl<T: Value> Value for &T {
	+impl<T: Value + ?Sized> Value for &T {
	     fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
		 <T as Value>::serialize(*self, buf)
	     }
	 }
	```

	...then it will automatically start to work.

- `Vec<Box<dyn Value>>`

	This is just a matter of following the previous point and adding impl for `Box`:
	
	```diff
	+impl<T: Value + ?Sized> Value for Box<T> {
	+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
	+        <T as Value>::serialize(self.as_ref(), buf)
	+    }
	+}
	```

rukai:
That is indeed a lot better, thanks for the feedback!
I've made the requested changes.

piodul (approved):
LGTM. Thanks.

PR #593, opened 2022-11-14
make: do not use --verbose flag with cargo build

havaker:
## Description

Running `make build` is too verbose to use it as a convenient `cargo build` replacement. Removing `--verbose` flag helps a lot - the terminal is not flooded with messages about which compiler flags are being used etc.

## Pre-review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ dockerized-test: up wait-for-cluster
 
 .PHONY: build
 build:
-	cargo build --verbose --examples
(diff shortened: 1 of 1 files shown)


cvybhu:
It looks like new version of clippy broke the CI, we will have to merge a fix first.

LGTM, `--verbose` has its place in the CI, but for normal builds it just clutters the output.

cvybhu:
#594 has been merged, please rebase on top of it

PR #600, opened 2022-11-19
Fix #599

### LINKED ISSUE #599: Batches aren't usable in tokio tasks
Consider this example which spawns 10 tasks and, in each one of them, tries to execute a batch:

```rust
use anyhow::Result;
use scylla::batch::{Batch, BatchType};
use scylla::query::Query;
use scylla::transport::session::Session;
use scylla::SessionBuilder;
use std::env;
use std::sync::Arc;

#[tokio::main]
async fn main() -> Result<()> {
    let uri = env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());

    println!("Connecting to {} ...", uri);

    let session: Session = SessionBuilder::new().known_node(uri).build().await?;
    let session = Arc::new(session);

    // Spawn 10 tasks to execute some batches in parallel
    let mut tasks = Vec::new();
    for pk in 0..10 {
        let session = session.clone();
        let cks = (0..10).into_iter().collect::<Vec<_>>();
        tasks.push(tokio::task::spawn(async move {
            let q = Query::new("INSERT INTO ks.t (pk, ck) VALUES (?, ?)");
            let mut batch = Batch::new(BatchType::Unlogged);
            let mut vals: Vec<(i64, &i64)> = Vec::with_capacity(cks.len());

            // Deliberately use .iter() here instead of .into_iter()
            for ck in cks.iter() {
                batch.append_statement(q.clone());
                vals.push((pk, ck));
            }
            session.batch(&batch, vals).await.unwrap();
        }));
    }

    for task in tasks {
        task.await.unwrap();
    }

    Ok(())
}

```

This works on v0.6.1, but on current main (23d799414bc2caa1db9e4b0b1d1b459ab3efbdb6) it fails with the following error:

```
error: implementation of `BatchValuesGatWorkaround` is not general enough
  --> examples/problematic_batches.rs:23:20
   |
23 |           tasks.push(tokio::task::spawn(async move {
   |  ____________________^
24 | |             let q = Query::new("INSERT INTO ks.t (pk, ck) VALUES (?, ?)");
25 | |             let mut batch = Batch::new(BatchType::Unlogged);
26 | |             let mut vals: Vec<(i64, &i64)> = Vec::with_capacity(cks.len());
...  |
33 | |             session.batch(&batch, vals).await.unwrap();
34 | |         }));
   | |__________^ implementation of `BatchValuesGatWorkaround` is not general enough
   |
   = note: `Vec<(i64, &i64)>` must implement `BatchValuesGatWorkaround<'0, &'1 Vec<(i64, &i64)>>`, for any two lifetimes `'0` and `'1`...
   = note: ...but it actually implements `BatchValuesGatWorkaround<'2>`, for some specific lifetime `'2`

error: could not compile `examples` due to previous error
```

Applying the PR which switches to real GATs (https://github.com/scylladb/scylla-rust-driver/pull/596) just changes the error:

```
error: implementation of `Value` is not general enough
  --> examples/problematic_batches.rs:23:20
   |
23 |           tasks.push(tokio::task::spawn(async move {
   |  ____________________^
24 | |             let q = Query::new("INSERT INTO ks.t (pk, ck) VALUES (?, ?)");
25 | |             let mut batch = Batch::new(BatchType::Unlogged);
26 | |             let mut vals: Vec<(i64, &i64)> = Vec::with_capacity(cks.len());
...  |
33 | |             session.batch(&batch, vals).await.unwrap();
34 | |         }));
   | |__________^ implementation of `Value` is not general enough
   |
   = note: `Value` would have to be implemented for the type `&'0 i64`, for any lifetime `'0`...
   = note: ...but `Value` is actually implemented for the type `&'1 i64`, for some specific lifetime `'1`

error: could not compile `examples` due to previous error
```

If I change the type of `values` to be `Vec<(i64, i64)>` - so, no borrowing - then it works with the "real GATs" PR, but _still_ fails on main, with a similar error to the first one:

```
error: implementation of `BatchValuesGatWorkaround` is not general enough
  --> examples/problematic_batches.rs:23:20
   |
23 |           tasks.push(tokio::task::spawn(async move {
   |  ____________________^
24 | |             let q = Query::new("INSERT INTO ks.t (pk, ck) VALUES (?, ?)");
25 | |             let mut batch = Batch::new(BatchType::Unlogged);
26 | |             let mut vals: Vec<(i64, i64)> = Vec::with_capacity(cks.len());
...  |
33 | |             session.batch(&batch, vals).await.unwrap();
34 | |         }));
   | |__________^ implementation of `BatchValuesGatWorkaround` is not general enough
   |
   = note: `Vec<(i64, i64)>` must implement `BatchValuesGatWorkaround<'0, &'1 Vec<(i64, i64)>>`, for any two lifetimes `'0` and `'1`...
   = note: ...but it actually implements `BatchValuesGatWorkaround<'2>`, for some specific lifetime `'2`

error: could not compile `examples` due to previous error
```

Ten0:
Fixes #599

For some unknown reason, the `BatchValuesIter` associated type, when not resolved to a concrete type for a given async function, cannot live across await boundaries while maintaining the corresponding future `Send`, unless `'r: 'static`

It is consequently required to drop that variable before any `await` in `Session::batch`.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -250,7 +250,7 @@ pub trait BatchValuesIterator<'a> {
     fn skip_next(&mut self) -> Option<()>;
 }
 
-/// Implements `BatchValuesIterator` from an `Iterator` over things that implement `ValueList`
+/// Implements `BatchValuesIterator` from an `Iterator` over references to things that implement `ValueList`
 ///
 /// Essentially used internally by this lib to provide implementors of `BatchValuesIterator` for cases
 /// that always serialize the same concrete `ValueList` type
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -223,6 +223,10 @@ pub trait BatchValues: for<'r> BatchValuesGatWorkaround<'r> {}
 impl<T: for<'r> BatchValuesGatWorkaround<'r> + ?Sized> BatchValues for T {}
 
 pub trait BatchValuesGatWorkaround<'r, ImplicitBounds = &'r Self> {
+    /// For some unknown reason, this type, when not resolved to a concrete type for a given async function,
+    /// cannot live across await boundaries while maintaining the corresponding future `Send`, unless `'r: 'static`
+    ///
+    /// See <https://github.com/scylladb/scylla-rust-driver/issues/599> for more details
     type BatchValuesIter: BatchValuesIterator<'r>;
     fn batch_values_iter(&'r self) -> Self::BatchValuesIter;
 }
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -1030,10 +1030,7 @@ impl Session {
         // If users batch statements by shard, they will be rewarded with full shard awareness
 
         // Extract first serialized_value
-        let mut batch_values_iter_for_first_serialized_value = values.batch_values_iter();
-        let first_serialized_value = batch_values_iter_for_first_serialized_value
-            .next_serialized()
-            .transpose()?;
+        let first_serialized_value = values.batch_values_iter().next_serialized().transpose()?;
         let first_serialized_value = first_serialized_value.as_deref();
         let statement_info = match (first_serialized_value, batch.statements.first()) {
             (Some(first_serialized_value), Some(BatchStatement::PreparedStatement(ps))) => {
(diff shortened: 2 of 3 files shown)


piodul (approved):
Thank you for the quick investigation and the fix! LGTM.

PR #601, opened 2022-11-21
cql_to_rust: expose impl_from_cql_value_from_method macro

piodul:
This commit exposes the existing `impl_from_cql_val` macro as `impl_from_cql_value_from_method` macro. It was originally used to define implementations of the FromCqlValue macro based on CqlValue's existing methods, but it can be useful outside the crate itself too - namely, users can define more methods for CqlValue via extension traits.

## Pre-review checklist

### DIFF
--- a/examples/custom_deserialization.rs
+++ b/examples/custom_deserialization.rs
@@ -1,0 +1,73 @@
+use anyhow::Result;
+use scylla::cql_to_rust::{FromCqlVal, FromCqlValError};
+use scylla::frame::response::result::CqlValue;
+use scylla::macros::impl_from_cql_value_from_method;
+use scylla::{Session, SessionBuilder};
+use std::env;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let uri = env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+
+    println!("Connecting to {} ...", uri);
+
+    let session: Session = SessionBuilder::new().known_node(uri).build().await?;
+
+    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1}", &[]).await?;
+    session
+        .query(
+            "CREATE TABLE IF NOT EXISTS ks.t (pk int PRIMARY KEY, v text)",
+            &[],
+        )
+        .await?;
+
+    session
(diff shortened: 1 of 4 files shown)


examples/custom_deserialization.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    impl_from_cql_value_from_method!(MyOtherType, into_my_other_type);
+
+    let (v,) = session
+        .query("SELECT v FROM ks.t WHERE pk = 1", ())
+        .await?
+        .single_row_typed::<(MyOtherType,)>()?;
+    assert_eq!(v, MyOtherType("asdf".to_owned()));
```
cvybhu, line 68: Man this is cumbersome to use.

At first I didn't know what's going on here, why the extra trait???.

Maybe we could modify the macro so that it takes a free function instead of expecting a method on `CqlValue`.

scylla-cql/src/frame/response/cql_to_rust.rs
```diff
@@ -69,33 +69,67 @@ impl<T: FromCqlVal<CqlValue>> FromCqlVal<Option<CqlValue>> for Option<T> {
...
-macro_rules! impl_from_cql_val {
+/// This macro implements FromCqlVal given a type and method of CqlValue that returns this type.
+///
+/// It can be useful in client code in case you have an extension trait for CqlValue
+/// and you would like to convert one of its methods into a FromCqlVal impl.
+/// The conversion method must return an `Option<T>`. `None` values will be
+/// converted to `CqlValue::BadCqlType`.
+///
+/// # Example
+/// ```
```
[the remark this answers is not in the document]
piodul: It doesn't seem to be required. We have lots of examples in `session_builder.rs` that don't have the `rust` word in the first line and they seem to be rendered fine: https://docs.rs/scylla/0.6.1/scylla/transport/session_builder/struct.SessionBuilder.html

examples/custom_deserialization.rs
piodul: > At first I didn't know what's going on here, why the extra trait???.

This is the extension trait pattern. You can't directly add more methods to types or traits defined outside your project. However, you can work around this limitation by defining your own trait and implementing it for the type/trait you wanted to extend.

> Man this is cumbersome to use.

This macro is meant for users who already defined their own extension trait with conversion methods and want to easily turn those methods into `FromCqlVal` implementation so that you can use the converted-to types in `into_typed` method and friends.

piodul:
@cvybhu review ping

PR #602, opened 2022-11-22
Cassandra 3 nodes CI & proxy features enhancements

wprzytula:
I've made our CI Cassandra cluster consist of 3 nodes (instead of only 1). This will let us test freely on both C* and Scylla in case that the test assumes 3 nodes present; we will no longer have to opt those tests out for C*.

Moveover, I've made several improvements to proxy:
- rule changing is no longer async
- Reaction prettier printing
- Support frames can be conveniently forged
- ShardAwareness::QueryNode falls back to Unaware in case of shard info not present at node

## Pre-review checklist

### DIFF
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -13,18 +13,16 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
-    services:
-      cassandra:
-        image: cassandra
-        ports:
-          - 9042:9042
-    env:
-      CDC: 'disabled'
     steps:
     - uses: actions/checkout@v2
+    - name: Setup 3-node Cassandra cluster
+      run: |
+        docker-compose -f test/cluster/cassandra/docker-compose.yml up -d
       # A separate step for building to separate measuring time of compilation and testing
     - name: Build the project
       run: cargo build --verbose --tests
     - name: Run tests on cassandra
-      # test threads must be one because else database tests will run in parallel and will result in flaky tests
-      run: cargo test --verbose -- --test-threads=1 --skip test_views_in_schema_info --skip retries_occur --skip speculative_execution_is_fired
+      run: |
+        CDC='disabled' SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose -- --skip test_views_in_schema_info
+    - name: Stop cluster
+      run: docker-compose -f test/cluster/cassandra/docker-compose.yml down
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -10,11 +10,11 @@ use std::collections::HashMap;
 use std::fmt::Display;
 use std::future::Future;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, TcpSocket, TcpStream};
 use tokio::sync::mpsc::error::TryRecvError;
-use tokio::sync::{broadcast, mpsc, Mutex};
+use tokio::sync::{broadcast, mpsc};
 use tracing::{debug, error, info, trace, warn};
 
 // Used to notify the user that the proxy finished - this happens when all Senders are dropped.
@@ -376,18 +376,18 @@ pub struct RunningNode {
 
 impl RunningNode {
     /// Replaces the previous request rules with the new ones.
-    pub async fn change_request_rules(&mut self, rules: Option<Vec<RequestRule>>) {
-        *self.request_rules.lock().await = rules.unwrap_or_default();
+    pub fn change_request_rules(&mut self, rules: Option<Vec<RequestRule>>) {
+        *self.request_rules.lock().unwrap() = rules.unwrap_or_default();
     }
 
     /// Replaces the previous response rules with the new ones.
-    pub async fn change_response_rules(&mut self, rules: Option<Vec<ResponseRule>>) {
+    pub fn change_response_rules(&mut self, rules: Option<Vec<ResponseRule>>) {
         *self
             .response_rules
             .as_ref()
             .expect("No response rules on a simulated node!")
             .lock()
-            .await = rules.unwrap_or_default();
+            .unwrap() = rules.unwrap_or_default();
     }
 }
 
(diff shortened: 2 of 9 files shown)


scylla-proxy/src/proxy.rs
```diff
@@ -681,22 +683,27 @@ impl Doorkeeper {
...
+            if let shards @ Some(_) = self.shards_number {
+                shards
+            } else {
+                let temporary_stream = TcpStream::connect(real_addr)
+                    .await
+                    .map_err(|err| DoorkeeperError::NodeConnectionAttempt(real_addr, err))?;
+                let shards = match self.obtain_shards_number(temporary_stream, real_addr).await {
+                    Ok(shards) => Some(shards),
+                    // If a node offers no sharding info, change proxy ShardAwareness to Unaware.
+                    Err(DoorkeeperError::ObtainingShardNumberNoShardInfo) => None,
```
piodul, line 696: It's a bit strange that `obtain_shards_number` considers a situation where there is no shard info in the SUPPORTED response to be an error. This just indicates that the cluster behind the proxy is a Cassandra cluster (or a Scylla cluster which does not advertise its shard count - there is an obscure configuration option for this which nobody uses).

scylla-proxy/src/proxy.rs
```diff
@@ -681,22 +683,27 @@ impl Doorkeeper {
...
+                    Err(e) => return Err(e),
+                };
+                self.shards_number = shards;
```
piodul, line 699: By the way, I think it's better to avoid the word "number" for variables which indicate a quantity of items. The `shards_number` variable could be interpreted both as "the number of shards" and "shard's number" (a.k.a. "shard's ID"). The word "count" is a better replacement as it's less ambiguous - so, e.g. `shard_count`.

scylla/src/transport/connection.rs
```diff
@@ -1550,6 +1553,22 @@ mod tests {
...
+        let in_agreement = versions.into_iter().all(|v| v == local_version);
+        Ok(in_agreement)
+    }
+
+    async fn await_schema_agreement(connections: &[&Connection]) -> Result<(), QueryError> {
+        while !check_schema_agreement(connections).await? {
+            tokio::time::sleep(Duration::from_millis(200)).await
+        }
+        Ok(())
+    }
```
piodul, line 1570: There is no need to reimplement a feature that the `Session` already supports. You should just create a session and then use it to send the CREATE TABLE query, the session will automatically wait for schema agreement afterwards. Don't duplicate logic without a good reason.

test/cluster/cassandra/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    healthcheck:
+        test: ["CMD", "cqlsh", "-e", "describe keyspaces" ]
+        interval: 5s
+        timeout: 5s
+        retries: 60
+    networks:
+      public:
+        ipv4_address: 172.42.0.4
+    environment:
+      - CDC='disabled'
```
piodul, line 53: I don't think it is required to pass this environment variable for the Cassandra node. This is just information for the tests that use Scylla's CDC to skip themselves.

.github/workflows/cassandra.yml
```diff
@@ -13,18 +13,17 @@ env:
...
+      run: |
+        docker-compose -f test/cluster/cassandra/docker-compose.yml up -d
       # A separate step for building to separate measuring time of compilation and testing
     - name: Build the project
       run: cargo build --verbose --tests
     - name: Run tests on cassandra
       # test threads must be one because else database tests will run in parallel and will result in flaky tests
-      run: cargo test --verbose -- --test-threads=1 --skip test_views_in_schema_info --skip retries_occur --skip speculative_execution_is_fired
+      run: |
+        CDC='disabled' SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose -- --test-threads=1 --skip test_views_in_schema_info
```
piodul, line 27: Could we remove the `--test-threads=1`? Long time ago our tests weren't safe to run concurrently, but since then we fixed this and, apparently, we forgot to remove the flag from here.

scylla-proxy/src/proxy.rs
piodul: Just ranting, it's not a big issue, it's fine to fix this later.

scylla-proxy/src/proxy.rs
piodul: Just ranting, it's not a big issue, it's fine to fix this later.

scylla-proxy/src/proxy.rs
```diff
@@ -681,22 +683,27 @@ impl Doorkeeper {
...
-
-                    shards
-                }
-                Some(shards) => shards,
-            };
+        let shards = if shard_awareness.is_aware() {
+            if let shards @ Some(_) = self.shards_number {
+                shards
+            } else {
+                let temporary_stream = TcpStream::connect(real_addr)
```
piodul, line 690: Do I understand correctly that, in case of Cassandra nodes, `make_cluster_stream` will keep issuing probe connections because `self.shards_number` will keep being `None`? The driver can be configured to establish multiple connections to the same node, even with a Cassandra cluster (it doesn't do it by default though, save for the control connection). It's not a big issue, but one that is probably easy to avoid.

Besides, why doesn't the Doorkeeper try to establish the number of shards at the moment when it's created?

scylla/src/transport/connection.rs
wprzytula: So what you suggest here is using Session for all queries? Or using it for CREATE TABLE prologue only and then resorting to bare Connection? If so, should those two coexist independently or should I somehow use a connection from inside the Session?

piodul: The test is supposed to test Connection::execute_all/query_all, so those methods need to use a bare connection (there aren't any equivalent methods in Session). Just use a Session to create a table. It's perfectly fine for both Session and bare Connection to coexist. If you wish, you can also drop the Session before creating the Connection, although I doubt it make any difference.

scylla-proxy/src/proxy.rs
wprzytula: It is an error at least in `obtain_shards_count`'s sense: after all, it has failed to obtain shards count.

scylla-proxy/src/proxy.rs
wprzytula: Your right, and so I've moved querying the number of shards to the beginning of Doorkeepers operation.

scylla/src/transport/connection.rs
```diff
@@ -1564,24 +1564,34 @@ mod tests {
...
+            let session = SessionBuilder::new()
+                .known_node_addr(addr)
+                .build()
+                .await
+                .unwrap();
+            session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'SimpleStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
+            session
+                .query("DROP TABLE IF EXISTS connection_query_all_tab", &[])
+                .await
+                .unwrap();
```
piodul, line 1581: You should do `use_keyspace` before this query, or provide fully qualified names of the keyspaces in queries.

piodul (approved):

PR #607, opened 2022-11-25
transport: reset the node up/down marker on topology update

havaker:
For node up/down markers to work reliably, a source of truth other than CQL events is needed. That's because the CQL events delivery is prone to connection failures. By quering `system.cluster_status` table from time to time, we are able to get the true information about the status of nodes in the cluster.

This pull request implements this quering and resets each node up/down marker value to the information received from the periodic quering.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -472,8 +472,8 @@ impl ClusterWorker {
                                 // later as planned.
 
                                 match status {
-                                    StatusChangeEvent::Down(addr) => self.change_node_down_marker(addr, true),
-                                    StatusChangeEvent::Up(addr) => self.change_node_down_marker(addr, false),
+                                    StatusChangeEvent::Down(addr) => self.change_node_up_marker(addr, false),
+                                    StatusChangeEvent::Up(addr) => self.change_node_up_marker(addr, true),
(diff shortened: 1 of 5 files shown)


havaker:
Oops, looks like `system.cluster_status` appeared in ScyllaDB only recently (https://github.com/scylladb/scylladb/commit/7c95bd334313bd5df7cfecff22d597491efd0c35), and it does not exists at all in Cassandra.

In the absence of `system.cluster_status` table, we will be forced to use another way to detect down nodes. I think that periodically sending a keepalive query to each node would be a good substitute - the infrastructure to do so already exists (https://github.com/scylladb/scylla-rust-driver/pull/395), but the question is how can it be used.

As for now, I'm converting this PR into a draft.

wprzytula:
This effort is waiting for more time resources available.

Lorak-mmk:
Time resources are not everything. I think someone told me that `cluster_status` is not the data source we are looking for (don't remember why) - maybe it was @piodul ? We would also need to implement different behavior for Scylla and Cassandra.
Even if there exists some correct data source, should we stop sending queries to node A if we still have open working connections to it, just because node B thinks that A is down? Maybe it's better to just rely on the driver?

Issue #608, opened 2022-11-26
Driver panics while paging because it calls `Option::unwrap()` on a `None` value


Jasperav:
I have a flaky test and while I tried debugging the problem, I noticed the problem is inside the driver. I attached some screenshots:

Inside iterator.rs:
![image](https://user-images.githubusercontent.com/20363565/204089016-57f30e31-54ac-42c7-9f70-588e923fc557.png)

Relevant stacktrace:
![image](https://user-images.githubusercontent.com/20363565/204089043-9e6b4ef7-41e5-41b0-8b4f-2a5410825123.png)

The problem:
![image](https://user-images.githubusercontent.com/20363565/204089060-4ef38755-0cef-46d5-879a-eb2579ce1cb1.png)

I am paging in a distributed environment where rows are removed and added all the time. Maybe there are more places in this driver which calls `unwrap` while it can fail. 

I looked a bit inside the code. I don't understand this logic:

![image](https://user-images.githubusercontent.com/20363565/204089827-ae0e52ea-0d37-47d9-9d52-d1218eab583f.png)

A tokio task is spawned and right after that, we wait for the result of that task. I am not sure why the tokio task is spawn at all. I think the spawning can be removed and just await the `worker.work` result. If that is okay, the whole `work` method can be refactored by pulling out the `sender` variable and just make the method return `mpsc::Sender<Result<ReceivedPage, QueryError>>`. That is a lot safer than `unwrap` and forces returning the correct values.

Jasperav:
@piodul I don't see that warning before the crash. I don't see any weird logging before the crash. Why is there a `sender` and `receiver` and not just a return value? It looks easier to me that `work` returns a value, rather than dealing with a `sender` and `receiver`.

piodul:
The `execute_iter` method keeps some the state related to retries through control flow. You can see there are `nodes_in_plan` and `same_node_retries` loops. If we wanted to change the design so that RowIteratorWorker::work() just returns the next page, we would have to put this state into the RowIteratorWorker and basically reimplement the work() and query_pages() methods. I'm not sure if it would result in clearer code - personally I find `async` functions easier to reason about than manually written state machines.

Besides, I'm not sure such rework would give us any more safety. For the sake of RowIterator::get_column_specs, the driver requires that at least one page should be returned - and the new interface still wouldn't guarantee this. The first invocation of work() still can return `None`, and the unwrap problem would still be there.

[...]

PR #609, opened 2022-11-29
iterator: fix panic when retry policy returns IgnoreWriteError in iterator

piodul:
Currently, the RowIterator expects that its background worker will always send at least one page (or an error) to the iterator. However, there is an edge case where it may not be true: if the first request executed via query_iter/execute_iter fails and the retry policy returns an IgnoreWriteError decision (currently returned only by downgrading consistency policy in case of WriteTimeout). In such case, those methods will trigger a panic.

This PR fixes this issue by sending a dummy, empty page to the iterator in the aforementioned case. Moreover, in order to make it less possible for such bugs to appear in this area of the code in the future, the RowIterator::work function is now forced to return an object which serves as a proof of the fact that it sent (or at least attempted to send) at least one page or error to the RowIterator.

Fixes: #608

### DIFF
--- a/scylla/src/transport/iterator.rs
+++ b/scylla/src/transport/iterator.rs
@@ -157,7 +157,7 @@ impl RowIterator {
             };
 
             let worker = RowIteratorWorker {
-                sender,
+                sender: sender.into(),
                 choose_connection,
                 page_query,
                 statement_info: Statement::default(),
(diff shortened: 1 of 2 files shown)


wprzytula (approved):
I'm truly mesmerised by how you've used Rust strong type system to enforce stronger guarantees in iter interface. **Utmost respect**, @piodul!

By the way, let's think if there are more places in the codebase when we assume more than the type system guarantees. If there are any, we could harden them in a similar manner (by stronger types).

piodul:
@cvybhu review ping

cvybhu (approved):
Very cool!

PR #611, opened 2022-12-02
connection: send DRIVER_VERSION in STARTUP message

### LINKED ISSUE #610: Driver does not send `DRIVER_VERSION` in Startup message
When the driver connects to Scylla, it sends some data (for example compression type, driver name) in Startup message. One of available fields to send is `DRIVER_VERSION`. Thanks to this, `SELECT * FROM system.clients` will show the version of Scylla Rust Driver that was used to connect.

Scylla Rust Driver does not send `DRIVER_VERSION` right now.

avelanarius:
When sending a `STARTUP` message while establishing the connection, this change properly populates the `DRIVER_VERSION` field.

Before this change, `SELECT * FROM system.clients` would allow a user to see that Scylla Rust Driver has connected to the cluster, but the version number was `null`.

The version number is fetched compile-time by invoking option_env! and reading a standard `CARGO_PKG_VERSION` environment variable.

Example contents of system.clients after change:
```
>> `select driver_name, driver_version from system.clients;`
| Text("scylla-rust-driver")| Text("0.6.1")   |
| Text("scylla-rust-driver")| Text("0.6.1")   |
```

Fixes #610

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -1146,6 +1146,7 @@ pub async fn open_connection(
         source_port,
         config,
         Some("scylla-rust-driver".to_string()),
+        option_env!("CARGO_PKG_VERSION").map(|v| v.to_string()),
     )
     .await
 }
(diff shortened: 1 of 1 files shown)


cvybhu (approved):
LGTM.

Tested manually.
Created a separate binary crate that used `scylla-rust-driver` and the driver version was correct.
For a moment I was worried that `CARGO_PKG_VERSION` might refer to the top level crate instead of `scylla-rust-driver`

PR #617, opened 2022-12-30
Identify nodes by host id instead of IPs

wprzytula:
As a corresponding change has been merged into Scylla itself, we should adjust as well.
Moreover, for Serverless Cloud host IDs are the only valid way of identifying nodes.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/topology.rs
+++ b/scylla/src/transport/topology.rs
@@ -49,6 +49,7 @@ pub struct Metadata {
     pub keyspaces: HashMap<String, Keyspace>,
 }
 
+#[non_exhaustive] // <- so that we can add more fields in a backwards-compatible way
 pub struct Peer {
     pub host_id: Uuid,
     pub address: SocketAddr,
--- a/scylla/src/transport/topology.rs
+++ b/scylla/src/transport/topology.rs
@@ -10,6 +10,7 @@ use crate::utils::parse::{ParseErrorCause, ParseResult, ParserState};
 
 use crate::QueryResult;
 use futures::future::try_join_all;
+use itertools::Itertools;
 use rand::seq::SliceRandom;
 use rand::{thread_rng, Rng};
 use std::borrow::BorrowMut;
@@ -501,7 +502,7 @@ async fn query_peers(
     ))?;
 
     let typed_peers_rows = peers_rows.into_typed::<(
-        Uuid,
+        Option<Uuid>,
         IpAddr,
         Option<String>,
         Option<String>,
@@ -512,7 +513,7 @@ async fn query_peers(
     let local_address = SocketAddr::new(local_ip, connect_port);
 
     let typed_local_rows = local_rows.into_typed::<(
-        Uuid,
+        Option<Uuid>,
         IpAddr,
         Option<String>,
         Option<String>,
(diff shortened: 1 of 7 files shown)


piodul (commented):
How does the new behavior interact with old clusters that don't identify nodes by host ID? What about Cassandra? Is this backwards compatible?

wprzytula:
> How does the new behavior interact with old clusters that don't identify nodes by host ID? What about Cassandra? Is this backwards compatible?

@avelanarius Could you join in?

avelanarius:
> > How does the new behavior interact with old clusters that don't identify nodes by host ID? What about Cassandra? Is this backwards compatible?
> 
> @avelanarius Could you join in?

I think that the cover letter is slightly misleading: this PR was inspired by recent changes in Scylla to unify different "node ID" mechanisms (in snitch, Raft, gossip, etc.) to use only host ID. But this PR does not require those server-side (Scylla/Cassandra) changes, but rather it is an internal driver change.

Refs https://github.com/scylladb/python-driver/pull/199 - in Python Driver now it's possible for a node to change its IP while maintaining the same host id. The driver should handle such situation correctly. Note that this might be out-of-scope of this PR, but eventually Rust Driver should support it correctly.

cc @kbr-scylla

scylla/src/transport/topology.rs
```diff
@@ -495,21 +498,31 @@ async fn query_peers(
         "system.local query response was not Rows",
     ))?;
 
-    let typed_peers_rows =
-        peers_rows.into_typed::<(IpAddr, Option<String>, Option<String>, Option<Vec<String>>)>();
+    let typed_peers_rows = peers_rows.into_typed::<(
+        Uuid,
```
piodul, line 504: `host_id` is not a key column, so it is allowed to be null. While I don't expect it should be ever null in normal circumstances, we have at least one report which mentions that information in the peers table may be "glitched": https://github.com/scylladb/scylla-rust-driver/issues/235.

Let's be defensive and fetch it as `Option<Uuid>`, then filter out nodes for which the `host_id` is null.

piodul:
> I think that the cover letter is slightly misleading: this PR was inspired by recent changes in Scylla to unify different "node ID" mechanisms (in snitch, Raft, gossip, etc.) to use only host ID. But this PR does not require those server-side (Scylla/Cassandra) changes, but rather it is an internal driver change.

I just meant to ask whether there will be issues if you use a driver that identifies by host ID but the cluster relies on identification by IP. For example, if host ID uniqueness is not enforced then a cluster could have two nodes with the same host ID (driver would connect to only one of them), or a node could change its host ID and would be considered a different node from that point of time. Now that I think of it, even if it is possible then it doesn't sound very probable nor harmful so I guess it's fine.

piodul (changes requested):
What will happen if a node changes its IP? Will the connection pool for that node react and start connecting to the new address?

mykaul:
> What will happen if a node changes its IP? Will the connection pool for that node react and start connecting to the new address?

I'd be very surprised if a node just changes its IP all the sudden. I would assume some topology change taking place (node going down and back up, etc.) has to happen.

piodul:
> I'd be very surprised if a node just changes its IP all the sudden. I would assume some topology change taking place (node going down and back up, etc.) has to happen.

The driver updates topology information in a poll-based manner - it fetches full topology information with a configurable interval (1 minute by default) and sees what has changed since the last poll. If the driver is unlucky it might not observe the node going down, only observe the state before and after IP change.

That's fine, my concern is that we need to make sure that the connection pool for that node is reconfigured to use the new IP.

piodul:
> That's fine, my concern is that we need to make sure that the connection pool for that node is reconfigured to use the new IP.

Upon closer inspection, it looks like the PR doesn't have this problem after all. In `ClusterData` we are still identifying nodes by their connect address, so if the IP changes then the old connection pool will be dropped and a new `Node` object will be created with a new connection pool.

The `ClusterData::new` function should be modified to consider the case when a node with a host ID takes over an IP of another node. We should detect it and create a new `Node` object in that case instead of reusing the old one.

wprzytula:
@piodul Addressed:
> The ClusterData::new function should be modified to consider the case when a node with a host ID takes over an IP of another node. We should detect it and create a new Node object in that case instead of reusing the old one.

as well as the small nit above.

Issue #586, opened 2022-11-04
Assign addresses for the test proxy endpoints automatically


piodul:
Recently, we have introduced a proxy for use in tests which is a man-in-the-middle between the driver and a real Scylla cluster and allows to intercept/modify some of the requests and responses. For each node in the real cluster, the proxy needs a free local IP so that it can listen on it. We cannot reuse the same set of IPs in each test because tests are usually run in parallel.

Currently, this problem is solved by statically assigning localhost IPs in tests and making sure that other tests are not using them. However, this may become a bit tedious and error prone if the number of tests grow. It would be best to make the process of assigning IPs for the proxy automatic.

We essentially need some kind of a global, atomic counter for this. We need to investigate how `cargo test` runs tests in parallel - if it only runs one process at a time and uses multiple threads, then an `AtomicU32` will suffice. If multiple processes can run in parallel, then `AtomicU32` won't suffice and we should consider a different approach.

PR #621, opened 2023-01-05
Automated proxy addresses in tests

wprzytula:
## Pre-review checklist
As tests are run concurrently on a single process, so addresses for proxy instances used in them can be assigned automatically in a non-colliding way. To this end, one atomic suffices.

Fixes: #586

### DIFF
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -10,6 +10,7 @@ use std::collections::HashMap;
 use std::fmt::Display;
 use std::future::Future;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::{Arc, Mutex};
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, TcpSocket, TcpStream};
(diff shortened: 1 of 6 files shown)


scylla-proxy/src/proxy.rs
```diff
@@ -1174,6 +1175,17 @@ impl ProxyWorker {
...
+// Useful for concurrent testing.
+#[doc(hidden)]
+pub fn get_exclusive_address_last_octet() -> u8 {
+    static ADDRESS_LAST_OCTET: AtomicU16 = AtomicU16::new(42);
+    ADDRESS_LAST_OCTET
+        .fetch_add(1, Ordering::Relaxed)
+        .try_into()
+        .expect("Loopback address pool for tests depleted")
+}
+
```
piodul, line 1188: I guess you are using a 16-bit atomic here to detect overflow?

I think this function unnecessarily restricts itself to just one byte. AFAIK the loopback subnet is quite large - the first octet has to be 127 and the rest is fairly arbitrary, so we have 2^24 - 2 addresses to choose from. While it is realistic to deplete a pool of `256 - 42` addresses in tests, with U32 we won't have to worry about overflow.

piodul: I also think this function should just return a whole address to make life simpler to the callers.

cvybhu:

piodul (approved):

PR #622, opened 2023-01-05
secrecy: Features and dependencies cleanup

wprzytula:
First, secrecy is now built as `scylla-cql` dependency only if the corresponding feature is requested. Secondly, the corresponding feature that turns on the secrecy-related serialization has been added to scylla crate.

## Pre-review checklist


cvybhu:

piodul (commented):
Looks good, however it appears that the `secret` feature is not tested in our CI at all. Please add a `cargo check` step with this feature enabled in the `rust.yml` workflow.

The number of features slowly starts to grow (ssl, secret, soon: cloud) so we should consider checking various (all?) combinations of the features in the future.

piodul (approved):

PR #625, opened 2023-01-12
caching_session: use custom hasher instead of ahash in tests

piodul:
Currently, we are importing the `ahash` crate for only one reason: to test that CachingSession works with custom hasher implementations. The custom hasher quality doesn't matter at all in this test, so it's very easy to write own implementation for testing purposes, rendering the dependency on `ahash` obsolete.

The change was prompted by a report from one of our users which use a fork of the driver and managed to trigger a bug in cargo - most likely this one: https://github.com/rust-lang/cargo/issues/7463 . The most likely cause was that the project specified a different, non-compatible version of ahash in the dependencies than the one in scylla's dev-dependencies. While it doesn't fix the cargo bug and it can still happen with other dev-dependencies, the bug should no longer happen with conflicting `ahash`. Moreover, removing an unneeded depdencency from a project is a good motivation in itself.

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -51,7 +51,6 @@ ntest = "0.8.1"
 criterion = "0.3"
 tracing-subscriber = { version = "0.3.14", features = ["env-filter"] }
 assert_matches = "1.5.0"
-ahash = { version = "0.8.0", default-features = false, features=["compile-time-rng"] }
 
 [[bench]]
 name = "benchmark"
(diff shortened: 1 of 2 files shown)


cvybhu:

piodul:
@cvybhu rebased, ping

piodul:
@avelanarius No, it was actually a bug in cargo's resolver which can prevent a dependent project from building at all. Added more info to the commit description.

avelanarius:
In the updated commit message:
> The change was prompted by a report from one of our users which use a
> fork of the driver and managed to trigger a bug in cargo - most likely
> this one: https://github.com/scylladb/scylla-rust-driver/pull/625 .

Shouldn't the link lead somewhere else?

piodul:
> Shouldn't the link lead somewhere else?

Yes, I realized that and already pushed a commit with corrected message.

piodul:
@cvybhu ping

cvybhu (approved):

PR #627, opened 2023-01-13
CI: bump MSRV to 1.65

piodul:
Bumping MSRV due to those reasons:

- The newest version of tokio (1.24.1) doesn't seem to compile with Rust 1.59 and `-D warnings`. Fortunately, 1.65 doesn't seem to be affected.
- #596 needs GATs, which were introduced in 1.65
- #614 would like to use the new `dep:` syntax for optional features.

The version 1.65 is also not the newest, released about 2.5 months ago since the time this commit message was written. Although our MSRV policy allows us to bump MSRV in major releases - and I understand going from 0.x to 0.(x+1) as such a change - choosing an older version has the benefit that, at the time of release of the next driver version, some of the users will probably already have migrated to a Rust version supported by the driver.

## Pre-review checklist

- [ ] ~All commits compile, pass static checks and pass test.~ not really, but it should make the situation better

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,7 +9,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   RUSTFLAGS: -Dwarnings
-  rust_min: 1.59.0 # <- Update this when bumping up MSRV
+  rust_min: 1.65.0 # <- Update this when bumping up MSRV
 
 jobs:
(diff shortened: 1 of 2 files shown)


cvybhu:
Lgtm

Some CI checks failed, but that's because they require separate PRs.
Merging all of the CI PRs should fix it.

PR #628, opened 2023-01-13
scylla-macros: increase hygiene

wmitros:
While trying to make my macros at https://github.com/wmitros/scylla-rust-udf/pull/1 more hygienic, I noticed that I also need the macros here to be more hygienic.
I'm posting this mostly to find out if I'm heading in the right direction with the changes, although this patch does pass tests if it turns out being close to a state ready for merging

~~- [ ] I added appropriate `Fixes:` annotations to PR description.~~

### DIFF
--- a/scylla-macros/src/from_row.rs
+++ b/scylla-macros/src/from_row.rs
@@ -22,8 +22,8 @@ pub fn from_row_derive(tokens_input: TokenStream) -> TokenStream {
                     .unwrap(); // vals_iter size is checked before this code is reached, so
                                // it is safe to unwrap
 
-                <#field_type as FromCqlVal<Option<CqlValue>>>::from_cql(col_value)
-                    .map_err(|e| FromRowError::BadCqlVal {
+                <#field_type as scylla::cql_to_rust::FromCqlVal<::std::option::Option<scylla::frame::response::result::CqlValue>>>::from_cql(col_value)
+                    .map_err(|e| scylla::cql_to_rust::FromRowError::BadCqlVal {
                         err: e,
                         column: col_ix,
                     })?
@@ -35,9 +35,10 @@ pub fn from_row_derive(tokens_input: TokenStream) -> TokenStream {
     let generated = quote! {
         impl #impl_generics scylla::cql_to_rust::FromRow for #struct_name #ty_generics #where_clause {
             fn from_row(row: scylla::frame::response::result::Row)
-            -> Result<Self, scylla::cql_to_rust::FromRowError> {
+            -> ::std::result::Result<Self, scylla::cql_to_rust::FromRowError> {
                 use scylla::frame::response::result::CqlValue;
                 use scylla::cql_to_rust::{FromCqlVal, FromRow, FromRowError};
+                use ::std::result::Result::{Ok, Err};
 
                 if #fields_count != row.columns.len() {
                     return Err(FromRowError::WrongRowSize {
--- a/scylla-macros/src/from_row.rs
+++ b/scylla-macros/src/from_row.rs
@@ -39,6 +39,7 @@ pub fn from_row_derive(tokens_input: TokenStream) -> TokenStream {
             -> ::std::result::Result<Self, #path::FromRowError> {
                 use #path::{CqlValue, FromCqlVal, FromRow, FromRowError};
                 use ::std::result::Result::{Ok, Err};
+                use ::std::iter::{Iterator, IntoIterator};
 
                 if #fields_count != row.columns.len() {
                     return Err(FromRowError::WrongRowSize {
--- a/scylla-macros/src/from_row.rs
+++ b/scylla-macros/src/from_row.rs
@@ -5,6 +5,7 @@ use syn::{spanned::Spanned, DeriveInput};
 /// #[derive(FromRow)] derives FromRow for struct
 pub fn from_row_derive(tokens_input: TokenStream) -> TokenStream {
     let item = syn::parse::<DeriveInput>(tokens_input).expect("No DeriveInput");
+    let path = crate::parser::get_path(&item).expect("No path");
     let struct_fields = crate::parser::parse_named_fields(&item, "FromRow");
 
     let struct_name = &item.ident;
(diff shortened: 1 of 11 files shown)


cvybhu:

scylla-macros/src/into_user_type.rs
```diff
@@ -14,20 +14,20 @@ pub fn into_user_type_derive(tokens_input: TokenStream) -> TokenStream {
...
         }
     });
 
     let generated = quote! {
         impl #impl_generics scylla::frame::value::Value for #struct_name #ty_generics #where_clause {
-            fn serialize(&self, buf: &mut Vec<u8>) -> std::result::Result<(), scylla::frame::value::ValueTooBig> {
+            fn serialize(&self, buf: &mut ::std::vec::Vec<::core::primitive::u8>) -> ::std::result::Result<(), scylla::frame::value::ValueTooBig> {
                 use scylla::frame::value::{Value, ValueTooBig};
                 use scylla::macros::BufMut;
                 use ::std::convert::TryInto;
```
piodul: Would importing the primitive types, just like `BufMut/TryInto/etc.` here work?

```rust
use ::std::primitive::{i32, usize};
```

scylla-macros/src/from_row.rs
```diff
@@ -22,8 +22,8 @@ pub fn from_row_derive(tokens_input: TokenStream) -> TokenStream {
                     .unwrap(); // vals_iter size is checked before this code is reached, so
                                // it is safe to unwrap
 
-                <#field_type as FromCqlVal<Option<CqlValue>>>::from_cql(col_value)
-                    .map_err(|e| FromRowError::BadCqlVal {
+                <#field_type as scylla::cql_to_rust::FromCqlVal<::std::option::Option<scylla::frame::response::result::CqlValue>>>::from_cql(col_value)
+                    .map_err(|e| scylla::cql_to_rust::FromRowError::BadCqlVal {
```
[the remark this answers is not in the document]
piodul: Ah, I think there are some tests in scylla-cql that need to use the macros and do `use crate as scylla` in order for the macros to work... I'm not sure what would be the best way to approach this at the moment, so I think we can make an exception for the `scylla` crate at the moment.

piodul (changes requested):
I think the amount of changes that need to be done here can be reduced. Instead of fully qualifying everything, which is tedious, you could import some items with the `use` directive. As long as you are importing in a scope created by the macro which does not pollute the global namespace, it should be fine. For example, importing inside a method body created by the derive macro is OK, but importing items just before the `impl` block is not.

It would also be nice to have a test that verifies that the macros are indeed hygienic. It shouldn't be hard to add, just create an integration test (in scylla-cql or scylla crate) that uses `#[no_implicit_prelude]` and use the derive macros there. The `scylla` crate would be an exception, you will have to import it. Ref: https://gist.github.com/Kestrer/8c05ebd4e0e9347eb05f265dfb7252e1#test-your-macro-with-no_implicit_prelude

scylla-macros/src/from_row.rs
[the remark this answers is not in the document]
wmitros: Yes, this is an issue that I also had

scylla-macros/src/into_user_type.rs
wmitros: Yes, though I used `core` instead of `std` for no particular reason

wmitros:
In the last rebase, I replaced the full paths with `use` statements in every place it shortened the code, adjusted the `get_path` function so that it does not allow renaming the crate multiple times, and added a `no_implicit_prelude` test for renaming and just using the derive macros

scylla-macros/src/parser.rs
```diff
@@ -17,3 +18,48 @@ pub(crate) fn parse_named_fields<'a>(
...
+            }
+            Err(err) => {
+                return Err(err);
+            }
+        }
+    }
+    match this_path {
+        Some(path) => Ok(path),
+        None => Ok(quote::quote!(scylla)),
+    }
```
piodul, line 64: Nit: `this_path.unwrap_or_else(|| quote::quote!(scylla));`

scylla/tests/hygiene.rs
```diff
@@ -0,0 +1,64 @@
...
+    let mut buf = Vec::<u8>::new();
+    test_struct.serialize(&mut buf).unwrap();
+    let mut buf_assert = Vec::<u8>::new();
+    let tuple_with_same_layout: (i32,) = (16,);
+    tuple_with_same_layout.serialize(&mut buf_assert).unwrap();
+    assert_eq!(buf, buf_assert);
+
+    let sv = test_struct.serialized().unwrap().into_owned();
+    let sv2 = tuple_with_same_layout.serialized().unwrap().into_owned();
+    assert_eq!(sv, sv2,);
```
piodul, line 63: Nit: comma after `sv2` is unnecessary

scylla/tests/hygiene.rs
```diff
@@ -0,0 +1,64 @@
...
+    use ::std::assert_eq;
+    use ::std::option::Option::Some;
+    use ::std::vec::Vec;
+    use _scylla::_macro_internal::{CqlValue, Row, Value, ValueList};
+    use _scylla::cql_to_rust::FromRow;
+
+    let test_struct = TestStruct { a: 16 };
+    fn get_row() -> Row {
+        Row {
+            columns: ::std::vec![Some(CqlValue::Int(16)),],
```
piodul, line 44: Nit: comma before the `]` is unnecessary

scylla-macros/src/parser.rs
```diff
@@ -17,3 +18,48 @@ pub(crate) fn parse_named_fields<'a>(
...
+                    return Err(syn::Error::new_spanned(
+                        &meta_name_value.lit,
+                        "the `scylla_crate` attribute should be a string literal",
+                    ));
+                }
+            }
+            Ok(other) => {
+                if !other.path().is_ident("scylla_crate") {
+                    continue;
+                }
```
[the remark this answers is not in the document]
wmitros: Ok

scylla/tests/hygiene.rs
```diff
@@ -0,0 +1,64 @@
+#![no_implicit_prelude]
+
+extern crate scylla as _scylla;
```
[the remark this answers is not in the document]
wmitros: I added a test for the `scylla_cql` crate, but I'm not completely sure it was what you expected, please take a look

scylla-macros/src/parser.rs
wmitros: Ok(Ok)

scylla/tests/hygiene.rs
wmitros: Ok

scylla/tests/hygiene.rs
wmitros: Ok, I wonder why `cargo fmt` didn't catch that

piodul (approved):

PR #638, opened 2023-02-02
cql_to_rust: impl FromCqlValue for Date

piodul:
The driver supports serializing and deserializing between the `date` CQL type and the `chrono::NaiveDate` type. However, the `NaiveDate` type is not always sufficient as it can't represent the whole range of values the `date` CQL type can. For that purpose, we provide a `Date` type that is able to represent all CQL values of `date` type.

Until now, it was only possible to serialize `Date`. This commit adds a missing FromCqlValue impl necessary for deserialization.

## Pre-review checklist


piodul:
Sigh... Clippy version update strikes again...

I sent another PR to disable this particular lint, I think there is a good reason to disable it - at least temporarily: https://github.com/scylladb/scylla-rust-driver/pull/639

wprzytula (approved):

piodul:
@cvybhu ping

piodul:
@cvybhu ping

PR #639, opened 2023-02-02
CI/Makefile: disable `uninlined_format_args` clippy lint

piodul:
New version of clippy introduced the `uninlined_format_args` lint which recommends putting variable names directly into format strings, like this:

    let what = "world";
    println!("Hello, {what}!");

    // Instead of:
    println!("Hello, {}!", what);

The problem with this feature is that rust-analyzer does not support renaming variables inside format strings yet. The lint has been downgraded to pedantic level and will most likely appear as such in the next release.

This commit disables the new lint in the CI and local Makefile. We can turn it on again after this feature gets better support from rust-analyzer.

## Pre-review checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Format check
       run: cargo fmt --verbose --all -- --check
     - name: Clippy check
-      run: cargo clippy --verbose --examples --tests
+      run: cargo clippy --verbose --examples --tests -- -Aclippy::uninlined_format_args
     - name: Cargo check without features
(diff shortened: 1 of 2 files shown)


wprzytula:
> `println!("Hello, {}!", world);`


You've probably meant:
`println!("Hello, {}!", what);`

wprzytula (approved):

piodul:
> > `println!("Hello, {}!", world);`
> 
> You've probably meant: `println!("Hello, {}!", what);`

Thanks, I corrected the PR description and the commit message.

cvybhu (approved):

Issue #640, opened 2023-02-03
Make address translation happen only upon opening new connection


wprzytula:
### Motivation
The logical route to a node being subject to address translation might undergo frequent changes. For instance, an IP of a proxy that the node resides behind may change. The user-provided `AddressTranslator` instance may handle this situation correctly, and this is good. However, the driver's implementation forces `Node` struct to be immutable, disallowing for adjusting the change inside `Node.address` field. To mitigate the problem, we recreate the `Node` instance if its translated IP changes. This necessarily leads, however, to recreation of the whole `NodeConnectionPool` and establishing all connections once again.

The issue is especially detrimental in case of Serverless, where a new DNS query is issued for each peer upon Metadata query, in order to get current address of proxy in the node's DC. Then, we likely receive a different IP each time (thanks to DNS round robin, for example), which leads to the recreation of node connection pool described above. This behaviour was observed during tests on serverless cluster.

### Proposed solution
Move address translation from Metadata query (`query_peers()`) to `PoolRefiller::start_opening_connection()`. This way, `Node` struct will not require any alterations if only a translated address changes.
In case of Serverless, a proxy of one DC will have a single `NodeConnectionPool` and `PoolRefiller`, which will contact it through possibly multiple IPs.

wprzytula:
Work on this is being done, a PR will be opened shortly.

mykaul:
@wprzytula quick question - if DNS returns *multiple* address (and indeed, in different order), shouldn't we walk over all addresses and store them?

wprzytula:
> @wprzytula quick question - if DNS returns *multiple* address (and indeed, in different order), shouldn't we walk over all addresses and store them?

We could, but how would we benefit from it?

mykaul:
> > @wprzytula quick question - if DNS returns _multiple_ address (and indeed, in different order), shouldn't we walk over all addresses and store them?
> 
> We could, but how would we benefit from it?

You'd connect to all its IPs, instead of only some of them? (long term, MPTCP could be cool, but that's futuristic)

PR #641, opened 2023-02-03
Move address translation to opening new connection

wprzytula:
Motivations are described in the referenced issue.
Correctness of the changes is tested by the existing tests making use of address translation. 
Fixes: #640 

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -10,7 +10,6 @@ use crate::transport::{
     errors::QueryError,
     node::Node,
     partitioner::PartitionerName,
-    session::AddressTranslator,
     topology::{Keyspace, Metadata, MetadataReader},
 };
 
@@ -134,7 +133,6 @@ impl Cluster {
         pool_config: PoolConfig,
         keyspaces_to_fetch: Vec<String>,
         fetch_schema_metadata: bool,
-        address_translator: &Option<Arc<dyn AddressTranslator>>,
         host_filter: &Option<Arc<dyn HostFilter>>,
     ) -> Result<Cluster, QueryError> {
         let (refresh_sender, refresh_receiver) = tokio::sync::mpsc::channel(32);
@@ -148,7 +146,6 @@ impl Cluster {
             server_events_sender,
             keyspaces_to_fetch,
             fetch_schema_metadata,
-            address_translator,
             host_filter,
         );
 
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -26,6 +26,8 @@ use std::time::Duration;
 use tracing::{debug, warn};
 use uuid::Uuid;
 
+use super::node::NodeAddr;
+
 #[derive(Debug, Clone)]
 #[non_exhaustive]
 pub struct ContactPoint {
@@ -532,7 +534,7 @@ impl ClusterWorker {
         let node = match cluster_data
             .known_peers
             .values()
-            .find(|&peer| peer.address == addr)
+            .find(|&peer| peer.address == NodeAddr::Translatable(addr))
         {
             Some(node) => node,
             None => {
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -310,12 +310,13 @@ impl ClusterData {
             // Changing rack/datacenter but not ip address seems improbable
             // so we can just create new node and connections then
             let node: Arc<Node> = match known_peers.get(&peer.host_id) {
-                Some(node)
-                    if node.datacenter == peer.datacenter
-                        && node.rack == peer.rack
-                        && node.address == peer.address =>
-                {
-                    node.clone()
+                Some(node) if node.datacenter == peer.datacenter && node.rack == peer.rack => {
+                    if node.address == peer.address {
+                        node.clone()
+                    } else {
+                        // If IP changes, the Node struct is recreated, but the underlying pool is preserved and notified about the IP change.
+                        Arc::new(Node::inherit_with_ip_changed(node, peer.to_peer_endpoint()))
+                    }
                 }
                 _ => {
                     let is_enabled = host_filter.map_or(true, |f| f.accept(&peer));
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -26,6 +26,12 @@ use std::time::Duration;
 use tracing::{debug, warn};
 use uuid::Uuid;
 
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct ContactPoint {
+    pub address: SocketAddr,
+}
+
 /// Cluster manages up to date information and connections to database nodes.
 /// All data can be accessed by cloning Arc<ClusterData> in the `data` field
 pub struct Cluster {
@@ -130,7 +136,7 @@ struct UseKeyspaceRequest {
 
 impl Cluster {
     pub async fn new(
-        initial_peers: Vec<SocketAddr>,
+        initial_peers: Vec<ContactPoint>,
         pool_config: PoolConfig,
         keyspaces_to_fetch: Vec<String>,
         fetch_schema_metadata: bool,
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -129,11 +129,11 @@ struct UseKeyspaceRequest {
 
 impl Cluster {
     pub async fn new(
-        initial_peers: &[SocketAddr],
+        initial_peers: Vec<SocketAddr>,
         pool_config: PoolConfig,
         keyspaces_to_fetch: Vec<String>,
         fetch_schema_metadata: bool,
-        host_filter: &Option<Arc<dyn HostFilter>>,
+        host_filter: Option<Arc<dyn HostFilter>>,
     ) -> Result<Cluster, QueryError> {
         let (refresh_sender, refresh_receiver) = tokio::sync::mpsc::channel(32);
         let (use_keyspace_sender, use_keyspace_receiver) = tokio::sync::mpsc::channel(32);
@@ -146,7 +146,7 @@ impl Cluster {
             server_events_sender,
             keyspaces_to_fetch,
             fetch_schema_metadata,
-            host_filter,
+            &host_filter,
         );
 
         let metadata = metadata_reader.read_metadata(true).await?;
@@ -173,7 +173,7 @@ impl Cluster {
             use_keyspace_channel: use_keyspace_receiver,
             used_keyspace: None,
 
-            host_filter: host_filter.clone(),
+            host_filter,
         };
 
         let (fut, worker_handle) = worker.work().remote_handle();
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -320,11 +320,8 @@ impl ClusterData {
                 _ => {
                     let is_enabled = host_filter.map_or(true, |f| f.accept(&peer));
                     Arc::new(Node::new(
-                        peer.host_id,
-                        peer.address,
+                        peer.to_peer_endpoint(),
                         pool_config.clone(),
-                        peer.datacenter,
-                        peer.rack,
                         used_keyspace.clone(),
                         is_enabled,
                     ))
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -24,6 +24,7 @@ use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Duration;
 use tracing::{debug, warn};
+use uuid::Uuid;
 
 /// Cluster manages up to date information and connections to database nodes.
 /// All data can be accessed by cloning Arc<ClusterData> in the `data` field
@@ -58,8 +59,8 @@ pub struct Datacenter {
 
 #[derive(Clone)]
 pub struct ClusterData {
-    pub(crate) known_peers: HashMap<SocketAddr, Arc<Node>>, // Invariant: nonempty after Cluster::new()
-    pub(crate) ring: BTreeMap<Token, Arc<Node>>, // Invariant: nonempty after Cluster::new()
+    pub(crate) known_peers: HashMap<Uuid, Arc<Node>>, // Invariant: nonempty after Cluster::new()
+    pub(crate) ring: BTreeMap<Token, Arc<Node>>,      // Invariant: nonempty after Cluster::new()
     pub(crate) keyspaces: HashMap<String, Keyspace>,
     pub(crate) all_nodes: Vec<Arc<Node>>,
     pub(crate) datacenters: HashMap<String, Datacenter>,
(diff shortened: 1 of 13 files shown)


scylla/src/transport/node.rs
```diff
@@ -55,6 +57,72 @@ impl TimestampedAverage {
...
+/// in system.local to 0.0.0.0. Mitigation involves replacing the faulty address
+/// with connection's address, but then that address must not be subject to AddressTranslator,
+/// so we carry that information using this enum. Address translation is never performed
+/// on InitiallyKnown and FixedLocal variants.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy)]
+pub enum NodeAddr {
+    RealPeer(SocketAddr),       // fetched in Metadata with query_peers()
+    InitiallyKnown(SocketAddr), // provided by user in SessionConfig
+    FixedLocal(SocketAddr), // built from control connection's address in order to mitigate the bug described above
```
piodul, line 72: Please document each of the enum variants with a docstring.

scylla/src/transport/host_filter.rs
```diff
@@ -54,7 +54,7 @@ impl AllowListHostFilter {
 
 impl HostFilter for AllowListHostFilter {
     fn accept(&self, peer: &Peer) -> bool {
-        self.allowed.contains(&peer.address)
+        self.allowed.contains(&peer.address.inner())
```
piodul, line 57: Hmm, so now we are consulting the host filter for all types of addresses (RealPeer, FixedLocal, InitiallyKnown). I think that we should only ask the hashmap about RealPeer addresses and return true for others:

- If the address is InitiallyKnown, the user put the address into the initial peers list, so they most likely want to connect to this node,
- If the address is FixedLocal, then the node either was originally an InitiallyKnown node, or a RealPeer node for which the host filter returned true.

What do you think?

scylla/src/transport/connection_pool.rs
```diff
@@ -603,7 +605,9 @@ impl PoolRefiller {
                 }
 
                 evt = self.ready_connections.select_next_some(), if !self.ready_connections.is_empty() => {
-                    self.handle_ready_connection(evt);
+                    if let Ok(evt) = evt {
+                        self.handle_ready_connection(evt);
+                    }
```
piodul, line 611: You must call `self.handle_ready_connection` in case of failure. It updates the state of the pool as a reaction to the error.

`OpenedConnectionEvent` already holds a result, so you should put the translation error there (probably converted to a QueryError?).

scylla/src/transport/connection_pool.rs
```diff
@@ -907,45 +908,66 @@ impl PoolRefiller {
...
+                    // Known nodes, a.k.a. contact points, are user-provied addresses that do not require translation.
+                    Ok(addr)
                 }
-                .boxed()
+                NodeAddr::FixedLocal(addr) => {
+                    // In this case, addr is considered to be translated, as it is the control connection's address.
+                    Ok(addr)
+                }
+            }
+        };
```
piodul, line 942: I think this code could be pushed down to an even lower layer - maybe put the address translator to the connection config and then use it in `connection::open_conection`?

Moreover, you aren't considering the shard-aware port in this code, but you could.

piodul (changes requested):

scylla/src/transport/node.rs
```diff
@@ -55,6 +57,72 @@ impl TimestampedAverage {
...
+impl Ord for NodeAddr {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.inner().cmp(&other.inner())
+    }
+}
+impl Hash for NodeAddr {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.inner().hash(state);
+    }
+}
```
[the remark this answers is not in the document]
wprzytula: I can't recall what made me choose this way. One idea: in case of no AddressTranslator provided, with current `known_peers` implementation, `Node`s will be recycled instead of being rebuilt if a control connection changes to another node: both previous and new control connection's `Node`s will be spared.

scylla/src/transport/node.rs
```diff
@@ -55,6 +57,72 @@ impl TimestampedAverage {
...
+    InitiallyKnown(SocketAddr), // provided by user in SessionConfig
+    FixedLocal(SocketAddr), // built from control connection's address in order to mitigate the bug described above
+}
+impl NodeAddr {
+    pub(crate) fn inner(&self) -> SocketAddr {
+        match self {
+            NodeAddr::RealPeer(addr)
+            | NodeAddr::InitiallyKnown(addr)
+            | NodeAddr::FixedLocal(addr) => *addr,
+        }
```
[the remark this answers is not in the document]
wprzytula: `non_exhaustive` restricts some actions only beyond the defining crate.

scylla/src/transport/cluster.rs
```diff
@@ -60,7 +61,7 @@ pub struct Datacenter {
 
 #[derive(Clone)]
 pub struct ClusterData {
-    pub(crate) known_peers: HashMap<SocketAddr, Arc<Node>>, // Invariant: nonempty after Cluster::new()
+    pub(crate) known_peers: HashMap<NodeAddr, Arc<Node>>, // Invariant: nonempty after Cluster::new()
```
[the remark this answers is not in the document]
wprzytula: Do you propose releasing the constraint that, in order for a Node to be recycled, IPs need to match? I.e. now only Host ID, DC and rack must be the same to recycle a Node?

scylla/src/transport/cluster.rs
```diff
@@ -531,7 +532,7 @@ impl ClusterWorker {
     fn change_node_down_marker(&mut self, addr: SocketAddr, is_down: bool) {
         let cluster_data = self.cluster_data.load_full();
 
-        let node = match cluster_data.known_peers.get(&addr) {
+        let node = match cluster_data.known_peers.get(&NodeAddr::RealPeer(addr)) {
```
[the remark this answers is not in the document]
wprzytula: I've moved to the default `Eq` etc. impls.

scylla/src/transport/topology.rs
```diff
@@ -61,7 +61,6 @@ pub struct Peer {
 
 #[non_exhaustive] // <- so that we can add more fields in a backwards-compatible way
 pub struct UntranslatedPeer {
-    pub host_id: Uuid,
```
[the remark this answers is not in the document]
wprzytula: I've kept it back and cherry picked the commit introducing `host_id` into ConnectionPool from serverless PR.

scylla/src/transport/host_filter.rs
wprzytula: Does this preserve the HostFilter's behaviour regarding initially known nodes? Did the filter accept them all previously?

scylla/src/transport/connection_pool.rs
wprzytula: Do you find `QueryError` suitable for containing variants regarding _translating addresses on establishing connection to nodes_ ? For me, this is highly counterintuitive. I'm getting more and more convinced that it's high time we considered refactor of our errors.

wprzytula: After all, there wasn't any query attempted yet, and the user gets a `QueryError`.

scylla/src/transport/node.rs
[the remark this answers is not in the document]
piodul: I see. Makes sense.

scylla/src/transport/connection_pool.rs
piodul: > Do you find `QueryError` suitable for containing variants regarding _translating addresses on establishing connection to nodes_ ? For me, this is highly counterintuitive. I'm getting more and more convinced that it's high time we considered refactor of our errors.

I agree that our error types need redesign, but that is not something that we can fix in this PR.

In this case though it wouldn't hurt to introduce an OpenConnectionError or sth. The error is converted to a string [here](https://github.com/scylladb/scylla-rust-driver/blob/4964b4471c3d14bfea75e532b2e7c24cbd48cd65/scylla/src/transport/connection_pool.rs#L348-L356) (again, not ideal), but the new error won't appear in the API and will slightly improve the clarity of the internal code.

> After all, there wasn't any query attempted yet, and the user gets a `QueryError`.

To be fair, using QueryError to represent an error related to a failed connection is not perfect, either.

scylla/src/transport/host_filter.rs
piodul: Previously, the AllowListHostFilter was able to filter out initial nodes, but the driver would still be able to establish a control connection to one of them - if it happened, it would print a warning about it and switch the control connection to another node.

It's not really possible to preserve the semantics. Previously, AllowListHostFilter was always given a translated address, and now we don't do that - and we can't because the notion of a translated address makes less and less sense: node's address can change, multiple nodes can be behind a proxy, etc.

Maybe it's better to remove the AllowListHostFilter. It's not clear what its semantics should be. It's also quite simple for users to implement something like this if they need it and it will force them to think about how to handle different address types. Anyway, the main motivation for the host filter interface was introducing the DcHostFilter, which should still work.

scylla/src/transport/cluster.rs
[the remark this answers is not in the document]
piodul: > Do you propose releasing the constraint that, in order for a Node to be recycled, IPs need to match? I.e. now only Host ID, DC and rack must be the same to recycle a Node?

Hm, right, it would indeed be nice to update the IP field if it changes, so I think we should indeed create a new Node instance.

This leaves us with the problem of NodeConnectionPool being recreated. However, I realize now that this problem can be avoided:

1. Add `#[derive(Clone)]` for NodeConnectionPool. Most (if not all) fields are already cloneable, and if not we could put `refiller_handle` and `keepaliver_handle` into an Arc and it should work.
2. Make it possible to update the pool refiller with the new details about the node - in this case we only need to tell it about the new IP. We could do it with an ArcSwap or a tokio broadcast channel.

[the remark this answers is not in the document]
wprzytula: So we drop the important invariant of NodeConnectionPool and PoolRefiller, which is constant IP. In fact we have already decided to do so because of serverless proxy hostname resolution.

wprzytula:
v2:
- addressed all comments,
- considerable change in the last commits: when a `node_addr` changes, a new Node is created but the pool from the old Node is preserved and updated with the new address.

wprzytula:
Rebased on main.

wprzytula:
v3:
- refactored topology, connection_pool and connection layers so that they employ new `UntranslatedEndpoint` enum to carry the data necessary for opening new connections, including information whether to perform address translation or not.

wprzytula:
@piodul @havaker @cvybhu Ready for review.

scylla/src/transport/topology.rs
```diff
@@ -61,6 +61,66 @@ pub struct Peer {
...
+impl From<Peer> for PeerEndpoint {
+    fn from(peer: Peer) -> Self {
+        Self {
+            host_id: peer.host_id,
+            address: peer.address,
+            datacenter: peer.datacenter,
+            rack: peer.rack,
+        }
+    }
+}
```
piodul, line 112: Do we need to expose this conversion? Is it used anywhere? I'm not sure whether it is useful to convert a Peer by value into a PeerEndpoint.

I'd rather not introduce such impls unnecessarily as they are a part of the API.

scylla/src/transport/topology.rs
```diff
@@ -267,7 +268,7 @@ impl Metadata {
                 let token = ((id as u128) << 64) / initial_peers.len() as u128;
 
                 Peer {
-                    address: endpoint.address(),
+                    address: NodeAddr::RealPeer(endpoint.address()),
```
piodul, line 271: I don't think we can treat an initial address as a `RealPeer`. This function will be called with contact point addresses because it is only called on first metadata fetch, so those addresses shouldn't be translated. If address translation is enabled then, in case of a failed first metadata fetch, the session will not work anymore.

The semantics of `FixedLocal` are more appropriate here, though the name isn't. I propose to change the variant names of `NodeAddr` to something more generic - something that only refers to whether the contained address should be translated or not.

By the way, it's too easy to construct a wrong `NodeAddr` from an `UntranslatedEndpoint` via the `.address()` method.

scylla/src/transport/topology.rs
```diff
@@ -463,7 +464,7 @@ impl MetadataReader {
         let control_connection_peer = metadata
             .peers
             .iter()
-            .find(|peer| peer.address == self.control_connection_endpoint.address());
+            .find(|peer| peer.address.into_inner() == self.control_connection_endpoint.address());
```
piodul, line 467: I think this is wrong. We shouldn't compare SocketAddrs but rather NodeAddrs. We cannot consider `RealPeer(x)` to be equal to `FixedLocal(x)` for any `x`.

scylla/src/transport/topology.rs
wprzytula: Right, it's currently unused.

scylla/src/transport/topology.rs
wprzytula: I adjusted the names and usage as suggested. However, I can't see any sensible solution the to problem of creating erroneuously wrong `NodeAddr` via `.address()`.

scylla/src/transport/connection_pool.rs
```diff
@@ -203,6 +204,13 @@ impl NodeConnectionPool {
         }
     }
 
+    pub(crate) fn update_node_addr(&self, new_node_addr: NodeAddr) {
```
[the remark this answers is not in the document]
wprzytula: OK, it seems that any use case of `update_node_addr` should have access to a whole `Peer`, which imples `PeerEndpoint` as well. We could do so then. However, what bothers me is:
- (minor thing) pointless clones of DC and Rack ,
- (major thing) Pool's API is now not immune to alterations of node parameters other than only address. From the perspective of the Pool, it is now possible that different rack, host ID or DC are passed.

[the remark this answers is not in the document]
piodul: > * (minor thing) pointless clones of DC and Rack ,

This happens very infrequently so it's not a big problem.

> * (major thing) Pool's API is now not immune to alterations of node parameters other than only address. From the perspective of the Pool, it is now possible that different rack, host ID or DC are passed.

Why is it a major problem?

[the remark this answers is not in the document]
wprzytula: > Why is it a major problem?

Alterations (especially silent) of the above stated node properties are not something we designed ConnectionPool for.

scylla/src/transport/topology.rs
wprzytula: Do you have any ideas @piodul?

scylla/src/transport/connection_pool.rs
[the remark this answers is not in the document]
wprzytula: But OK, the Pool won't break in such a case, and we don't intend to cause such situation anyway, so let's enable complete endpoint updates.

scylla/src/transport/topology.rs
wprzytula: I've rewritten the `address()` method to return `NodeAddr` always. In case of a `ContactPoint`, a `NodeAddr::Untranslatable` is returned.

piodul (approved):

Issue #631, opened 2023-01-24
`execute_iter`/`query_iter` error on `UPDATE` or `INSERT` query.


0e4ef622:
It would be more reasonable if it returned an empty iterator. Instead, I get `ProtocolError("Unexpected response to next page query")`.

wprzytula:
@piodul WDYT?

piodul:
Although the `_iter` methods are not really meant for sending such queries, I think we change the behavior to what the issue proposes - I don't see a reason against this.

PR #644, opened 2023-02-06
iterator: Return empty iterator on modification stmt

wprzytula:
...instead of ProtocolError as before. This was requested by #631. Also, added a test for this behaviour.

Fixes: #631

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/iterator.rs
+++ b/scylla/src/transport/iterator.rs
@@ -540,6 +540,28 @@ where
                     self.metrics.inc_failed_paged_queries();
                     return Err(err);
                 }
+                Ok(NonErrorQueryResponse {
+                    response: NonErrorResponse::Result(_),
+                    tracing_id,
+                    ..
+                }) => {
+                    // We have most probably sent a modification statement (e.g. INSERT or UPDATE),
+                    // so let's return an empty iterator as suggested in #631.
+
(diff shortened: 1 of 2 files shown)


scylla/src/transport/iterator.rs
```diff
@@ -540,6 +540,27 @@ where
...
+                    // We must attempt to send something because the iterator expects it.
+                    let (proof, _) = self
+                        .sender
+                        .send(Ok(ReceivedPage {
+                            rows: Rows {
+                                metadata: Default::default(),
+                                rows_count: 0,
+                                rows: Vec::new(),
+                            },
+                            tracing_id: None,
```
piodul, line 559: `tracing_id` is available in the `NonErrorQueryResponse`, so let's use it.

piodul (commented):
One tiny issue (and a test for it would be nice), otherwise LGTM.

scylla/src/transport/iterator.rs
wprzytula: How do I test that? (as you mentioned in the general comment)

piodul: Run `query_iter` with tracing enabled (`Query::set_tracing`) and then verify that the list of tracing ids reported by the iterator is not empty.

Issue #462, opened 2022-06-24
Switch deserialization to an iterator-based interface to avoid allocations


piodul:
Currently, after receiving a `Rows` response from the database, the driver unconditionally parses the row data into an equivalent of `Vec<Vec<Option<CqlValue>>`. This results in at least `N + 1` allocations (where `N` is the number of rows) and possibly more if there are some composite `CqlValue`s in the result such as collections or UDTs.

The most common way to interpret the data - I suppose it is the most common way, at least - is to convert the rows into some typed representation before consuming them. For example, if the data is fetched from a table which has three `int` columns, then they can be converted to `(i32, i32, i32)`. Note that this representation contains all data in-line without any allocation, so `Vec<Option<CqlValue>>` is quite a costly intermediate representation.

I propose to cut the middleman out and adjust our deserialization framework to avoid allocations. Instead of eagerly materializing all rows into a vector, let's keep them in an unserialized form and introduce an iterator interface to avoid allocations.

Here are some ideas on how the new interface could look like:

```rust
pub struct QueryResult {
    pub rows: Option<RawRows>,
    // We will probably need to include information about the result type
    // ...
}

// RawRows represents raw, unserialized rows
struct RawRows {
    frame_data: Bytes,
    row_data_offset: usize,
    row_data_size: usize,
    // ...
}

impl RawRows {
    pub fn iter(&self) -> ResultIterator { /* ... */ }
}

// Iterates over rows in result
struct ResultIterator<'rows> { /* ... */ }

impl<'rows> Iterator for ResultIterator<'rows> {
    type Item = RowIterator<'rows>;
}

// Iterates over columns in a row
struct RowIterator<'row> { /* ... */ }

impl<'row> Iterator for RowIterator<'row> {
    type Item = &'row [u8];
}

// Helper iterators for consuming compound types
// We can consider typed variants of those, i.e. such that they automatically convert items before returning
struct SequenceIterator<'column> { /* ... */ } // Item = &'column [u8], could probably be used for lists and sets
struct MapIterator<'column> { /* ... */ } // Item = (&'column [u8], &'column [u8])
struct UDTIterator<'col_type, 'column> { /* ... */ } // Item = (&'col_type str, &'column [u8])
```

We should also introduce a new trait which allows deserializing values directly from a byte slice:

```rust
// The lifetime in the trait definition will allow us to return types which borrow from the data,
// will be useful e.g. for MapIterator
trait WipName<'cell> {
[...]

PR #645, opened 2023-02-17
Replace Connection::query_all with Connection::query_iter

piodul:
For internal queries that can return a large number of rows (topology info, schema) we use currently use `Connection::query_all` to perform them. That method fetches results of the query in multiple pages and then returns it as a single combined `QueryResult`. This PR replaces it with a new `Connection::query_iter` method which has semantics more similar to the iterator-based query methods from the public interface (`Session::query_iter`, `Session::execute_iter`).

The motivations for this change are as follows:

- Although the new methods will use slightly more memory if the query returns only one page (because of the need to spawn a new task/channels/etc.), it should consume less memory if multiple pages are fetched because rows are processed on-line as new pages appear and only 1-2 pages need to exist at a given time.
- The logic that combines multiple pages into a single `QueryResult` will be hard to translate to use the upcoming iterator-based deserialization interface (https://github.com/scylladb/scylla-rust-driver/issues/462). Although the old interface will still be available as a deprecated fallback, I would like to adapt those methods as a form of dogfooding test for the new interface.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -30,6 +30,7 @@ use std::{
 };
 
 use super::errors::{BadKeyspaceName, BadQuery, DbError, QueryError};
+use super::iterator::RowIterator;
 
 use crate::batch::{Batch, BatchStatement};
 use crate::frame::protocol_features::ProtocolFeatures;
@@ -626,6 +627,30 @@ impl Connection {
                 return Ok(final_result);
             }
         }
+    }
+
+    /// Executes a query and fetches its results over multiple pages, using
+    /// the asynchronous iterator interface.
+    pub(crate) async fn query_iter(
+        self: Arc<Self>,
+        query: Query,
+        values: impl ValueList,
+    ) -> Result<RowIterator, QueryError> {
+        let serialized_values = values.serialized()?.into_owned();
+
+        let consistency = query
+            .config
+            .determine_consistency(self.config.default_consistency);
+        let serial_consistency = query.config.serial_consistency.flatten();
+
+        RowIterator::new_for_connection_query_iter(
+            query,
+            self,
+            serialized_values,
+            consistency,
+            serial_consistency,
+        )
+        .await
     }
 
     #[allow(dead_code)]
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -521,35 +521,6 @@ impl Connection {
         }
     }
 
-    pub async fn execute_single_page(
-        &self,
-        prepared_statement: &PreparedStatement,
-        values: impl ValueList,
-        paging_state: Option<Bytes>,
-    ) -> Result<QueryResult, QueryError> {
-        self.execute(prepared_statement, values, paging_state)
-            .await?
-            .into_query_result()
-    }
-
-    pub async fn execute(
-        &self,
-        prepared_statement: &PreparedStatement,
-        values: impl ValueList,
-        paging_state: Option<Bytes>,
-    ) -> Result<QueryResponse, QueryError> {
-        self.execute_with_consistency(
-            prepared_statement,
-            values,
-            prepared_statement
-                .config
-                .determine_consistency(self.config.default_consistency),
-            prepared_statement.config.serial_consistency.flatten(),
-            paging_state,
-        )
-        .await
-    }
-
     pub async fn execute_with_consistency(
         &self,
         prepared_statement: &PreparedStatement,
@@ -589,43 +560,6 @@ impl Connection {
                     .await
             }
             _ => Ok(query_response),
-        }
-    }
-
-    /// Performs execute_single_page multiple times to fetch all available pages
-    #[allow(dead_code)]
-    pub async fn execute_all(
-        &self,
-        prepared_statement: &PreparedStatement,
-        values: impl ValueList,
-    ) -> Result<QueryResult, QueryError> {
-        if prepared_statement.get_page_size().is_none() {
-            return Err(QueryError::BadQuery(BadQuery::Other(
-                "Called Connection::execute_all without page size set!".to_string(),
-            )));
-        }
-
-        let mut final_result = QueryResult::default();
-
-        let serialized_values = values.serialized()?;
-        let mut paging_state: Option<Bytes> = None;
-
-        loop {
-            // Send next paged query
-            let mut cur_result: QueryResult = self
-                .execute_single_page(prepared_statement, &serialized_values, paging_state)
-                .await?;
-
-            // Set paging_state for the next query
-            paging_state = cur_result.paging_state.take();
-
-            // Add current query results to the final_result
-            final_result.merge_with_next_page_res(cur_result);
-
-            if paging_state.is_none() {
-                // No more pages to query, we can return the final result
-                return Ok(final_result);
-            }
         }
     }
 
@@ -1621,14 +1555,14 @@ mod tests {
         }
     }
 
-    /// Tests for Connection::query_all and Connection::execute_all
+    /// Tests for Connection::query_all
     /// 1. SELECT from an empty table.
     /// 2. Create table and insert ints 0..100.
-    ///    Then use query_all and execute_all with page_size set to 7 to select all 100 rows.
+    ///    Then use query_all with page_size set to 7 to select all 100 rows.
     /// 3. INSERT query_all should have None in result rows.
     /// 4. Calling query_all with a Query that doesn't have page_size set should result in an error.
     #[tokio::test]
-    async fn connection_query_all_execute_all_test() {
+    async fn connection_query_all_test() {
         let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
         let addr: SocketAddr = resolve_hostname(&uri).await;
 
@@ -1670,11 +1604,6 @@ mod tests {
         let empty_res = connection.query_all(&select_query, &[]).await.unwrap();
         assert!(empty_res.rows.unwrap().is_empty());
 
-        let mut prepared_select = connection.prepare(&select_query).await.unwrap();
-        prepared_select.set_page_size(7);
-        let empty_res_prepared = connection.execute_all(&prepared_select, &[]).await.unwrap();
-        assert!(empty_res_prepared.rows.unwrap().is_empty());
-
         // 2. Insert 100 and select using query_all with page_size 7
         let values: Vec<i32> = (0..100).collect();
         let mut insert_futures = Vec::new();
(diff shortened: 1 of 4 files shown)


piodul:
```
running 2 tests
test retries_occur ... ok
test speculative_execution_is_fired ... FAILED

failures:

---- speculative_execution_is_fired stdout ----
Unique name: test_rust_1676631776_1
thread '<unnamed>' panicked at 'Worker failed: Node 172.42.0.2:9042 disconnected', scylla/tests/retries.rs:95:21
thread 'speculative_execution_is_fired' panicked at 'timeout: the function call took 3772 ms. Max time 30000 ms', scylla/tests/retries.rs:20:1


failures:
    speculative_execution_is_fired

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.77s
```

Strange, sounds a little bit as if a node has crashed. It's hard to tell for sure because we don't print logs from the scylla docker containers at all.

I'll try to improve the CI so that it does print the logs from the cluster. In the meantime, I'll just re-run and hope that the error was an unimportant fluke.

scylla/src/transport/connection.rs
```diff
@@ -1559,8 +1558,7 @@ mod tests {
     /// 1. SELECT from an empty table.
     /// 2. Create table and insert ints 0..100.
     ///    Then use query_all with page_size set to 7 to select all 100 rows.
-    /// 3. INSERT query_all should have None in result rows.
-    /// 4. Calling query_all with a Query that doesn't have page_size set should result in an error.
+    /// 3. INSERT query_all should work and not return any rows.
```
wprzytula, line 1498: I can't see any commit replacing refs to `query_all` in comments with refs to `query_iter`.

scylla/src/transport/connection.rs
```diff
@@ -1559,8 +1558,7 @@ mod tests {
...
     #[tokio::test]
     async fn connection_query_all_test() {
```
wprzytula, line 1500: Ditto with the test name: shouldn't it change its to name to `connection_query_iter_test` ? I understand that it still tests querying _all_ entities of some kind, but the function we test here is `query_iter`, so we should rather keep the name consistent with the function tested.

wprzytula (changes requested):
Very good overall, but some old naming is still left.

scylla/src/transport/iterator.rs
```diff
@@ -683,6 +715,66 @@ where
...
+/// any complicated logic related to retries, it just fetches pages from
+/// a single connection.
+struct SingleConnectionRowIteratorWorker<Fetcher> {
+    sender: ProvingSender<Result<ReceivedPage, QueryError>>,
+    fetcher: Fetcher,
+}
+
+impl<Fetcher, FetchFut> SingleConnectionRowIteratorWorker<Fetcher>
+where
+    Fetcher: Fn(Option<Bytes>) -> FetchFut + Send + Sync,
```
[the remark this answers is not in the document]
piodul: Yes, it is conventional not to place constraints in type definitions unless they are absolutely necessary. I learned about this convention while learning Haskell, but when I usually look at the struct definitions in the Rust ecosystem I can see the same principle is used, too.

There are some good arguments in favor of this convention in this SO answer: https://stackoverflow.com/a/66369912

scylla/src/transport/topology.rs
```diff
@@ -510,42 +510,40 @@ async fn query_peers(
...
+        let (source, raw_row) = row_result?;
+        let row = raw_row.into_typed().map_err(|_| {
+            QueryError::ProtocolError("system.peers or system.local has invalid column type")
+        })?;
         create_peer_from_row(source, row, local_address, address_translator).await
     });
 
-    let peers = try_join_all(translated_peers_futures).await?;
+    let peers = translated_peers_futures
+        .buffer_unordered(64)
```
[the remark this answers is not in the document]
piodul: It's an intuitive guess, there is no deep reasoning behind it... The `buffer_unordered` method required me to choose some number.

This number will limit the number of concurrent address translations (previously it could be unbounded). I don't think it is a bad idea to limit it, though it usually won't matter as `64` is higher than most cluster sizes so it won't effectively change the old behavior for most users. I can add a comment that this number is just a guess.

Do you have a suggestion for a better number?

scylla/src/transport/connection.rs
piodul: Yeah, I just forgot to update the comments and the tests name. It doesn't make sense to have `query_all` anywhere in the repo at this point.

scylla/src/transport/topology.rs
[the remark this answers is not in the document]
wprzytula: I would aim to choose a number that would be an upper size bound for a vast majority of clusters. If it were me who were to guess, it would be 256, but 64 is IMO OK too.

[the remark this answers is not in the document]
piodul: All right, I changed to 256. I was a bit worried that it would increase memory usage, but the implementation doesn't seem to pre-allocate anything.

scylla/src/transport/connection.rs
piodul: Done (in the commit which adjusts the test to use query_iter).

scylla/src/transport/connection.rs
piodul: Fixed, I also replaced the `query_all` substrings in the names of the tables created during the test.

piodul:
v2:
- Changed the `buffer_unordered` parameter from `64` -> `256` to allow more concurrency when translating addresses,
- Updated some leftover comments/function names/string literals not to mention `query_all` but rather `query_iter`.

wprzytula (approved):

piodul:
@cvybhu ping

scylla/src/transport/iterator.rs
```diff
@@ -683,6 +715,66 @@ where
     }
 }
 
+/// A massively simplified version of the RowIteratorWorker. It does not have
+/// any complicated logic related to retries, it just fetches pages from
+/// a single connection.
```
cvybhu, line 720: It would be nice to use a `RetryPolicy` for retries. I think it wouldn't be too hard to incorporate it.

cvybhu (commented):
Code looks ok, left a few nits.

Given that, I'm not a fan of this change. The streams are very fancy, but IMO they are much harder to read and reason about. Previously there were a few clearly defined steps where one thing happened at a time and I could easily follow it step after step. With streams it feels different - it's like first we prepare what to do and then do everything at once.

With normal code there are usually a few intermediate variables with clearly defined types, but here all the chained stream combinators make it impossible to be sure about the type of things at each step. The error handling is also much more complicated, all the errors must be propagated in the stream, and then all the `try_something` functions hopefully handle them, but I find it very hard to reason about. Then there are also the complications with `.buffer_unordered(256)`

IMO it would be better to keep using the `query_all` function. It could be implemented using `query_iter`, although I guess it doesn't make much sense.
I'm a big fan of KISS.

I might be complaining too much, maybe streams are just a Rust thing that I'm not used to yet, but my initial feeling is that they introduce new complexity that could easily be avoided.

In the PR description you mentioned that it will be hard to keep `query_all` after the deserialization refactor, but why is that? Will it be impossible to simply collect the rows to a Vec?

piodul:
> Code looks ok, left a few nits.
> 
> Given that, I'm not a fan of this change. The streams are very fancy, but IMO they are much harder to read and reason about. Previously there were a few clearly defined steps where one thing happened at a time and I could easily follow it step after step. With streams it feels different - it's like first we prepare what to do and then do everything at once.

I assume that your comment is about the changes in `query_peers` from `topology.rs`.

I'm not sure if I agree. The biggest advantage of streams here is that they reduce the peak memory use. Instead of loading all results from a query into memory, we only need to keep 1-2 pages in memory at a given time. I'll admit that the benefit is not huge because we are constructing the metadata anyway, but it's not only about making the code look more "fancy".

> 
> With normal code there are usually a few intermediate variables with clearly defined types, but here all the chained stream combinators make it impossible to be sure about the type of things at each step. The error handling is also much more complicated, all the errors must be propagated in the stream, and then all the `try_something` functions hopefully handle them, but I find it very hard to reason about. Then there are also the complications with `.buffer_unordered(256)`

We are already using iterators extensively in the codebase, so I thought that streams would be familiar. They have a slightly different API due to their asynchronicity, but the concept is the same.

I'm not sure what do you mean with complications with `.buffer_unordered()`. Is it about the need to explicitly limit parallelism? Limiting parallelism is not a bad thing.

> 
> IMO it would be better to keep using the `query_all` function. It could be implemented using `query_iter`, although I guess it doesn't make much sense. I'm a big fan of KISS.

With `query_iter`, you can choose either to collect everything to a Vec and then process, or do the processing on the fly, as pages are fetched.

> 
> I might be complaining too much, maybe streams are just a Rust thing that I'm not used to yet, but my initial feeling is that they introduce new complexity that could easily be avoided.
> 
> In the PR description you mentioned that it will be hard to keep `query_all` after the deserialization refactor, but why is that? Will it be impossible to simply collect the rows to a Vec?

I guess it will be possible to change query_all to just return a Vec of parsed rows, not a full QueryResponse. I just felt that solving the problem in this way will prevent us from utilizing the new interface to its full potential and will deny us from testing its ergonomics on our internal code.

In general, the new deserialization interface will not return pre-parsed rows in the style of `Vec<Vec<Option<CqlValue>>>>`, but will return an iterator that lazily deserializes the contents of a response. If we wanted the `query_all` to keep returning QueryResponse, then we would have to concatenate the contents of the pages into a large blob, which is even worse with regards to memory use.

scylla/src/transport/iterator.rs
piodul: I'm not sure whether it makes sense for queries that are limited to a single connection and only read data from a single node. For example: what should happen on `RetryNextNode`?

scylla/src/transport/iterator.rs
```diff
@@ -265,6 +266,37 @@ impl RowIterator {
         Self::new_from_worker_future(worker_task, receiver).await
     }
 
+    pub(crate) async fn new_for_query_iter(
```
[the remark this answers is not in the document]
piodul: Good point, changed to `new_for_connection_query_iter`.

scylla/src/transport/iterator.rs
```diff
@@ -265,6 +266,37 @@ impl RowIterator {
...
+        }
+        let (sender, receiver) = mpsc::channel::<Result<ReceivedPage, QueryError>>(1);
+
+        let worker_task = async move {
+            let worker = SingleConnectionRowIteratorWorker {
+                sender: sender.into(),
+                fetcher: |paging_state| {
+                    connection.query_with_consistency(
+                        &query,
+                        values.clone(),
```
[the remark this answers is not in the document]
piodul: You are right, it wasn't needed. Changed.

piodul:
v3:
- renamed `new_for_query_iter` -> `new_for_connection_query_iter`
- got rid of an unnecessary `.clone()` in a lambda defined in `RowIterator::new_for_connection_query_iter`

piodul:
@cvybhu review ping

cvybhu (approved):

PR #646, opened 2023-02-17
CI: print logs from the cluster as a build step

piodul:
Some test failures may be caused due to issues in Cassandra or Scylla and it is hard to identify them if they are silent during the whole CI run.

This PR modifies our CI jobs that use docker-compose so that the logs from the cluster are always printed.

## Pre-review checklist

### DIFF
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -24,5 +24,11 @@ jobs:
     - name: Run tests on cassandra
       run: |
         CDC='disabled' SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose -- --skip test_views_in_schema_info
-    - name: Stop cluster
+    - name: Stop the cluster
+      if: ${{ always() }}
+      run: docker-compose -f test/cluster/cassandra/docker-compose.yml stop
+    - name: Print the cluster logs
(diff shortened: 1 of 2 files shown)


piodul:
It looks like printing the logs after failure works correctly, e.g. here: https://github.com/scylladb/scylla-rust-driver/actions/runs/4204000245/jobs/7294090832

.github/workflows/cassandra.yml
```diff
@@ -24,5 +24,11 @@ jobs:
     - name: Run tests on cassandra
       run: |
         CDC='disabled' SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose -- --skip test_views_in_schema_info
-    - name: Stop cluster
+    - name: Stop the cluster
+      if: ${{ always() }}
+      run: docker-compose -f test/cluster/cassandra/docker-compose.yml stop
```
havaker, line 29: Why did you add the `stop` command? Was this to see all the cluster logs, including those produced during Scylla shutdowns?

piodul: Yes. I can't do it after `docker-compose shutdown` because it deletes the containers along with the logs.

havaker (approved):

piodul:
@cvybhu ping

cvybhu (approved):

PR #647, opened 2023-02-17
docker-compose.yml: improve settings and the boot order

piodul:
Up until now, our docker-compose file that sets up Scylla cluster was starting up all nodes in parallel. This is not something that is supported in Scylla and it can sometimes fail. We didn't see those failures until now because we didn't print the cluster logs. This commit makes sure that nodes are booted up in order.

In addition to this, now nodes start with a fixed number of shards (2 each) and memory (1GB each). I'm not sure a more powerful cluster is really needed, and - by default - the shard count is equal to the number of cores of the host machine. A large number of shards can quickly eat up the /proc/sys/fs/aio-nr resource in the system, so the cluster wasn't really usable on machines with a large number of cores.

## Pre-review checklist

### DIFF
--- a/test/cluster/docker-compose.yml
+++ b/test/cluster/docker-compose.yml
@@ -14,16 +14,58 @@ services:
     networks:
       public:
         ipv4_address: 172.42.0.2
-    command: --rpc-address 172.42.0.2 --listen-address 172.42.0.2 --seeds 172.42.0.2 --skip-wait-for-gossip-to-settle 0
+    command: |
+      --rpc-address 172.42.0.2
+      --listen-address 172.42.0.2
+      --seeds 172.42.0.2
+      --skip-wait-for-gossip-to-settle 0
+      --ring-delay-ms 0
+      --smp 2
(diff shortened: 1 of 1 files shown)


piodul:
It looks like the CI runner has only two cores and Scylla refuses to boot with more shards than there are cores - even though Scylla inside the container is being run with the `--overprovisioned` flag. I'll reduce the number of cores.

piodul:
Note: I based this PR's branch on https://github.com/scylladb/scylla-rust-driver/pull/646 , that PR should probably go first.

wprzytula (approved):

piodul:
@cvybhu ping (after https://github.com/scylladb/scylla-rust-driver/pull/646 is merged)

cvybhu (approved):
LGTM, #646 got merged, please rebase onto main.

piodul:
> LGTM, #646 got merged, please rebase onto main.

Thanks, done

cvybhu (approved):

PR #649, opened 2023-02-22
topology: UDTs hold complete definition of themselves

wprzytula:
It has been requested by a customer to make `CqlType::UserDefinedType` hold a complete definition of the type. 

Before, the definition was distributed among multiple code locations: the type name was present inside the mentioned variant, the keyspace name was to be deduced from the context, and the fields definitions were only stored in a hash map in each `Keyspace`.

This PR:
- introduces `UserDefinedType` struct as a complete UDT definition,
- makes the `Keyspace` hold the hash map from UDT name to  `Arc<UserDefinedType>`,
- makes `CqlType::UserDefinedType` store a complete definition of a UDT (`Result<Arc<UserDefinedType>, MissingUserDefinedType>`, the `Err` variant is used in case when an unknown UDT is defined as a field of another UDT),
- introduces `UdtRow` struct to represent a row with UDT definition,
- adds a topological sort of `UdtRow` to avoid processing references to unknown UDTs. The topological order is always achievable, as UDTs circular dependencies are forbidden and rejected by Scylla.

This PR is based on #645, because it is very likely that it will be merged soon and otherwise it would introduce a lot of merge conflicts.

## Pre-review checklist

- ~~[ ] I added appropriate `Fixes:` annotations to PR description~~.

### DIFF
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -1312,6 +1312,66 @@ async fn test_prepared_config() {
     assert_eq!(prepared_statement.get_page_size(), Some(42));
 }
 
+fn udt_type_a_def(ks: &str) -> Vec<(String, CqlType)> {
+    vec![
+        (
+            "a".to_string(),
+            CqlType::Collection {
+                frozen: false,
+                type_: CollectionType::Map(
+                    Box::new(CqlType::Collection {
+                        frozen: true,
+                        type_: CollectionType::List(Box::new(CqlType::Native(NativeType::Int))),
+                    }),
+                    Box::new(CqlType::Native(NativeType::Text)),
+                ),
+            },
+        ),
+        (
+            "b".to_string(),
+            CqlType::Collection {
+                frozen: true,
+                type_: CollectionType::Map(
+                    Box::new(CqlType::Collection {
+                        frozen: true,
+                        type_: CollectionType::List(Box::new(CqlType::Native(NativeType::Int))),
+                    }),
+                    Box::new(CqlType::Collection {
+                        frozen: true,
+                        type_: CollectionType::Set(Box::new(CqlType::Native(NativeType::Text))),
+                    }),
+                ),
+            },
+        ),
+    ]
+}
+
+fn udt_type_b_def(ks: &str) -> Vec<(String, CqlType)> {
+    vec![
+        ("a".to_string(), CqlType::Native(NativeType::Int)),
+        ("b".to_string(), CqlType::Native(NativeType::Text)),
+    ]
+}
+
+fn udt_type_c_def(ks: &str) -> Vec<(String, CqlType)> {
+    vec![(
+        "a".to_string(),
+        CqlType::Collection {
+            frozen: false,
+            type_: CollectionType::Map(
+                Box::new(CqlType::Collection {
+                    frozen: true,
+                    type_: CollectionType::Set(Box::new(CqlType::Native(NativeType::Text))),
+                }),
+                Box::new(CqlType::UserDefinedType {
+                    frozen: true,
+                    name: "type_b".to_string(),
+                }),
+            ),
+        },
+    )]
+}
+
 #[tokio::test]
 async fn test_schema_types_in_metadata() {
     let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
@@ -1398,7 +1458,7 @@ async fn test_schema_types_in_metadata() {
         a.type_,
         CqlType::UserDefinedType {
             name: "type_a".to_string(),
-            frozen: true
+            frozen: true,
         }
     );
 
@@ -1518,72 +1578,15 @@ async fn test_user_defined_types_in_metadata() {
 
     let type_a = &user_defined_types["type_a"];
 
-    assert_eq!(
-        type_a,
-        &vec![
-            (
-                "a".to_string(),
-                CqlType::Collection {
-                    frozen: false,
-                    type_: CollectionType::Map(
-                        Box::new(CqlType::Collection {
-                            frozen: true,
-                            type_: CollectionType::List(Box::new(CqlType::Native(NativeType::Int)))
-                        }),
-                        Box::new(CqlType::Native(NativeType::Text))
-                    )
-                }
-            ),
-            (
-                "b".to_string(),
-                CqlType::Collection {
-                    frozen: true,
-                    type_: CollectionType::Map(
-                        Box::new(CqlType::Collection {
-                            frozen: true,
-                            type_: CollectionType::List(Box::new(CqlType::Native(NativeType::Int)))
-                        }),
-                        Box::new(CqlType::Collection {
-                            frozen: true,
-                            type_: CollectionType::Set(Box::new(CqlType::Native(NativeType::Text)))
-                        })
-                    )
-                }
-            )
-        ]
-    );
+    assert_eq!(*type_a, udt_type_a_def(&ks));
 
     let type_b = &user_defined_types["type_b"];
 
-    assert_eq!(
-        type_b,
-        &vec![
-            ("a".to_string(), CqlType::Native(NativeType::Int)),
-            ("b".to_string(), CqlType::Native(NativeType::Text))
-        ]
-    );
+    assert_eq!(*type_b, udt_type_b_def(&ks));
 
     let type_c = &user_defined_types["type_c"];
 
-    assert_eq!(
-        type_c,
-        &vec![(
-            "a".to_string(),
-            CqlType::Collection {
-                frozen: false,
-                type_: CollectionType::Map(
-                    Box::new(CqlType::Collection {
-                        frozen: true,
-                        type_: CollectionType::Set(Box::new(CqlType::Native(NativeType::Text)))
-                    }),
-                    Box::new(CqlType::UserDefinedType {
-                        frozen: true,
-                        name: "type_b".to_string()
-                    })
-                )
-            }
-        )]
-    );
+    assert_eq!(*type_c, udt_type_c_def(&ks));
 }
 
 #[tokio::test]
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -11,7 +11,9 @@ use crate::tracing::{GetTracingConfig, TracingInfo};
 use crate::transport::errors::{BadKeyspaceName, BadQuery, DbError, QueryError};
 use crate::transport::partitioner::{Murmur3Partitioner, Partitioner, PartitionerName};
 use crate::transport::topology::Strategy::SimpleStrategy;
-use crate::transport::topology::{CollectionType, ColumnKind, CqlType, NativeType};
+use crate::transport::topology::{
+    CollectionType, ColumnKind, CqlType, NativeType, UserDefinedType,
+};
 use crate::utils::test_utils::{supports_feature, unique_keyspace_name};
 use crate::CachingSession;
 use crate::ExecutionProfile;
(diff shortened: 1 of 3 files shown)


scylla/src/transport/topology.rs
```diff
@@ -870,6 +871,106 @@ async fn query_user_defined_types(
...
+            let mut decrement_referred_udts = |type_name: &str| {
+                *indegree
+                    .get_mut(&(new_leaf.keyspace_name.clone(), type_name.to_string()))
+                    .unwrap()
+                    .1
+                    .get_mut() -= 1;
+            };
+            // For each node that the currently removed node was dependant upon...
+            for field_type in new_leaf.field_types.iter() {
+                let pre_cql_type = map_string_to_cql_type(field_type);
```
piodul, line 964: Ouch, we are parsing the type here? Can't `topo_sort_udts` obtain a list of already parsed UDTs so that parsing multiple times is not necessary?

scylla/src/transport/topology.rs
```diff
@@ -870,6 +871,106 @@ async fn query_user_defined_types(
...
+                if let Ok(type_) = pre_cql_type {
+                    do_with_referenced_udts(&mut decrement_referred_udts, &type_);
+                }
+            }
+        }
+    }
+
+    leaves_first.reverse();
+    mem::swap(&mut leaves_first, udts);
+}
```
piodul, line 974: I'm not happy that the toposort algorithm is quadratic.

I think you are trying too hard to remove values from the map soon after their counter is decremented to zero. I don't think you have to do that - you can just operate on the hashmap keys.

Let's try this: keep doing what you did before up until line `925`, then something like this should follow (didn't compile it):

```rust
let mut sorted = Vec::with_capacity(indegree.len());
let mut next_idx = 0;

// Schedule keys that had an initial indeg of 0
for (key, (_, deg)) in indegree.iter() {
    if deg.get() == 0 {
        scheduled.push(&key);
    }
}

while let Some(key) = sorted.get(next_idx) {
    next_idx += 1;
    // Decrement the counters of all UDTs that this UDT depends upon
    // and then schedule them if their counter drops to 0
    let mut decrement_referred_udts = |type_name: &str| {
        let cnt = &indegree
            .get(key)
            .unwrap()
            .1;
        let new_cnt = cnt.get() - 1;
        cnt.set(new_cnt);
        if new_cnt == 0 {
            scheduled.push(key);
        }
    };
    let def = indegs.get(key).unwrap().0;
    for field_type in def.field_types.iter() {
        // get rid of parsing please
        if let Ok(type_) = map_string_to_cql_type(field_type) {
            do_with_referenced_udts(&mut decrement_referred_udts, &type_);
        }
    }
}
```

scylla/src/transport/topology.rs
```diff
@@ -971,6 +971,93 @@ fn topo_sort_udts(udts: &mut Vec<UdtRow>) {
     mem::swap(&mut leaves_first, udts);
 }
 
+#[test]
+#[ntest::timeout(100)]
```
piodul, line 977: Just in case, can we increase it to e.g. 1 second?

piodul (changes requested):

scylla/src/transport/topology.rs
```diff
@@ -1271,76 +1271,76 @@ mod tests {
     #[test]
     fn test_cql_type_parsing() {
         let test_cases = [
-            ("bigint", CqlType::Native(NativeType::BigInt)),
+            ("bigint", PreCqlType::Native(NativeType::BigInt)),
```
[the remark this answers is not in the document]
wprzytula: I reasoned this way: the whole commit is rather big, and I can't reasonably make it smaller. However, I can make it compile without tests in one commit, and fix tests (so they compile too) in the second.
I will squash them anyway.

scylla/src/transport/topology.rs
```diff
@@ -870,6 +871,106 @@ async fn query_user_defined_types(
...
+            } else {
+                indegree_dual.insert(key, (def, deg));
+            }
+        }
+
+        if leaves_first.len() == extend_beg {
+            panic!(
+                "Bug in fetched User Defined Types definitions: circular type dependency detected.
+                 Topological sort is thus impossible. Panicking to prevent an infinite loop."
+            )
```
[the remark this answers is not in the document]
wprzytula: What kind of error do you suggest? I would use `QueryError::ProtocolError` or `QueryError::InvalidMessage`.

scylla/src/transport/topology.rs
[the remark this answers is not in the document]
piodul: It is a good practice to make sure that all commits compile and pass tests - it's one of the points in the pre-review checklist (which you ticked off, by the way). This helps keep the project bisectable.

scylla/src/transport/topology.rs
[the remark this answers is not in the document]
piodul: Let's use `ProtocolError` here.

scylla/src/transport/topology.rs
wprzytula: Are `sorted` and `scheduled` the same?

piodul: > Are `sorted` and `scheduled` the same?

Yes, they should be the same. Updated the snippet.

wprzytula:
v2:
- addressed all @piodul's comments.

scylla/src/transport/topology.rs
```diff
@@ -902,6 +903,91 @@ async fn query_user_defined_types(
...
+                (def, Cell::new(0u32)),
+            )
+        })
+        .collect::<HashMap<_, _>>();
+
+    // For each node in the graph...
+    for (def, _) in indeg.values() {
+        let mut increment_referred_udts = |type_name: &str| {
+            let deg_cell = indeg
+                .get(&(def.keyspace_name.clone(), type_name.to_string()))
```
piodul, line 943: It sucks a bit that we have to clone the strings just to perform a lookup.

As far as I know, UDTs cannot refer to UDTs from other keyspaces, so probably we could just operate on a single keyspace at a time and only look up by type name. However, this would require making sure that `udts` is grouped by keyspace name - it probably is already, returned from the db, but I wouldn't like to depend on this.

Without some benchmarks it's probably not worth optimizing it as much at this point.

scylla/src/transport/topology.rs
```diff
@@ -902,6 +903,91 @@ async fn query_user_defined_types(
...
+        }
+    }
+
+    let owned_sorted = sorted.into_iter().cloned().collect::<Vec<_>>();
+    assert!(udts.is_empty());
+    for key in owned_sorted.into_iter().rev() {
+        udts.push(indeg.remove(&key).unwrap().0);
+    }
+
+    Ok(())
```
piodul, line 990: Do we detect reference cycles after all?

scylla/src/transport/topology.rs
```diff
@@ -988,6 +988,103 @@ fn topo_sort_udts(udts: &mut Vec<UdtRowWithParsedFieldTypes>) -> Result<(), Quer
...
+    assert!(get_udt_idx(&udts, KEYSPACE1, "B") < get_udt_idx(&udts, KEYSPACE1, "C"));
+    assert!(get_udt_idx(&udts, KEYSPACE1, "B") < get_udt_idx(&udts, KEYSPACE1, "F"));
+    assert!(get_udt_idx(&udts, KEYSPACE1, "F") < get_udt_idx(&udts, KEYSPACE1, "C"));
+    assert!(get_udt_idx(&udts, KEYSPACE1, "D") < get_udt_idx(&udts, KEYSPACE1, "B"));
+
+    assert!(get_udt_idx(&udts, KEYSPACE2, "B") < get_udt_idx(&udts, KEYSPACE2, "A"));
+    assert!(get_udt_idx(&udts, KEYSPACE2, "D") < get_udt_idx(&udts, KEYSPACE2, "B"));
+    assert!(get_udt_idx(&udts, KEYSPACE2, "D") < get_udt_idx(&udts, KEYSPACE2, "E"));
+    assert!(get_udt_idx(&udts, KEYSPACE2, "E") < get_udt_idx(&udts, KEYSPACE2, "A"));
+    assert!(get_udt_idx(&udts, KEYSPACE2, "A") < get_udt_idx(&udts, KEYSPACE2, "C"));
```
piodul, line 1087: Please add a test for detection of reference cycles.

scylla/src/transport/session_test.rs
```diff
@@ -1397,18 +1399,67 @@ async fn test_schema_types_in_metadata() {
     assert_eq!(
         a.type_,
         CqlType::UserDefinedType {
-            name: "type_a".to_string(),
-            frozen: true
+            frozen: true,
+            definition: Ok(Arc::new(UserDefinedType {
```
piodul, line 1403: This is the definition of type "type_a`, and I see that you provided it in two places - here and at line 1595. I think you should only define it once as a variable and then use it in both places.

scylla/src/transport/topology.rs
```diff
@@ -873,6 +873,8 @@ async fn query_user_defined_types(
         .try_collect()
         .await?;
 
+    topo_sort_udts(&mut udt_rows)?;
```
piodul, line 876: Small request: please measure the time it takes to toposort the UDT definitions and print it with `debug` level.

scylla/src/transport/topology.rs
```diff
@@ -902,6 +903,91 @@ async fn query_user_defined_types(
...
+        })
+        .collect::<HashMap<_, _>>();
+
+    // For each node in the graph...
+    for (def, _) in indeg.values() {
+        let mut increment_referred_udts = |type_name: &str| {
+            let deg_cell = indeg
+                .get(&(def.keyspace_name.clone(), type_name.to_string()))
+                .map(|(_, count)| count)
+                .unwrap();
```
piodul, line 945: Hmm, what if the metadata is inconsistent and some UDT definitions are missing? This will probably crash the application.

If the dependent UDT is not present in the metadata, then we should just ignore it and not increment anything.

Please add a test which verifies that `topo_sort_udts` correctly handles inconsistent metadata.

scylla/src/transport/topology.rs
```diff
@@ -902,6 +903,91 @@ async fn query_user_defined_types(
...
+    while let Some(key @ (keyspace, _type_name)) = sorted.get(next_idx).copied() {
+        next_idx += 1;
+        // Decrement the counters of all UDTs that this UDT depends upon
+        // and then schedule them if their counter drops to 0
+        let mut decrement_referred_udts = |type_name: &str| {
+            let ref_key = indeg
+                .get_key_value(&(keyspace.clone(), type_name.to_string()))
+                .unwrap()
+                .0;
+            let cnt = &indeg.get(ref_key).unwrap().1;
```
piodul, line 971: Why are we looking up twice in indeg? The first lookup doesn't seem to do anything useful (looking up a key with the same value as the provided key).

piodul (changes requested):
There are some issues with respect to robustness of `topo_sort_udts` - which, fortunately, shouldn't be difficult to fix. Please also add tests for cases with inconsistent metadata and reference cycles.

scylla/src/transport/topology.rs
wprzytula: This is a plain leftover.

wprzytula: The reason for that is in line 975:
`sorted.push(ref_key);`
In order to push a value to a vec, the lifetime of the value must be at least as long as the lifetime of the very vec; if we create a temporary stack tuple and store a reference to it in a vec, then we end up with a dangling reference (which, of course, the borrow checker wisely refuses). For acquring a valid reference with a long-enough lifetime, we call `get_key_value()` on `indeg` hashmap, yielding a reference (to the key) that lives as long as the whole hashmap, which is enough for the reference to be accepted as a vec's value.

In short, by performing `get_key_value()` we acquire a reference to the same key that we can have on the stack, but one with longer lifetime.

scylla/src/transport/topology.rs
wprzytula: Good catch, it got lost somewhere in the middle of a refactor.

scylla/src/transport/topology.rs
wprzytula: Actually, what we **can** do is getting rid of the _second_ lookup (as we already get both the key and the value in the first one), which I change atm.

wprzytula:
v3:
- made toposort more robust,
- avoided duplication in `session_test.rs`.

scylla/src/transport/topology.rs
piodul: Oh, right, `sorted` is a vector of references that point inside `indeg`, hence we need to look up the key, even if we know how the key actually looks like.

piodul (changes requested):
The implementation looks good, but it looks like the commits don't compile starting from `topology: Keyspace: UDTs held as complete defs`, and the first commit that compiles after that is `session_test: Adjust tests for new UDT form`. It looks like some tests need updating after some of the commits.

Please make sure that the individual commits compile (`cargo check --all-targets`) and pass tests (`make test`).

piodul (approved):

PR #663, opened 2023-03-17
session: add get_keyspace() method

piodul:
Adds the `get_keyspace()` method which allows to get the keyspace name currently set for the session. Initially, it is set to the keyspace name specified in the session configuration as the keyspace to be USE-d, but it will be updated when calling `Session::use_keyspace()` later. Because of that possibility, `get_keyspace` returns a shared pointer to the keyspace name instead of a reference.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -7,6 +7,7 @@ use crate::cloud::CloudConfig;
 use crate::frame::types::LegacyConsistency;
 use crate::history;
 use crate::history::HistoryListener;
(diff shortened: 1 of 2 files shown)


wprzytula (approved):

piodul:
v2: rebased on main to include a fix for the new clippy lints

piodul:
v3: rebased on current main

piodul:
I'll merge. The versions that were pushed after @wprzytula's approval were only rebases and didn't have any other changes.

PR #664, opened 2023-03-17
treewide: appease clippy

piodul:
The usual: clippy have been updated in GitHub's CI runner image that we use and it has new lints that our codebase doesn't currently pass.

This time, clippy complains about some `impl Default`s being unnecessarily hand-written where they could be easily implemented with an automatic derive. This commit removes the problematic impl for PartitionerName and replaces it with a derive, but the impl for Consistency is left as is and marked with an "allow" annotation. Unfortunately, implementing Default with a derive requires adding a #[default] annotation in the enum body which incorrectly interacts with num_enum::TryFromPrimitive macro.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/types.rs
+++ b/scylla-cql/src/frame/types.rs
@@ -46,6 +46,16 @@ pub enum LegacyConsistency {
     Serial(SerialConsistency),
 }
 
+// Although we could `impl Default for Consistency` with an automatic derive,
+// this would require adding a #[default] annotation on the default variant,
+// but - unfortunately - that annotation is also recognized by `TryFromPrimitive`
+// derive macro. If there is a #[default] variant then `TryFromPrimitive`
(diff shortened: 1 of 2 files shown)


havaker (approved):

havaker:
Oops, It looks like deriving `Default` and `TryFromPrimitive` interact in a weird way, failing `frame::types::type_consistency` test?

piodul:
> Oops, It looks like `Default` and `TryFromPrimitive` derives interact in a weird way, failing `frame::types::type_consistency` test?

Yes, `TryFromPrimitive` unnecessarily interacts with the `#[default]` annotation. For now I just marked the impl for Consistency with `allow`, but it will be possible to fix this properly after https://github.com/illicitonion/num_enum/issues/75 is fixed.

wprzytula (approved):

Issue #604, opened 2022-11-25
Rack-aware load balancing


piodul:
Current cloud providers tend to organize their resources in a two-tier hierarchy: physical datacenters are called _availability zones_ which are grouped into _regions_. Scylla/Cassandra snitches for AWS and GCE translate those terms as follows:

- _availability zone_ -> __rack__
- _region_ -> __DC__

We already have a DC aware policy which routes requests to the local DC first. However, in this case a __DC__ is not actually a physical datacenter and may consist of multiple _availability zones_. Traffic between _availability zones_ is expensive, both in terms of latency and money.

We should let users keep their traffic within the local _availability zone_, i.e. __rack__. There are two aspects of this:

- Enable our load balancing to prefer the local rack. In our current composable policy design we could introduce a "rack aware policy", similar to our current DC aware policy.
- Allow to prevent the driver from connecting to nodes from other racks. We already have the HostFilter mechanism and it should just be a matter of providing an implementation which filters by DC and rack.

References:
- https://github.com/gocql/gocql/pull/1565
- https://github.com/gocql/gocql/pull/1659

avelanarius:
Yes, I’ll assign somebody, but the customer said that gocql is their priority (this part assigned to @Lorak-mmk to merge the upstream to our fork). The customer also mentioned the C++ driver, where we have two options:
- C++ bindings which also require https://github.com/scylladb/scylla-rust-driver/issues/462 for the bindings to be able to get “raw bytes” unparsed response from driver (required by the customer)
- C++ driver, requiring some help from @jul-stas for the release process

mykaul:
@avelanarius - is this complete for 0.8? If not, please re-target to 0.8.1 or 0.9.

PR #666, opened 2023-03-20
Add rack-awareness to default LBP

Gor027:
Added rack-awareness option to prefer replicas in local rack while picking a coordinator. It will allow to create a top-level tier with local datacenter and local rack replicas in the fallback plan.

Fixes: #604 

## Pre-review checklist

### DIFF
--- a/docs/source/load-balancing/default-policy.md
+++ b/docs/source/load-balancing/default-policy.md
@@ -11,6 +11,7 @@ for queries with non-local consistency mode is also supported.
 `DefaultPolicyBuilder` with the following default values:
 
 - `preferred_datacenter`: `None`
+- `preferred_rack`: `None`
 - `is_token_aware`: `true`
 - `permit_dc_failover`: `false`
 - `latency_awareness`: `None`
@@ -25,6 +26,7 @@ use scylla::load_balancing::DefaultPolicy;
 
 let default_policy = DefaultPolicy::builder()
         .prefer_datacenter("dc1".to_string())
+        .prefer_rack("rack1".to_string())
         .token_aware(true)
         .permit_dc_failover(true)
         .build();
@@ -47,6 +49,14 @@ When datacenter failover is disabled (`permit_dc_failover` is set to
 false), the default policy will only include local nodes in load balancing
 plans. Remote nodes will be excluded, even if they are alive and available to
 serve requests.
+
+#### Preferred Rack
+
+The `preferred_rack` field in `DefaultPolicy` allows the load balancing policy to
+prioritize nodes based on their availability zones in the preferred datacenter.
+When a preferred rack is set, the policy will first return replicas in the local rack
+in the preferred datacenter, and then the other replicas in the datacenter.
+When a preferred datacenter is not set, setting preferred rack will not have any effect.
 
 #### Datacenter Failover
 
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -13,6 +13,13 @@ use scylla_cql::{errors::QueryError, frame::types::SerialConsistency, Consistenc
 use std::{fmt, sync::Arc, time::Duration};
 use tracing::warn;
 
+#[derive(Clone, Copy)]
+enum ReplicaLocation {
+    Any,
+    Datacenter,
+    DatacenterAndRack,
+}
+
 // TODO: LWT optimisation
 /// The default load balancing policy.
 ///
@@ -62,9 +69,29 @@ or refrain from preferring datacenters (which may ban all other datacenters, if 
             }
         }
         if let Some(ts) = &routing_info.token_with_strategy {
+            // Try to pick some alive local rack random replica.
+            // If preferred rack is not specified, try to pick local DC replica.
+            if self.preferred_rack.is_some() {
+                let local_rack_picked = self.pick_replica(
+                    ts,
+                    ReplicaLocation::DatacenterAndRack,
+                    &self.pick_predicate,
+                    cluster,
+                );
+
+                if let Some(alive_local_rack_replica) = local_rack_picked {
+                    return Some(alive_local_rack_replica);
+                }
+            }
+
             // Try to pick some alive local random replica.
             // If preferred datacenter is not specified, all replicas are treated as local.
-            let picked = self.pick_replica(ts, true, &self.pick_predicate, cluster);
+            let picked = self.pick_replica(
+                ts,
+                ReplicaLocation::Datacenter,
+                &self.pick_predicate,
+                cluster,
+            );
 
             if let Some(alive_local_replica) = picked {
                 return Some(alive_local_replica);
@@ -72,7 +99,8 @@ or refrain from preferring datacenters (which may ban all other datacenters, if 
 
             // If datacenter failover is possible, loosen restriction about locality.
             if self.is_datacenter_failover_possible(&routing_info) {
-                let picked = self.pick_replica(ts, false, &self.pick_predicate, cluster);
+                let picked =
+                    self.pick_replica(ts, ReplicaLocation::Any, &self.pick_predicate, cluster);
                 if let Some(alive_remote_replica) = picked {
                     return Some(alive_remote_replica);
                 }
@@ -125,11 +153,19 @@ or refrain from preferring datacenters (which may ban all other datacenters, if 
 
         // If token is available, get a shuffled list of alive replicas.
         let maybe_replicas = if let Some(ts) = &routing_info.token_with_strategy {
-            let local_replicas = self.shuffled_replicas(ts, true, Self::is_alive, cluster);
+            let local_rack_replicas = self.shuffled_replicas(
+                ts,
+                ReplicaLocation::DatacenterAndRack,
+                Self::is_alive,
+                cluster,
+            );
+            let local_replicas =
+                self.shuffled_replicas(ts, ReplicaLocation::Datacenter, Self::is_alive, cluster);
 
             // If a datacenter failover is possible, loosen restriction about locality.
             let maybe_remote_replicas = if self.is_datacenter_failover_possible(&routing_info) {
-                let remote_replicas = self.shuffled_replicas(ts, false, Self::is_alive, cluster);
+                let remote_replicas =
+                    self.shuffled_replicas(ts, ReplicaLocation::Any, Self::is_alive, cluster);
                 Either::Left(remote_replicas)
             } else {
                 Either::Right(std::iter::empty())
@@ -137,7 +173,11 @@ or refrain from preferring datacenters (which may ban all other datacenters, if 
 
             // Produce an iterator, prioritizes local replicas.
             // If preferred datacenter is not specified, every replica is treated as a local one.
-            Either::Left(local_replicas.chain(maybe_remote_replicas))
+            Either::Left(
+                local_rack_replicas
+                    .chain(local_replicas)
+                    .chain(maybe_remote_replicas),
+            )
         } else {
             Either::Right(std::iter::empty::<NodeRef<'a>>())
         };
@@ -265,39 +305,59 @@ impl DefaultPolicy {
     fn replicas<'a>(
         &'a self,
         ts: &TokenWithStrategy<'a>,
-        should_be_local: bool,
+        replica_location: ReplicaLocation,
         predicate: impl Fn(&NodeRef<'a>) -> bool,
         cluster: &'a ClusterData,
     ) -> impl Iterator<Item = NodeRef<'a>> {
+        let predicate = move |node| match replica_location {
+            ReplicaLocation::Any | ReplicaLocation::Datacenter => predicate(&node),
+            ReplicaLocation::DatacenterAndRack => {
+                predicate(&node) && node.rack == self.preferred_rack
+            }
+        };
+        let should_be_local = match replica_location {
+            ReplicaLocation::Any => false,
+            ReplicaLocation::Datacenter | ReplicaLocation::DatacenterAndRack => true,
+        };
         self.nonfiltered_replica_set(ts, should_be_local, cluster)
             .into_iter()
-            .filter(predicate)
+            .filter(move |node: &NodeRef<'a>| predicate(node))
     }
 
     fn pick_replica<'a>(
         &'a self,
         ts: &TokenWithStrategy<'a>,
-        should_be_local: bool,
+        replica_location: ReplicaLocation,
         predicate: &impl Fn(&NodeRef<'a>) -> bool,
         cluster: &'a ClusterData,
     ) -> Option<NodeRef<'a>> {
+        let predicate = |node| match replica_location {
+            ReplicaLocation::Any | ReplicaLocation::Datacenter => predicate(&node),
+            ReplicaLocation::DatacenterAndRack => {
+                predicate(&node) && node.rack == self.preferred_rack
+            }
+        };
+        let should_be_local = match replica_location {
+            ReplicaLocation::Any => false,
+            ReplicaLocation::Datacenter | ReplicaLocation::DatacenterAndRack => true,
+        };
         let replica_set = self.nonfiltered_replica_set(ts, should_be_local, cluster);
         if let Some(fixed) = self.fixed_shuffle_seed {
             let mut gen = Pcg32::new(fixed, 0);
-            replica_set.choose_filtered(&mut gen, |node| predicate(&node))
+            replica_set.choose_filtered(&mut gen, predicate)
         } else {
-            replica_set.choose_filtered(&mut thread_rng(), |node| predicate(&node))
+            replica_set.choose_filtered(&mut thread_rng(), predicate)
         }
     }
 
     fn shuffled_replicas<'a>(
         &'a self,
         ts: &TokenWithStrategy<'a>,
-        should_be_local: bool,
+        replica_location: ReplicaLocation,
         predicate: impl Fn(&NodeRef<'a>) -> bool,
         cluster: &'a ClusterData,
     ) -> impl Iterator<Item = NodeRef<'a>> {
-        let replicas = self.replicas(ts, should_be_local, predicate, cluster);
+        let replicas = self.replicas(ts, replica_location, predicate, cluster);
 
         self.shuffle(replicas)
     }
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -1313,6 +1313,102 @@ mod tests {
                     .group([B, C, E]) // remote nodes
                     .build(),
             },
+            // Keyspace SS with RF=2 with enabled DC failover and rack-awareness
+            Test {
+                policy: DefaultPolicy {
+                    preferred_datacenter: Some("eu".to_owned()),
+                    preferred_rack: Some("r1".to_owned()),
+                    is_token_aware: true,
+                    permit_dc_failover: true,
+                    ..Default::default()
+                },
+                routing_info: RoutingInfo {
+                    token: Some(Token { value: 160 }),
+                    keyspace: Some(KEYSPACE_NTS_RF_3),
+                    consistency: Consistency::One,
+                    ..Default::default()
+                },
+                // going though the ring, we get order: F , A , C , D , G , B , E
+                //                                      us  eu  eu  us  eu  eu  us
+                //                                      r2  r1  r1  r1  r2  r1  r1
+                expected_groups: ExpectedGroupsBuilder::new()
+                    .group([A, C]) // pick local rack replicas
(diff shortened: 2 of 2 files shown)


scylla/src/transport/load_balancing/default.rs
```diff
@@ -393,16 +423,24 @@ impl DefaultPolicyBuilder {
...
+                DefaultPolicy::is_alive(node)
+                    && latency_predicate(node)
+                    && DefaultPolicy::rack_predicate(node, &preferred_rack)
+            }) as Box<dyn Fn(&NodeRef) -> bool + Send + Sync + 'static>
         } else {
-            Box::new(DefaultPolicy::is_alive)
+            Box::new(move |node: &NodeRef<'_>| {
+                DefaultPolicy::is_alive(node)
+                    && DefaultPolicy::rack_predicate(node, &preferred_rack)
+            })
```
wprzytula, line 438: I'm afraid this defies the whole pick mechanism. Please note that `rack_predicate()` evaluates to false for `preferred_rack` set to `None`. Having this in mind, we can see that if no rack is preferred, then `pick_predicate` is always false, so the `pick()` function will be able to choose a node only not using the `pick_predicate()`, in line with comment: 
```
// Previous checks imply that every node we could have selected is down.
// Let's try to return a down node that wasn't disabled.
```
...which is surely not the desired behaviour.

Hence, either `pick_predicate` must not call `rack_predicate` or, even better, `rack_predicate` shall be true if no rack is preferred.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -120,7 +122,24 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
...
+            );
+            let local_replicas = self.shuffled_replicas(
+                ts,
+                true,
+                |node| {
+                    Self::is_alive(node)
+                        && !DefaultPolicy::rack_predicate(node, &self.preferred_rack)
+                },
+                cluster,
+            );
```
wprzytula, line 142: Judging from your `fallback()` implementation, you suggest such a load-balancing order:
1. local-rack local-DC replicas,
2. non-local-rack local-DC replicas,
3. remote replicas,
4. local nodes,
5. remote nodes.

As you haven't modified `pick()`, however, it chooses nodes in such an order:
1. local-rack local-DC replicas (because only then `pick_predicate() is true),
2. all other nodes.

It is probable that there are no replicas in the preferred rack. What then? I believe we should still prioritise local DC replicas, which currently is not true.

wprzytula (changes requested):

piodul (commented):
You added the implementation of rack awareness, but I can't see a way to enable it from the user's perspective. The `DefaultPolicyBuilder` should be adjusted.

Gor027:
V2:

* The `pick()` function will prefer local rack replicas, and if there are none, it will prioritize local DC replicas.
* The `fallback()` function will have a new tier in the load balancing order prioritizing local DC local rack replicas in the fallback plan.

Fixed all suggestions mentioned above.

piodul:
Please rebase on main. The PR doesn't seem to account for the newly added option that disables replica shuffling.

wprzytula:
Now that we want filter replicas not only by DC, but also by rack, we should consider adding this option to the replica locator, similarly to how filtering by DC is currently implemented. WDYT @havaker @cvybhu?
Alternatively, we can filter replicas by rack on the higher level (as this commit attempts atm), but this is an inconsistency, and a probable source of inefficiency.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -272,6 +306,24 @@ impl DefaultPolicy {
...
+            return self
+                .nonfiltered_replica_set(ts, should_be_local, cluster)
+                .choose_filtered(&mut thread_rng(), |node| {
+                    predicate(&node) && node.rack.as_deref() == Some(rack)
+                });
+        }
+
+        None
+    }
+
```
havaker, line 326: In my opinion, creating a distinct function for selecting local rack replicas is unnecessary. Instead, the desired outcome can be accomplished by utilizing a suitable predicate with the `pick_replicas` function.

Furthermore, it appears that the modifications introduced by rebasing were not taken into account. While `pick_replica` has the ability to select from an unshuffled sequence, this capability is not present in `pick_local_rack_replica`, which results in a lack of support for the `enable_shuffling_replicas(false)` option.

Overall, I believe this reinforces the argument against replicating the functionality of `pick_replica` and instead utilizing a custom predicate to choose local rack replicas.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -1234,6 +1307,54 @@ mod tests {
...
+                    token: Some(Token { value: 760 }),
+                    keyspace: Some(KEYSPACE_SS_RF_2),
+                    consistency: Consistency::One,
+                    ..Default::default()
+                },
+                // going though the ring, we get order: F , A , C , D , G , B , E
+                //                                      us  eu  eu  us  eu  eu  us
+                //                                      r2  r1  r1  r1  r2  r1  r1
+                expected_groups: ExpectedGroupsBuilder::new()
+                    .group([G]) // pick local rack replicas
```
havaker, line 1329: Having a single element rack group fails to test shuffling.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -123,7 +134,26 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
...
         // If token is available, get a shuffled list of alive replicas.
         let maybe_replicas = if let Some(ts) = &routing_info.token_with_strategy {
-            let local_replicas = self.shuffled_replicas(ts, true, Self::is_alive, cluster);
+            let local_rack_replicas = self.replicas(
+                ts,
+                true,
+                |node| {
+                    Self::is_alive(node)
+                        && self.preferred_rack.is_some()
+                        && DefaultPolicy::rack_predicate(node, &self.preferred_rack)
```
havaker, line 143: I would prefer to include checking if `preferred_rack` is not `None` to `rack_predicate`. It would reduce the complexity of this expression.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -353,6 +405,10 @@ impl DefaultPolicy {
         node.is_enabled()
     }
 
+    fn rack_predicate(node: &NodeRef<'_>, preferred_rack: &Option<String>) -> bool {
```
havaker, line 408: The other predicates are named after the condition they check, such as `is_alive` which verifies if a node is alive (it doesn't at the moment but will be;))). It may be worthwhile to rename `rack_predicate` to adhere to this convention, e.g. `is_rack_local`.

havaker (changes requested):

Gor027:
V3:

* Picking a local rack replica is implemented using `pick_replica` as suggested above:
   As @wprzytula suggested above, the `should_be_local` parameter in `pick_replica` function is replaced by `ReplicaLocation` enum which resulted in the following change in the function:
   ```
           let (predicate, should_be_local): (Box<dyn Fn(&NodeRef<'a>) -> bool>, bool) =
            match replica_location {
                ReplicaLocation::Any => (Box::new(|node| predicate(node)), false),
                ReplicaLocation::Datacenter => (Box::new(|node| predicate(node)), true),
                ReplicaLocation::DatacenterAndRack => (
                    Box::new(|node| predicate(node) && node.rack == self.preferred_rack),
                    true,
                ),
            };

   ```
   Even though the signatures of the closures in the above match arms are the same, however, the types are different so I could not find a more elegant solution than to wrap the returned predicates in Box. As a consequence, the `predicate` variable has a complex type and Clippy started complaining about it, so for now, the warnings are suppressed with the `allow` attribute: `#[allow(clippy::type_complexity)]`
* Added tests for rack awareness and fixed shuffling enabled
* Added documentation in the docs book

scylla/src/transport/load_balancing/default.rs
```diff
@@ -272,13 +321,23 @@ impl DefaultPolicy {
...
     ) -> Option<NodeRef<'a>> {
+        let (predicate, should_be_local): (Box<dyn Fn(&NodeRef<'a>) -> bool>, bool) =
+            match replica_location {
+                ReplicaLocation::Any => (Box::new(|node| predicate(node)), false),
+                ReplicaLocation::Datacenter => (Box::new(|node| predicate(node)), true),
+                ReplicaLocation::DatacenterAndRack => (
+                    Box::new(|node| predicate(node) && node.rack == self.preferred_rack),
+                    true,
+                ),
+            };
```
wprzytula, line 340: As the "boxing problem" is a possible inefficiency, this is my proposed solution:
(instead of having three closures with one behaviour specified, let's have a single predicate closure that chooses its behaviour with a `match`)
```rust
  fn pick_replica<'a>(
      &'a self,
      ts: &TokenWithStrategy<'a>,
      replica_location: ReplicaLocation,
      predicate: &impl Fn(&NodeRef<'a>) -> bool,
      cluster: &'a ClusterData,
  ) -> Option<NodeRef<'a>> {
      let predicate = |node| match replica_location {
          ReplicaLocation::Any |
          ReplicaLocation::Datacenter => predicate(&node),
          ReplicaLocation::DatacenterAndRack => predicate(&node) && node.rack == self.preferred_rack,
      };
      let should_be_local = match replica_location {
          ReplicaLocation::Any => false,
          ReplicaLocation::Datacenter |
          ReplicaLocation::DatacenterAndRack => true,
      };
      let replica_set = self.nonfiltered_replica_set(ts, should_be_local, cluster);
      if let Some(fixed) = self.fixed_shuffle_seed {
          let mut gen = Pcg32::new(fixed, 0);
          replica_set.choose_filtered(&mut gen, predicate)
      } else {
          replica_set.choose_filtered(&mut thread_rng(), predicate)
      }
  }
```

Gor027: Done. Thanks for the hint.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -13,6 +13,12 @@ use scylla_cql::{errors::QueryError, frame::types::SerialConsistency, Consistenc
 use std::{fmt, sync::Arc, time::Duration};
 use tracing::warn;
 
+pub enum ReplicaLocation {
```
havaker, line 16: Why is it public?

scylla/src/transport/load_balancing/default.rs
```diff
@@ -125,7 +150,25 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
...
+            );
+            let local_replicas = self.shuffled_replicas(
+                ts,
+                true,
+                |node| {
+                    Self::is_alive(node)
+                        && (self.preferred_rack.is_none() || node.rack != self.preferred_rack)
+                },
+                cluster,
+            );
```
havaker, line 171: `self.replicas`, `self.shuffled_replicas` and `self.pick_replica` had a very similar signature. I feel that it's a bit inconsistent that you introduced a `ReplicaLocation` enum only in `self.pick_replica`. Was there any specific reason behind it?

scylla/src/transport/load_balancing/default.rs
```diff
@@ -62,17 +68,36 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
...
+            let local_rack_picked = self.pick_replica(
+                ts,
+                ReplicaLocation::DatacenterAndRack,
+                &self.pick_predicate,
+                cluster,
+            );
+
+            if let Some(alive_local_rack_replica) = local_rack_picked {
+                return Some(alive_local_rack_replica);
+            }
```
havaker, line 82: I'd like to see this fragment of code enclosed in an If statement that would check if a `preferred_rack` field is `Some`.

In its current form, we would unnecessarily iterate through all local replicas in the case when `preferred_rack` is `None`. This is due to the way `self.pick_replica` works if provided with always-false filter.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -123,7 +134,26 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
...
+            let local_replicas = self.shuffled_replicas(
+                ts,
+                true,
+                |node| {
+                    Self::is_alive(node)
+                        && (self.preferred_rack.is_none()
+                            || !DefaultPolicy::rack_predicate(node, &self.preferred_rack))
+                },
+                cluster,
+            );
```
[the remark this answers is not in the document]
havaker: @Gor027

scylla/src/transport/load_balancing/default.rs
havaker: Please add a test, that tests shuffling of local-rack replicas.

docs/source/load-balancing/default-policy.md
havaker, line 1: Please update the `Creating a DefaultPolicy` section.

havaker (changes requested):

docs/source/load-balancing/default-policy.md
Gor027: Done.

scylla/src/transport/load_balancing/default.rs
Gor027: Done, thanks for the hint.

scylla/src/transport/load_balancing/default.rs
havaker: I think it does test shuffling.
https://github.com/scylladb/scylla-rust-driver/blob/758ec3e17e248a77ec7d269d9c2e9a3d82a402c3/scylla/src/transport/load_balancing/default.rs#L707-L715

After some thought though, I realized that it would be quite difficult to test if returned local-rack replicas as shuffled. It is because local-rack replicas are always first in the plan (if available), an hence the first element from the plan is a picked local-rack replica. If `pick` chooses nodes in a random way but `fallback` doesn't, the returned group in plan is not deterministic anyway. Our testing framework has no way of detecting that a group is only partially shuffled.

I don't know if adding new functionality to the testing framework is worth it. I guess that I'm withdrawing my request for a test that checks shuffling of local-rack replicas. @wprzytula any thoughts on this?

scylla/src/transport/load_balancing/default.rs
havaker: After your changes, semantics of `self.shuffled_replicas` and `self.pick_replica` differ. `self.pick_replica` modifies its predicate according to the value of `ReplicaLocation` argument, but `self.shuffled_replicas` does not. Why?

havaker:
Our convention regarding commit names is something like `module_name: short description of changes`. I don't see you sticking to it:/

scylla/src/transport/load_balancing/default.rs
Gor027: You're right, the predicate logic was passed in the `fallback` function, so I thought it is redundant to modify the predicate in `self.replicas`. I fixed it, but the predicate modification required the `ReplicaLocation` enum to implement `Clone` and `Copy`.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -68,6 +69,21 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
             }
         }
         if let Some(ts) = &routing_info.token_with_strategy {
+            // Try to pick some alive local rack random replica.
+            // If preferred rack is not specified, try to pick local DC replica.
+            if let Some(_local_rack) = &self.preferred_rack {
```
havaker, line 74: `if let` is a bit superfluous here, you are creating an unused `_local_rack` variable. `if self.preferred_rack.is_some()` would be a better choice there probably.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -283,13 +309,19 @@ impl DefaultPolicy {
         predicate: impl Fn(&NodeRef<'a>) -> bool,
         cluster: &'a ClusterData,
     ) -> impl Iterator<Item = NodeRef<'a>> {
+        let predicate = move |node| match replica_location {
+            ReplicaLocation::Any | ReplicaLocation::Datacenter => predicate(&node),
+            ReplicaLocation::DatacenterAndRack => {
+                predicate(&node) && node.rack == self.preferred_rack
+            }
+        };
```
havaker, line 317: I think that it belongs to the previous commit, where you added `ReplicaLocation` parameter to `pick_replica`, `replicas` and `shuffled_replicas`. 

This is because without this predicate modification, those functions are not fully respecting `ReplicaLocation` values - `replicas(..., ReplicaLocation::DatacenterAndRack, ...)` won't return a set that is restricted to local rack.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -299,6 +331,12 @@ impl DefaultPolicy {
         predicate: &impl Fn(&NodeRef<'a>) -> bool,
         cluster: &'a ClusterData,
     ) -> Option<NodeRef<'a>> {
+        let predicate = |node| match replica_location {
+            ReplicaLocation::Any | ReplicaLocation::Datacenter => predicate(&node),
+            ReplicaLocation::DatacenterAndRack => {
+                predicate(&node) && node.rack == self.preferred_rack
+            }
+        };
```
havaker, line 339: ditto

scylla/src/transport/load_balancing/default.rs
```diff
@@ -283,13 +309,19 @@ impl DefaultPolicy {
...
         let should_be_local = match replica_location {
             ReplicaLocation::Any => false,
             ReplicaLocation::Datacenter | ReplicaLocation::DatacenterAndRack => true,
         };
         self.nonfiltered_replica_set(ts, should_be_local, cluster)
             .into_iter()
-            .filter(predicate)
+            .filter(move |node: &NodeRef<'a>| predicate(node))
```
havaker, line 324: This change should also be applied to the previous commit - it does not compile without it.

docs/source/load-balancing/default-policy.md
```diff
@@ -48,6 +50,14 @@ false), the default policy will only include local nodes in load balancing
 plans. Remote nodes will be excluded, even if they are alive and available to
 serve requests.
 
+#### Preferred Rack
+
+The `preferred_rack` field in `DefaultPolicy` allows the load balancing policy to
+prioritize nodes based on their availability zones in the preferred datacenter.
+When a preferred rack is set, the policy will first return replicas in the local rack
+in the preferred datacenter, and then the other replicas in the datacenter.
+When a preferred datacenter is not set, setting preferred rack will not have any effect.
```
havaker, line 59: Please add a test that checks this behavior.

havaker (changes requested):

scylla/src/transport/load_balancing/default.rs
piodul: I think it would be nice to at least verify that `pick` chooses rack-local replicas randomly. Having a test with the "local rack replicas" group of size 2 should check this.

Later we can consider adjusting the tests - for example, in addition to checking plans that combine `pick` and `fallback` we could also check plans returned by `fallback` only.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -1313,6 +1313,101 @@ mod tests {
...
+                    token: Some(Token { value: 560 }),
+                    keyspace: Some(KEYSPACE_SS_RF_2),
+                    consistency: Consistency::Two,
+                    ..Default::default()
+                },
+                // going though the ring, we get order: B , C , E , G , A , F , D
+                //                                      eu  eu  us  eu  eu  us  us
+                //                                      r1  r1  r1  r2  r1  r2  r1
+                expected_groups: ExpectedGroupsBuilder::new()
+                    .deterministic([B, C]) // pick local rack + fallback replicas
```
piodul, line 1361: Nit: instead of this single `.deterministic`, you should do:

```rust
.deterministic([B])
.deterministic([C])
```

The `.deterministic()` method does not check that the load balancing policy always returns in the specified order - e.g. in your case it does not check that it always return `B` and then `C`. It only checks that the nodes within the group are always returned in an _unspecified_, but _always the same order_.

In your case, both `[B, C]` and `[C, B]` orders will be accepted by the test as long as the load balancing policy always sticks to the same order. I assume that you wanted to check that `B` then `C` is always returned - in such case you should use two groups like I suggested.

havaker (approved):
LGTM

There's one minor thing left that I think is worth fixing - the main thing `transport: load_balancing: Add ReplicationLocation enum ` commit introduces is rack-awareness to the default policy internals, but it's not mentioned anywhere in the commit message.

Gor027:
> LGTM
> 
> There's one minor thing left that I think is worth fixing - the main thing `transport: load_balancing: Add ReplicationLocation enum ` commit introduces is rack-awareness to the default policy internals, but it's not mentioned anywhere in the commit message.

It's mentioned in a short sentence: `Modified pick and fallback plan to prefer local rack replicas.`

piodul (approved):

Issue #634, opened 2023-01-27
Allow for token-aware routing without shuffling


piodul:
In some scenarios it's beneficial to have token-aware routing _without_ the round-robin behavior. For example, if a client does many reads with CL=ONE and the driver always chooses the same coordinator for a particular partition, each partition will be cached on one node only. This is currently not supported by the load balancing policies provided by the driver as RoundRobinPolicy/DcAwareRoundRobing policy must always be a part of the constructed load balancing policy.

A similar feature already exists in the cpp driver: https://github.com/scylladb/cpp-driver/blob/eb6b70d138752c5e6a487eeaf69e6baa4d6652a4/include/cassandra.h#L1113-L1133

This can be solved in one of the following ways, depending on whether it is done before or after the load balancing policy refactor:

- Make it possible to disable shuffling in round robin policies when they handle a token aware query,
- Add this option to the default load balancing policy that will be introduced during the refactor.

PR #667, opened 2023-03-20
load_balancing: allow to disable shuffling replicas

piodul:
## Pre-review checklist

When token awareness is enabled in the default load balancing policy, it chooses a random replica to send the query to. This is supposed to balance the load however in some scenarios it might be more beneficial to always choose the same replica - for example, reads might benefit from better caching. This PR adds an option to disable this behavior, which will cause the load balancing policy to always choose nodes in the same order for each replica set.

This feature was inspired by similar functionality in the cpp driver.

Fixes: #634


scylla/src/transport/load_balancing/default.rs
```diff
@@ -323,11 +330,18 @@ impl DefaultPolicy {
...
+        &self,
+        iter: impl Iterator<Item = NodeRef<'a>>,
+    ) -> impl Iterator<Item = NodeRef<'a>> {
         let mut vec: Vec<NodeRef<'a>> = iter.collect();
 
-        let mut rng = thread_rng();
-        vec.shuffle(&mut rng);
+        if let Some(fixed) = self.fixed_shuffle_seed {
+            let mut gen = Pcg32::new(fixed as u64, 0);
+            vec.shuffle(&mut gen);
```
wprzytula, line 341: Is the cost of calling `Pcg32::new()` on each query negligible, or should we take it into account?

piodul:
Accidentally marked as draft, un-drafting.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -651,13 +681,37 @@ mod tests {
...
-                            }
+
+                        // Verify that the group has the same nodes as the
+                        // expected one
+                        let got_set: HashSet<_> = got_group.iter().copied().collect();
+                        let expected_set = expected.nodes();
+                        assert_eq!(&got_set, expected_set);
+
+                        // Put the group into sets_of_groups
+                        sets_of_groups[group_id].insert(got_group.clone());
```
wprzytula, line 698: Isn't this `clone()` unnecessary?

scylla/src/transport/load_balancing/default.rs
```diff
@@ -606,12 +606,21 @@ mod tests {
 
         enum ExpectedGroup {
             AnyOrder(HashSet<u16>),
+            Deterministic(HashSet<u16>),
         }
```
wprzytula: I like this design very much! The types rule.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -984,6 +1009,31 @@ mod tests {
...
+                    keyspace: Some(KEYSPACE_NTS_RF_3),
+                    consistency: Consistency::Quorum,
+                    ..Default::default()
+                },
+                // going though the ring, we get order: F , A , C , D , G , B , E
+                //                                      us  eu  eu  us  eu  eu  us
+                //                                      r2  r1  r1  r1  r2  r1  r1
+                expected_groups: ExpectedGroupsBuilder::new()
+                    .deterministic([A, C, G]) // pick + fallback local replicas
+                    .deterministic([F, D, E]) // remote replicas
```
wprzytula, line 1040: How did you find out the output order for the given seed? Purely empirically?

scylla/src/transport/load_balancing/default.rs
```diff
@@ -605,21 +605,21 @@ mod tests {
         };
 
         enum ExpectedGroup {
-            AnyOrder(HashSet<u16>),
+            NonDeterministic(HashSet<u16>),
             Deterministic(HashSet<u16>),
         }
```
wprzytula: So now we expect nondeterminism both for replicas and for non-replicas, which is a generalisation. Replicas are shuffled, whereas non-replicas are randomly rotated. If we want to consider both as nondeterministic, *dot*, then we can leave it as you propose; if, however, the distinction between them is meaningful, we should highlight that distinction at least in naming.

scylla/src/transport/load_balancing/default.rs
piodul: It should be pretty cheap: https://docs.rs/rand_pcg/latest/src/rand_pcg/pcg64.rs.html#88-92. I can see one multiplication, two additions and some operations on bits.

scylla/src/transport/load_balancing/default.rs
piodul: Yes, it is, I will get rid of it.

scylla/src/transport/load_balancing/default.rs
piodul: I don't assume that the nodes within the group will be in some particular order. I just go over every plan that is being checked and verify that they all have the same ordering of nodes within that group.

wprzytula: Oh, right.

scylla/src/transport/load_balancing/default.rs
piodul: I'd rather leave this as "non-deterministic". If we wanted to differentiate "shuffling" and "rotating" then we probably could provide different assertions for both types. I'd rather not do it in this PR.

piodul:
v2:
- Changed the `fixed_shuffle_seed` size from `u32` to `u64`
- Got rid of an unnecessary clone in `assert_proper_grouping_in_plans`
- Added `fixed_shuffle_seed` to the `impl Debug`

wprzytula (approved):

PR #685, opened 2023-03-28
Docs: Add serverless cluster connect code example

### LINKED ISSUE #677: Docs: missing documentation on how to connect to Serverless
The example is @ https://github.com/scylladb/scylla-rust-driver/blob/main/examples/cloud.rs but it is missing from https://rust-driver.docs.scylladb.com/stable/connecting/connecting.html

Gor027:
Fixes #677

## Pre-review checklist

### DIFF
--- a/docs/source/connecting/connecting.md
+++ b/docs/source/connecting/connecting.md
@@ -34,6 +34,53 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+Scylla Serverless is an elastic and dynamic deployment model. When creating a `Session` you need to
+specify the secure connection bundle as follows:
+
+```rust
+# extern crate scylla;
+# extern crate tokio;
+# fn check_only_compiles() {
+use std::path::Path;
+use std::error::Error;
+use scylla::CloudSessionBuilder;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let session = CloudSessionBuilder::new(Path::new("config_data.yaml"))
+        .unwrap()
+        .build()
+        .await
+        .unwrap();
+
+    Ok(())
+}
+# }
+```
+
+Note that the bundle file will be provided after the serverless cluster is created. Here is an example of a
+configuration file for a serverless cluster:
+
+```yaml
+datacenters:
+  datacenter1:
+    certificateAuthorityData: CERTIFICATE_DATA
+    server: 127.0.1.1:9142
+    nodeDomain: cql.cluster-id.scylla.com
+    insecureSkipTlsVerify: false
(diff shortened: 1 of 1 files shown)


docs/source/connecting/connecting.md
```diff
@@ -34,6 +34,52 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+async fn main() -> Result<(), Box<dyn Error>> {
+    let session = CloudSessionBuilder::new(Path::new("config_data.yaml"))
+        .unwrap()
+        .build()
+        .await
+        .unwrap();
+
+    Ok(())
+}
+
```
piodul, line 57: Nit: unnecessary empty line at the end of the snippet (the previous example doesn't have it).

docs/source/connecting/connecting.md
```diff
@@ -34,6 +34,52 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+Scylla Serverless is an elastic and dynamic deployment model. When creating a `Session` you need to
+specify the secure connection bundle as follows:
+
+```rust
```
piodul, line 40: The CI fails because it compiles and runs examples from the documentation, and this particular example tries to use a yaml configuration that doesn't exist to actually connect to Scylla Cloud.

AFAIK we don't simulate Scylla Cloud in CI, so until we do we should use the `no_run` [attribute](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes):

    ```rust,no_run
    // Contents of the example go here
    ```

wprzytula (approved):

docs/source/connecting/connecting.md
piodul: > As @cvybhu suggested, the example is wrapped under a `check_only_compiles` function which ensures the example is compiled but not run.

Ok, such a solution makes sense as well.

docs/source/connecting/connecting.md
```diff
@@ -34,6 +34,52 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+        .build()
+        .await
+        .unwrap();
+
+    Ok(())
+}
+# }
+```
+
+Here is an example of a configuration file for a serverless cluster:
```
piodul, line 61: I think it would be good to mention here that the configuration file is provided by us, not something that the user is supposed to write themself.

Are there any docs, tutorials or other learning resources that explain the concept of the "connection bundle"? If so, then we could link them here to guide users that might not be familiar with serverless.

Gor027: I added a note about the configuration file but unfortunately, I could not find any specific tutorial or documentation explaining the concept of a bundle file and that it will be provided after creating a serverless cluster.

piodul (approved):

PR #686, opened 2023-03-30
Handle errors that occur during handshake

rukai:
## Pre-review checklist

This allows users to see the errors that occurred during the handshake, previously the user just gets a generic `Wrong response to OPTIONS message was received`


wprzytula (approved):
In case of handling server-side errors, transparency is paramount. Thank you, @rukai!

Issue #688, opened 2023-03-31
Adjust existing tests for Serverless


wprzytula:
We would like to run existing tests against both ordinary and Serverless clusters. Once we add a Serverless cluster to our CI (#687), as Serverless clusters require different Session setup than ordinary ones, we need to adjust the exisiting tests (at least a sensibly chosen subset of them) so that they can be generic over the cluster type.

wprzytula:
One idea how to do that is by using generics:

the chosen tests are made generic over `GenericSessionBuilder` so that they can be run with thin runner entries that consist only of `SessionBuilder` configuration. The entries for serverless are `ignore`d by default, and they are only run explicitly by the CI workflow for Serverless. 

The screenshot below is a PoC.

![Adjust_tests_for_Serverless](https://user-images.githubusercontent.com/59482568/229071641-6a97cf3a-0827-4336-a967-277aba6902ad.png)

wprzytula:
Another idea (by @piodul), which I find cleaner than the above one, is to use conditional compilation to make tests _serverless-agnostic_. This is a sketch how to achieve that:
- `--cfg serverless-tests` would be passed when building and running tests for Serverless,
- write `fn create_new_session()`, whose behaviour would depend on `#[cfg(serverless-tests)]` / `#[cfg(not(serverless-tests))]`,
- refactor the chosen tests to use `create_new_session()`.

PR #689, opened 2023-04-03
Serverless ci

### LINKED ISSUE #687: Add Serverless cluster to CI
We need automated testing of our driver on Serverless clusters as well.
As CCM already supports setting up a simple Serverless cluster, this shouldn't be complicated.

Gor027:
## Pre-review checklist

This PR adds a test workflow to CI for a serverless cluster. For now, only the `examples/cloud` example is added to test serverless support, but the existing tests will be adjusted to be run with a serverless cluster, see #688.

Fixes #687

### DIFF
--- a/.github/workflows/serverless.yaml
+++ b/.github/workflows/serverless.yaml
@@ -1,0 +1,35 @@
+name: Serverless
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python 2.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+      - name: Install scylla-ccm
+        run: pip3 install https://github.com/scylladb/scylla-ccm/archive/master.zip
+
+      - name: Create serverless cluster
+        run: |
+          ccm create serverless -i 127.0.1. -n 1 --scylla -v release:5.1.6
+          ccm start  --sni-proxy --sni-port 7777
+
+      - name: Check
+        run: cargo check --verbose
+      - name: Run cloud example
+        run: cargo run --example cloud -- $HOME/.ccm/serverless/config_data.yaml
+
+      - name: Remove serverless cluster
+        run: ccm remove serverless
--- a/examples/cloud.rs
+++ b/examples/cloud.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::path::Path;
 
 use anyhow::Result;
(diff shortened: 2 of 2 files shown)


.github/workflows/serverless.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Setup Python 3
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+      - name: Install scylla-ccm
+        run: pip3 install https://github.com/scylladb/scylla-ccm/archive/master.zip
+
+      - name: Create serverless cluster
+        run: |
+          ccm create serverless -i 127.0.1. -n 1 --scylla -v release:4.6.9
```
mykaul, line 26: 4.6.9? Why such an old release?

.github/workflows/serverless.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python 3
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
```
piodul, line 20: The step is called "Setup Python 3", but it looks like it installs Python 2.7.

Does ccm really need python 2?

.github/workflows/serverless.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Install scylla-ccm
+        run: pip3 install https://github.com/scylladb/scylla-ccm/archive/master.zip
+
+      - name: Create serverless cluster
+        run: |
+          ccm create serverless -i 127.0.1. -n 1 --scylla -v release:5.1.6
```
piodul, line 26: Can ccm pick the latest released version instead of choosing a particular one? Otherwise we will have to remember to update the version from time to time.

.github/workflows/serverless.yaml
```diff
@@ -0,0 +1,35 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python 3
+        uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+      - name: Install scylla-ccm
+        run: pip3 install https://github.com/scylladb/scylla-ccm/archive/master.zip
```
piodul, line 22: And here we use pip3? I'm confused now, do we need python 2 at all?

piodul (changes requested):

.github/workflows/serverless.yaml
Gor027: Previously I was testing with python version `3.x` on my fork, which caused some problems: https://github.com/Gor027/scylla-rust-driver/actions/runs/4591382447/jobs/8107506338

I am not sure what is the root cause of the problem, however, as `cqlsh` is compatible with Python 2.7, perhaps it is safer to prefer version 2.7 over 3.x

.github/workflows/serverless.yaml
Gor027: I think `pip3` just refers to the pip module of the intended python version.

.github/workflows/serverless.yaml
piodul: Understood, thanks.

.github/workflows/serverless.yaml
piodul: Could you try again here with python 3? It looks like ccm tried to run some script from the downloaded "unified package" which required python 2. Maybe this problem was fixed in 5.1.6.

piodul: OK, I see that it failed with the same error. In that case let's stick to python 2.

piodul (approved):

PR #690, opened 2023-04-03
Execution profiles API extensions

wprzytula:
In order to implement execution profiles in Cpp-Rust-driver project, the public API of execution profiles needs to be extended. To sum up, this PR:
- derives `Clone` and `Default` for `ExecutionProfileBuilder`,
- reexports `ExecutionProfile` directly from crate root instead of `transport::` - this is just for user convenience and it is similarly done with policies etc.,
- makes `Session::get_default_execution_profile_handle()` public,
- adds `ExecutionProfileHandle::pointee_to_builder()`, which creates a builder with configuration equivalent to one of the profile pointed by provided handle.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -1815,7 +1815,9 @@ impl Session {
         }
     }
 
-    fn get_default_execution_profile_handle(&self) -> &ExecutionProfileHandle {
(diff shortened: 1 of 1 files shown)


Gor027 (approved):
LGTM

PR #694, opened 2023-04-11
Serverless tests

### LINKED ISSUE #688: Adjust existing tests for Serverless

Gor027:
## Pre-review checklist

This PR adjusts session tests to run with a serverless cluster in CI. As a serverless cluster requires a different Session setup, a generic way of session builder creation is added that creates a `CloudSessionBuilder` if the "cloud" feature is enabled for session tests.
Fixes #688 

EDIT:
Modified CI workflows to run cloud tests separately, as examples required the "cloud" feature to be enabled, `cargo test` run session tests for serverless mode.

### DIFF
--- a/.github/workflows/serverless.yaml
+++ b/.github/workflows/serverless.yaml
@@ -30,6 +30,8 @@ jobs:
         run: cargo check --verbose
       - name: Run cloud example
         run: cargo run --example cloud -- $HOME/.ccm/serverless/config_data.yaml
+      - name: Run cloud tests
+        run: CLOUD_CONFIG_PATH=$HOME/.ccm/serverless/config_data.yaml RUSTFLAGS="--cfg scylla_cloud_tests" cargo test --verbose
 
       - name: Remove serverless cluster
         run: ccm remove serverless
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -1628,6 +1628,7 @@ mod tests {
     ///    Then use query_iter with page_size set to 7 to select all 100 rows.
     /// 3. INSERT query_iter should work and not return any rows.
     #[tokio::test]
+    #[cfg(not(scylla_cloud_tests))]
     async fn connection_query_iter_test() {
         let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
         let addr: SocketAddr = resolve_hostname(&uri).await;
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -14,7 +14,9 @@ use crate::transport::topology::Strategy::SimpleStrategy;
 use crate::transport::topology::{
     CollectionType, ColumnKind, CqlType, NativeType, UserDefinedType,
 };
-use crate::utils::test_utils::{supports_feature, unique_keyspace_name};
+use crate::utils::test_utils::{
+    create_new_session_builder, supports_feature, unique_keyspace_name,
+};
 use crate::CachingSession;
 use crate::ExecutionProfile;
 use crate::QueryResult;
@@ -59,8 +61,7 @@ async fn test_connection_failure() {
 
 #[tokio::test]
 async fn test_unprepared_statement() {
-    let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
-    let session = SessionBuilder::new().known_node(uri).build().await.unwrap();
+    let session = create_new_session_builder().build().await.unwrap();
     let ks = unique_keyspace_name();
 
     session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'SimpleStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
@@ -158,8 +159,7 @@ async fn test_unprepared_statement() {
 
 #[tokio::test]
 async fn test_prepared_statement() {
-    let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
-    let session = SessionBuilder::new().known_node(uri).build().await.unwrap();
+    let session = create_new_session_builder().build().await.unwrap();
     let ks = unique_keyspace_name();
 
     session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'SimpleStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
@@ -365,8 +365,7 @@ async fn test_prepared_statement() {
 
 #[tokio::test]
 async fn test_batch() {
-    let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
-    let session = Arc::new(SessionBuilder::new().known_node(uri).build().await.unwrap());
+    let session = Arc::new(create_new_session_builder().build().await.unwrap());
     let ks = unique_keyspace_name();
 
     session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'SimpleStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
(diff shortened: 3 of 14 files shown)


scylla/src/transport/session_test.rs
```diff
@@ -33,6 +38,38 @@ use std::time::Duration;
...
+        let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+        let session_builder: GenericSessionBuilder<DefaultMode> =
+            SessionBuilder::new().known_node(uri);
+        session_builder
+    }
+
+    #[cfg(feature = "cloud")]
+    {
+        use crate::CloudSessionBuilder;
+        use std::path::Path;
```
wprzytula, line 63: Why are these imports here and the remaining ones are on the top of the file?

scylla/src/transport/session_test.rs
```diff
@@ -33,6 +38,38 @@ use std::time::Duration;
...
+
+        let session_builder: GenericSessionBuilder<CloudMode> = std::env::var("CLOUD_CONFIG_PATH")
+            .map(|config_path| CloudSessionBuilder::new(Path::new(&config_path)))
+            .unwrap()
+            .unwrap();
```
wprzytula, line 68: As this code is going to be called frequently (on each test), it might be helpful in debug to use `.expect(SomeMeaningfulMessage)` instead of plain `.unwrap()`.

.github/workflows/rust.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     - name: Build
       run: cargo build --verbose --examples
     - name: Run tests
-      run: SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose
+      run: SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose --package scylla
```
wprzytula, line 36: AFAIU, this means that we no longer run tests for other crates (`scylla-cql`, `scylla-proxy`, `scylla-macros`) in the CI. This probably isn't the desired state.

wprzytula (changes requested):
We should definitely keep running tests from crates other than `scylla` in our CI.

scylla/src/transport/session_test.rs
Gor027: The imports are only used in this block, the others on the top of the file allow to create the conditional type `ConditionalSessionBuilderKind`.

scylla/src/transport/session_test.rs
Gor027: Fixed it, thanks.

.github/workflows/rust.yml
Gor027: Enabled tests for other crates in `rust` and `cassandra` workflows.

scylla/src/transport/session_test.rs
```diff
@@ -33,6 +38,38 @@ use std::time::Duration;
...
+#[cfg(feature = "cloud")]
+type ConditionalSessionBuilderKind = CloudMode;
+
+/*
+  Creates a generic session builder based on conditional compilation configuration
+  For SessionBuilder of DefaultMode type, adds localhost to known hosts, as all of the tests
+  connect to localhost.
+*/
+fn create_new_session_builder() -> GenericSessionBuilder<ConditionalSessionBuilderKind> {
+    #[cfg(not(feature = "cloud"))]
```
piodul, line 52: I'm not sure whether we should use features for this. It's a bit fragile to depend on some feature _not_ being enabled - I remember that there were some issue with one example requiring the "cloud" feature, and compiling tests enabled this feature for everybody, right?

Instead, I think that we should use conditional compilation. You can provide a `--cfg my-thing-here` parameter when compiling and add a bunch of attributes that enable/disable compilation based on those parameters - like you did with features as of now.

You could use a `--cfg cloud-tests` parameter. Then:

- Tests that are "cloud_agnostic" would use `create_new_session_builder()` which creates either a cloud or non-cloud session, depending on the parameter,
- Tests that require cloud would be compiled only if the parameter is set in commandline - so, they would be prepended with `#[cfg_attr(cloud_agnostic)]`,
- Tests that must not run with serverless would be decorated with `#[not(cfg_attr(cloud_agnostic))]`.

This method will work even if `cloud` feature is enabled in both cases, so I think it is safer.

Read more about conditional compilation here: https://doc.rust-lang.org/reference/conditional-compilation.html

.github/workflows/serverless.yaml
```diff
@@ -30,6 +30,10 @@ jobs:
         run: cargo check --verbose
       - name: Run cloud example
         run: cargo run --example cloud -- $HOME/.ccm/serverless/config_data.yaml
+      - name: Run session tests
+        run: CLOUD_CONFIG_PATH=$HOME/.ccm/serverless/config_data.yaml cargo test --features cloud --package scylla --lib transport::session_test
+      - name: Run cloud tests
+        run: CLOUD_CONFIG_PATH=$HOME/.ccm/serverless/config_data.yaml cargo test --features cloud --package scylla --lib cloud
```
piodul, line 36: I don't like this approach. Now, developers have to remember to add more serverless tests if they appear. I think we should have a single `cargo test` invocation which passes `--cfg serverless-tests` and everything works correctly. It's much more reasonable to require developers to use `#[cfg_attr]` annotations and `create_new_session_builder` in tests IMO.

piodul (changes requested):
IMO all tests in the `scylla` crate should be adjusted, not only those from `session_test`.

scylla/src/transport/session_test.rs
Gor027: My initial approach was to use the `--cfg` flag, however, `cargo test` does not support it and using the `cloud` feature is meant to be a workaround to conditionally run tests. As the `--cfg` is a `rustc` flag, I can pass it through the `RUSTFLAGS` environment variable but this may affect the dependencies as well. So I found the workaround with a feature as a better approach.

piodul: Passing `--cfg` via `RUSTFLAGS` is fine.

I don't think that any of the dependencies will react to `cloud_tests`, but if you are worried about that then we can prefix it with the name of the crate - so, `--cfg scylla_cloud_tests`. For example, tokio does exactly that - see here: https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml

scylla/src/transport/session_test.rs
```diff
@@ -33,6 +38,38 @@ use std::time::Duration;
...
+type ConditionalSessionBuilderKind = DefaultMode;
+#[cfg(feature = "cloud")]
+type ConditionalSessionBuilderKind = CloudMode;
+
+/*
+  Creates a generic session builder based on conditional compilation configuration
+  For SessionBuilder of DefaultMode type, adds localhost to known hosts, as all of the tests
+  connect to localhost.
+*/
+fn create_new_session_builder() -> GenericSessionBuilder<ConditionalSessionBuilderKind> {
```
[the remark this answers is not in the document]
wprzytula: Very good catch!

scylla/src/utils/test_utils.rs
```diff
@@ -52,3 +54,35 @@ pub(crate) async fn supports_feature(session: &Session, feature: &str) -> bool {
         .split(',')
         .any(|f| f == feature)
 }
+
+/*
+  Creates a generic session builder based on conditional compilation configuration
+  For SessionBuilder of DefaultMode type, adds localhost to known hosts, as all of the tests
+  connect to localhost.
+*/
```
piodul, line 62: Nit: we usually use `// line comments` for item descriptions.

scylla/src/transport/connection.rs
```diff
@@ -1628,6 +1628,7 @@ mod tests {
     ///    Then use query_iter with page_size set to 7 to select all 100 rows.
     /// 3. INSERT query_iter should work and not return any rows.
     #[tokio::test]
+    #[cfg_attr(scylla_cloud_tests, ignore)]
```
piodul, line 1631: Ah, I suggested using `cfg_attr` not having read beforehand what it does precisely... While what you did here will work, I think it will be slightly clearer if you use `#[cfg(not(scylla_cloud_tests))]`.

piodul (commented):
Looks good after fixing some minor issues.

scylla/src/utils/test_utils.rs
Gor027: Done.

scylla/src/transport/connection.rs
Gor027: Ok, changed it to `cfg(not(...))`.

piodul (approved):

Issue #696, opened 2023-04-12
Deadlock if LatencyAwareness is used


wprzytula:
**Symptoms**
If `LatencyAwareness` is turned on in `DefaultPolicy`, the issued query never ends. `tokio-console` shows idle tokio tasks.

**Reproducer**

```
#[tokio::test]
async fn test_latency_aware() {
    let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());

    let policy = DefaultPolicy::builder()
        .latency_awareness(LatencyAwarenessBuilder::default())
        .build();
    let handle = ExecutionProfile::builder()
        .load_balancing_policy(policy)
        .build()
        .into_handle();

    let session = SessionBuilder::new()
        .known_node(uri)
        .default_execution_profile_handle(handle)
        .build()
        .await
        .unwrap();

    session.query("whatever", ()).await.unwrap_err();
}
```

**Found cause**
Rust, unfortunately, makes lock guards taken on right side of `if let` expression live for `else` branch duration, too. This results in taking write lock while holding read lock.
 ```
 pub(super) fn report_query(&self, node: &Node, latency: Duration) {
            if let Some(node_avg) = self.node_avgs.read().unwrap().get(&node.host_id) {
                ...
            } else {
                // We need to add the node to the map.
                self.node_avgs.write().unwrap() ...
            }
        }
```

A PR with fix is going to be opened soon.

PR #697, opened 2023-04-12
default_policy: Fix deadlock on LatencyAwareness used

wprzytula:
Described in #696.
Rust, unfortunately, makes lock guards taken on right side of `if let` expression live for else branch duration, too. This results in taking write lock while holding read lock.

In order to compile, I had to hint the `usize` type in one unrelated place, hence that change.

Added a regression test for the issue.

Fixes: #696

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -757,7 +757,7 @@ mod tests {
                 for got in gots {
                     // First, make sure that `got` has the right number of items,
                     // equal to the sum of sizes of all expected groups
-                    let combined_groups_len = self.groups.iter().map(|s| s.len()).sum();
+                    let combined_groups_len: usize = self.groups.iter().map(|s| s.len()).sum();
                     assert_eq!(got.len(), combined_groups_len);
 
                     // Now, split `got` into groups of expected sizes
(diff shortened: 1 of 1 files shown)


scylla/src/transport/load_balancing/default.rs
```diff
@@ -1619,16 +1619,27 @@ mod latency_awareness {
         }
 
         pub(super) fn report_query(&self, node: &Node, latency: Duration) {
-            if let Some(node_avg) = self.node_avgs.read().unwrap().get(&node.host_id) {
+            let guard = self.node_avgs.read().unwrap();
```
havaker, line 1622: How about naming it `node_avgs_guard`?

havaker (commented):
LGTM

scylla/src/transport/load_balancing/default.rs
```diff
@@ -757,7 +757,7 @@ mod tests {
                 for got in gots {
                     // First, make sure that `got` has the right number of items,
                     // equal to the sum of sizes of all expected groups
-                    let combined_groups_len = self.groups.iter().map(|s| s.len()).sum();
+                    let combined_groups_len: usize = self.groups.iter().map(|s| s.len()).sum();
```
cvybhu, line 760: I removed the `usize` and it still compiles with rust 1.68, but I'm a fan of having explicit type annotations so we could leave it as is.

cvybhu (approved):

cvybhu (approved):

scylla/src/transport/load_balancing/default.rs
```diff
@@ -2582,5 +2594,29 @@ mod latency_awareness {
...
+                .into_handle();
+
+            let session = SessionBuilder::new()
+                .known_node(uri)
+                .default_execution_profile_handle(handle)
+                .build()
+                .await
+                .unwrap();
+
+            session.query("whatever", ()).await.unwrap_err();
```
piodul, line 2619: This statement should send a valid query and check that it succeeded. If it starts failing due to some unrelated reason the test will stop being useful.

wprzytula: This test was meant as a regression test, so it only tests that the test completes. We could write some other sophisticated integration tests to ensure that LAP does not cause unpredictable behaviour.

Issue #699, opened 2023-04-14
Token awareness in `execute_iter` doesn't work


piodul:
Found by Martins Purins on ScyllaDB user slack.

https://github.com/scylladb/scylla-rust-driver/blob/cacb276f955d015c5fbb7be84349520016712fac/scylla/src/transport/iterator.rs#L227-L233

The `RoutingInfo` constructed by `RowIterator` does not set the `keyspace` field. The name of the keyspace is necessary to look up information about the replication strategy of the keyspace and calculate replicas correctly.

Until `0.7.0` it used to partially work because the token aware policy would fall back to `SimpleStrategy` with `rf = 1` which would just choose the primary replica. However, the new implementation of the load balancing module just falls back to choosing a random node instead.

PR #700, opened 2023-04-14
iterator: make token awareness work again with execute_iter

piodul:
Due to an oversight, when the RowIterator asks the load balancing policy to calculate the plan (sequence of nodes that will be tried), it passes a RoutingInfo structure without providing the keyspace relevant to the query. The old token aware policy used to fall back to SimpleStrategy and rf=1, which could provide an illusion of token-aware routing. However, the new default load balancing policy will, in the case described above, treat this as if no routing information is available and will choose a random node.

This commit fixes the simple issue and adds a test which verifies that token awareness works both in case of `execute` and `execute_iter`.

Fixes: #699

## Pre-review checklist

- [ ] ~I have provided docstrings for the public items that I want to introduce~.
- [ ] ~I have adjusted the documentation in `./docs/source/`~.

### DIFF
--- a/scylla/src/transport/iterator.rs
+++ b/scylla/src/transport/iterator.rs
@@ -224,16 +224,16 @@ impl RowIterator {
             .unwrap_or(config.execution_profile.serial_consistency);
         let retry_session = config.execution_profile.retry_policy.new_session();
 
-        let statement_info = RoutingInfo {
-            consistency,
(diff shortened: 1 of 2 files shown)


piodul:
@cvybhu review ping

wprzytula (approved):

PR #705, opened 2023-04-20
session_builder: known_nodes accept IntoIterator

wprzytula:
There is no reason why not accept arbitrary `impl IntoIterator` in `SessionBuilder::[known_nodes(), known_nodes_addr()]`. It will also save an allocation in bindings.

This is an API mere _extension_, so it can be considered as **non**-API-breaking.

## Pre-review checklist


piodul (changes requested):
I'm not sure whether it's super important to save a small amount of allocations during the initialization of session. I guess it doesn't hurt...

...provided that this is a backwards-compatible change. The original methods were generic, so it's not clear to me that the new versions are fully compatible.

If you want to convince me and for this to be merged in 0.8.1 then please add a compile-time regression check, e.g.:

```rust
// Reference the PR in this comment
fn check_known_nodes_compatibility(hostnames: impl IntoIterator<Item = impl AsRef<str>>) {
    let mut sb: SessionBuilder = todo!();
    sb.add_known_nodes(hostnames);
}
```

... and an analogous one for each changed function.

wprzytula:
v2: added a compile-time compatibility assertion.

scylla/src/transport/session_builder.rs
```diff
@@ -982,4 +986,10 @@ mod tests {
         assert!(builder.config.keyspace_case_sensitive);
         assert!(!builder.config.fetch_schema_metadata);
     }
+
+    // This is to assert that #705 does not break the API (i.e. it merely extends it).
+    fn _check_known_nodes_compatibility(hostnames: &[impl AsRef<str>]) {
+        let sb: SessionBuilder = SessionBuilder::new();
+        sb.known_nodes(hostnames);
+    }
```
piodul, line 994: What about a test for `SessionConfig::add_known_nodes_addr` and `SessionConfig::add_known_nodes`? At least the first one has a different signature than what is checked here.

piodul (approved):

PR #708, opened 2023-04-21
session_test: fix flakiness due to slow Cassandra tracing

piodul:
The driver's default configuration for fetching traces from the DB is quite unforgiving. After the query finishes, it tries 5 times to fetch tracing information, waiting 3ms between each attempt. This is too fast for Cassandra in CI and the `test_token_awareness` fails due to that.

This commit modifies the test so that it uses a much more relaxed configuration: 50 attempts with 200ms sleeps in between. As a long term solution, we should consider changing the default parameters.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -562,6 +562,14 @@ async fn test_token_awareness() {
         .unwrap();
     prepared_statement.set_tracing(true);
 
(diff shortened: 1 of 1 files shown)


piodul:
Example of a failed run: https://github.com/scylladb/scylla-rust-driver/actions/runs/4766264360/jobs/8473054921

havaker (approved):

PR #712, opened 2023-04-25
docs/batch: clarify the wording around what can be put in batches

piodul:
This commit clarifies the doc page about batches and performs two changes:

- Occurrences of "query" have been changed to "statement". Batches do not support putting queries into them, only data-modifying statements. The driver documentation interchangeably uses "query" and "statement" which is not really correct but is especially confusing in this page.
- The first paragraph now also lists `DELETE` as a possible operation for batch in addition to `INSERT` and `UPDATE`.

As a bonus, a single backslash that was unnecessarily rendered at the end of a line was removed.

## Pre-review checklist

- [ ] ~I have split my patch into logically separate commits.~

### DIFF
--- a/docs/source/queries/batch.md
+++ b/docs/source/queries/batch.md
@@ -1,8 +1,8 @@
 # Batch statement
 
-A batch statement allows to run many queries at once.\
-These queries can be [simple queries](simple.md) or [prepared queries](prepared.md).\
-Only queries like `INSERT` or `UPDATE` can be in a batch.
+A batch statement allows to execute many data-modifying statements at once.\
+These statements can be [simple](simple.md) or [prepared](prepared.md).\
+Only `INSERT`, `UPDATE` and `DELETE` statements are allowed.
 
 ```rust
 # extern crate scylla;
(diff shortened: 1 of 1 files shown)


docs/source/queries/batch.md
```diff
@@ -1,8 +1,8 @@
 # Batch statement
 
-A batch statement allows to run many queries at once.\
-These queries can be [simple queries](simple.md) or [prepared queries](prepared.md).\
-Only queries like `INSERT` or `UPDATE` can be in a batch.
+A batch statement allows to run many statements at once.\
+These statements can be [simple](simple.md) or [prepared](prepared.md).\
+Only `INSERT`, `UPDATE` and `DELETE` statements are allowed.
```
wprzytula, line 5: I'm not convinced about _"running statements"_. _Statement,_ as I understand it, is a passive object containing a string contents ("the statement"), whereas `running` implies some kind of a process, an action. A _query_, on the other hand, is a process of asking the DB for some information. To sum up, "_running_" goes better with some action word.
Another word denoting a related process is _request_.  _Request_ is wider than _query_ in that it allows both for `SELECT`s (which are essentially questions to the DB) and mutating statements.
I suggest replacing _"running statements"_ as well as _"running queries"_ with _"running requests"_ or _"issuing requests"_.

Another combination that I find acceptable is _"executing statements"_, but _"requests"_ are IMO more suitable.

piodul: > I'm not convinced about _"running statements"_. _Statement,_ as I understand it, is a passive object containing a string contents ("the statement"), whereas `running` implies some kind of a process, an action. A _query_, on the other hand, is a process of asking the DB for some information. To sum up, "_running_" goes better with some action word.

Yes, "running statements" sounds a bit strange.

> Another word denoting a related process is _request_. _Request_ is wider than _query_ in that it allows both for `SELECT`s (which are essentially questions to the DB) and mutating statements. I suggest replacing _"running statements"_ as well as _"running queries"_ with _"running requests"_ or _"issuing requests"_.

No, "request" is too broad in my opinion. "Statement" is more specific, and maybe not specific enough as "statement" can mean both "INSERT"/"UPDATE"/"DELETE" which can be sent in batches as well as "SELECT"s which cannot.

> Another combination that I find acceptable is _"executing statements"_, but _"requests"_ are IMO more suitable.

"Statement execution" is a phrase that appears in our docs (https://docs.scylladb.com/stable/cql/dml.html#batch-statement), so I'll change to "executing statements".

wprzytula (approved):

Issue #710, opened 2023-04-21
`latency_aware_default_policy_stops_penalising_after_min_average_increases_enough_only_after_update_rate_elapses` is flaky


piodul:
One of the latency aware policy tests seems to be flaky - a failure has been observed:

```
---- transport::load_balancing::default::latency_awareness::tests::latency_aware_default_policy_stops_penalising_after_min_average_increases_enough_only_after_update_rate_elapses stdout ----
Set latency: node 1, latency Some(TimestampedAverage { timestamp: Instant { tv_sec: 710, tv_nsec: 993332057 }, average: 21ms, num_measures: 50 }).
Set latency: node 3, latency Some(TimestampedAverage { timestamp: Instant { tv_sec: 710, tv_nsec: 993332857 }, average: 10ms, num_measures: 50 }).
Set node latency stats.
thread 'transport::load_balancing::default::latency_awareness::tests::latency_aware_default_policy_stops_penalising_after_min_average_increases_enough_only_after_update_rate_elapses' panicked at 'assertion failed: `(left == right)`
  left: `{1, 5}`,
 right: `{5, 4}`', scylla/src/transport/load_balancing/default.rs:775:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Example: https://github.com/scylladb/scylla-rust-driver/actions/runs/4766690919/jobs/8474024312

cc: @wprzytula

PR #714, opened 2023-04-25
latency_awareness: Combat tests flakiness by not relying on timing

wprzytula:
The test bearing an undeniably attractive name: `latency_aware_default_policy_stops_penalising_after_min_average_increases_enough_only_after_update_rate_elapses` has shown to be flaky. I suspect it is due to the fact that it relies on timing to be valid, and execution in CI is often extremely slow. The problem, most probably, lies in minimum average updates done periodically: the test code assumes that an update has not yet been triggered at some point, which in CI happens to not be true.
To combat that:
- min avg update logic is extracted into `MinAvgUpdater` struct
- it is made configurable upon LA creation whether updates are to be scheduled to be periodically issued (as before, used in real cases) _or_ triggered manually by calls on `MinAvgUpdater`, 
- the affected tests (those that have relied on timing) are changed to use the `MinAvgUpdater` explicitly instead of scheduling periodic updates. This prevents flakiness.

Fixes: #710 

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -1837,6 +1837,22 @@ mod latency_awareness {
                 minimum_measurements,
             )
         }
+
+        #[cfg(test)]
+        fn build_for_test(self) -> (LatencyAwareness, MinAvgUpdater) {
+            let Self {
+                exclusion_threshold,
+                retry_period,
+                update_rate,
+                minimum_measurements,
+            } = self;
+            LatencyAwareness::new_for_test(
+                exclusion_threshold,
+                retry_period,
+                update_rate,
+                minimum_measurements,
+            )
+        }
     }
 
     impl Default for LatencyAwarenessBuilder {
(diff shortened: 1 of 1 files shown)


piodul (changes requested):
I don't like this approach as it still relies on timing. You are making the test much slower to reduce the probability of failure, yet you still don't have a guarantee that a spurious failure won't occur.

I suggest changing the test case not to rely on timing at all. As I understand the problem here (btw, the PR description doesn't tell what is the _cause_ of the flakiness) is that the test waits for some period of time and hopes that a task spawned by the latency aware policy (the updater) will do some some work. Instead of waiting and praying, you can control when the updater does its work.

- Introduce a `LatencyAwareUpdater` struct with a single method `async fn tick()`,
- Introduce `fn new_for_test(...) -> (Self, LatencyAwareUpdater)` and reimplement `fn new() -> Self` using `new_for_test`,
- Use `new_for_test()` in the tests that need to control when the updater does the tick.

wprzytula:
> I don't like this approach as it still relies on timing. You are making the test much slower to reduce the probability of failure, yet you still don't have a guarantee that a spurious failure won't occur.
> 
> I suggest changing the test case not to rely on timing at all. As I understand the problem here (btw, the PR description doesn't tell what is the _cause_ of the flakiness) is that the test waits for some period of time and hopes that a task spawned by the latency aware policy (the updater) will do some some work. Instead of waiting and praying, you can control when the updater does its work.
> 
>     * Introduce a `LatencyAwareUpdater` struct with a single method `async fn tick()`,
> 
>     * Introduce `fn new_for_test(...) -> (Self, LatencyAwareUpdater)` and reimplement `fn new() -> Self` using `new_for_test`,
> 
>     * Use `new_for_test()` in the tests that need to control when the updater does the tick.

v2: taken the suggested approach:
- extracted update logic into `MinAvgUpdater`,
- implemented `LatencyAwareness::new_for_test()`,
- modified tests not to rely on timing, by disabling periodic min avg updates and invoking them manually instead.

piodul (approved):
LGTM. Please update the title and description of the PR because it is no longer describes what the PR really does. Then, I'll merge.

Issue #589, opened 2022-11-07
Logging example doesn't produce a warning


cvybhu:
The example `logging.rs` is supposed to produce a warning by doing a `USE KEYSPACE` query, but it doesn't.
There's a comment saying that this will print out a warning.
This particular warning must've been removed a while ago, other warning are printed without problems.
This can be really confusing, the user will think that they are doing something wrong.

```bash
$ RUST_LOG=info cargo run --example logging
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/examples/logging`
2022-11-07T10:24:32.035956Z  INFO logging: Connecting to 127.0.0.1:9042
$
```

cc @havaker

PR #716, opened 2023-04-26
examples: logging: removed obsolete comment

wprzytula:
The executed request no longer produces any warnings. This was changed in 2af7800c: `session: improve use keyspace queries detection`, since when `USE KEYSPACE ` statement is no longer parsed by the driver. As since then the parsing is delegated to the database (and therefore considered stable), the warning about sending raw `USE KEYSPACE` being experimental is no longer relevant.
Fixes: #589 
## Pre-review checklist

- ~~[ ] I have split my patch into logically separate commits.~~

### DIFF
--- a/examples/logging.rs
+++ b/examples/logging.rs
@@ -19,7 +19,6 @@ async fn main() -> Result<()> {
     let session: Session = SessionBuilder::new().known_node(uri).build().await?;
     session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1}", &[]).await?;
 
-    // This query should generate a warning message
     session.query("USE ks", &[]).await?;
 
     Ok(())


piodul:
I would appreciate if the description were more comprehensive. The description in the linked issue suggests that the warning is obsolete but doesn't explain why, only that it "must've been removed a while ago", which isn't a good argument - what if the warning is supposed to be printed but no longer is due to a bug?

Including a proper motivation in a pull request/commit shows that the submitter understands the problem that they solve. https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence

wprzytula:
> I would appreciate if the description were more comprehensive. The description in the linked issue suggests that the warning is obsolete but doesn't explain why, only that it "must've been removed a while ago", which isn't a good argument - what if the warning is supposed to be printed but no longer is due to a bug?
> 
> Including a proper motivation in a pull request/commit shows that the submitter understands the problem that they solve. https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence

As you requested, I did some research and found the commit that removed the warning, as well as provided relevant explanation.

PR #725, opened 2023-05-22
connection: make it possible to switch off write coalescing

piodul:
Each connection is managed by a tokio task that reads from and writes to a single socket. The task (called internally as "router") polls, amongst others, a future that collects write requests from a channel and pushes them to the socket. It currently has a heuristic that is meant to be a throughput optimization: if there are no more write requests on the channel, the future yields, hoping to give an opportunity for more write requests to arrive on the channel. This improves efficiency as it allows more requests to be batched together into a single write and reduces the number of syscalls issued.

The described optimization was implemented in:
e1c0b6f3b48eac2f60c043c05fc0b806189788c4

However, this optimization obviously can cause worse latency in some scenarios. For example, if there aren't many requests being issued but the application otherwise runs lots of unrelated tokio tasks, then a yielded task can be rescheduled again only after a significant delay. If the request rate is low enough then batching won't ever bring any benefit but will only make the latency worse.

In order to help in those scenarios, this commit adds a flag that allows turning off the write coalescing optimization. Write coalescing is kept turned on by default, but it can be easily disabled when constructing a new session.

Results of a test with modified cql-stress that runs 100 tasks which spin for 0.25ms and then yield, in a loop (https://github.com/piodul/cql-stress/commit/f20050e3a3514af20f431e0c4c187cd254500fbd):

```
cargo run --release --bin cql-stress-scylla-bench -- -max-rate 10000 -mode write -workload uniform -partition-count 128 -concurrency 16 -nodes 127.0.0.1 -duration 10s -noise-task-count 100 -noise-task-hog-duration 250us
```

With the coalescing optimization _enabled_:

```
  99h:          11.5ms
  95h:          7.21ms
  90h:          6.25ms
  median:       3.74ms
  mean:         4.11ms
```

With the coalescing optimization _disabled_:

```
  99h:          6.24ms
  95h:          4.89ms
  90h:          4.25ms
  median:       2.32ms
  mean:         2.44ms
```

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -300,6 +300,7 @@ pub struct ConnectionConfig {
     pub(crate) cloud_config: Option<Arc<CloudConfig>>,
     pub authenticator: Option<Arc<dyn AuthenticatorProvider>>,
     pub address_translator: Option<Arc<dyn AddressTranslator>>,
+    pub enable_write_coalescing: bool,
 }
 
 impl Default for ConnectionConfig {
@@ -316,6 +317,7 @@ impl Default for ConnectionConfig {
             address_translator: None,
             #[cfg(feature = "cloud")]
             cloud_config: None,
+            enable_write_coalescing: true,
         }
     }
 }
@@ -901,6 +903,8 @@ impl Connection {
         // across .await points. Therefore, it should not be too expensive.
         let handler_map = StdMutex::new(ResponseHandlerMap::new());
 
+        let enable_write_coalescing = config.enable_write_coalescing;
+
         let r = Self::reader(
             BufReader::with_capacity(8192, read_half),
             &handler_map,
@@ -910,6 +914,7 @@ impl Connection {
             BufWriter::with_capacity(8192, write_half),
             &handler_map,
             receiver,
+            enable_write_coalescing,
         );
         let o = Self::orphaner(&handler_map, orphan_notification_receiver);
 
@@ -1019,6 +1024,7 @@ impl Connection {
         mut write_half: (impl AsyncWrite + Unpin),
         handler_map: &StdMutex<ResponseHandlerMap>,
         mut task_receiver: mpsc::Receiver<Task>,
+        enable_write_coalescing: bool,
     ) -> Result<(), QueryError> {
         // When the Connection object is dropped, the sender half
         // of the channel will be dropped, this task will return an error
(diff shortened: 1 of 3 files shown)


piodul:
Ran the CI with coalescing disabled by default, all tests passed, e.g. here: https://github.com/scylladb/scylla-rust-driver/actions/runs/5047296411/jobs/9053966160?pr=725

piodul:
v1.1: added a test

scylla/src/transport/connection.rs
```diff
@@ -1721,6 +1728,112 @@ mod tests {
...
+                    futures::future::join_all(futs).await;
+                }));
+
+                tokio::task::yield_now().await;
+            }
+
+            futures::future::join_all(futs).await;
+
+            // Check that everything was written properly
+            let range_end = (10 - 1) * 10 / 2;
```
wprzytula, line 1812: It might be clearer IMO if you make `10` some meaningfully named variable instead of a magic constant.

scylla/src/transport/connection.rs
```diff
@@ -1721,6 +1728,112 @@ mod tests {
...
+            connection
+                .query(&"TRUNCATE t".into(), (), None)
+                .await
+                .unwrap();
+
+            let mut futs = Vec::new();
+
+            for batch_size in 0..10 {
+                // Each future should issue more and more queries in the first poll
+                let base = (batch_size - 1) * batch_size / 2;
```
wprzytula, line 1791: What does this formula mean?

piodul: `n * (n - 1) / 2` = `1 + 2 + .. n - 1`

I'll extract it to a helper function and name it.

scylla/src/transport/connection.rs
piodul: Done, extracted to a constant: `NUM_BATCHES`

piodul:
v2:
- Extracted the arithmetic sum expression in the test to a helper function `arithmetic_sequence_sum`,
- Introduced a name for the number of batches in the test (`NUM_BATCHES`).

wprzytula (approved):

piodul:
v2.1: no changes to the code, only updated the PR and commit descriptions with the result of my benchmark.

piodul:
v2.2 rephrased the comments about the newly added option - the previous "driver is run in a busy system and there is not much load" could be a bit confusing

wprzytula (approved):

Issue #613, opened 2022-12-13
Remove any uses of SimpleStrategy


wprzytula:
As described [here](https://github.com/scylladb/scylladb/issues/8638), there is an ongoing effort on banning use of SimpleStrategy in favour of NetworkTopologyStrategy. Thus, we need to examine all its uses in the driver tests/examples and remove them, because eventually they will stop working.

PR #726, opened 2023-05-22
treewide: remove usages of SimpleStrategy

cvybhu:
As described in https://github.com/scylladb/scylladb/issues/8638, we're moving away from `SimpleStrategy`, in the future it will become deprecated.

We should remove all uses of it and replace them with `NetworkTopologyStrategy`.

The typical way to use `NetworkTopologyStrategy` is to specify a replication factor for each datacenter. This could be a bit cumbersome, we would have to fetch the list of datacenters, set the repfactors, etc.

Luckily there is another way - we can just specify a replication factor to use for or each existing datacenter, like this:
```cql
CREATE KEYSPACE {} WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1};
```

This makes the change rather straightforward - just replace all instances of `'SimpleStrategy'`, with `'NetworkTopologyStrategy'`.

Fixes: #613

## Pre-review checklist

### DIFF
--- a/docs/source/logging/logging.md
+++ b/docs/source/logging/logging.md
@@ -28,7 +28,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     session
         .query(
             "CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = \
-            {'class' : 'SimpleStrategy', 'replication_factor' : 1}",
+            {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}",
             &[],
         )
         .await?;
--- a/docs/source/quickstart/example.md
+++ b/docs/source/quickstart/example.md
@@ -24,7 +24,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     session
         .query(
             "CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = \
-            {'class' : 'SimpleStrategy', 'replication_factor' : 1}",
+            {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}",
             &[],
         )
         .await?;
--- a/examples/allocations.rs
+++ b/examples/allocations.rs
@@ -131,7 +131,7 @@ async fn main() -> Result<()> {
     let session: Session = SessionBuilder::new().known_node(args.node).build().await?;
     let session = Arc::new(session);
 
-    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1}", &[]).await?;
+    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await?;
     session.await_schema_agreement().await.unwrap();
 
     session
--- a/examples/auth.rs
+++ b/examples/auth.rs
@@ -14,7 +14,7 @@ async fn main() -> Result<()> {
         .await
         .unwrap();
 
-    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1}", &[]).await.unwrap();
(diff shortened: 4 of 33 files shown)


cvybhu:
One sad consequence is that after this change we won't test how the driver interacts with `SimpleStrategy` keyspaces. There might be a regression for `SimpleStrategy` and we won't notice it :/

I guess we could add a test that runs a few queries on a `SimpleStrategy` keyspace, it would protect us until the `SimpleStrategy` is really deprecated.

cvybhu:
v2:
* add a single test with `SimpleStrategy`

wprzytula:
> One sad consequence is that after this change we won't test how the driver interacts with `SimpleStrategy` keyspaces. There might be a regression for `SimpleStrategy` and we won't notice it :/

For this reason I'm against removing usages of `SimpleStrategy` from tests until it is really deprecated.
Side question: is `SimpleStrategy` going to be still supported in Cassandra? If so, then for compatibility we should test it even after it gets deprecated in Scylla.  

> I guess we could add a test that runs a few queries on a `SimpleStrategy` keyspace, it would protect us until the `SimpleStrategy` is really deprecated.

Such a small bunch of tests is inherently weaker than the whole set that we have currently. Again, I'm calling for leaving tests untouched for the time being.

Of course, we **definitely should** remove `SimpleStrategy` from documentation and examples, and this is what this PR should contain.

wprzytula:
Any updates on this, @cvybhu?

cvybhu:
> Any updates on this, @cvybhu?

Not really, the quality week has ended and I have other high priority stuff to do.
I hope to come back to this in a free moment somewhere in the future.

mykaul:
> > Any updates on this, @cvybhu?
> 
> Not really, the quality week has ended and I have other high priority stuff to do. I hope to come back to this in a free moment somewhere in the future.

Other than customer issues, this is a very high priority item for us.

cvybhu:
> Other than customer issues, this is a very high priority item for us.

Hm ok, I'm moving it up in my queue.

ptrsmrn:
> > One sad consequence is that after this change we won't test how the driver interacts with `SimpleStrategy` keyspaces. There might be a regression for `SimpleStrategy` and we won't notice it :/
> 
> For this reason I'm against removing usages of `SimpleStrategy` from tests until it is really deprecated. Side question: is `SimpleStrategy` going to be still supported in Cassandra? If so, then for compatibility we should test it even after it gets deprecated in Scylla.

AFAIK `SimpleStrategy` will never be completely removed, because our clients/users started to use it and will continue to, and we cannot break their setups. Also, SimpleStrategy is considered to be deprecated, and we're just waiting for some official note. But it'll probably never get obsoleted, for the reason I explained above (clients).
Cassandra's compatibility isn't enough to keep our clients using SimpleStrategy.  

Most of the tests don't care about the strategy, it's there because it has to, and only a few really test the strategy itself, so switching from SimpleStrategy to NetworkTopologyStrategy isn't really a "big deal" from tests' perspective, meaning - it doesn't change what the tests test. As we want to rely on NetworkTopologyStrategy rather than SimpleStrategy, it's also a natural move for us to move away from SimpleStrategy.
> 
> > I guess we could add a test that runs a few queries on a `SimpleStrategy` keyspace, it would protect us until the `SimpleStrategy` is really deprecated.
> 
> Such a small bunch of tests is inherently weaker than the whole set that we have currently. Again, I'm calling for leaving tests untouched for the time being.

The tests that really validate strategy should be left untouched, but for others it doesn't make a real difference. We should also ensure we really have tests that test both strategies, but this is a next step.
> Of course, we **definitely should** remove `SimpleStrategy` from documentation and examples, and this is what this PR should contain.

This is an ongoing process. It'll take time to remove most of the mentions, I don't expect we'll ever going to remove all of them.

cvybhu:
v2.1:
* rebased onto the latest main

ptrsmrn (approved):

piodul (commented):
Nit: in the second commit, "it's still a case that is supported" -> "it's still the case that is supported". Additionally, please adjust the comment according to @ptrsmrn's request.

Apart from that, LGTM.

cvybhu:
v3:
* Updated the comment

piodul (approved):

Issue #727, opened 2023-05-23
Remove clone.rs example


cvybhu:
The example: https://github.com/scylladb/scylla-rust-driver/blob/main/examples/clone.rs
This example creates many sessions using a single `SessionBuilder`.

I think it would be better to remove it, it doesn't really do anything interesting, it just uses a Builder to Build.
The example is a bit harmful because it promotes usage of multiple `Sessions`, which is usually a bad idea.
The users might take a look at the example and start using multiple `Sessions` because of it (e.g. https://github.com/scylladb/scylla-rust-driver/issues/724#issuecomment-1557645457)

Issue #724, opened 2023-05-21
Scylla DB Session: Technical Questions Regarding Sessions


WarrenN1:
I know you guys are busy and your time is valuable so I will keep this short. My questions are as follows:
1. Does SessionBuilder::new() manage an underlying connection pool or does session builder only have 1 connection?
2. Does SessionBuilder allow for sending multiple queries from multiple threads to the same session which can be completed at different times or does it have a query queue that is blocking in nature ie. sends 1 query, waits for response, sends next query etc.
3. Any recommendations for handling Sessions that makes use of async for proper web server connection pooling.

cvybhu:
Hi,

The general idea is that each application should have a single instance of `Session`. At the beginning you can create a `Session` using `SessionBuilder`, this should be done only once. During `SessionBuilder::build()` the `Session` creates its own pool of connections to the cluster, which is managed automatically. To share a `Session` between many threads it can be wrapped in an `Arc` (`Arc<Session>`). 

> 2. Does SessionBuilder allow for sending multiple queries from multiple threads to the same session which can be completed at different times or does it have a query queue that is blocking in nature ie. sends 1 query, waits for response, sends next query etc.

It allows sending multiple queries from multiple threads. Each call to `query` starts a new request in parallel to the other ones.

To sum up, create a single `Arc<Session>` and share it around your application. Connection pooling and query parallelism is taken care of automatically.

WarrenN1:
Wonderful design. Love to hear it. Wish more libraries/products did this.

cvybhu:
Have fun :)

Closing the issue then.

WarrenN1:
One more quick question, you said to wrap it in an Arc. Does your Session implement Send, Sync in its design or do I need to implement that (via wrapping it in a mutex)?

WarrenN1:
@cvybhu Ok wonderful. Was just checking given the example implied a different underlying structure. Could you point me to any resource in the documentation on the particulars of the implementation/event loop handling?

cvybhu:
> Could you point me to any resource in the documentation on the particulars of the implementation/event loop handling?

There isn't any such documentation. You can try reading the code in `connection.rs` and the adjacent modules.
This is an internal implementation detail that is abstracted away from driver users.

PR #728, opened 2023-05-24
examples: remove clone.rs example

cvybhu:
`clone.rs` is an old example that uses `SessionBuilder` to build many copies of `Session`.

I think this example is harmful because it promotes the use of multiple Sessions, which is usually a bad idea.

One user started using multiple `Sessions` instead of `Arc<Session>` because they saw such approach in this example: https://github.com/scylladb/scylla-rust-driver/issues/724#issuecomment-1557645457

The example itself isn't very useful. It just uses a Builder to Build sessions, this isn't anything tricky, it's what's expected of the Builder pattern.

I think it would be better to remove it to avoid misleading the users.

Fixes: #727

## Pre-review checklist


wprzytula (commented):
Although I agree that the particular example is misleading, I believe the aspect of using `Session` from many threads deserves better explanation in docs. We should:
- state that although building multiple `Session`s using a single `SessionBuilder` is possible, shared access to single `Session` through an `Arc` is prefered, 
- compare both options in terms of inner working (disadvantages of multiple `Session`s: duplicated connection pools, precomputed replicas etc., but (as advantages) possibly distinct configuration and clean separation of two workloads (?)).

Before this is described in docs, I oppose the mere removal of this example.

wprzytula:
Taken a look at #216, I think we should alter the `clone.rs` example to put `Session` inside an `Arc` and then use it concurrently from multiple threads. This would give us both:
- no confusion induced by creating multiple `Session`s,
- show the suggested way of sharing `Session`.
WDYT @cvybhu @piodul?

wprzytula (changes requested):
See #216.

avelanarius:
After a second thought, maybe another example is not really neccessary as https://github.com/scylladb/scylla-rust-driver/blob/main/examples/parallel.rs this already shows using `Session` in an `Arc`.

piodul:
I am in favor of removing the example. As @avelanarius wrote, we have the "parallel" example already. We don't need to change "clone" to look like "parallel".

We already mention that the session is able to handle multiple requests in parallel in the documentation, at the end of this page: https://rust-driver.docs.scylladb.com/stable/queries/queries.html

Perhaps we could also add it to the docstring for `Session` so that it appears in the API docs (on docs.rs) - currently the docstring is very laconic. That can be left for a followup.

piodul (approved):

piodul:
@wprzytula please approve and merge if you agree to remove the example.

Issue #718, opened 2023-05-09
Close connections if they don't respond to keepalives


piodul:
Currently, we send keepalives to prevent connections from being automatically closed by intermediate network devices due to inactivity. However, it might also be a good idea to close connections from the driver side if if is unable to send a successful keepalive within some period of time, as it might indicate a broken TCP connection.

Some other drivers already implement this:

- cpp: https://github.com/scylladb/cpp-driver/blob/eb6b70d138752c5e6a487eeaf69e6baa4d6652a4/include/cassandra.h#L2585-L2587
- python: https://github.com/scylladb/python-driver/blob/00329a3b34cc43d357afffaab52e80fa316119b7/cassandra/cluster.py#L890-L894

siningma:
1. What is the equivalent of `cass_cluster_set_connection_heartbeat_interval` ?
2. What is the equivalent of `cass_cluster_set_tcp_keepalive` ?

piodul:
> 1. What is the equivalent of `cass_cluster_set_connection_heartbeat_interval` ?

https://docs.rs/scylla/0.8.1/scylla/transport/session_builder/struct.GenericSessionBuilder.html#method.keepalive_interval

> 2. What is the equivalent of `cass_cluster_set_tcp_keepalive` ?

There is no equivalent. We considered it long time ago, but there was no way to set a TCP keepalive on a tokio socket. It looks like the situation has changed since then, so I reopened https://github.com/scylladb/scylla-rust-driver/issues/132 .

PR #731, opened 2023-05-26
Close connections on no response to keepalives 

wprzytula:
This PR aims to add a feature that has been supported in other drivers: closing connections when not receiving responses to keepalives. To do so:
- `SessionBuilder`, `SessionConfig` and `ConnectionConfig` get new parameter: `keepalive_timeout`,
- `Keepaliver` is refactored so that it no longer resides in the connection pool layer, but...
- ...it is moved into `Connection` layer, more specifically into `router`. This enables clean closing connection when the keepalive timeout elapses without receiving a response,
- a test is added for the new feature.

Fixes: #718

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -70,8 +70,8 @@ const LOCAL_VERSION: &str = "SELECT schema_version FROM system.local WHERE key='
 // FIXME: Make this constants configurable
 // The term "orphan" refers to stream ids, that were allocated for a {request, response} that no
 // one is waiting anymore (due to cancellation of `Connection::send_request`). Old orphan refers to
-// a stream id, that is orphaned for a long time. This long time is defined below
-// (`OLD_AGE_ORPHAN_THRESHOLD`). Connection, that has a big number (`OLD_ORPHAN_COUNT_THRESHOLD`)
+// a stream id that is orphaned for a long time. This long time is defined below
+// (`OLD_AGE_ORPHAN_THRESHOLD`). Connection that has a big number (`OLD_ORPHAN_COUNT_THRESHOLD`)
 // of old orphans is shut down (and created again by a connection management layer).
 const OLD_ORPHAN_COUNT_THRESHOLD: usize = 1024;
 const OLD_AGE_ORPHAN_THRESHOLD: std::time::Duration = std::time::Duration::from_secs(1);
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -77,12 +77,16 @@ const OLD_ORPHAN_COUNT_THRESHOLD: usize = 1024;
 const OLD_AGE_ORPHAN_THRESHOLD: std::time::Duration = std::time::Duration::from_secs(1);
 
 pub struct Connection {
-    submit_channel: mpsc::Sender<Task>,
     _worker_handle: RemoteHandle<()>,
 
     connect_address: SocketAddr,
     config: ConnectionConfig,
     features: ConnectionFeatures,
+    router_handle: Arc<RouterHandle>,
+}
+
+struct RouterHandle {
+    submit_channel: mpsc::Sender<Task>,
 
     // Each request send by `Connection::send_request` needs a unique request id.
     // This field is a monotonic generator of such ids.
@@ -93,6 +97,60 @@ pub struct Connection {
     // pushing values in a synchronous way (without an `.await`), which is
     // needed for pushing values in `Drop` implementations.
     orphan_notification_sender: mpsc::UnboundedSender<RequestId>,
+}
+
+impl RouterHandle {
+    fn allocate_request_id(&self) -> RequestId {
+        self.request_id_generator
+            .fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+    }
+
+    async fn send_request(
+        &self,
+        request: &impl Request,
+        compression: Option<Compression>,
+        tracing: bool,
+    ) -> Result<TaskResponse, QueryError> {
+        let serialized_request = SerializedRequest::make(request, compression, tracing)?;
+        let request_id = self.allocate_request_id();
+
+        let (response_sender, receiver) = oneshot::channel();
+        let response_handler = ResponseHandler {
+            response_sender,
+            request_id,
+        };
+
+        // Dropping `notifier` (before calling `notifier.disable()`) will send a notification to
+        // `Connection::router`. This notification is then used to mark a `stream_id` associated
+        // with this request as orphaned and free associated resources.
+        let notifier = OrphanhoodNotifier::new(request_id, &self.orphan_notification_sender);
+
+        self.submit_channel
+            .send(Task {
+                serialized_request,
+                response_handler,
+            })
+            .await
+            .map_err(|_| {
+                QueryError::IoError(Arc::new(std::io::Error::new(
+                    ErrorKind::Other,
+                    "Connection broken",
+                )))
+            })?;
+
+        let task_response = receiver.await.map_err(|_| {
+            QueryError::IoError(Arc::new(std::io::Error::new(
+                ErrorKind::Other,
+                "Connection broken",
+            )))
+        })?;
+
+        // Response was successfully received, so it's time to disable
+        // notification about orphaning.
+        notifier.disable();
+
+        task_response
+    }
 }
 
 #[derive(Default)]
@@ -422,6 +480,12 @@ impl Connection {
         // Unbounded because it allows for synchronous pushes
         let (orphan_notification_sender, orphan_notification_receiver) = mpsc::unbounded_channel();
 
+        let router_handle = Arc::new(RouterHandle {
+            submit_channel: sender,
+            request_id_generator: AtomicU64::new(0),
+            orphan_notification_sender,
+        });
+
         let _worker_handle = Self::run_router(
             config.clone(),
             stream,
@@ -432,13 +496,11 @@ impl Connection {
         .await?;
 
         let connection = Connection {
-            submit_channel: sender,
             _worker_handle,
             config,
             features: Default::default(),
             connect_address: addr,
-            request_id_generator: AtomicU64::new(0),
-            orphan_notification_sender,
+            router_handle,
         };
 
         Ok((connection, error_receiver))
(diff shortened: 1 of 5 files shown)


scylla/src/transport/connection.rs
```diff
@@ -1136,6 +1144,64 @@ impl Connection {
...
+                                // Sending keepalives finished before receiving new tick.
+                                // Wait for the new tick and start over.
+                                interval.tick().await;
+                            }
+                            Err(err) => {
+                                warn!(
+                                    "Failed to execute keepalive request on a connection to node {} - {}",
+                                    node_address, err
+                                );
+                                return Err(QueryError::NoResponseToKeepalive)
```
piodul, line 1220: Nit: missing semicolon at the end of line

scylla/src/transport/connection.rs
```diff
@@ -1157,6 +1157,31 @@ impl Connection {
...
+                    std::future::pending::<()>().await;
+                }
+            }
+
+            fn reset(&mut self) {
+                if let Some(interval) = &mut self.0 {
+                    interval.reset();
+                }
+            }
+        }
```
piodul, line 1183: I don't think there is a need for an ad-hoc abstraction like `Timeouter`. I can see two ways to express it differently and more simply:

- Use an `if` guard in the `select` statement - requires an unwrap:
  ```rust
  tokio::select! {
      () = tokio::time::sleep(keepalive_interval.unwrap()) if keepalive_interval.is_some() => {
          // ...
      },
  }
  ```
- Change `issue_keepalive_query` so that it accepts an optional timeout and enforces it.

scylla/src/transport/connection.rs
```diff
@@ -1168,11 +1193,21 @@ impl Connection {
...
+            let mut timeouter = Timeouter::new(keepalive_timeout);
             loop {
+                let timeout = timeouter.tick();
                 tokio::select! {
+                    () = timeout => {
+                        warn!(
+                            "Timeout while waiting for response to keepalive request on a connection to node {}",
+                            node_address
+                        );
+                        return Err(QueryError::NoResponseToKeepalive)
```
piodul, line 1205: Nit: missing semicolon at the end of line

scylla/src/transport/connection.rs
```diff
@@ -93,6 +97,60 @@ pub struct Connection {
     orphan_notification_sender: mpsc::UnboundedSender<RequestId>,
 }
 
+impl RouterHandle {
+    fn allocate_request_id(&self) -> RequestId {
+        self.request_id_generator
+            .fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+    }
+
+    async fn issue_request(
```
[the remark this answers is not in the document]
wprzytula: `send_request()` is a method on `Connection`, whereas `issue_request()` is a new method on `RouterHandle` (which is essentially the extracted core of old `Connection::send_request()`) . I chose a distinct name to make them easier distinguishable.

scylla/src/transport/connection.rs
```diff
@@ -437,6 +437,8 @@ impl Connection {
             receiver,
             error_sender,
             orphan_notification_receiver,
+            router_handle.clone(),
+            addr.ip(),
```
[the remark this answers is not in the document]
wprzytula: It will be used for better log messages from keepaliver.

scylla/src/transport/connection.rs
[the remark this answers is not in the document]
piodul: If there are two of them and they do the same thing (send a request), then they should have the same name. Using a different word suggests that there might be some difference in semantics.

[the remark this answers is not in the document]
piodul: I don't think this issue was resolved. I still think this method should be named `send_request`. I can still see it still named `issue_request` in the PR.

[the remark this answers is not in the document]
wprzytula: I have a habit of marking an issue as resolved as soon as it is done in my offline version - this allows me cleaner management over issues. I agree that the version online is still unresolved.

scylla-cql/src/errors.rs
```diff
@@ -49,6 +49,10 @@ pub enum QueryError {
     /// Address translation failed
     #[error("Address translation failed: {0}")]
     TranslationError(#[from] TranslationError),
+
+    /// Server didn't respond to a keepalive.
+    #[error("Server didn't respond to a keepalive")]
+    NoResponseToKeepalive,
```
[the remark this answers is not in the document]
piodul: > Both Java and CPP drivers make the connection defunct on no response to keepalive.

Isn't that what the current version of the PR is doing? The error from keepaliver will interrupt in flight requests - an error will be sent to each one of them, the one that the keepaliver returns.

My concern wasn't about that. I was mostly concerned about the error variant returned and its error message. Could you check what kind of error other drivers return? Do they have a dedicated variant? What is the text of the returned message?

wprzytula:
It'd like to discuss whether it is worth the effort to implement sending keepalives only if the connection was idle for some period.

mykaul:
> It'd like to discuss whether it is worth the effort to implement sending keepalives only if the connection was idle for some period. WDYT @piodul @cvybhu @havaker ?

Unneeded optimization, IMHO. Keep-alive should be cheap, just send it always.

scylla/src/transport/connection.rs
wprzytula: > * Use an `if` guard in the `select` statement - requires an unwrap:
>       ```rust
>       tokio::select! {
>           () = tokio::time::sleep(keepalive_interval.unwrap()) if keepalive_interval.is_some() => {
>               // ...
>           },
>       }
>       ```

I'm afraid this isn't semantically equivalent to the current code.
- in your suggestion, the sleep is interrupted and dropped once another `select!` branch completes, and this happens after every `keepalive_interval`. 
- in my version, the state (how much time has already elapsed _since sending the first keepalive that hasn't been followed by any response_ - sorry for the complicated phrase here; I dunno how to put it into easier words) is saved between iterations. 
Do you agree @piodul?

piodul: OK, I see why my suggestion is incorrect, but I think the first bullet point is easy to fix: introduce a mutable variable that keeps the time of the last successful keepalive, and then use `tokio::time::sleep_until(last_successful_keepalive_time + keepalive_timeout)`.

piodul:
> > It'd like to discuss whether it is worth the effort to implement sending keepalives only if the connection was idle for some period. WDYT @piodul @cvybhu @havaker ?
> 
> Unneeded optimization, IMHO. Keep-alive should be cheap, just send it always.

I agree, adjusting the current code is not worth the effort right now.

scylla-cql/src/errors.rs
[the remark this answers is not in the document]
piodul: Let's use an existing error variant, the `IoError` with an appropriate error message sounds like a good fit. Introducing a new error variant would be a breaking change.

scylla/src/transport/connection.rs
wprzytula: I've realised that both your suggestion and the current implementation are semantically different that what is desired (and what I stated above). The difference is that you suggest to count time since _last successful keepalive response_ (and this is done by the code), whereas what should be done is, most probably, to count time since _last keepalive not yet having faced any keepalive response_.

The code can be fixed by swapping lines, from:
```rust
timeouter.reset();
// Sending keepalives finished before receiving new tick.
// Wait for the new tick and start over.
interval.tick().await;

// implicit continue
```
to:
```rust
// Sending keepalives finished before receiving new tick.
// Wait for the new tick and start over.
interval.tick().await;
timeouter.reset();

// implicit continue
```

To illustrate the difference:
```
                     <-----------------------------> timeout len
           <-----------------------------> timeout len
    a      b         c                   d         e = d + (c - b)
----|------|---------|-------------------|---------|------------------>
a - keepalive 1 sent
b - received response to keepalive 1
c - keepalive 2 sent (no response arrived)
d - timeout by your suggestion and previous code
e - timeout by my description and updated code
```

wprzytula: I'm still in favor of `Timeouter`. Such abstractions, even if "ad-hoc", enable delegating details, while leaving only meaningful logic.

piodul: > ```
>                      <-----------------------------> timeout len
>            <-----------------------------> timeout len
>     a      b         c                   d         e = d + (c - b)
> ----|------|---------|-------------------|---------|------------------>
> a - keepalive 1 sent
> b - received response to keepalive 1
> c - keepalive 2 sent (no response arrived)
> d - timeout by your suggestion and previous code
> e - timeout by my description and updated code
> ```

What are the arguments in favor of your suggestion?

I think about it like this: a successful response gives us a reason to think that the connection works right now, therefore we should count the idle timeout starting from the last response. Glancing at the CPP driver implementation this [seems to be the case](https://github.com/scylladb/cpp-driver/blob/eb6b70d138752c5e6a487eeaf69e6baa4d6652a4/src/connection.cpp#L237), and the docstring seems to agree with it (emphasis mine):

* Sets the amount of time a connection is allowed to be _without a successful heartbeat response_ before being terminated and scheduled for reconnection.

I'd vote for this interpretation as it will make it easier for us to have the same behavior in cpp wrapper over the rust driver.

piodul: > I'm still in favor of `Timeouter`. Such abstractions, even if "ad-hoc", enable delegating details, while leaving only meaningful logic.

I am still not. I'm not convinced that this particular abstraction improves the readability rather than obfuscating the code. The amount of complexity it hides is tiny compared to the size of the type definition and overhead related to reading and understanding it. I think in this particular case it helps to be explicit about the details, and by using names such as `last_successful_keepalive_time` instead of non-descript `timeouter` it will actually _improve_ readability.

wprzytula: I refactored the code to remove `Timeouter` altogether.

piodul (commented):
Looks _almost_ good. The default keepalive timeout seems to effectively be 60s, but the docstring says that it's 30s. Please adjust the default to 30s to be aligned with Python driver's defaults.

piodul (approved):

Issue #650, opened 2023-02-28
Add `scale` parameter to `LatencyAwarePolicy`


Gor027:
Since a host's performance may vary over time and a time-weighted average score is computed after each request, a `scale` parameter would allow users to customize how fast the weight of older latencies should decrease over time. The configurable parameter exists in Java 3.x and C++ drivers, too. As the average latency score is calculated with a similar algorithm in the driver, the modified algorithm will look like the following:

```
delay = (now - prev_avg.timestamp) / scale
prev_weight = 1 - ln((delay + 1)) / delay
average = prev_weight * last_latency + (1 - prev_weight * prev_avg.average)
```

PR #733, opened 2023-05-31
Latency awareness: add `scale` parameter

wprzytula:
The parameter was omited accidentally while porting.

To the best of my understanding of the previous version of the code, the scale that was implicitly used before was equal to **1 second**. The new default value is, in line with other drivers, **100 milliseconds**. This implies that this is an API-breaking change.

Fixes: #650

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -1476,7 +1476,11 @@ mod latency_awareness {
     }
 
     impl TimestampedAverage {
-        pub(crate) fn compute_next(previous: Option<Self>, last_latency: Duration) -> Option<Self> {
+        pub(crate) fn compute_next(
+            previous: Option<Self>,
+            last_latency: Duration,
+            scale_secs: f64,
+        ) -> Option<Self> {
             let now = Instant::now();
             match previous {
                 prev if last_latency.is_zero() => prev,
@@ -1487,7 +1491,8 @@ mod latency_awareness {
                 }),
                 Some(prev_avg) => Some({
                     let delay = (now - prev_avg.timestamp).as_secs_f64();
-                    let prev_weight = (delay + 1.).ln() / delay;
+                    let scaled_delay = delay / scale_secs;
+                    let prev_weight = (scaled_delay + 1.).ln() / scaled_delay;
 
                     let last_latency_secs = last_latency.as_secs_f64();
                     let prev_avg_secs = prev_avg.average.as_secs_f64();
@@ -1511,6 +1516,7 @@ mod latency_awareness {
         pub(super) retry_period: Duration,
         pub(super) _update_rate: Duration,
         pub(super) minimum_measurements: usize,
+        pub(super) scale_secs: f64,
 
         /// Last minimum average latency that was noted among the nodes. It is updated every
         /// [update_rate](Self::_update_rate).
@@ -1534,6 +1540,7 @@ mod latency_awareness {
             retry_period: Duration,
             update_rate: Duration,
             minimum_measurements: usize,
+            scale: Duration,
         ) -> (Self, MinAvgUpdater) {
             let min_latency = Arc::new(AtomicDuration::new());
 
@@ -1553,6 +1560,7 @@ mod latency_awareness {
                     retry_period,
                     _update_rate: update_rate,
                     minimum_measurements,
+                    scale_secs: scale.as_secs_f64(),
                     last_min_latency: min_latency_clone,
                     node_avgs: node_avgs_clone,
                     _updater_handle: None,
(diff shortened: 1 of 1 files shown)


Lorak-mmk (approved):

wprzytula:
@piodul @cvybhu ping

scylla/src/transport/load_balancing/default.rs
```diff
@@ -1487,12 +1491,13 @@ mod latency_awareness {
...
                 Some(prev_avg) => Some({
                     let delay = (now - prev_avg.timestamp).as_secs_f64();
-                    let prev_weight = (delay + 1.).ln() / delay;
-                    let last_latency_nanos = last_latency.as_nanos() as f64;
-                    let prev_avg_nanos = prev_avg.average.as_nanos() as f64;
-                    let average = Duration::from_nanos(
-                        ((1. - prev_weight) * last_latency_nanos + prev_weight * prev_avg_nanos)
-                            .round() as u64,
+                    let scaled_delay = delay / scale_secs;
+                    let prev_weight = (delay + 1.).ln() / scaled_delay;
```
piodul, line 1495: Shouldn't this be `let prev_weight = (scaled_delay + 1.).ln() / scaled_delay`;?

scylla/src/transport/load_balancing/default.rs
```diff
@@ -1845,12 +1891,14 @@ mod latency_awareness {
...
                 update_rate,
                 minimum_measurements,
+                scale,
             } = self;
             LatencyAwareness::new_for_test(
                 exclusion_threshold,
                 retry_period,
                 update_rate,
                 minimum_measurements,
+                scale,
```
piodul, line 1901: Would it make sense to add some tests for the `scale` parameter?

scylla/src/transport/load_balancing/default.rs
```diff
@@ -1823,18 +1840,47 @@ mod latency_awareness {
...
+        ///
+        /// The `scale` provides control on how the weight given to older latencies decreases
+        /// over time. For a given host, if a new latency `l` is received at time `t`, and
+        /// the previously calculated average is `prev` calculated at time `t'`, then the
+        /// newly calculated average `avg` for that host is calculated thusly:
+        ///
+        /// ```text
+        /// d = (t - t') / scale
+        /// alpha = 1 - (ln(d+1) / d)
+        /// avg = alpha * l + (1 - alpha * prev)
```
piodul, line 1853: This looks wrong, shouldn't it be `avg = alpha * l + (1 - alpha) * prev` ? It looks like the closing paren is too far.

piodul (changes requested):

scylla/src/transport/load_balancing/default.rs
wprzytula: Indeed it should... thanks.

scylla/src/transport/load_balancing/default.rs
wprzytula: This is taken directly from Java driver 3.x docstring. Why do you think it is erroneous?

scylla/src/transport/load_balancing/default.rs
wprzytula: It would probably involve some meticulous calculations to find the penalisation threshold for given parameters (involving scale) and then testing that a node is penalised iff the threshold is crossed. I don't consider it worth our effort.

scylla/src/transport/load_balancing/default.rs
piodul: It looks like it was supposed to represent exponential smoothing from `l` to `prev` with an `alpha` weight. This neither corresponds to our code:

```rust
let average = Duration::from_secs_f64(
    (1. - prev_weight) * last_latency_secs + prev_weight * prev_avg_secs,
);
```

nor the original code from the cpp driver:

```c++
current_.average =
    static_cast<int64_t>((1.0 - weight) * latency_ns + weight * previous.average);
```

The code is correct, the docstring is just wrong.

wprzytula: I'll fix the docstring then. Should we bother with fixing the docstring in Java driver?

avelanarius: @wprzytula Yes, please send the PR to Java Driver. The easier path is to send it only to our fork, but if you're willing to take some more time (dealing with "Contributor License Agreement") you can also send to the upstream - according to https://github.com/datastax/python-driver/pull/1114#issuecomment-953564923 we have already signed all the neccessary documents to do this.

wprzytula: Done.

piodul (approved):

PR #735, opened 2023-06-01
connection: Add configurable TCP keepalives

wprzytula:
Similarly to other drivers, support for TCP-layer keepalives is added. This is possible, because tokio::net::TcpStream finally implements `AsFd`. This enables creating `socket2::SockRef`, which in turn allows for configuring the underlying socket.
Fixes: #132 

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -16,14 +16,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 default = []
 ssl = ["dep:tokio-openssl", "dep:openssl"]
-cloud = [
-    "ssl",
-    "scylla-cql/serde",
-    "dep:serde_yaml",
-    "dep:serde",
-    "dep:url",
-    "dep:base64",
-]
+cloud = ["ssl", "scylla-cql/serde", "dep:serde_yaml", "dep:serde", "dep:url", "dep:base64"]
 secret = ["scylla-cql/secret"]
 
 [dependencies]
@@ -34,7 +27,7 @@ bytes = "1.0.1"
 futures = "0.3.6"
 histogram = "0.6.9"
 num_enum = "0.5"
-tokio = { version = "1.12", features = ["net", "time", "io-util", "sync", "rt", "macros"] }
+tokio = { version = "1.27", features = ["net", "time", "io-util", "sync", "rt", "macros"] }
 snap = "1.0"
 uuid = { version = "1.0", features = ["v4"] }
 rand = "0.8.3"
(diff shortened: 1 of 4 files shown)


wprzytula:
I have no idea how to check TCP keepalives properly.

scylla/Cargo.toml
```diff
@@ -34,7 +27,7 @@ bytes = "1.0.1"
 futures = "0.3.6"
 histogram = "0.6.9"
 num_enum = "0.5"
-tokio = { version = "1.12", features = ["net", "time", "io-util", "sync", "rt", "macros"] }
+tokio = { version = "1.28", features = ["net", "time", "io-util", "sync", "rt", "macros"] }
```
piodul, line 30: What was the tokio version bump needed for? The code seems to compile without that.

piodul: Never mind, of course I changed it to `version = "1.12"` here and Cargo resolved it to version `1.28.2` anyway...

Testing it now again, it looks like at least version `1.27` is required.

piodul: I guess that a _minor_ version bump of a dependency that appears in the public interface is not considered to be a breaking change, although I wish the PR were more explicit about it in the commit message/PR description.

Let's change it to the lowest version that supports operating on sockets with the `socket2` crate (`1.27`?). The library shouldn't impose more constraints on the dependencies than necessary.

piodul:
I'm afraid that it won't be enough. The option that you are setting here only configures the period of inactivity after which keepalives start being sent. After that, system will start sending keepalives using a different interval and will close the connection after a particular number of retries. According to the [`tcp`](https://man7.org/linux/man-pages/man7/tcp.7.html) manpage, the default interval is 75 seconds and 9 attempts, meaning that the connection will be closed after additional 11 minutes. Those defaults are definitely too high for us.

I looked at how the cpp driver does it, and it uses the [`uv_tcp_keepalive`](https://github.com/libuv/libuv/blob/65c1402ee66a37d8cdd294f64bfd6def4b9fcf3f/src/unix/tcp.c#L380-L391) function which configures an interval of 1 second and 10 retries (the same as on Windows, apparently). I think those values are much more sensible and we can start with that.

The `socket2` library exposes `TcpKeepalive::with_interval` and `TcpKeepalive::with_retries` functions which can be used to implement the remaining functionality. It appears that those are not available on every system, so it looks like we will have to do conditional compilation based on `target_os` - look at how the TcpKeepalive methods are defined in the library, they have appropriate `#[cfg()]` directives.

piodul (approved):

PR #738, opened 2023-06-03
Add Session::first_shard_for_statement

### LINKED ISSUE #468: Exposing `Shard` ID publicly

Ten0:
Resolves #468

This is a follow-up on #508 and #658:
- To minimize CPU usage related to network operations when inserting a very large number of lines, it is relevant to batch.
- To batch in the most efficient manner, these batches have to be shard-aware. Since #508, `batch` will pick the shard of the first statement to send the query to. However it is left to the user to constitute the batches in such a way that the target shard is the same for all the elements of the batch.
- This was made *possible* by #658, but it was still very boilerplate-ish. I was waiting for #612 to be merged (amazing work btw! 😃) to implement a more direct and factored API (as that would use it).
- This new ~`Session::first_shard_for_statement(self, &PreparedStatement, &SerializedValues) -> Option<(Node, Option<Shard>)>` makes shard-aware batching easy on the users, by providing access to the first node and shard of the query plan.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -8,6 +8,7 @@ use crate::frame::types::LegacyConsistency;
 use crate::history;
 use crate::history::HistoryListener;
 use crate::retry_policy::RetryPolicy;
+use crate::routing;
 use arc_swap::ArcSwapOption;
 use async_trait::async_trait;
 use bytes::Bytes;
@@ -898,15 +899,7 @@ impl Session {
             .as_ref()
             .map(|pk| prepared.get_partitioner_name().hash(pk));
 
-        let statement_info = RoutingInfo {
-            consistency: prepared
-                .get_consistency()
-                .unwrap_or(self.default_execution_profile_handle.access().consistency),
-            serial_consistency: prepared.get_serial_consistency(),
-            token,
-            keyspace: prepared.get_keyspace_name(),
-            is_confirmed_lwt: prepared.is_confirmed_lwt(),
-        };
+        let statement_info = self.routing_info(prepared, token);
 
         let span =
             RequestSpan::new_prepared(partition_key.as_ref(), token, serialized_values.size());
@@ -1814,13 +1807,63 @@ impl Session {
         prepared: &PreparedStatement,
         serialized_values: &SerializedValues,
     ) -> Result<Option<Token>, QueryError> {
-        match self.calculate_partition_key(prepared, serialized_values) {
-            Ok(Some(partition_key)) => {
-                let partitioner_name = prepared.get_partitioner_name();
-                Ok(Some(partitioner_name.hash(&partition_key)))
-            }
-            Ok(None) => Ok(None),
-            Err(err) => Err(err),
+        Ok(self
+            .calculate_partition_key(prepared, serialized_values)?
+            .map(|partition_key| prepared.get_partitioner_name().hash(&partition_key)))
+    }
+
+    /// Get the first node/shard that the load balancer would target if running this query
+    ///
+    /// This may help constituting shard-aware batches
+    pub fn first_shard_for_statement(
+        &self,
+        prepared: &PreparedStatement,
+        serialized_values: &SerializedValues,
+    ) -> Result<Option<(Arc<Node>, Option<routing::Shard>)>, QueryError> {
+        let token = match self.calculate_token(prepared, serialized_values)? {
+            Some(token) => token,
+            None => return Ok(None),
+        };
+        let routing_info = self.routing_info(prepared, Some(token));
+        let cluster_data = self.cluster.get_data();
+        let execution_profile = prepared
+            .config
+            .execution_profile_handle
+            .as_ref()
+            .unwrap_or_else(|| self.get_default_execution_profile_handle())
+            .access();
+        let mut query_plan = load_balancing::Plan::new(
+            &*execution_profile.load_balancing_policy,
+            &routing_info,
+            &cluster_data,
+        );
+        // We can't return the full iterator here because the iterator borrows from local variables.
+        // In order to achieve that, two designs would be possible:
+        // - Construct a self-referential struct and implement iterator on it via e.g. Ouroboros
+        // - Take a closure as a parameter that will take the local iterator and return anything, and
+        //   this function would return directly what the closure returns
+        // Most likely though, people would use this for some kind of shard-awareness optimization for batching,
+        // and are consequently not interested in subsequent nodes.
+        // Until then, let's just expose this, as it is simpler
+        Ok(query_plan.next().map(move |node| {
+            let token = node.sharder().map(|sharder| sharder.shard_of(token));
+            (node.clone(), token)
+        }))
+    }
+
+    fn routing_info<'p>(
+        &self,
+        prepared: &'p PreparedStatement,
+        token: Option<Token>,
+    ) -> RoutingInfo<'p> {
+        RoutingInfo {
+            consistency: prepared
+                .get_consistency()
+                .unwrap_or(self.default_execution_profile_handle.access().consistency),
+            serial_consistency: prepared.get_serial_consistency(),
+            token,
+            keyspace: prepared.get_keyspace_name(),
+            is_confirmed_lwt: prepared.is_confirmed_lwt(),
         }
     }
 


Ten0:
One more clippy lint I don't agree with -.-
Type seems clear enough. Fine if I silence that one?

scylla/src/transport/session.rs
```diff
@@ -1814,13 +1807,63 @@ impl Session {
...
+    ) -> RoutingInfo<'p> {
+        RoutingInfo {
+            consistency: prepared
+                .get_consistency()
+                .unwrap_or(self.default_execution_profile_handle.access().consistency),
+            serial_consistency: prepared.get_serial_consistency(),
+            token,
+            keyspace: prepared.get_keyspace_name(),
+            is_confirmed_lwt: prepared.is_confirmed_lwt(),
         }
```
wprzytula, line 1868: The name of this method should clearly indicate that this is intended specifically for prepared statements, as `Session` constructs `RoutingInfo` for other request types as well.

scylla/src/transport/session.rs
```diff
@@ -1814,13 +1807,63 @@ impl Session {
...
-            }
-            Ok(None) => Ok(None),
-            Err(err) => Err(err),
+        Ok(self
+            .calculate_partition_key(prepared, serialized_values)?
+            .map(|partition_key| prepared.get_partitioner_name().hash(&partition_key)))
+    }
+
+    /// Get the first node/shard that the load balancer would target if running this query
+    ///
```
wprzytula, line 1816: Although the shard is fixed for a given token (at least until resharding occurs), the first node returned by a load balancer may, and likely will, vary from call to call. Taking a look at the default load balancer, `DefaultPolicy`, we can see that we choose a random replica for `pick()`, so `first_shard_for_statement()` may return a replica different that one that will be queried later, while executing the batch of interest.

scylla/src/transport/session.rs
```diff
@@ -1814,13 +1807,63 @@ impl Session {
...
+    /// This may help constituting shard-aware batches
+    pub fn first_shard_for_statement(
+        &self,
+        prepared: &PreparedStatement,
+        serialized_values: &SerializedValues,
+    ) -> Result<Option<(Arc<Node>, Option<routing::Shard>)>, QueryError> {
```
wprzytula, line 1823: Naming nit: as this returns a node along with a shard, please mention it in the function's name, e.g. `first_node_and_shard_for_statement()`.

scylla/src/transport/session.rs
```diff
@@ -1815,6 +1815,7 @@ impl Session {
     /// Get the first node/shard that the load balancer would target if running this query
     ///
     /// This may help constituting shard-aware batches
+    #[allow(clippy::type_complexity)]
     pub fn first_shard_for_statement(
         &self,
         prepared: &PreparedStatement,
```
wprzytula: Personally, I don't oppose silencing this lint. The return type is readable enough.

scylla/src/transport/session.rs
Ten0: The way I understand it a shard in a cluster is only entirely determined by the `(Node, ShardIdxOnNode)` tuple, which is what's returned here. It seems that it's rather `routing::Shard` that's poorly named, because it's named as if it uniquely determines a shard, but in fact it's only a `ShardIdxOnNode` `u32` and it needs to be associated with its `Node` to do so.

scylla/src/transport/session.rs
Ten0: Ah, that seems to be a problem indeed for [this approach](https://github.com/scylladb/scylla-rust-driver/issues/448#issuecomment-1125008817).
It seems that the only thing left for it to work ideally and provide the benefits of immensely reducing network-management CPU-cost while preserving shard awareness would be to make it deterministic (at least on the second run).

The main option I see for this would be to use some other part of the token to choose among the replicas.

scylla/src/transport/session.rs
Ten0: Since both involve prepared statements I've opted to name it routing_info_from_prepared_statement in hope to help distinguish that it's built as much as possible "only from this prepared statement" (all its components are extracted from there except `token`, unlike the other scenario (which is precisely for routing the batches towards their first replica shard)).

scylla/src/transport/session.rs
Ten0: It looks like another option would be to enforce the target shard the second time around by setting a custom `LoadBalancerPolicy` in `Batch.config.execution_profile_handle.load_balancing_policy`.

Ten0: I've implemented that second option.

scylla/src/transport/session.rs
wprzytula: You're right, our naming here is inaccurate. It's probable that in a future release we will rename `Shard` to `ShardIdx` or something similar.

scylla/src/transport/load_balancing/enforce_node.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::sync::Arc;
+
+/// This policy will always return the same node, unless it is not available anymore, in which case it will
+/// fallback to the provided policy.
+///
+/// This is meant to be used for shard-aware batching.
+#[derive(Debug)]
+pub struct EnforceTargetNodePolicy {
+    target_node: uuid::Uuid,
+    fallback: Arc<dyn LoadBalancingPolicy>,
```
wprzytula, line 12: `use` is preferred over qualifying `Uuid`.

scylla/src/transport/session.rs
```diff
@@ -1839,13 +1832,64 @@ impl Session {
...
-        match self.calculate_partition_key(prepared, serialized_values) {
-            Ok(Some(partition_key)) => {
-                let partitioner_name = prepared.get_partitioner_name();
-                Ok(Some(partitioner_name.hash(&partition_key)))
-            }
-            Ok(None) => Ok(None),
-            Err(err) => Err(err),
+        Ok(self
+            .calculate_partition_key(prepared, serialized_values)?
+            .map(|partition_key| prepared.get_partitioner_name().hash(&partition_key)))
```
wprzytula, line 1837: Thanks for making this more functional.

wprzytula:
Have you tested your solution? How? Could you write some unit tests?

wprzytula:
I can see that you merged `main` into your branch. As our policy is to rebase, could you instead rebase your branch on `main`, please?

wprzytula:
The approach with `enforce_target_node()` is very interesting and it seems to be working, at least conceptually. I'm a bit worried, however, that it may turn out to be somewhat fragile. For instance, consider a situation when an enforced node is removed from the cluster. Not only will the correct shard not be targeted, but each query will involve an expensive call to `fallback()`. Even worse, in order to repair this, the application will have to be restarted.

scylla/src/transport/execution_profile.rs
```diff
@@ -485,4 +485,8 @@ impl ExecutionProfileHandle {
     pub fn map_to_another_profile(&mut self, profile: ExecutionProfile) {
         self.0 .0.store(profile.0)
     }
+
+    pub fn load_balancing_policy(&self) -> Arc<dyn LoadBalancingPolicy> {
+        self.0 .0.load().load_balancing_policy.clone()
+    }
```
wprzytula, line 491: I'd prefer to have this `pub(crate)`, if possible.

wprzytula:
> I'm making the call to the non-fallback function of self.fallback inside the pick function, so that should be pretty inexpensive for each message whose shard/batching task has already been picked before the node was removed.

Correct, I missed it.

> If the node is removed from the cluster, new lines on which one would call shard_for_cluster wouldn't associate with this shard anymore, so the task dedicated to sending batches to this shard wouldn't receive any more messages to send [...]

It seems that I don't quite understand your use case. As I imagine it, one would once (upon application initialisation) consistute/associate a batch with a specific shard using `shard_for_node()` and `enforce_target_node()`, and then keep calling `Session::batch()` with that batch for the whole lifetime of the application. That would mean incorrect node being enforced forever if the target node is removed. How does your use case prevent that?

scylla/src/transport/execution_profile.rs
wprzytula: A docstring providing rationale behind exposing this would be welcome, then. A consise rephrasing of the above  explanation should do.

Ten0:
> As I imagine it, one would once (upon application initialisation) consistute/associate a batch with a specific shard

No, rather I have an app that receives millions of lines per second to insert. These lines are processed, spread over multiple threads, in a manner that involves just identifying the shard, and then looking up in a `HashMap<(NodeUuid, ShardIdxOnNode), Sender>` for the lazily instantiated channel that sends to a task that constitutes and sends the batches for a given target shard. This allows getting rid of ~90% of the scylla driver CPU consumption.

Now if a node is removed, the task will be left dangling, and a few messages will be sent to a default-policy-chosen shard by the time its channel is drained, but new messages will be assigned to an alive shard (and its corresponding channel).

wprzytula:
> No, rather I have an app that receives millions of lines per second to insert. These lines are processed, spread over multiple threads, in a manner that involves just identifying the shard, and then looking up in a `HashMap<(NodeUuid, ShardIdxOnNode), Sender>` for the lazily instantiated channel that sends to a task that constitutes and sends the batches for a given target shard.

Let me dig deeper. What is the lifetime of your `Batch` structs? Do they contain `PreparedStatement`s? What is their lifetime?
Are the lines that you get every second merely bind values to be sent with prepared statements with fixed query string?

Ten0:
> Are the lines that you get every second merely bind values to be sent with prepared statements with fixed query string?

Basically yes. There's a bunch of different prepared statements but those are prepared at application startup.
A `Batch` object is initialized once in each task dedicated to a shard (with its overridden execution policy from this PR). When it has constituted a set of lines to insert, it sets the corresponding prepared statements in the `Batch`, then sends it.

scylla/src/statement/batch.rs
```diff
@@ -144,6 +145,82 @@ impl Batch {
...
+            base_execution_profile_from_session.get_default_execution_profile_handle()
+        });
+        self.set_execution_profile_handle(Some(
+            execution_profile_handle
+                .pointee_to_builder()
+                .load_balancing_policy(Arc::new(
+                    crate::load_balancing::EnforceTargetNodePolicy::new(
+                        node,
+                        execution_profile_handle.load_balancing_policy(),
+                    ),
```
wprzytula, line 218: Ditto: better to `use` than to qualify.

wprzytula:
Maybe you could write an example called, say, `shard_aware_batching.rs`, and present a complete minimal way to utilise `shard_for_statement()` and `enforce_target_node()` - similar to what you put in the docstring, but complete? This would let others follow your design more easily and adapt it to their workloads.

scylla/src/statement/batch.rs
Ten0: 84ee6aa

scylla/src/transport/execution_profile.rs
Ten0: a484a4c

scylla/src/transport/session.rs
Ten0: done as part of [128f741](https://github.com/scylladb/scylla-rust-driver/pull/738/commits/128f741e013d35a2b8d66da16e32403048f6e6aa)

scylla/tests/integration/shard_aware_batching.rs
```diff
@@ -0,0 +1,193 @@
...
+    // In a production setting these dynamically instantiated tasks may be monitored more easily
+    // by using e.g. `tokio_tasks_shutdown`
+    std::mem::drop(channels_for_shards);
+    for task in batching_tasks {
+        task.await.unwrap();
+    }
+
+    // finally check that batching was indeed shard-aware.
+
+    // TODO
```
Ten0, line 176: I'm not exactly sure how to implement this, help would be welcome

Ten0:
> Have you tested your solution? How? Could you write some unit tests?

> Maybe you could write an example

I have added an test that may also serve as example in 59ced8f72420c0b2dae61f18ff83b969f7cdf2d7.
However I'm not sure exactly how to control that the queries are indeed directed to the proper shard.

I have also tested the solution in a multi node cluster at 1M lines/s and could see a much reduced metric of non-shard-aware queries in the monitoring (that metric is documented in Grafana as being wrong if there are batches but the fact it went from 60% to 10% probably still says that it was more targeted, so properly targeted).

scylla/tests/integration/shard_aware_batching.rs
wprzytula: Please take a look at #788. Does this solve you issue?

wprzytula:
@Ten0 We are currently working over a load balancing refactor such that instead returning an iterator of nodes, an LB policy would return pairs (node, shard number). How would this affect your idea of shard-aware batching? Does that help you, or rather obstruct?

Ten0:
I guess that would simplify the code. 🙂👍

wprzytula:
> I guess that would simplify the code. slightly_smiling_face+1

Do you mind waiting with this PR until we merge the mentioned refactor? This way we would avoid merging this and then throwing away with the refactor introduced.

wprzytula:
@Ten0
https://github.com/scylladb/scylla-rust-driver/pull/944 has been merged. If you are still interested in bringing shard-aware batching to the driver, now it should be much easier.

Ten0:
I don't see why it wouldn't be relevant. I may work on this still if there's a change at getting it merged.

lvboudre:
> I don't see why it wouldn't be relevant. I may work on this still if there's a change at getting it merged.

Sorry, "relevant" was not the good word .

 I meant if the PR is still "active" since they have not been any commit since a long time. I do think the issue is still relevant.

Ten0:
@wprzytula 

> This is also related to #974 where a batch that is token-aware would be much more efficient.

Updated the PR. It looks like it indeed does solve #974. 😊
Also #788 did indeed solve [this issue](https://github.com/scylladb/scylla-rust-driver/pull/738#discussion_r1285209510), thanks! 😊

(NB: I know this will need squash because oldest commits are very old and we don't want spaghetti on `main` - can we merge this by using Github's squash-and-merge feature rather than force-pushing here and losing history?)

wprzytula:
After discussion with @Lorak-mmk, our impression is that we still don't understand the approach taken by this PR. Why would we want to hardwire the destination node and shard manually, instead of relying on the default load balancer?
There is a heuristics that chooses the target node and shard based on token of the first statement in a batch. Why is it not enough? Once other statements in a batch are destined to the same node and shard, then performance will be optimal.

Hardwiring the target node and shard has significant drawbacks. RetryPolicy and SpeculativeExecutionPolicy are going to malfunction on single-target query plans. It's the responsibility of the user to select the correct node and shard.

Clearly, it's unclear to us why the problem of shard-aware batches is approached this way.

Ten0:
> Why is it not enough?

Thanks for the review.

I initially thought it was enough as well, however

> As [pointed out in the first review](https://github.com/scylladb/scylla-rust-driver/pull/738#discussion_r1243336565), this was not enough considering the load balancer (now?) picks a random replica at each call.

This means that if we check which node a statement would pick to put it in a batch, then send the batch, the batch may randomly be sent to a different shard that is appropriate for that statement, but not necessarily for all other statements of the batch, resulting in them having one chance out of `nb_replicas` to actually hit in shard-aware.

[Following this change](https://github.com/scylladb/scylla-rust-driver/pull/738#issuecomment-1666839724):

> I have also tested the solution in a multi node cluster at 1M lines/s and could see a much reduced metric of non-shard-aware queries in the monitoring (that metric is documented in Grafana as being wrong if there are batches but the fact it went from 60% to 10% probably still says that it was more targeted, so properly targeted).

Lorak-mmk:
> > Why is it not enough?
> 
> Thanks for the review.
> 
> I initially thought it was enough as well, however
> 
> > As [pointed out in the first review](https://github.com/scylladb/scylla-rust-driver/pull/738#discussion_r1243336565), this was not enough considering the load balancer (now?) picks a random replica at each call.
> 
> This means that if we check which node a statement would pick to put it in a batch, then send the batch, the batch may randomly be sent to a different shard that is appropriate for that statement, but not necessarily for all other statements of the batch, resulting in them having one chance out of `nb_replicas` to actually hit in shard-aware.

The solution is to batch by token, not by "first" replica.

> 
> [Following this change](https://github.com/scylladb/scylla-rust-driver/pull/738#issuecomment-1666839724):
> 
> > I have also tested the solution in a multi node cluster at 1M lines/s and could see a much reduced metric of non-shard-aware queries in the monitoring (that metric is documented in Grafana as being wrong if there are batches but the fact it went from 60% to 10% probably still says that it was more targeted, so properly targeted).

Ten0:
> The solution is to batch by token, not by "first" replica.

Unless I'm mistaken because the space of tokens is large this would imply that we very rarely are able to batch, whereas batching by replica would allow to group by "one random acceptable replica" for each statement.

Lorak-mmk:
> > The solution is to batch by token, not by "first" replica.
> 
> Unless I'm mistaken because the space of tokens is large this would imply that we very rarely are able to batch, whereas batching by replica allows to group by "one random acceptable replica" for each statement.

Afaik it doesn't really make sense to perform multi partition batches. This incurs additional work for the coordinator node and you would be better off just using single, non-batched writes.
So if your data to insert is spread across partitions so that there is no significant amount of data to update in any single partition - just use single non-batched writes.
If you have data that is more grouped by partitions (meaning you have many updates to a partition) then it may make sense to use batches (but you should benchmark if it gives you performance advantage over single writes), and you can group them by Token.

Ten0:
> Afaik it doesn't really make sense to perform multi partition batches. This incurs additional work for the coordinator node and you would be better off just using single, non-batched writes.

The main issue with not batching is that going through the heavy Tokio & networking machinery for every statement puts x5 CPU load on the server that makes the scylla calls using the Rust driver, whereas with batching that load is shared across coordinators (and probably less significant than managing Tokio futures is even if ignoring the fact that work is shared, because Tokio turned out to be heavy in my benchmarks). 
I'm aware that the driver currently attempts to batch networking at low level to some extent (and have read the relevant blog post), but that 1. doesn't cut off Tokio heaviness and 2. Only waits for a single Tokio yield to try to append more, whereas I significantly benefit from waiting more to build larger network batches. 

Using batches enabled me to x5 my insertion speed due to this.

lvboudre:
> Afaik it doesn't really make sense to perform multi partition batches. This incurs additional work for the coordinator node and you would be better off just using single, non-batched writes.

I forge my own batches and I do see increase in performance and lower latency. I did this couple months ago, I don't have the number with me, but it was a no brainer to code more in order to do batching locally before sending off to the scylla driver.

Lorak-mmk:
> > Afaik it doesn't really make sense to perform multi partition batches. This incurs additional work for the coordinator node and you would be better off just using single, non-batched writes.
> 
> The main issue with not batching is that going through the heavy Tokio & networking machinery for every statement puts x5 CPU load on the server that makes the scylla calls using the Rust driver, whereas with batching that load is shared across coordinators (and probably less significant than managing Tokio futures is even if ignoring the fact that work is shared, because Tokio turned out to be heavy in my benchmarks). I'm aware that the driver currently attempts to batch networking at low level to some extent (and have read the relevant blog post), but that 1. doesn't cut off Tokio heaviness and 2. Only waits for a single Tokio yield to try to append more, whereas I significantly benefit from waiting more to build larger network batches.

Would you be able to share some benchmark we could reproduce? If the problem is driver's performance then my preferred solution would be to improve driver's performance instead of doing hacky workarounds.
As for the solution for you for now: `shard_for_token` makes no sense because of tablets and mixed-shards clusters. If I understand correctly you want to batch together statements that have 1 common replica, and send them to this replica. I think you can do this today. You can use `ClusterData::replica_locator` to get `ReplicaLocator` struct, which has [replicas_for_token](https://docs.rs/scylla/latest/scylla/transport/locator/struct.ReplicaLocator.html#method.replicas_for_token) method you can call. This method returns `ReplicaSet` which implements `IntoIterator` of replicas for a given token.
You can then use whatever heuristics you like to batch statements (IIUC you will want to use first element of this iterator as grouping key), and create your own [LoadBalancingPolicy](https://docs.rs/scylla/latest/scylla/transport/load_balancing/trait.LoadBalancingPolicy.html) to use on such batch. This LBP would return plan consisting of 1 element.

> 
> Using batches enabled me to x5 my insertion speed due to this.

I'd really like to see and reproduce the benchmark to investigate if we can improve this.

Ten0:
Thanks for your answer.

> If the problem is driver's performance then my preferred solution would be to improve driver's performance instead of doing hacky workarounds.

I sure see your point - if there's a way to reduce the load in the driver without putting extra load on the servers it's definitely best to do that instead.

> I'd really like to see and reproduce the benchmark to investigate if we can improve this.

I don't think I still have the `perf` graphs but I can tell you what did show up on it.

Bear in mind that for each statement, the work I do is otherwise reading a buffered line, some minor parsing to get a partition key, pushing the parsed line into a batch of say 64 lines, and then doing the operations below only 1 out of 64 times.

What shows up on `perf` is:
- Statements being handled individually and independently, independent tokio tasks keep [going to sleep and be woken up for each statement](https://github.com/tokio-rs/tokio/blob/06582776a564c88a7b4c6f9e3d7c0ebd0ef3f34b/tokio/src/runtime/task/harness.rs#L94-L230).
- It does for every statement [all](https://github.com/scylladb/scylla-rust-driver/blob/1fcadc99055fe21a47ec623a8f1af1bb8790d186/scylla/src/transport/session.rs#L1006-L1065) [operations](https://github.com/scylladb/scylla-rust-driver/blob/1fcadc99055fe21a47ec623a8f1af1bb8790d186/scylla/src/transport/session.rs#L1512-L1657) [under](https://github.com/scylladb/scylla-rust-driver/blob/1fcadc99055fe21a47ec623a8f1af1bb8790d186/scylla/src/transport/session.rs#L1673-L1778) [`execute`](https://github.com/scylladb/scylla-rust-driver/blob/1fcadc99055fe21a47ec623a8f1af1bb8790d186/scylla/src/transport/connection.rs#L698-L742) (notably [dedicated channel](https://github.com/scylladb/scylla-rust-driver/blob/1fcadc99055fe21a47ec623a8f1af1bb8790d186/scylla/src/transport/connection.rs#L116-L154)).

NB: We should probably merge these discussions because they seem to be about the same thing, but I'll copy it here because it's also relevant to this particular sub-discussion (from https://github.com/scylladb/scylla-rust-driver/issues/468#issuecomment-2188539114): for the use-case of performing High-Frequency Append-Log Style Operations like is my case and that of the author of https://github.com/scylladb/scylla-rust-driver/issues/974, it seems clear that more scylla servers will be required to store & perform maintenance operations than client servers (~1/10 ratio), so all things otherwise equal, being able to offload some work to the servers doesn't seem necessarily bad.

To answer the question from the same thread:
> OK, could you re-iterate what kind of API you need specifically access to in order to get such client-side performance gains?

Fundamentally what I need is to send statements in a processing queue, where I don't necessarily care about getting each individual result as an `async` return (which seems to be the heavy part), but I do need that once a line is processed it gets sent to the next step of the pipeline.

What allows to achieve optimal performance is that there is nowhere in the pipeline async sleep/wakeups that occur for each statement, because:
- Scylla part is batched, and batches are always sent to a single node (so always 1 networking call, result channel...)
- Everything else in the pipeline uses [`relaxed_channel`](https://docs.rs/relaxed_channel/latest/relaxed_channel/index.html)

Wakeups do otherwise cost a lot, *relatively*, as I tried to explain in the [`relaxed_channel`](https://docs.rs/relaxed_channel) documentation. (Please let me know if the explanation there is unclear as I could also forward my talk on this topic from the Nov. 2023 Paris Rust Meetup.)

Currently I achieve this no-per-statement-wakeup by:
- Looking up a target shard (that's only the first step of the *relatively* heavy call tree above, and notably [it's not async](https://github.com/scylladb/scylla-rust-driver/pull/738/files#diff-80ba8e49418bed2ad375905c49b6784a23c454eac059d99c3e18679a8525d9feR1826).)
- Sending to a channel dedicated to that shard (reasonably lightweight thanks [`relaxed_channel`](https://docs.rs/relaxed_channel/latest/relaxed_channel/index.html))
- The task dedicated to that shard makes batches from that channel using [`chunks_timeout`](https://docs.rs/futures-batch/latest/futures_batch/trait.ChunksTimeoutStreamExt.html#method.chunks_timeout) (~64 size)
- If batch is successfully inserted, send all the lines to the next step of the pipeline.
- Otherwise, retry the whole batch (soon, shouldn't go at the back of the queue because of the requirements of the next pipeline steps) (errors happens seldom so shows no perf impact)
- In case of unrecoverable error, error the shard task, which stops the program [~gracefully](https://crates.io/crates/tokio_tasks_shutdown).

Indeed I don't care about the other properties of batches that Scylla may try to uphold. My understanding is that if batch is `Unlogged` they are minimal - am I mistaken ?

> `shard_for_token` makes no sense because of tablets and mixed-shards clusters

My understanding is that this would result in *most of the time*, hitting a shard that will be one of those that holds one of the copies of this particular data. Am I mistaken?
If I am indeed mistaken here, then I guess a smaller version of batch that would uphold even smaller guarantees would be a useful API.

I don't care that I hit a correct shard 100% of the time: if the Scylla nodes need to transfer the call to another node 1% of the time because there was a tablet split it doesn't look like it's going to have any impact.

Actually even hitting a correct shard almost every time probably has almost no impact: my understanding is that the cluster will still perform as many extra network jumps for each statement as there are replicas, so that's just one more, so I could probably just random-batch.
What I'm trying to optimize here compared to the random-batch solution is that I hit a relevant shard most of the time, as opposed to statistically never, requiring one extra jump all the time.

My understanding is that since the Scylla servers will already perform some per-statement jumps whatever happens, letting it perform the per-statement work instead of doing it via per-statement async management in `tokio` enables to reduce the number of clients without *significantly* increasing the load on the servers in a way that would require provisioning more of them. Please let me know if I am mistaken.

Thanks!

mykaul:
> Thanks for your answer.
> 
> > If the problem is driver's performance then my preferred solution would be to improve driver's performance instead of doing hacky workarounds.
> 
> I sure see your point - if there's a way to reduce the load in the driver without putting extra load on the servers and/or exposing less type-safe APIs it's definitely best to do that instead.

General comment, not specific to this issue - I think it makes sense to always offload more to the client, if possible, rather than push more work to the servers. There are most likely (in most scenarios) far more clients and they can be scaled out and so on than server resources.

Ten0:
> I think it makes sense to always offload more to the client - There are most likely (in most scenarios) far more clients and they can be scaled out

Definitely. Although it seems that unless I'm mistaken, in this particular case (append-log style operations) the load difference if a user chooses to use the API that offloads (namely, the batch API) may be insignificant on the server, but an order of magnitude of difference on the client, so maybe the approach of using batches for that purpose may be relevant, and if that's indeed the case, then proper shard awareness for batches (to the extent possible) is just uncompromising (perf-wise) load reduction on the server.

avikivity:
> > Indeed I don't care about the other properties of batches that Scylla may try to uphold. My understanding is that if batch is `Unlogged` they are minimal - am I mistaken ?
> 
> There must be some overhead for multi-partition unlogged batches, but I'm not sure how big it is in practice. Coordinator will have to split the batch and execute it's statements separatly iiuc. Maybe @avikivity would know how big the overhead typically is?

It's better to coordinate on the client. The overhead isn't large but it's an extra hop.

Ten0:
> I understand the explanation, but the talk sounds interesting anyway. Could you share it?

Alright, I'll take the time to finish cleaning the auto-generated subtitles that are a disaster 😅 then post the link. 🙂

> I understand that you have N nodes, each has K shards, and on the driver side you have K queues (not N, not N*K, just K)

Ah it looks like we had a misunderstanding here: I do create N*K queues, that is, one per actual shard, not one per shard-index or one per number-of-shards-per-node. (I'm using the naming discussed [here](https://github.com/scylladb/scylla-rust-driver/pull/738#discussion_r1250245107), but maybe I should have said *replica*? I thought that was a different concept from node, associated to a particular statement.)

> With tablets it's possible (and it's a normal case, not some edge case) for replicas to have different shards, meaning your replicas could be (A, n), (B, l), (C, j) with n != l != j - so there isn't any shard that you could call THE shard of this statement.

I already don't make such assumptions. Apologies if I said "**the** shard for a statement" anywhere above, I really only ever meant "**a** shard that would match this statement": I only ever manipulate pairs of (Node, ShardIndex), hence the signature of the main new function of this PR `shard_for_statement`. (Note to self: If `shard` is not the term but `replica` I assume it should be named `replica_for_statement` then.)

> Take replica set from https://docs.rs/scylla/latest/scylla/transport/locator/struct.ReplicaLocator.html#method.replicas_for_token , use the first element as batching key

> None of the above implementations requires changes in the driver, you can do them from user code.

I know that it's already possible, it's even written in the PR's description:
> This was made possible by #658, but it was still very boilerplate-ish. I was waiting for #612 to be merged (amazing work btw! 😃) to implement a more direct and factored API (as that would use it).

The issues that I have with implementing this outside of the driver are:
1. It's not very direct: I feel like this would often break from one scylla release to another, or worse, become inaccurate, compared to using the query planning code: if not using the query planner, it may not take into account nodes being down, or other things from the load balancing (node being overloaded, node's queue being more full...) that may come out from one release to another.
2. It's very boilerplate-ish to calculate the token for the statement, notably because if we want to take into account the load balancer strategy, however complex, methods that the function this PR proposes use are not currently public. Rather than exposing them and leaving it to all users who want to do the same thing as I do to reimplement and exposing functions that are largely internal, it looks like it may be preferable to instead just add the one function that does what people actually may want to do with it.

To be clear: for me adding this function is the main point of this PR:
https://github.com/scylladb/scylla-rust-driver/blob/d5f71a93ef73cac47c4d513ca4f2d231c168f7e5/scylla/src/transport/session.rs#L1826-L1830

making the code that picks **a** shard for a given statement easily accessible and *correct* and *consistent with load balancing*, as efficient load balancing is effectively what I'm looking for.

Now as pointed out by @wprzytula [here](https://github.com/scylladb/scylla-rust-driver/pull/738#discussion_r1243336565) any code that uses this doesn't make sense without also enforcing the target (node,shard) afterwards, and it's nice to have usage examples, which is why the PR also adds a policy to enforce the target shard afterwards, and an test that also constitutes an example of how to write a system such as the one I was describing in my previous message.

Ten0:
> I understand the explanation, but the talk sounds interesting anyway. Could you share it?

Finally finished the subtitles (I recommend enabling them): https://youtu.be/cBElzSO4A_s

PR #752, opened 2023-06-29
CI: use ubuntu-20.04 in serverless workflow

avelanarius:
ccm requries Python 2 when creating a serverless cluster. Installing Python 2 with actions/setup-python@v2 is now deprecated, so we solve it by explicitly using ubuntu-20.04 runners, which have it installed.

Closes #751

## Pre-review checklist


Gor027:
@wprzytula Could you please review this as it is necessary for #750.

Issue #746, opened 2023-06-16
Provide a way to calculate token without the need for prepared statement


piodul:
(Requested by a user)

Currently, the driver provides a way to calculate token based on a `PreparedStatement` and values for that statement. However, if a user knows:

1. The order of the columns in the partition key,
2. The values of the columns of the partition key,
3. The partitioner of the table that the statement operates on,

then having a `PreparedStatement` is not necessary and the token can be calculated based on that information. We should provide a provide a function which does that calculation.

PR #757, opened 2023-07-05
Calculate token without preparing

wprzytula:
The function `Session::calculate_token_for_partition_key()` is added and tested. It enables computing the token without prior preparing, straight from a given partitioner and a serialized partition key.

Fixes: #746

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -2798,3 +2798,98 @@ async fn simple_strategy_test() {
 
     assert_eq!(rows, vec![(1, 2, 3), (4, 5, 6), (7, 8, 9)]);
 }
+
+#[tokio::test]
+async fn test_manual_primary_key_computation() {
+    // Setup session
+    let ks = unique_keyspace_name();
+    let session = create_new_session_builder().build().await.unwrap();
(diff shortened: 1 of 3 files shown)


scylla/src/transport/session_test.rs
```diff
@@ -2798,3 +2798,98 @@ async fn simple_strategy_test() {
 
     assert_eq!(rows, vec![(1, 2, 3), (4, 5, 6), (7, 8, 9)]);
 }
+
+#[tokio::test]
+async fn manual_primary_key_computation() {
```
piodul, line 2803: Nit: add `_test` at the end, or `test_` at the beginning (I prefer the latter). Let's have a rule to name tests this way; it will make them stand out more and will reduce the risk of name collision with non-test functions.

piodul (changes requested):
Looks good, some small nits only.

piodul (approved):

avelanarius:
@piodul Apart from the fact that we have 3 separate functions to calculate token from keys, why did this function:

```
pub fn calculate_token_for_partition_key<P: Partitioner>(
    serialized_partition_key_values: &SerializedValues,
    _partitioner: &P,
) -> Result<Token, PartitionKeyError>
```

end up in `Session`? Wouldn't it have made more sense to put it in `partitioner.rs`?

BTW I'm now looking into how those 3 different functions should look like (less of them? in other places?).

PR #758, opened 2023-07-07
Stateful partitioners

wprzytula:
## Motivation
Currently, partitioners require the whole input to be passed in a single contiguous slice, which often forces an allocation that could else be avoided.

## What's done
1. **Partitioners' API is refactored** so it resembles `std::hash` API:
    - `Partitioner` corresponds to `BuildHasher` as it becomes kind of a factory,
    - `PartitionerHasher` is introduced and corresponds to `Hasher`. It contains the hashing state, which enables feeding it with multiple chunks of data separately.
A **test** is added that **ensures consistent hashing result** no matter how the partition key is partitioned into chunks.
2. After the partitioners gain ability to partition by chunks, the codebase is modified so that **all needless allocations of partition key are avoided**, i.e., allocations that only served the purpose of calculating token.
To this end, `PartitionKey` struct is introduced, whose motivation is as follows: the algorithm for computing partition key for a given prepared statement and bound values can be divided into steps. The first of them involves extracting values that consistute the partition key and putting them in proper partition key order. `PartitionKey` performs this step on construction, and serves as an entry point for further steps, with token calculation among others. For those further steps to be done, an iterator over partition key values is accessible, which yields pairs _(value, spec)_ for each column that constitutes partition key, in partition key order.
3. As **there was one more place** where materialised partition key was used - by decoding it into the produced trace - an adaptor is introduced that operates on `PartitionKey`'s iterator provides values to be deserialised by the mechanism added recently in #766. **Allocation is then saved there**, too.
4. **An API cleanup is done** regarding partition key and token computation. `calculate_token_from_partition_key()` is moved to `partitioner.rs`, `calculate_token()` to `prepared_statement.rs`, and a number of helper functions are deleted as no longer needed.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -34,7 +34,7 @@ pub struct SchemaChange {
     pub event: SchemaChangeEvent,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TableSpec {
     pub ks_name: String,
     pub table_name: String,
@@ -329,7 +329,7 @@ impl CqlValue {
     // TODO
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ColumnSpec {
     pub table_spec: TableSpec,
     pub name: String,
--- a/scylla/benches/benchmark.rs
+++ b/scylla/benches/benchmark.rs
@@ -1,7 +1,11 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 
 use bytes::BytesMut;
-use scylla::frame::types;
+use scylla::{
+    frame::types,
+    frame::value::ValueList,
+    transport::partitioner::{calculate_token_for_partition_key, Murmur3Partitioner},
+};
 
 fn types_benchmark(c: &mut Criterion) {
     let mut buf = BytesMut::with_capacity(64);
@@ -35,5 +39,57 @@ fn types_benchmark(c: &mut Criterion) {
     });
 }
 
-criterion_group!(benches, types_benchmark);
+fn calculate_token_bench(c: &mut Criterion) {
+    let simple_pk = ("I'm prepared!!!",);
+    let serialized_simple_pk = simple_pk.serialized().unwrap().into_owned();
+    let simple_pk_long_column = (
+        17_i32,
+        16_i32,
+        String::from_iter(std::iter::repeat('.').take(2000)),
+    );
+    let serialized_simple_pk_long_column = simple_pk_long_column.serialized().unwrap().into_owned();
+
+    let complex_pk = (17_i32, 16_i32, "I'm prepared!!!");
+    let serialized_complex_pk = complex_pk.serialized().unwrap().into_owned();
+    let complex_pk_long_column = (
+        17_i32,
+        16_i32,
+        String::from_iter(std::iter::repeat('.').take(2000)),
+    );
+    let serialized_values_long_column = complex_pk_long_column.serialized().unwrap().into_owned();
+
+    c.bench_function("calculate_token_from_partition_key simple pk", |b| {
+        b.iter(|| calculate_token_for_partition_key(&serialized_simple_pk, &Murmur3Partitioner))
+    });
+
+    c.bench_function(
+        "calculate_token_from_partition_key simple pk long column",
+        |b| {
+            b.iter(|| {
+                calculate_token_for_partition_key(
+                    &serialized_simple_pk_long_column,
+                    &Murmur3Partitioner,
+                )
+            })
+        },
+    );
+
+    c.bench_function("calculate_token_from_partition_key complex pk", |b| {
+        b.iter(|| calculate_token_for_partition_key(&serialized_complex_pk, &Murmur3Partitioner))
+    });
+
+    c.bench_function(
+        "calculate_token_from_partition_key complex pk long column",
+        |b| {
+            b.iter(|| {
+                calculate_token_for_partition_key(
+                    &serialized_values_long_column,
+                    &Murmur3Partitioner,
+                )
+            })
+        },
+    );
+}
+
+criterion_group!(benches, types_benchmark, calculate_token_bench);
 criterion_main!(benches);
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -1,4 +1,3 @@
-use byteorder::{BigEndian, ReadBytesExt};
 use bytes::{Bytes, BytesMut};
 use scylla_cql::errors::{BadQuery, QueryError};
 use smallvec::{smallvec, SmallVec};
@@ -8,8 +7,7 @@ use std::time::Duration;
 use thiserror::Error;
 use uuid::Uuid;
 
-use scylla_cql::frame::frame_errors::ParseError;
-use scylla_cql::frame::response::result::{deser_cql_value, ColumnSpec, CqlValue};
+use scylla_cql::frame::response::result::ColumnSpec;
 
 use super::StatementConfig;
 use crate::frame::response::result::PreparedMetadata;
@@ -450,96 +448,15 @@ impl<'ps> PartitionKey<'ps> {
     }
 }
 
-// The PartitionKeyDecoder reverses the process of PreparedStatement::compute_partition_key:
-// it returns the consecutive values of partition key column that were encoded
-// by the function into the Bytes object, additionally decoding them to CqlValue.
-//
-// The format follows the description here:
-// <https://github.com/scylladb/scylla/blob/40adf38915b6d8f5314c621a94d694d172360833/compound_compat.hh#L33-L47>
-//
-// TODO: Currently, if there is a null value specified for a partition column,
-// it will be skipped when creating the serialized partition key. We should
-// not create such partition keys in the future, i.e. fail the request or
-// route it to a random replica instead and let the DB reject it. In the
-// meantime, this struct will return some garbage data while it tries to
-// decode the key, but nothing bad like a panic should happen otherwise.
-// The struct is currently only used for printing the partition key, so that's
-// completely fine.
-#[derive(Clone, Copy)]
-pub(crate) struct PartitionKeyDecoder<'pk> {
-    prepared_metadata: &'pk PreparedMetadata,
-    partition_key: &'pk [u8],
-    value_index: usize,
-}
-
-impl<'pk> PartitionKeyDecoder<'pk> {
-    pub(crate) fn new(prepared_metadata: &'pk PreparedMetadata, partition_key: &'pk [u8]) -> Self {
-        Self {
-            prepared_metadata,
-            partition_key,
-            value_index: 0,
-        }
-    }
-}
-
-impl<'pk> Iterator for PartitionKeyDecoder<'pk> {
-    type Item = Result<CqlValue, ParseError>;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        if self.value_index >= self.prepared_metadata.pk_indexes.len() {
-            return None;
-        }
-
-        // It should be safe to unwrap because PartitionKeyIndex::sequence
-        // fields of the pk_indexes form a permutation, but let's err
-        // on the safe side in case of bugs.
-        let pk_index = self
-            .prepared_metadata
-            .pk_indexes
-            .iter()
-            .find(|pki| pki.sequence == self.value_index as u16)?;
-
-        let col_idx = pk_index.index as usize;
-        let spec = &self.prepared_metadata.col_specs[col_idx];
-
-        let cell = if self.prepared_metadata.pk_indexes.len() == 1 {
-            Ok(self.partition_key)
-        } else {
-            self.parse_cell()
-        };
-        self.value_index += 1;
-        Some(cell.and_then(|mut cell| deser_cql_value(&spec.typ, &mut cell)))
-    }
-}
-
-impl<'pk> PartitionKeyDecoder<'pk> {
-    fn parse_cell(&mut self) -> Result<&'pk [u8], ParseError> {
-        let buf = &mut self.partition_key;
-        let len = buf.read_u16::<BigEndian>()? as usize;
-        if buf.len() < len {
-            return Err(ParseError::IoError(std::io::Error::new(
-                std::io::ErrorKind::UnexpectedEof,
-                "value too short",
-            )));
-        }
-        let col = &buf[..len];
-        *buf = &buf[len..];
-        buf.read_u8()?;
-        Ok(col)
-    }
-}
-
 #[cfg(test)]
 mod tests {
-    use bytes::{BufMut, Bytes, BytesMut};
     use scylla_cql::frame::{
         response::result::{
-            ColumnSpec, ColumnType, CqlValue, PartitionKeyIndex, PreparedMetadata, TableSpec,
+            ColumnSpec, ColumnType, PartitionKeyIndex, PreparedMetadata, TableSpec,
         },
         value::SerializedValues,
     };
 
-    use super::PartitionKeyDecoder;
     use crate::prepared_statement::PartitionKey;
 
     fn make_meta(
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -124,7 +124,7 @@ impl PreparedStatement {
 
     /// Computes the partition key of the target table from given values —
     /// it assumes that all partition key columns are passed in values.
-    /// Partition keys have a specific serialization rules.
+    /// Partition keys have specific serialization rules.
     /// Ref: <https://github.com/scylladb/scylla/blob/40adf38915b6d8f5314c621a94d694d172360833/compound_compat.hh#L33-L47>
     pub fn compute_partition_key(
         &self,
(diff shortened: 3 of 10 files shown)


mykaul:
Any measurable performance impact?

avelanarius (changes requested):
As a part of this change, you should also tackle issues with public API for calculating tokens.

Currently, there are some problems with that: There are 3 functions to calculate Token: `Session::calculate_token`, `Session::calculate_token_for_partition_key` and `ClusterData::compute_token`. It doesn't make sense to put them inside `Session` or `ClusterData` - wrong level of abstraction. A similar case for functions to calculate partition key.

More context: https://github.com/scylladb/scylla-rust-driver/pull/757#issuecomment-1643766615
My initial exploration on fixing it: https://github.com/avelanarius/scylla-rust-driver/tree/token_calc_cleanup

wprzytula:
> Any measurable performance impact?

Number of allocations has dropped (6 -> 5 per insert, 19 -> 18 per select).

**Before:**
```
Args { mode: All, node: "127.0.0.1:9042", parallelism: 256, requests: 100000 }
Connecting to 127.0.0.1:9042 ...
Sending 100000 inserts, hold tight ..........
----------
Inserts:
----------
allocs/req:                 6.08
reallocs/req:               6.00
frees/req:                  6.07
bytes allocated/req:      272.75
bytes reallocated/req:     73.06
bytes freed/req:          267.77
----------
Sending 100000 selects, hold tight ..........
----------
Selects:
----------
allocs/req:                19.00
reallocs/req:               6.00
frees/req:                 19.00
bytes allocated/req:     1123.09
bytes reallocated/req:     73.00
bytes freed/req:         1123.03
----------
```

**After:**
```
Args { mode: All, node: "127.0.0.1:9042", parallelism: 256, requests: 100000 }
Connecting to 127.0.0.1:9042 ...
Sending 100000 inserts, hold tight ..........
----------
Inserts:
----------
allocs/req:                 5.08
reallocs/req:               5.00
frees/req:                  5.07
bytes allocated/req:      265.09
bytes reallocated/req:     77.06
bytes freed/req:          260.06
----------
Sending 100000 selects, hold tight ..........
----------
Selects:
----------
allocs/req:                18.00
reallocs/req:               5.00
frees/req:                 18.00
bytes allocated/req:     1115.19
bytes reallocated/req:     77.00
bytes freed/req:         1115.05
----------

```

piodul:
> > Any measurable performance impact?
> 
> Number of allocations has dropped (6 -> 5 per insert, 19 -> 18 per select).

Please prepare a microbenchmark for this. While an allocation is now avoided, the new implementation has more branches, so it will be interesting to see whether the tradeoff was worth it.

We are basically interested in the `PreparedStatement::compute_partition_key` function. Interesting cases are single-column pks with different lengths and multi-column pks with different lengths.

We already have a small amount of benchmarks in `scylla` and `scylla-cql`, you can look at them for reference (and at the external `criterion` crate that those benchmarks use).

scylla/src/transport/partitioner.rs
```diff
@@ -325,6 +330,76 @@ impl PartitionerHasher for CDCPartitionerHasher {
...
 }
 
+/// Calculates the token for given prepared statement and serialized values.
+///
+/// Returns the token that would be computed for executing the provided
+/// prepared statement with the provided values.
+pub fn calculate_token(
+    prepared: &PreparedStatement,
+    serialized_values: &SerializedValues,
+) -> Result<Option<Token>, QueryError> {
```
avelanarius, line 340: Wouldn't it be better if this was in `PreparedStatement`?

`prepared_statement.calculate_token(...)`

```rust
impl PreparedStatement {
  pub fn calculate_token(
      &self,
      serialized_values: &SerializedValues,
  ) -> Result<Option<Token>, QueryError>
}
```

mykaul:
> > Please prepare a microbenchmark for this. While an allocation is now avoided, the new implementation has more branches, so it will be interesting to see whether the tradeoff was worth it.
> 
> After benchmarks and some changes to the algorithm, the new way seems to outperform the old way by 15% to 40% in various cases. I believe this suffices to merge this.

NICE! Well done! Would be happy to see details of this tests.

wprzytula:
CI fails due to use of `Option::unzip()`, which is stable since Rust 1.66, whereas our MSRV is 1.65. Can we bump it @piodul?
Apart from that, ready for review.

wprzytula:
I reimplemented `Option::unzip()` (technically, copied its code from the standard library) temporarily, not to force MSRV bump yet.

wprzytula:
Rebased on `main`.

scylla/src/statement/prepared_statement.rs
```diff
@@ -132,59 +134,68 @@ impl PreparedStatement {
...
-                        PartitionKeyError::NoPkIndexValue(pk_index.index, bound_values.len())
-                    })?;
-                // Add it in sequence order to pk_values
-                if let Some(v) = next_val {
-                    pk_values[pk_index.sequence as usize] = Some(v);
-                    buf_size += std::mem::size_of::<u16>() + v.len() + std::mem::size_of::<u8>();
+        let partition_key = self.extract_partition_key(bound_values)?;
+        let mut pk_val_iter = partition_key.iter().map(|(val, _spec)| val);
+        if let Some(first_value) = pk_val_iter.next() {
+            if let Some(second_value) = pk_val_iter.next() {
```
piodul, line 142: Why do we need to do it like this? We already know the partition length, don't we? Why can't we check the length and then handle single-value or composite cases in separate branches of the `if`?

scylla/src/statement/prepared_statement.rs
```diff
@@ -335,6 +346,83 @@ pub enum PartitionKeyError {
...
+                }
+            } else {
+                // Single-value partition key case
+                partitioner_hasher.write(first_value);
+            }
+        }
+
+        Ok(partitioner_hasher.finish())
+    }
+}
```
piodul: It looks like you duplicate the logic related to handling single-value and composite cases with `PreparedStatement::compute_partition_key`. Can we abstract them away? E.g. you could pass a `FnMut` closure that gets fed with the bytes.

scylla/src/statement/prepared_statement.rs
```diff
@@ -335,6 +346,83 @@ pub enum PartitionKeyError {
     ValueTooLong(usize),
 }
 
+pub(crate) type PartitionKeyValue<'ps> = (&'ps [u8], &'ps ColumnSpec);
+
+pub(crate) struct PartitionKey<'ps> {
+    partitioner_name: &'ps PartitionerName,
+    pk_values: SmallVec<[Option<PartitionKeyValue<'ps>>; PartitionKey::SMALLVEC_ON_STACK_SIZE]>,
+}
```
piodul: It's a large object that now gets also constructed in a single-value partition key case. Did you re-do benchmarks after introducing it?

scylla/src/statement/prepared_statement.rs
```diff
@@ -610,4 +614,59 @@ mod tests {
...
+            ],
+            [4, 0, 3],
+        );
+        let values = ValuesMaker::new()
+            .value(67i8)
+            .value(42i16)
+            .value(23i32)
+            .value(89i64)
+            .value([1u8, 2, 3, 4, 5])
+            .finish();
```
piodul, line 529: Do we really need to introduce a builder for this?

Why not just:

```rust
let mut values = SerializedValues::new();
values.add_value(&67i8).unwrap();
values.add_value(&42i16).unwrap();
values.add_value(&23i32).unwrap();
values.add_value(&89i64).unwrap();
values.add_value(&[1u8, 2, 3, 4, 5]).unwrap();
```

scylla/src/statement/prepared_statement.rs
```diff
@@ -610,4 +614,59 @@ mod tests {
...
+    }
+
+    #[derive(Debug)]
+    struct PtrEq<'t, T>(&'t T);
+    impl<'t, T> PartialEq for PtrEq<'t, T> {
+        fn eq(&self, other: &Self) -> bool {
+            std::ptr::eq(self.0, other.0)
+        }
+    }
+    impl<'t, T> Eq for PtrEq<'t, T> {}
```
piodul: Do we need this? As far as I see, it's used in conjunction with `&ColumnSpec`, and for those you have derived the `Eq` trait.

scylla/src/utils/mod.rs
```diff
@@ -2,3 +2,11 @@ pub(crate) mod parse;
...
 pub(crate) mod pretty;
 pub mod test_utils;
+
+// FIXME: replace this with `Option::unzip()` once MSRV is bumped to at least 1.66
+pub(crate) fn unzip_option<T, U>(opt: Option<(T, U)>) -> (Option<T>, Option<U>) {
+    match opt {
+        Some((a, b)) => (Some(a), Some(b)),
+        None => (None, None),
+    }
+}
```
piodul, line 12: Instead of a standalone function, you can implement an extension trait for `Option` which defines an `unzip` function for it. The code will look like you used a real `Option::unzip` in the standard library. When we bump the MSRV you will only have to drop the trait import.

scylla/src/transport/session.rs
```diff
@@ -2237,12 +2226,16 @@ impl Drop for RequestSpan {
...
+fn partition_key_displayer<'ps, 'res>(
+    mut pk_values_iter: impl Iterator<Item = (&'ps [u8], &'ps ColumnSpec)> + 'res + Clone,
+) -> impl Display + 'res {
     CommaSeparatedDisplayer(
-        PartitionKeyDecoder::new(prepared_metadata, partition_key).map(|c| match c {
+        std::iter::from_fn(move || {
+            pk_values_iter
+                .next()
+                .map(|(mut cell, spec)| deser_cql_value(&spec.typ, &mut cell))
+        })
```
piodul, line 2150: Why is `std::iter::from_fn` needed? Isn't just `pk_values_iter.map(|(mut cell, spec)| deser_cql_value(&spec.typ, &mut cell))` sufficient?

scylla/src/transport/iterator.rs
```diff
@@ -239,20 +237,35 @@ impl RowIterator {
             .unwrap_or(&*config.execution_profile.retry_policy)
             .new_session();
 
+        // We have to extract partition key here, because here we can return Err in case it is returned
+        // we can't do it in the worker future). However, we can't preserve the extracted partition key
+        // until it's needed inside the worker future, because in the meantime `config` is moved,
+        // and this invalidates the PartitionKey (as it holds refs to prepared statement and values).
+        // As a result, we have to extract partition key twice: here and inside the worker future.
+        // This shouldn't make a noticeable difference though.
```
piodul, line 245: I don't understand why we can't return Err inside the worker future. When you encounter an error, you should send it using the channel that returns the "proof" of value being sent.

scylla/benches/benchmark.rs
```diff
@@ -1,7 +1,11 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 
 use bytes::BytesMut;
-use scylla::frame::types;
+use scylla::{
+    frame::types,
+    frame::value::ValueList,
+    transport::partitioner::{calculate_token_for_partition_key, Murmur3Partitioner},
+};
```
piodul, line 8: Nit: putting the commit with the benchmark at the beginning of the PR should make it easier to use it to compare performance before and after the changes.

scylla/src/statement/prepared_statement.rs
```diff
@@ -126,6 +126,10 @@ impl PreparedStatement {
     /// it assumes that all partition key columns are passed in values.
     /// Partition keys have specific serialization rules.
     /// Ref: <https://github.com/scylladb/scylla/blob/40adf38915b6d8f5314c621a94d694d172360833/compound_compat.hh#L33-L47>
+    ///
+    /// Note: this is no longer required to compute a token. This is because partitioners are now stateful and stream-based,
+    /// so they do not require materialised partition key.
+    /// The function is solely kept for backward compatibility purposes.
```
piodul, line 132: I'm not sure that the only purpose of this function is to compute the token.

I'd change this paragraph so that it only explains that if one wants to calculate the token, there are more efficient methods than computing partition key and then using it to compute the token.

piodul (changes requested):

scylla/src/statement/prepared_statement.rs
wprzytula: I though the same. However, `test` attribute (as in `cfg(test)`) does not propagate through crate boundary. That is, if I compile tests for `scylla` crate, `PartialEq` is not implemented for structs living in `scylla_cql`.

piodul: Then let's not overcomplicate. `ColumnSpec` is a part of the API so we can just derive `Eq` and `PartialEq` for it always.

scylla/src/utils/mod.rs
wprzytula: The idea is great, but not achievable...
```
error[E0658]: use of unstable library feature 'unzip_option': recently added
   --> scylla/src/transport/iterator.rs:249:14
    |
249 |             .unzip();
    |              ^^^^^
    |
    = note: see issue #87800 <https://github.com/rust-lang/rust/issues/87800> for more information
```
even though the trait is imported by the module. For this to work, fully qualified syntax is needed:
```rust
<Option<(PartitionKey, Token)> as UnzipOption<PartitionKey, Token>>::unzip(
            prepared
                .extract_partition_key_and_calculate_token(&serialized_values)?);
```

So, I'm keeping the standalone function.

piodul: Bummer.

scylla/src/transport/session.rs
wprzytula: I did overengineer this, I admit, due to my fascination with `from_fn` once I've recently discovered it. Enexpectedly, however, I run into strange lifetime issues once I change it to `map`. Could you yourself try to refactor it as you suggest?

scylla/src/transport/cluster.rs
```diff
@@ -409,31 +410,17 @@ impl ClusterData {
...
+            &partitioner,
+        )
+        .map_err(|err| match err {
+            PartitionKeyError::NoPkIndexValue(_, _) => unreachable!(), // calculate_token_for_partition_key() never returns this variant
+            PartitionKeyError::ValueTooLong(values_len) => {
+                BadQuery::ValuesTooLongForKey(values_len, u16::MAX.into())
             }
-        };
-        Ok(partitioner.hash(&serialized_pk.unwrap_or_default()))
+        })
```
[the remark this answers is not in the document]
wprzytula: Regarding the first bullet: this is temporary. Later in this PR `calculate_token_for_partition_key()` is moved to `partitioner.rs`, breaking the cycle.

scylla/src/statement/prepared_statement.rs
wprzytula: I did, and it seems that it's even better now than previously.

scylla/src/transport/iterator.rs
wprzytula: I reasoned that there is a semantic difference between returning an error _before_ the Iterator Worker is spawned and _after_. It seems that, however, there is no such difference, so I'll use the channel as you suggest.

scylla/src/statement/prepared_statement.rs
```diff
@@ -142,48 +142,47 @@ impl PreparedStatement {
...
-            let next_val = values_iter
-                .nth((pk_index.index - values_iter_offset) as usize)
-                .ok_or_else(|| {
-                    PartitionKeyError::NoPkIndexValue(pk_index.index, bound_values.len())
-                })?;
-            // Add it in sequence order to pk_values
-            if let Some(v) = next_val {
-                pk_values[pk_index.sequence as usize] = Some(v);
-                buf_size += std::mem::size_of::<u16>() + v.len() + std::mem::size_of::<u8>();
+        } else {
```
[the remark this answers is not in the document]
wprzytula: I believe the source of our disagreement is that while you treat the single-value partition key as a special case and a multi-value one as an ordinary case, I treat them as equally important cases. Neither of them is an exception of any kind; there's just a fork of treatment: different for both cases. I insist thus on making them two branches of an `if` statement.

scylla/src/transport/cluster.rs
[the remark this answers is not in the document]
wprzytula: I refactored the error types so that `unreachable!()` can be avoided.

scylla/src/statement/prepared_statement.rs
wprzytula: This method is arguably more type-safe. No unwrap nor array indexing here, so no panics are possible.

scylla/src/statement/prepared_statement.rs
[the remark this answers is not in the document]
piodul: > I believe the source of our disagreement is that while you treat the single-value partition key as a special case and a multi-value one as an ordinary case, I consider them to be equally important cases.

No, I didn't mean that special == less important. Personally, I just find code that is less indented to be easier to read.

Let's not waste more time on this. I won't insist on reverting the indentation change in order not to add more work.

scylla/src/transport/partitioner.rs
```diff
@@ -24,84 +24,198 @@ impl PartitionerName {
...
+        // If the buffer is nonempty and can be filled completely, so that we can fetch two i64s from it,
+        // fill it and hash its contents, then make it empty.
+        if buf_len > 0 && Self::BUF_CAPACITY - buf_len <= pk_part.len() {
+            // First phase: populate buffer until full, then consume two i64s.
+            let to_write = Ord::min(Self::BUF_CAPACITY - buf_len, pk_part.len());
+            self.buf[buf_len..buf_len + to_write].copy_from_slice(&pk_part[..to_write]);
+            pk_part.advance(to_write);
+            buf_len += to_write;
+
+            assert_eq!(buf_len, Self::BUF_CAPACITY);
```
piodul, line 186: Why not a debug assert?

scylla/src/statement/prepared_statement.rs
```diff
@@ -527,22 +533,22 @@ mod tests {
                 ColumnType::BigInt,
                 ColumnType::Blob,
             ],
-            [4, 0, 3],
+            [0, 3, 4],
```
piodul, line 519: Sorry, I got confused previously. We should explicitly sort the `PartitionKeyIndex` structures by `index` after all. The order `4, 0, 3` was intentional here - you can have a query which expects partition key column to be specified out of order (e.g. `UPDATE tbl SET v = ? WHERE pk3 = ? AND pk1 = ? AND pk2 = ?`).

scylla-cql/src/frame/response/result.rs
```diff
@@ -34,7 +34,7 @@ pub struct SchemaChange {
     pub event: SchemaChangeEvent,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
piodul, line 37: > prepared: add tests for PartitionKey
>
> The tests require PartialEq,Eq on TableSpec and ColumnSpec, so these
> trait were derived conditionally (if cfg(test)).

Please update the commit message - the traits are derived _unconditionally_.

scylla/src/statement/prepared_statement.rs
```diff
@@ -360,6 +351,93 @@ impl From<TokenCalculationError> for PartitionKeyError {
     }
 }
 
+pub(crate) type PartitionKeyValue<'ps> = (&'ps [u8], &'ps ColumnSpec);
+
+pub(crate) struct PartitionKey<'ps> {
+    partitioner_name: &'ps PartitionerName,
```
piodul, line 375: This doesn't feel like it belongs here. Not every user of this struct will need the partitioner. I think it should just be an argument for the `calculate_token` method.

scylla/src/statement/prepared_statement.rs
```diff
@@ -613,21 +572,22 @@ mod tests {
...
-            .unwrap();
+        let mut values = SerializedValues::new();
+        values.add_value(&67i8).unwrap();
+        values.add_value(&42i16).unwrap();
+        values.add_value(&23i32).unwrap();
+        values.add_value(&89i64).unwrap();
+        values.add_value(&[1u8, 2, 3, 4, 5]).unwrap();
+
+        let pk = PartitionKey::new(&meta, &PartitionerName::Murmur3, &values).unwrap();
+        let pk_cols = Vec::from_iter(pk.iter().map(|(value, spec)| (value, spec)));
```
piodul, line 529: `.map(|(value, spec)| (value, spec))`

Does this part do anything?

scylla/src/transport/session.rs
```diff
@@ -2117,9 +2108,8 @@ impl RequestSpan {
         }
     }
 
-    pub(crate) fn new_prepared(
-        prepared_metadata: &PreparedMetadata,
-        partition_key: Option<&Bytes>,
+    pub(crate) fn new_prepared<'ps>(
+        partition_key: Option<impl Iterator<Item = (&'ps [u8], &'ps ColumnSpec)> + Clone>,
```
piodul, line 2018: Does this need to be that much generic? Does it really make sense to put here anything that isn't an `Option<PartitionKey>`? Perhaps doing so will make it easier to fix the lifetime issue in `partition_key_displayer`.

scylla/src/transport/session.rs
piodul: I tried, but I gave up quickly.

piodul (changes requested):
Another batch of comments - hopefully the last one this time.

scylla/src/statement/prepared_statement.rs
wprzytula: I don't think I get your point.
How should I adjust the code then? Before it was ..., now it has to be ... ?

scylla/src/statement/prepared_statement.rs
wprzytula: It's a quite funny remnant.

scylla/src/statement/prepared_statement.rs
piodul: From `frame/response/result.rs`:

```rust
pub struct PreparedMetadata {
    // ...
    /// pk_indexes are sorted by `index` and can be reordered in partition key order
    /// using `sequence` field
    pub pk_indexes: Vec<PartitionKeyIndex>,
    // ...
}
```

wprzytula: I knew the snippet, yet I'm still confused.

scylla/src/transport/session.rs
wprzytula: I've tried it up and gave up. Apparently, the `Iterator` returned by `PartitionKey::iter()` is less problematic in terms of lifetime issue than `PartitionKey` itself.

scylla/src/statement/prepared_statement.rs
piodul:
```rust
let meta = make_meta(
    [
        ColumnType::TinyInt,
        ColumnType::SmallInt,
        ColumnType::Int,
        ColumnType::BigInt,
        ColumnType::Blob,
    ],
    [4, 0, 3],
);
```

The above metadata was supposed to simulate the case where you have a query that constructs the partition key from 4th, 0th and 3rd bind marker - e.g.:

```cql
INSERT INTO tbl (pk2, v1, v2, pk3, pk1) VALUES (?, ?, ?, ?, ?);
```

So, the order of the numbers in the second argument was intentional, and requiring the numbers to be sorted is wrong. The `make_meta` is incorrect and should sort the generated pk indexes itself, as it is suggested by the comment in `PreparedMetadata` and as it is actually done in the deserialization code.

The test currently asserts the wrong thing. As a result, there is a bug in `PartitionKeyDecoder` which manifests when you have multiple partition key columns with different types and they specified out of order in the bind markers. I'll send a fix, then let's rebase.

wprzytula: Rebased.

piodul (approved):

Issue #660, opened 2023-03-15
Decrease excessive visibility from `pub` throughout the crate 


wprzytula:
Our codebase suffers from the prevalent light-hearted use of the `pub` qualifier. In order to reduce chances of having to incur backward-incompatible changes in the future, we should tackle with this problem once and (hopefully) forever.
This should definitely be done before we release 1.0.

Some suggestions have already been issued (e.g. #497).

wprzytula:
Most work regarding this issue has already been done. Keeping this open, however, to remember us about revisiting the whole API before we release 1.0.

PR #763, opened 2023-07-19
connection: remove re-export of QueryResult

avelanarius:
Remove re-export of `QueryResult` (from `scylla::transport::query_result::QueryResult` to `scylla::transport::connection::QueryResult`). The motivation for this change is to reduce the number of re-exports in the codebase (some of them don't make sense and are there for legacy reasons), as a part of our effort of stabilizing the API.

It probably makes sense to follow this up with moving `QueryResult` from `transport/` module to a more appropriate module (`query/` module?).

Refs #660

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -60,10 +60,7 @@ use crate::statement::prepared_statement::PreparedStatement;
 use crate::statement::Consistency;
 use crate::transport::session::IntoTypedRows;
 use crate::transport::Compression;
-
-// Existing code imports scylla::transport::connection::QueryResult because it used to be located in this file.
-// Reexport QueryResult to avoid breaking the existing code.
-pub use crate::QueryResult;
+use crate::QueryResult;
 
 // Queries for schema agreement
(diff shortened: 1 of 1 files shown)


piodul (approved):

piodul:
I approved it, but... should we really merge it? The module hierarchy needs some redesign and, as you noted, we should ultimately move `QueryResult` to a different module, so why force users to update their import path twice (first from `scylla::QueryResult` to `scylla::transport::query_result::QueryResult`, then to the correct path) instead of once?

avelanarius:
> I approved it, but... should we really merge it? The module hierarchy needs some redesign and, as you noted, we should ultimately move `QueryResult` to a different module, so why force users to update their import path twice (first from `Scylla::QueryResult` to `scylla::transport::query_result::QueryResult`, then to the correct path) instead of once?

The users won’t have to update their import from `scylla::QueryResult` to `scylla::transport::query_result::QueryResult` - `scylla::QueryResult`  is not affected, only (more obscure?) `scylla::transport::connection::QueryResult`. As you can see in this patch, I didn’t have to adjust any examples/other files, so I guess this probably will be the case for a typical user code.

Issue #720, opened 2023-05-10
Change the partition key's representation in tokio tracing to a human-readable form


piodul:
Currently, in `RequestSpan`, we are including a "routing key" - which is the partition key serialized in a particular way. The driver computes it in order to apply MurmurHash3 on it and compute the token, but it is technically also possible for users to decode a partition key from it. However, while this was simple to implement, this form is hard to parse manually and one needs to know the types of partition key's columns.

This information is available to the driver at the moment when the routing key is included in tracing, so for the sake users' convenience, the driver should decode it itself. In most of the cases it can be done without any allocations (especially after #462).

PR #766, opened 2023-07-21
session: include partition key in RequestSpan in human readable form

piodul:
This PR changes the format used by `RequestSpan` to represent the partition key for a given query. Previously, it was the same representation that the driver creates before applying MurmurHash3 to it, encoded in base64 - the format is documented so technically users can decode it themselves, but it's quite inconvenient. Now, the driver decodes the contents of the partition key and represents it in a human-readable form, similar to the CQL syntax.

For example, given the schema

```cql
create table ks.t (pk1 int, pk2 time, pk3 frozen<set<text>>, primary key(pk1, pk2, pk3));
```

and the request

```rust
session
    .execute(
        &prepared,
        (
            6622i32,
            Time(chrono::Duration::seconds(12323)),
            vec!["quick", "brown", "fox"],
        ),
    )
    .await
    .unwrap();
```

previously, the encoded key would look like this:

```
partition_key=0004000019de00000800000b352c099e0000001d0000000300000005717569636b0000000562726f776e00000003666f7800
```

but now it looks like this:

```
partition_key=6622,'03:25:23.000000000',{'quick','brown','fox'}
```

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/720

## Pre-review checklist

### DIFF
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -1,3 +1,4 @@
+use byteorder::{BigEndian, ReadBytesExt};
 use bytes::{BufMut, Bytes, BytesMut};
 use smallvec::{smallvec, SmallVec};
 use std::convert::TryInto;
@@ -5,6 +6,9 @@ use std::sync::Arc;
 use std::time::Duration;
 use thiserror::Error;
 use uuid::Uuid;
+
+use scylla_cql::frame::frame_errors::ParseError;
+use scylla_cql::frame::response::result::{deser_cql_value, CqlValue};
 
 use super::StatementConfig;
 use crate::frame::response::result::PreparedMetadata;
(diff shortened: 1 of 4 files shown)


scylla/src/utils/pretty.rs
```diff
@@ -19,3 +23,310 @@ impl<'a> UpperHex for HexBytes<'a> {
...
+            format!("{}", CqlValueDisplayer(CqlValue::List(list_or_set.clone()))),
+            "[1,3,2]"
+        );
+        assert_eq!(
+            format!("{}", CqlValueDisplayer(CqlValue::Set(list_or_set.clone()))),
+            "{1,3,2}"
+        );
+
+        let mut tuple: Vec<_> = list_or_set.into_iter().map(Some).collect();
+        tuple.push(None);
```
wprzytula, line 301: Possibly cleaner:

wprzytula:
```suggestion
        let tuple = Vec::from_iter(list_or_set.into_iter().map(Some).chain(std::iter::once(None)));
```

wprzytula: `FromIterator` should be encouraged, as it reduces verbosity induced by `collect()` (either type hint or turbofish).

scylla/src/statement/prepared_statement.rs
```diff
@@ -337,3 +341,132 @@ pub enum PartitionKeyError {
...
+
+impl<'pk> Iterator for PartitionKeyDecoder<'pk> {
+    type Item = Result<CqlValue, ParseError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.index >= self.prepared_metadata.pk_indexes.len() {
+            return None;
+        }
+
+        let idx = self.prepared_metadata.pk_indexes[self.index].index as usize;
```
wprzytula, line 378: Please rename it to `col_idx` for less ambiguity.

scylla/src/statement/prepared_statement.rs
```diff
@@ -337,3 +341,132 @@ pub enum PartitionKeyError {
...
+// route it to a random replica instead and let the DB reject it. In the
+// meantime, this struct will return some garbage data while it tries to
+// decode the key, but nothing bad like a panic should happen otherwise.
+// The struct is currently only used for printing the partition key, so that's
+// completely fine.
+#[derive(Clone, Copy)]
+pub(crate) struct PartitionKeyDecoder<'pk> {
+    prepared_metadata: &'pk PreparedMetadata,
+    partition_key: &'pk [u8],
+    index: usize,
```
wprzytula, line 357: Perhaps `value_index` or a similar name could help distinguish this conceptually from column indices, which are involved too.

scylla/src/statement/prepared_statement.rs
```diff
@@ -337,3 +341,132 @@ pub enum PartitionKeyError {
...
+                std::io::ErrorKind::UnexpectedEof,
+                "value too short",
+            )));
+        }
+        let col = &buf[..len];
+        *buf = &buf[len..];
+        buf.read_u8()?;
+        Ok(col)
+    }
+}
```
wprzytula, line 406: Please add some comments for future readers, explaining the format, or at least provide a link to the source describing it.

scylla/src/statement/prepared_statement.rs
```diff
@@ -337,3 +341,132 @@ pub enum PartitionKeyError {
...
+                    typ: ColumnType::Int,
+                })
+                .collect(),
+            pk_indexes: (0..col_count)
+                .map(|i| PartitionKeyIndex {
+                    index: i as u16,
+                    sequence: i as u16,
+                })
+                .collect(),
+        };
```
wprzytula, line 438: Is such `PreparedMetadata` general enough to cover the most important test cases? I mean, column type is always `Int`, and sequence numbers are equal to indices. I'm not sure though if we need to test other cases _here_.

wprzytula (commented):
Looks good, only some naming/comments nits.

scylla/src/utils/pretty.rs
piodul: > `FromIterator` should be encouraged, as it reduces verbosity induced by `collect()` (either type hint or turbofish).

Why? I quite like `.collect()` to be fair. It leads to more natural flow in the code. When you work with iterators, you perform a series of steps (such as map, filter, etc.) and then finally you `.collect()` the results.

wprzytula: This is a valid counterargument.

piodul:
v2:

- `std::iter::chain` instead of collect to vector then `push`,
- Rename `idx` -> `col_idx`,
- Rename `index` -> `value_index`,
- Added a link to the explanation of the format that `PartitionKeyDecoder` decodes from,
- Reworked the tests for `PartitionKeyDecoder` a bit and added a more complicated case that involves out-of-order columns and different column types.

wprzytula (approved):

PR #768, opened 2023-07-21
session: remove estimate_replicas_for_query, get_tracing_info_custom

avelanarius:
This PR removes two public functions from `Session`: `estimate_replicas_for_query` and `get_tracing_info_custom`.

Those changes are a part of our effort to stabilize the API and reduce the number of pub functions.

Refs https://github.com/scylladb/scylla-rust-driver/issues/660

#### `estimate_replicas_for_query`
Remove `estimate_replicas_for_query` from Session. A user of this function can switch to using `get_token_endpoints()` from `ClusterData` directly.

It did not make sense for `estimate_replicas_for_query()` to be in `Session`, as this is too much low-level function for `Session` and doesn't fit the abstraction of `Session`.

#### `get_tracing_info_custom`
Remove `Session::get_tracing_info_custom` - a variant of `Session::get_tracing_info` that allows for additional customization of the fetch process (number of attempts, interval between attempts, consistency level of fetching tracing info). This customization is now moved to `SessionConfig` (as new parameters) and new helper functions were introduced to `SessionBuilder`.

Since now this customization is on `SessionConfig`, `create_new_session_builder()` in `test_utils.rs` returns a `SessionBuilder` with pre-configured large number of `TracingInfo` fetch attempts, longer delays (to account for a slow CI). Previously each tracing test had to configure it on its own by using `get_tracing_info_custom`.

## Pre-review checklist

### DIFF
--- a/examples/tracing.rs
+++ b/examples/tracing.rs
@@ -7,7 +7,7 @@ use scylla::batch::Batch;
 use scylla::statement::{
     prepared_statement::PreparedStatement, query::Query, Consistency, SerialConsistency,
 };
-use scylla::tracing::{GetTracingConfig, TracingInfo};
+use scylla::tracing::TracingInfo;
 use scylla::transport::iterator::RowIterator;
 use scylla::QueryResult;
 use scylla::{Session, SessionBuilder};
@@ -21,7 +21,10 @@ async fn main() -> Result<()> {
     let uri = env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
 
     println!("Connecting to {} ...", uri);
-    let session: Session = SessionBuilder::new().known_node(uri).build().await?;
+    let session: Session = SessionBuilder::new()
+        .known_node(uri.as_str())
+        .build()
+        .await?;
 
     session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await?;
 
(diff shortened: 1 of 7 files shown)


scylla/src/utils/test_utils.rs
```diff
@@ -60,27 +61,34 @@ pub(crate) async fn supports_feature(session: &Session, feature: &str) -> bool {
...
+        }
+    };
+
+    // The reason why we enable so long waiting for TracingInfo is... Cassandra. (Yes, again.)
+    // In Cassandra Java Driver, the wait time for tracing info is 10 seconds, so here we do the same.
+    // However, as Scylla usually gets TracingInfo ready really fast (our default interval is hence 3ms),
+    // we stick to a not-so-much-terribly-long interval here.
+    session_builder
+        .tracing_info_fetch_attempts(NonZeroU32::new(50).unwrap())
+        .tracing_info_fetch_interval(Duration::from_millis(200))
```
piodul, line 93: To be fair, our defaults are awfully low (5 times, 3ms) - I think we should increase them. Not in this PR, of course.

piodul (approved):

scylla/src/utils/test_utils.rs
avelanarius: Curiously this is the same default as in Java Driver 3.x and I never saw a CI failure in that driver (we test also Cassandra 3.x) caused by that.

examples/tracing.rs
```diff
@@ -21,7 +21,10 @@ async fn main() -> Result<()> {
     let uri = env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
 
     println!("Connecting to {} ...", uri);
-    let session: Session = SessionBuilder::new().known_node(uri).build().await?;
+    let session: Session = SessionBuilder::new()
+        .known_node(uri.clone())
```
wprzytula, line 25: Better to pass `uri.as_str()` and save a clone.

wprzytula (approved):
Apart from two nits, LGTM.

scylla/src/utils/test_utils.rs
```diff
@@ -60,27 +61,34 @@ pub(crate) async fn supports_feature(session: &Session, feature: &str) -> bool {
...
+        #[cfg(scylla_cloud_tests)]
+        {
+            use crate::transport::session_builder::CloudMode;
+            use crate::CloudSessionBuilder;
+            use std::path::Path;
+
+            std::env::var("CLOUD_CONFIG_PATH")
+                .map(|config_path| CloudSessionBuilder::new(Path::new(&config_path)))
+                .unwrap()
+                .expect("failed to create a session in cloud mode")
```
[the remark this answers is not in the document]
avelanarius: Added two `expect()` with better error messages in the ammended commit. Ultimately, I didn't add a `format!()` with path in `.expect("Failed to initialize CloudSessionBuilder")` as I would have had to put `config_path` in some temporary variable, making the code longer.

avelanarius:
Addressed @wprzytula comments in the ammended commit.

Issue #756, opened 2023-07-03
Re-resolve hostnames as fallback when all hosts are unreachable


wprzytula:
If all nodes in the cluster change their IPs at one time, driver can no longer ever contact the cluster; the only solution is to restart the driver. A fallback should be added to the control connection reconnect logic so that when no known host is reachable, all hostnames provided in `SessionConfig` (initial contact points) are reresolved and control connection is attempted to be opened to any of them. If this succeeds, a metadata fetch is issued normally and the whole cluster is discovered with its new IPs.

An analogous feature has been merged to gocql lastly: https://github.com/gocql/gocql/pull/1708

PR #770, opened 2023-07-26
Re-resolve hostnames as fallback when control connection is broken

wprzytula:
## Motivation

See the referenced issue (#756).

### Main matter
- **fallback logic is introduced** to `read_metadata()`.
This is the crucial step in boosting the driver's robustness in case of sudden IP changes in the cluster. In case that the control connection fails to fetch metadata, after all known peers fail to be queried, the fallback phase is entered: addresses of all initially known nodes are resolved again and fetching metadata from them is attempted.
Until now, if all nodes in the cluster changed their IPs at once (which is particularly easy in a single-node cluster), the driver was unable to contact the cluster ever more. Now, the driver reconnects the control connection with the next metadata fetch attempt (after at most 60 seconds, as configured in `ClusterWorker`), fetches new topology, discovers all nodes' addresses and finally connects to them.
- **control connection reconnect attempts are more frequent.**
If the control connection is faulty and hence metadata fetch fails, it is advisable that further attempts to reconnect and fetch take place more frequently. The motivation is: if the control connection fails, it is possible that the node has changed its IP and hence we need to fetch new metadata ASAP to discover its new address. Therefore, the `ClusterWorker`'s sleep time is changed from 60 seconds to 1 second once a metadata fetch fails, and is only reverted back to 60 seconds after a fetch succeeds.
- **control connection reconnect attempt begins immediately after it breaks.**
Until now, if the control connection node changed it IP, we would discover that node only after the next metadata fetch is issued, which would possibly happen only after 60 seconds (if previous fetch succeeded). To make the rediscovery faster, **immediate signalling** that the control connection got broken **is introduced**, so that `ClusterWorker` begins instantly its every-1-second-attempt phase.

### Bonus
- `ContactPoint` is **renamed** to `ResolvedContactPoint`, to stress its difference from `KnownNode` (which is an unresolved contact point, essentially).
- various node-related structs (`ResolvedContactPoint`, `KnownNode`, `CloudEndpoint`) and hostname resolution-related routines are **moved** from `cluster` and `session` modules to the `node` module. This contributes to making the `session` module cleaner, as well as keeps those entities in the most conceptually related module.

## Evaluation
Manual tests were conducted. They were identical to ones performed before merging the analogous enhancement to gocql. Both single- and multi-node clusters were tested. 
The tests involved random connection breaks between arbitrary nodes and the driver, as well as changes of IPs of all the nodes in the cluster at once.
The implemented solution showed to be very robust in the tests: immediately after losing control connection, the reconnect-attempt-phase begins. As soon as any node (either one known initially or one known from recently fetched metadata) becomes reachable, a control connection is opened and metadata is fetched successfully, so the whole cluster is discoverable. The driver performed stable, with no fluctuations in behaviour over time.

Fixes: #756 

## Pre-review checklist


scylla/src/transport/node.rs
```diff
@@ -215,6 +223,110 @@ impl Hash for Node {
...
+
+    ret.ok_or_else(|| {
+        io::Error::new(
+            io::ErrorKind::Other,
+            format!("Empty address list returned by DNS for {}", hostname),
+        )
+    })
+}
+
+/// Transforms the given [`KnownNode`]s into [`ContactPoint`]s.
```
avelanarius, line 282:
```suggestion
/// Transforms the given [`KnownNode`]s into [`ResolvedContactPoint`]s.
```

wprzytula:
Rebased on main.

wprzytula:
Rebased on `main`.

scylla/src/transport/topology.rs
```diff
@@ -380,6 +384,7 @@ impl MetadataReader {
     /// Creates new MetadataReader, which connects to initially_known_peers in the background
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn new(
+        initial_known_nodes: Vec<KnownNode>,
         initially_known_peers: Vec<ResolvedContactPoint>,
```
piodul: What is the difference between `initial_known_nodes` and `initially_known_peers`? Their names are very similar.

scylla/src/transport/topology.rs
```diff
@@ -469,6 +470,23 @@ impl MetadataReader {
...
+                .retry_fetch_metadata_on_nodes(
+                    initial,
+                    initial_peers
+                        .into_iter()
+                        .map(UntranslatedEndpoint::ContactPoint),
+                    prev_err,
+                )
+                .await;
+        }
+
```
piodul, line 496: Does it make sense to do this in case when this is an initial metadata fetch? E.g. if none of the addresses are reachable? Won't it make the connection process (which will ultimately result in a failure) twice as long?

scylla/src/transport/cluster.rs
```diff
@@ -542,6 +549,25 @@ impl ClusterWorker {
...
+                            // so subsequent attempts will be issued every second.
+                        }
+                        Err(tokio::sync::broadcast::error::RecvError::Lagged(_)) => {
+                            // This is very unlikely; we would have to have a lot of concurrent
+                            // control connections opened and broken at the same time.
+                            // The best we can do is ignoring this.
+                        }
+                        Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                            // If control_connection_repair_channel was closed then MetadataReader was dropped,
+                            // we can stop working.
```
piodul, line 567: In that case, let's `return`.

scylla/src/transport/topology.rs
wprzytula: Good catch. Let's `if` the initial fetch case out.

scylla/src/transport/topology.rs
wprzytula: Perhaps a better naming could be: `initial_nodes_unresolved` and `initial_nodes_resolved`. WDYT?

piodul: It's strange that we need to pass both resolved and unresolved nodes, it sounds redundant. Why can't `MetadataReader` do the resolution itself?

wprzytula: We resolve early to know that we have at least one succesfully resolved address upon creating `Session`.

piodul: How is that in conflict with resolving addresses in `MetadataReader::new` or `MetadataReader::read_metadata`? Both functions are called in the process of creating a `Session`.

wprzytula: That's right, we can alter `Cluster::new()` return type from `Result<_, QueryError>` to `Result<_, NewSessionError>` and then we can indeed resolve hostnames in `MetadataReader::new()`. Good idea!

wprzytula:
@piodul Rebased on main.

Issue #701, opened 2023-04-19
Un-pub `StatementConfig`


piodul:
The `statement::StatementConfig` type is public and has public fields, but it is only used internally and does not appear in any public API. In the current state it is impossible to change it without introducing a breaking change. We should make it and its fields private so that we have more flexibility in changing it.

Refs: https://github.com/scylladb/scylla-rust-driver/issues/660

PR #772, opened 2023-07-28
statement: unpub StatementConfig

wprzytula:
`StatementConfig` is only used internally and does not appear in any public API.

- It is made `pub(crate)` together with its fields so that we have more flexibility in changing it without introducing a breaking change.
- After that, `retry_policy` field is moved from `Query`, `PreparedStatement` and `Batch` into `StatementConfig`, as it suits there best.

Refs: #660
Fixes: #701 
## Pre-review checklist


piodul (approved):
A small, but nice cleanup. Thanks.

PR #773, opened 2023-07-30
codewide: Decrease excessive visibility from `pub` - nonbreaking changes

wprzytula:
In spirit of the referenced issue, `scylla` crate is saved from the hazardous siege of `pub` qualifiers. This PR collects the changes that I believe to be non-breaking, so that they can be merged immediately.

Refs: #660 

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/caching_session.rs
+++ b/scylla/src/transport/caching_session.rs
@@ -19,10 +19,10 @@ use std::hash::BuildHasher;
 /// to the `CachingSession::execute` family of methods.
 #[derive(Debug)]
 struct RawPreparedStatementData {
-    pub id: Bytes,
-    pub is_confirmed_lwt: bool,
-    pub metadata: PreparedMetadata,
-    pub partitioner_name: PartitionerName,
+    id: Bytes,
+    is_confirmed_lwt: bool,
+    metadata: PreparedMetadata,
+    partitioner_name: PartitionerName,
 }
 
 /// Provides auto caching while executing queries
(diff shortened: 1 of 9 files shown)


piodul (commented):
Looks good apart from one comment.

I'm worried that we might introduce more items that superficially look public but are actually not. I hope that, in the future, with a module structure that is more thought out than the current one, it will become more obvious whether `pub` items are actually visible based on their location...

wprzytula:
> Looks good apart from one comment.
> 
> I'm worried that we might introduce more items that superficially look public but are actually not. I hope that, in the future, with a module structure that is more thought out than the current one, it will become more obvious whether `pub` items are actually visible based on their location...

A simple way to prevent _false `pub`s_ is to stop using them without a strong need to do so. I can see this has become the new way in the driver recently, as before it was popular to keep methods or fields `pub` regardless of the level of user's intended access to them.

wprzytula:
v2: dropped the controversial commit regarding `MetadataReader` associated functions.

piodul:
> v2: dropped the controversial commit regarding MetadataReader associated functions.

@wprzytula I think you dropped the wrong commit - the one that unpubs methods of the metadata reader, not the ones that convert some free functions to static methods.

wprzytula:
> > v2: dropped the controversial commit regarding MetadataReader associated functions.
> 
> @wprzytula I think you dropped the wrong commit - the one that unpubs methods of the metadata reader, not the ones that convert some free functions to static methods.

I should be OK now.

piodul (approved):

PR #775, opened 2023-07-30
default_policy: Builder: forbid preferring rack without specifing datacenter

wprzytula:
This is an API restriction continuing of what has already been done in bb09756a1d97da4a870f87f00c5b2c081e8079b2: as the case of providing preferred rack without specifying the preferred datacenter is invalid, in spirit of the principle _“Make illegal states unrepresentable”_, `DefaultBuilder`'s API is changed accordingly to disallow creating such configuration.

## Pre-review checklist

### DIFF
--- a/docs/source/load-balancing/default-policy.md
+++ b/docs/source/load-balancing/default-policy.md
@@ -10,8 +10,7 @@ for queries with non-local consistency mode is also supported.
 `builder()` method of `DefaultPolicy` returns a new instance of
 `DefaultPolicyBuilder` with the following default values:
 
-- `preferred_datacenter`: `None`
-- `preferred_rack`: `None`
+- `preferences`: no particular datacenter/rack preference
 - `is_token_aware`: `true`
 - `permit_dc_failover`: `false`
 - `latency_awareness`: `None`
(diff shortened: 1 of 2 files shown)


docs/source/load-balancing/default-policy.md
```diff
@@ -10,8 +10,7 @@ for queries with non-local consistency mode is also supported.
 `builder()` method of `DefaultPolicy` returns a new instance of
 `DefaultPolicyBuilder` with the following default values:
 
-- `preferred_datacenter`: `None`
-- `preferred_rack`: `None`
+- `preferences`: `ReplicaLocationPreference::Any`
 - `is_token_aware`: `true`
 - `permit_dc_failover`: `false`
 - `latency_awareness`: `None`
```
piodul, line 16: It's not clear at a first glance what "`preferences`: `ReplicaLocationPreference::Any`" means. I think we should change this section to explain the default parameters in plain English and not refer to the names of internal fields of the load balancing policy.

docs/source/load-balancing/default-policy.md
```diff
@@ -35,36 +33,34 @@ let default_policy = DefaultPolicy::builder()
...
-is set, the policy will treat nodes in that datacenter as "local" nodes, and
-nodes in other datacenters as "remote" nodes. This affects the order in which
-nodes are returned by the policy when selecting replicas for read or write
-operations. If no preferred datacenter is specified, the policy will treat all
-nodes as local nodes.
+`preferences` allow the load balancing policy to prioritize nodes based on their
+availability zones in the preferred datacenter, too. When `preferences` are set
+to `DatacenterAndRack(my_dc, my_rack)`, the policy will first return replicas
+in the local rack in the preferred datacenter, and then the other replicas
+in the datacenter (followed by remote nodes).
```
piodul, line 50: I think this paragraph could be more clear if it didn't mention implementation details so much. It should explain that the load balancing policy has three modes and they behave such and such without referring to `preferences` and `ReplicaLocationPreference`.

piodul (approved):

PR #776, opened 2023-07-30
configurable topology refresh duration

rishabharyal:
This pull requests a configurable way to topology refresh duration. Two files session.rs and cluster.rs have been modified to set the session configuration and consume the duration configuration respectively.

Note: I am not sure if we need any test cases in this case as the duration is still provided from where its called. 

Fixes: #232

### DIFF
--- a/docs/source/connecting/connecting.md
+++ b/docs/source/connecting/connecting.md
@@ -20,7 +20,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
         .known_node("127.0.0.72:4321")
         .known_node("localhost:8000")
         .connection_timeout(Duration::from_secs(3))
-        .cluster_topology_refresh_interval(Duration::from_secs(10))
+        .cluster_metadata_refresh_interval(Duration::from_secs(10))
         .known_node_addr(SocketAddr::new(
             IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
             9000,
@@ -36,7 +36,7 @@ After successfully connecting to some specified node the driver will fetch topol
 other nodes in this cluster and connect to them as well.
 
 The driver refreshes the cluster metadata periodically, which contains information about cluster topology as well as the cluster schema. By default, the driver refreshes the cluster metadata every 60 seconds. 
-However, you can set the `cluster_topology_refresh_interval` to a non-negative value to periodically refresh the cluster topology. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
+However, you can set the `cluster_metadata_refresh_interval` to a non-negative value to periodically refresh the cluster topology. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
 
 Scylla Serverless is an elastic and dynamic deployment model. When creating a `Session` you need to
 specify the secure connection bundle as follows:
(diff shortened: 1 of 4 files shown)


docs/source/connecting/connecting.md
```diff
@@ -34,6 +35,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+You can also set the `cluster_topology_refresh_interval` to a non-zero value to periodically refresh the
+cluster topology. This is useful when you do not have unexpected amount of traffic or when you
+have an extra traffic causing topology to change frequently. The default value is 60,
+which means that the driver will refresh the topology every 60 seconds.
+
```
piodul: IMO, the way the first sentence is worded ("You can also set the `cluster_topology_refresh_interval` to a non-zero value to periodically refresh the cluster topology") implicitly suggests that the driver does not refresh the cluster topology if you don't set this parameter. This is clarified at the end of the paragraph, but somebody that just skims over the docmentation and reads the beginnings of the paragraphs might miss it.

Personally, I'd structure it like this:

- First mention that the driver periodically refreshes cluster topology,
- Then mention that the interval is 60s by default, but can be changed via `cluster_topology_refresh_interval`.

piodul (changes requested):
To be precise, the parameter that you introduced affects how frequently the whole _cluster metadata_ is refreshed. Cluster metadata not only contains information about the cluster topology, but also information about the cluster schema. The issue #232 was created long ago, when the only metadata that the driver used to fetch in `ClusterWorker::perform_refresh()` was indeed cluster topology, but it has changed since.

I think the current behavior of the parameter is fine and it can control fetching of the whole metadata, but the names used in the code and in the documentation should be adjusted.

Apart from that and some other small comments - looks good.

piodul:
> Note: I am not sure if we need any test cases in this case as the duration is still provided from where its called.

I don't have a good idea on how to write a good test for this. Usually, a `SessionBuilder` test + a manual test to see whether the parameter is properly propagated is sufficient.

rishabharyal:
Hi @piodul, are any further changes needed? Let me know in case you missed this.

wprzytula (changes requested):
There are a number of places with names still containing "topology refresh" instead of "metadata refresh". Please fix them.

docs/source/connecting/connecting.md
```diff
@@ -36,7 +36,7 @@ After successfully connecting to some specified node the driver will fetch topol
 other nodes in this cluster and connect to them as well.
 
 The driver refreshes the cluster metadata periodically, which contains information about cluster topology as well as the cluster schema. By default, the driver refreshes the cluster metadata every 60 seconds. 
-However, you can set the `cluster_topology_refresh_interval` to a non-negative value to periodically refresh the cluster topology. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
+However, you can set the `cluster_metadata_refresh_interval` to a non-negative value to periodically refresh the cluster topology. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
```
wprzytula, line 39: Still two uses of `topology` where `metadata` would be more appropriate.

wprzytula (changes requested):

docs/source/connecting/connecting.md
rishabharyal: Yep, took care of that.

wprzytula (approved):
Thanks, LGTM.
Whoever maintainer is going to merge this, I highly recommend squash&merge.

piodul:
> Hi @piodul, are any further changes needed? Let me know in case you missed this.

@rishabharyal Sorry, I was out on vacation for the last two weeks.

PR #777, opened 2023-07-30
default policy: Rack awareness without token awareness

wprzytula:
Until now, the default policy respected rack preference only when it came to replicas.
Considering non-token-aware queries (e.g. unprepared statements), they would never prioritize the preferred rack.

Now, local rack nodes are always preferred before other nodes, both in `pick()` and in `fallback()`.
Existing tests adjusted for new behaviour.

The PR contains breaking changes, as it slightly changes signature of `ReplicaSet::choose_filtered()` (note the reference added).

## Pre-review checklist

### DIFF
--- a/docs/source/load-balancing/default-policy.md
+++ b/docs/source/load-balancing/default-policy.md
@@ -44,7 +44,7 @@ policy to prioritize nodes based on their location. It has three modes:
 
 When a datacenter `"my_dc"` is preferred, the policy will treat nodes in `"my_dc"`
 as "local" nodes, and nodes in other datacenters as "remote" nodes. This affects
-the order in which nodes are returned by the policy when selecting replicas for
+the order in which nodes are returned by the policy when selecting nodes for
 read or write operations. If no datacenter is preferred, the policy will treat
 all nodes as local nodes.
 
@@ -52,7 +52,8 @@ all nodes as local nodes.
 availability zones (racks) in the preferred datacenter, too. When a datacenter
 and a rack are preferred, the policy will first return replicas in the local rack
 in the preferred datacenter, and then the other replicas in the datacenter
-(followed by remote replicas).
+(followed by remote replicas). After replicas, the other node will be ordered
+similarly, too (local rack nodes, local datacenter nodes, remote nodes).
 
 When datacenter failover is disabled (`permit_dc_failover` is set to
 false), the default policy will only include local nodes in load balancing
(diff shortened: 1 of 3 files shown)


scylla/src/transport/load_balancing/default.rs
```diff
@@ -170,8 +170,22 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
         // some alive local node.
         // If there was no preferred datacenter specified, all nodes are treated as local.
         let nodes = self.preferred_node_set(cluster);
-        let picked = Self::pick_node(nodes, &self.pick_predicate);
-        if let Some(alive_local) = picked {
+
```
piodul, line 173: The `token_with_strategy.is_some()` and `token_with_strategy_is_none()` paths look very similar. Perhaps we could unify them somehow? Not necessarily here, just food for thought for future improvements.

piodul (approved):
LGTM

@havaker please take a look as well

scylla/src/transport/load_balancing/default.rs
```diff
@@ -1575,7 +1601,8 @@ mod tests {
                 expected_groups: ExpectedGroupsBuilder::new()
                     .deterministic([B]) // pick local rack replicas
                     .deterministic([C]) // fallback replicas
-                    .group([A, G]) // local nodes
+                    .group([A]) // local rack nodes
```
havaker, line 1604: > Considering non-token-aware queries (e.g. unprepared statements), they would never prioritize the preferred rack.

I'd like to see a test which has `routing_info` with a `token` field set to `None`.

docs/source/load-balancing/default-policy.md
```diff
@@ -54,8 +54,8 @@ serve requests.
 
 The `preferred_rack` field in `DefaultPolicy` allows the load balancing policy to
 prioritize nodes based on their availability zones in the preferred datacenter.
-When a preferred rack is set, the policy will first return replicas in the local rack
-in the preferred datacenter, and then the other replicas in the datacenter.
+When a preferred rack is set, the policy will first return nodes in the local rack
```
havaker, line 57: kudos for updating the documentation!

havaker (changes requested):

wprzytula:
v2:
- renamed `ReplicaLocation[Criteria|Location]` to `Node...`
- provided a test for rack awareness when no token is known
- rebased on `main`

havaker (approved):

piodul (approved):

PR #779, opened 2023-07-31
Slim down schema agreement API

wprzytula:
Schema agreement API is another place in the driver's code that can be polished towards API stability.

## What's done

1. `fetch_schema_version() ` is removed. 
It can be a nasty source of race bugs when schema version is fetched without awaiting for schema agreement. It is especially risky when one awaits schema agreement and subsequently fetches schema version, because the schema version can change in the meantime and schema may no longer be in agreement. To cope with the problem, `await_schema_agreement()` now returns the agreed schema version, and  `fetch_schema_version()` is removed alltogether.
2. `await_schema_agreement()` gets implicit timeout.
    Similarly to
    - what Java driver 3 does in regard to awaiting schema agreement's API,
    - what was recently done in Rust driver in regard to tracing info's API,
    
    the API of awaiting schema agreement is altered so that:
    - `await_timed_schema_agreement()`, which takes explicit timeout, is
      removed,
    - `await_schema_agreement()` is now bound with a timeout that is set
      globally per-`Session` in `SessionConfig`.

    The motivation is that it could lead to application's deadlock when `await_schema_agreement()` was called with no timeout given and the cluster stoped synchronising (so schema agreement would be never reached), It is rarely desirable to block the calling application for arbitrarily long time, especially that some extreme situations are possible, such as a network partition between nodes.
    
3. `SessionBuilder`s API is accomodated to the changes. Docs are updated as well.

### Bonus
- missing docstrings were added for `AddressTranslator`.

Refs: #660 

## Pre-review checklist

### DIFF
--- a/docs/source/queries/schema-agreement.md
+++ b/docs/source/queries/schema-agreement.md
@@ -4,24 +4,11 @@ Sometimes after performing queries some nodes have not been updated, so we need 
 There is a number of methods in `Session` that assist us.
 Every method raise `QueryError` if something goes wrong, but they should never raise any errors, unless there is a DB or connection malfunction.
 
-### Checking schema version
-`Session::fetch_schema_version` returns an `Uuid` of local node's schema version.
-
-```rust
-# extern crate scylla;
-# use scylla::Session;
-# use std::error::Error;
-# async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
-println!("Local schema version is: {}", session.fetch_schema_version().await?);
-# Ok(())
-# }
-```
-
 ### Awaiting schema agreement
 
 `Session::await_schema_agreement` returns a `Future` that can be `await`ed as long as schema is not in an agreement.
 However, it won't wait forever; `SessionConfig` defines a timeout that limits the time of waiting. If the timeout elapses,
-the return value is `false`, otherwise it is `true`.
+the return value is `Err(QueryError::RequestTimeout)`, otherwise it is `Ok(schema_version)`.
 
 ```rust
 # extern crate scylla;
@@ -62,7 +49,7 @@ If you want to check if schema is in agreement now, without retrying after failu
 # use scylla::Session;
 # use std::error::Error;
 # async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
-if session.check_schema_agreement().await? {
+if session.check_schema_agreement().await?.is_some() {
     println!("SCHEMA AGREED");
 } else {
     println!("SCHEMA IS NOT IN AGREEMENT");
--- a/docs/source/queries/schema-agreement.md
+++ b/docs/source/queries/schema-agreement.md
@@ -1,11 +1,11 @@
 # Schema agreement
 
-Sometimes after performing queries some nodes have not been updated so we need a mechanism that checks if every node have agreed schema version.
-There are four methods in `Session` that assist us. 
+Sometimes after performing queries some nodes have not been updated, so we need a mechanism that checks if every node have agreed on schema version.
+There is a number of methods in `Session` that assist us.
 Every method raise `QueryError` if something goes wrong, but they should never raise any errors, unless there is a DB or connection malfunction.
 
 ### Checking schema version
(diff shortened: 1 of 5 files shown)


scylla/src/transport/session.rs
```diff
@@ -1821,77 +1821,33 @@ impl Session {
...
-        Ok(())
     }
 
-    pub async fn await_timed_schema_agreement(
+    pub async fn await_schema_agreement(
         &self,
         timeout_duration: Duration,
-    ) -> Result<bool, QueryError> {
-        timeout(timeout_duration, self.await_schema_agreement())
+    ) -> Result<Option<Uuid>, QueryError> {
```
avelanarius, line 1836:
```suggestion
    ) -> Result<Uuid, QueryError> {
```

maybe the timeout should be represented as a timeout `QueryError`?

wprzytula: Good idea!

wprzytula: The only problem I see here is that we lose a way to conveniently distinguish cases of timeout (not yet in agreement) from cases of server-side or connection-related errors.

wprzytula: I've implemented it anyway.

docs/source/queries/schema-agreement.md
```diff
@@ -1,25 +1,14 @@
 # Schema agreement
```
avelanarius, line 1: This was a problem before your PR, but this documentation page is missing any mention about automatic schema agreement after schema-changing queries. 

(cc: @muzarski - what we talked about today: the driver does automatic schema agreement after such queries, no need to do it youself)

docs/source/queries/schema-agreement.md
```diff
@@ -87,4 +58,23 @@ if session.check_schema_agreement().await? {
...
+# use std::error::Error;
+# async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
+let session = SessionBuilder::new()
+    .known_node("127.0.0.1:9042")
+    .auto_await_schema_agreement(false)
+    .build()
+    .await?;
+# Ok(())
+# }
+```
```
avelanarius, line 80: I think it'd be better if this was at the start of this doc page (as a sort of warning: "You probably don't need to read the remaining portion of the doc - the driver does it for you")

avelanarius (approved):

scylla/src/transport/session.rs
```diff
@@ -83,6 +83,24 @@ use crate::authentication::AuthenticatorProvider;
...
+ * directly by the driver or should not be the preferred address to use to reach the node (for
+ * instance, the `rpc_address` set on ScyllaDB nodes might be a private IP, but some clients
+ * may have to use a public IP, or pass by a router, e.g. through NAT, to reach that node).
+ * This interface allows to deal with such cases, by allowing to translate an address as sent
+ * by a ScyllaDB node to another address to be used by the driver for connection.
+ *
+ * Please note that the "known nodes" addresses provided while creating the [`Session`]
+ * instance are not translated, only IP address retrieved from or sent by Cassandra nodes
+ * to the driver are.
+ */
```
piodul, line 103: How is this related to slimming down schema agreement API? This should be sent separately.

piodul (approved):

piodul:
Requires a rebase and dropping one of the commits, but otherwise LGTM.

scylla/src/transport/session.rs
wprzytula: This is a bonus, as described in the PR's description. Opening PRs have nontrivial overhead, so I prefer to add such tiny things to existing ones.

piodul: > Opening PRs has nontrivial overhead

Is it really that much of an overhead to send a single commit PR? You send it, then somebody reviews it quickly (as it usually is with such simple improvements) and later we can all forget about it before the rest is reviewed.

> I prefer to add such tiny things to existing ones.

No, please, for the sake of order let's not mix up things. Each PR should be concentrated on one topic.

wprzytula:
Rebased and got rid of irrelevant commit.

piodul:
Thanks. I'll merge after the CI checks finish.

PR #780, opened 2023-07-31
cluster: small performance optimisations

wprzytula:
This is a bunch of performance-related small optimisation of the `cluster` module, mainly focused on avoiding cloning where possible.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -599,13 +599,10 @@ impl ClusterWorker {
         cluster_data: Arc<ClusterData>,
         keyspace_name: &VerifiedKeyspaceName,
     ) -> Result<(), QueryError> {
-        let mut use_keyspace_futures = Vec::new();
-
-        for node in cluster_data.known_peers.values() {
-            let fut = node.use_keyspace(keyspace_name.clone());
-            use_keyspace_futures.push(fut);
-        }
-
+        let use_keyspace_futures = cluster_data
+            .known_peers
+            .values()
+            .map(|node| node.use_keyspace(keyspace_name.clone()));
         let use_keyspace_results: Vec<Result<(), QueryError>> =
             join_all(use_keyspace_futures).await;
 
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -301,19 +301,27 @@ impl ClusterData {
             // Take existing Arc<Node> if possible, otherwise create new one
             // Changing rack/datacenter but not ip address seems improbable
             // so we can just create new node and connections then
-            let node: Arc<Node> = match known_peers.get(&peer.host_id) {
+            let peer_host_id = peer.host_id;
+            let peer_address = peer.address;
+            let peer_tokens;
+
+            let node: Arc<Node> = match known_peers.get(&peer_host_id) {
                 Some(node) if node.datacenter == peer.datacenter && node.rack == peer.rack => {
-                    if node.address == peer.address {
+                    let (peer_endpoint, tokens) = peer.into_peer_endpoint_and_tokens();
+                    peer_tokens = tokens;
+                    if node.address == peer_address {
                         node.clone()
                     } else {
                         // If IP changes, the Node struct is recreated, but the underlying pool is preserved and notified about the IP change.
-                        Arc::new(Node::inherit_with_ip_changed(node, peer.to_peer_endpoint()))
+                        Arc::new(Node::inherit_with_ip_changed(node, peer_endpoint))
                     }
                 }
                 _ => {
                     let is_enabled = host_filter.map_or(true, |f| f.accept(&peer));
+                    let (peer_endpoint, tokens) = peer.into_peer_endpoint_and_tokens();
+                    peer_tokens = tokens;
                     Arc::new(Node::new(
-                        peer.to_peer_endpoint(),
+                        peer_endpoint,
                         pool_config.clone(),
                         used_keyspace.clone(),
                         is_enabled,
@@ -321,7 +329,7 @@ impl ClusterData {
                 }
             };
 
-            new_known_peers.insert(peer.host_id, node.clone());
+            new_known_peers.insert(peer_host_id, node.clone());
 
             if let Some(dc) = &node.datacenter {
                 match datacenters.get_mut(dc) {
@@ -336,7 +344,7 @@ impl ClusterData {
                 }
             }
 
-            for token in peer.tokens {
+            for token in peer_tokens {
                 ring.push((token, node.clone()));
             }
 
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -345,21 +345,18 @@ impl ClusterData {
 
         Self::update_rack_count(&mut datacenters);
 
-        let keyspace_strategies: Vec<Strategy> = metadata
-            .keyspaces
-            .values()
-            .map(|ks| ks.strategy.clone())
-            .collect();
-
-        let locator = tokio::task::spawn_blocking(move || {
-            ReplicaLocator::new(ring.into_iter(), keyspace_strategies.iter())
+        let keyspaces = metadata.keyspaces;
+        let (locator, keyspaces) = tokio::task::spawn_blocking(move || {
+            let keyspace_strategies = keyspaces.values().map(|ks| &ks.strategy);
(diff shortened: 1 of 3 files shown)


wprzytula:
v2: rebased on `main`.

scylla/src/transport/session.rs
```diff
@@ -865,27 +865,19 @@ impl Session {
...
-
-            first_ok = Some(res);
-
-            if is_ok {
-                break;
-            }
-        }
-
-        let mut prepared: PreparedStatement = first_ok.unwrap()?;
+        // Safety: there is at least one node in the cluster, so there will be at least
```
havaker, line 879: I would clarify that `self.cluster.get_working_connections()?` returns at least one connection or an error. None empty cluster is not a good guarantee.

scylla/src/transport/cluster.rs
```diff
@@ -481,6 +458,17 @@ impl ClusterData {
     pub fn replica_locator(&self) -> &ReplicaLocator {
         &self.locator
     }
+
+    /// Returns iterator of working connections to all shards
+    pub(crate) fn iter_working_connections(
```
havaker, line 463: `get_working_connections` returned an error when there were no working connections, by changing that some methods are going to panic.

havaker (changes requested):

scylla/src/transport/cluster.rs
```diff
@@ -345,21 +345,18 @@ impl ClusterData {
...
 
-        let keyspace_strategies: Vec<Strategy> = metadata
-            .keyspaces
-            .values()
-            .map(|ks| ks.strategy.clone())
-            .collect();
-
-        let locator = tokio::task::spawn_blocking(move || {
-            ReplicaLocator::new(ring.into_iter(), keyspace_strategies.iter())
+        let keyspaces = metadata.keyspaces;
```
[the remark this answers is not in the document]
wprzytula: You can try passing `metadata.keyspaces` directly and see what happens :stuck_out_tongue: 
Another field is moved from `metadata` above, so `metadata` not movable anymore as a whole.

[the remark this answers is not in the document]
havaker: I did, and it `cargo check` passed on that commit Oo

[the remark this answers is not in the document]
wprzytula: `cargo check` sometimes lags behind. Make sure to do `cargo clean -p scylla` before.

scylla/src/transport/cluster.rs
wprzytula: Do you mean the case when there are some nodes, but no working connections, or the case when there are no nodes?
In the first case, I can't see any possibility to trigger panic in both places that call this function.
I don't think that the second case is permitted by our invariant.

havaker: If there are some nodes, but no working connections `iter_working_connections` will return an empty iterator.
```rust
        // Safety: there is at least one node in the cluster, and `Cluster::get_working_connections()`
        // returns at least one connection or an error, so there will be at least one result.
        let first_ok: Result<PreparedStatement, QueryError> =
            results.by_ref().find_or_first(Result::is_ok).unwrap();
```
[ref](https://github.com/scylladb/scylla-rust-driver/pull/780/files#diff-80ba8e49418bed2ad375905c49b6784a23c454eac059d99c3e18679a8525d9feR882)

I believe that:
- the safety comment is wrong here (`get_working_connections` was swapped for `iter_working_connections`)
- assumption about `iter_working_connections` returning at least one element is incorrect

scylla/src/transport/cluster.rs
```diff
@@ -476,6 +458,21 @@ impl ClusterData {
...
+    pub(crate) fn iter_working_connections(
+        &self,
+    ) -> impl Iterator<Item = Result<Arc<Connection>, QueryError>> + '_ {
+        let peers = &self.known_peers;
+        peers
+            .values()
+            .map(|node| node.get_working_connections())
+            .flatten_ok()
+        // By an invariant `self.known_peers` is nonempty, so the returned iterator
+        // is nonempty, too.
```
havaker, line 474: Why is that? If all node pools are in `MaybePoolConnections::Broken` state, every call to `node.get_working_connections()` will fail.

scylla/src/transport/session.rs
```diff
@@ -861,31 +861,29 @@ impl Session {
...
+        });
+        let mut results = join_all(handles).await.into_iter();
+
+        // If at least one prepare was successful, `prepare()` returns Ok.
+        // Find first result that is Ok, or Err if all failed.
 
-        let mut prepared: PreparedStatement = first_ok.unwrap()?;
+        // Safety: there is at least one node in the cluster, and `Cluster::get_working_connections()`
+        // returns at least one connection or an error, so there will be at least one result.
+        let first_ok: Result<PreparedStatement, QueryError> =
```
havaker, line 881: `iter_working_connections` is being used here, there is no `get_working_connections` call.

scylla/src/transport/session.rs
```diff
@@ -1870,9 +1868,15 @@ impl Session {
...
-        let handles = connections.iter().map(|c| c.fetch_schema_version());
+        let handles = connections_iter.map(|c| async {
+            match c {
+                Ok(c) => c.fetch_schema_version().await,
+                Err(err) => Err(err),
+            }
+        });
         let versions = try_join_all(handles).await?;
 
         let local_version: Uuid = versions[0];
```
havaker, line 1882: `versions` might be empty, this line might panic.

havaker (changes requested):

scylla/src/transport/cluster.rs
wprzytula: > If there are some nodes, but no working connections iter_working_connections will return an empty iterator.

This is incorrect. `iter_working_connections()` returns a `Result` for every node. It is either `Err` or `Ok`, but the iterator will definitely not be empty.

scylla/src/transport/cluster.rs
wprzytula: That's true, and hence the resulting iterator will consist of only `Err` variants.
`flatten_ok()` does not remove any `Err` variants; what it does is flattening the `Ok` variants; that is, `Ok(Ok(x))` becomes `Ok(x)`.

scylla/src/transport/session.rs
wprzytula: I disagree. See the other comments.

scylla/src/transport/cluster.rs
havaker: Oops, you're right! I wrongly assumed what `flatten_ok` does in `iter_working_connections`.

scylla/src/transport/cluster.rs
```diff
@@ -481,6 +458,21 @@ impl ClusterData {
     pub fn replica_locator(&self) -> &ReplicaLocator {
         &self.locator
     }
+
+    /// Returns iterator of connections to all shards.
+    ///
+    /// By nonemptiness invariant of `self.known_peers`, the returned iterator is nonempty.
+    pub(crate) fn iter_working_connections(
```
havaker, line 465: I have a slight concern about the naming choice in this piece of code. It's not immediately obvious to me why each element in the iterator that produces working connections needs to be a `Result` rather than an `Arc<Connection>`.

havaker (approved):
Left a nit, LGTM

piodul (changes requested):
One request, LGTM otherwise.

scylla/src/transport/session.rs
```diff
@@ -1862,9 +1868,15 @@ impl Session {
...
+        let cluster_data = self.get_cluster_data();
+        let connections_iter = cluster_data.iter_working_connections();
 
-        let handles = connections.iter().map(|c| c.fetch_schema_version());
+        let handles = connections_iter.map(|c| async {
+            match c {
+                Ok(c) => c.fetch_schema_version().await,
+                Err(err) => Err(err),
+            }
+        });
```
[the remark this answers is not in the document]
wprzytula: You probably mean "instead of `match`". Then I don't think so, because the operation is done inside an `async` block, which is a known limitation of functional idioms in Rust.

wprzytula:
Reworked `iter_working_connections()` so that now it returns either an error or an iterator with successful working connections (with at least one of them).

scylla/src/transport/cluster.rs
wprzytula: Solved.

piodul (approved):

PR #785, opened 2023-08-03
doc: document releasing Rust Driver documentation

annastuchlik:
This commit updates the release procedure to include a step to publish the docs for the new driver version.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,6 +13,10 @@ The following list includes information how to release version vX.Y.Z of scylla-
 6. Go to https://github.com/scylladb/scylla-rust-driver/releases , click the `Draft new release` button and follow the procedure to create a new release on GitHub. Use the release notes as its description.
 7. Publish the crate(s) on crates.io. For `scylla` crate, go directly to the crate directory, i.e. `/your/repo/path/scylla-rust-driver/scylla`, and run `cargo publish`. It will ask you for an access token, follow the instructions if you haven't set one up.
 8. Send the release notes to scylladb-users (https://groups.google.com/g/scylladb-users) as well as Cassandra users (https://lists.apache.org/list.html?user@cassandra.apache.org) mailing lists.
+9. Publish the documentation for the new version. In `docs/source/conf.py`,
+update the following:
+- TAGS = [' '] - add the new release tag to the list
+- LATEST_VERSION = ' ' - replace the previous tag with the new release tag
 
 You're done!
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -13,6 +13,10 @@ The following list includes information how to release version vX.Y.Z of scylla-
 6. Go to https://github.com/scylladb/scylla-rust-driver/releases , click the `Draft new release` button and follow the procedure to create a new release on GitHub. Use the release notes as its description.
 7. Publish the crate(s) on crates.io. For `scylla` crate, go directly to the crate directory, i.e. `/your/repo/path/scylla-rust-driver/scylla`, and run `cargo publish`. It will ask you for an access token, follow the instructions if you haven't set one up.
 8. Send the release notes to scylladb-users (https://groups.google.com/g/scylladb-users) as well as Cassandra users (https://lists.apache.org/list.html?user@cassandra.apache.org) mailing lists.
+9. Publish the documentation for the new version. In `docs/source/conf.py`,
+update the following:
+- TAGS = [' '] - add the new release tag to the list
+- LATEST_VERSION = ' ' - replace the previous tag with the new release tag
```
avelanarius, line 19:
```suggestion
- `TAGS = [' ']` - add the new release tag to the list
- `LATEST_VERSION = ' '` - replace the previous tag with the new release tag
```

wprzytula (approved):

RELEASE.md
```diff
@@ -13,6 +13,10 @@ The following list includes information how to release version vX.Y.Z of scylla-
 6. Go to https://github.com/scylladb/scylla-rust-driver/releases , click the `Draft new release` button and follow the procedure to create a new release on GitHub. Use the release notes as its description.
 7. Publish the crate(s) on crates.io. For `scylla` crate, go directly to the crate directory, i.e. `/your/repo/path/scylla-rust-driver/scylla`, and run `cargo publish`. It will ask you for an access token, follow the instructions if you haven't set one up.
 8. Send the release notes to scylladb-users (https://groups.google.com/g/scylladb-users) as well as Cassandra users (https://lists.apache.org/list.html?user@cassandra.apache.org) mailing lists.
+9. Publish the documentation for the new version. In `docs/source/conf.py`,
+update the following:
+- TAGS = [' '] - add the new release tag to the list
```
piodul: Which versions should be _kept_ in `TAGS`? Are we supposed to remove old releases from this list at some point (when we stop supporting them), or should we keep adding them without removing? At the moment, I can only see that this list only has the `'v0.9.0'` (the most recent release).

annastuchlik: @piodul 
- I've added `v0.9.0` to make sure we at least publish the docs for the latest version (previously, we only had the docs for branch `main`).  You can add earlier versions to the list. 
- If we stop supporting a version, it shouldn't be removed from `TAGS = [' ']`, but it should be added to the list of deprecated versions in the same section of the conf.py file: `DEPRECATED_VERSIONS = []`

See the [Sphinx theme documentation](https://sphinx-theme.scylladb.com/stable/configuration/multiversion.html) for details.

piodul: Thanks for the answer and for the link.

piodul (approved):

PR #797, opened 2023-08-24
proxy: add an example configurable from cmdline

wprzytula:
The example is useful for manual testing of drivers with regard to address translation of DNS related functionalities, such as hostname mapping alterations combined with address changes.

## Pre-review checklist

### DIFF
--- a/scylla-proxy/examples/cmdline.rs
+++ b/scylla-proxy/examples/cmdline.rs
@@ -1,0 +1,60 @@
+//! Command line-configurable transparent proxy.
+//!
+//! Useful for testing e.g. address translation or DNS related functionalities,
+//! such as alterations of hostname mapping combined with IP changes.
+//!
+//! How to use:
+//! cargo run --example cmdline -- [node_addr] [proxy_addr]
+//! This starts a proxy that forwards all traffic between a driver and [proxy_addr] to [node_addr].
+
+use std::{
+    net::{Ipv4Addr, SocketAddr},
+    str::FromStr,
+};
+
+use scylla_proxy::{Node, Proxy, ShardAwareness};
+use tracing::instrument::WithSubscriber;
+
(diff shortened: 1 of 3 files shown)


cmdline.rs
```diff
@@ -0,0 +1,69 @@
...
+    );
+    let node_proxy_addr = SocketAddr::new(
+        std::net::IpAddr::V4(Ipv4Addr::new(
+            127,
+            0,
+            0,
+            proxy_octet.as_ref().map(|s| s.parse().unwrap()).unwrap_or(66),
+        )),
+        9042,
+    );
```
piodul, line 55: I think it is a bit confusing that the `[node_addr]` and `[proxy_addr]` arguments are actually only the last octet of the IP address. I think we should try to parse `[node_addr]` and `[proxy_addr]` as full addresses.

cmdline.rs
```diff
@@ -0,0 +1,69 @@
...
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .without_time()
+        .init();
+}
+
+fn pause() {
+    println!("Press Enter to stop proxy...");
+    std::io::stdin().read_exact(&mut [0]).unwrap();
+    println!();
+}
```
piodul, line 29: Ouch, you are using a blocking API here. The `pause()` invocation in `main` will block one of the executor threads.

How about waiting for a Ctrl+C signal instead? AFAIR tokio already has some utilities for that which allow to wait for this signal asynchronously.

piodul (changes requested):

wprzytula:
v2:
- fixed waiting from blocking to nonblocking; the same fix is applied to examples already present in the codebase. 
- now the whole address is passed from cmdline (not merely the last octet).

piodul (approved):

PR #804, opened 2023-08-31
codewide: re-structure modules

wprzytula:
According to the new proposed design, the structure of modules and imports is changed to provide more rational code separation.
- `scylla-cql` is no longer exposed; only some entities are selectively re-exported using new `cql` module;
- `transport` module is decomposed and a number of new modules emerge instead.

## Pre-review checklist


Lorak-mmk:
v2: https://github.com/scylladb/scylla-rust-driver/pull/817

Issue #800, opened 2023-08-25
Serialization refactor: obtain information about column types when doing `Session::execute`


piodul:
Sub-task of #463.

Information about the types of the bind markers is necessary to make the serialization API safe (i.e. so that the user data won't be misinterpreted when sending it to the database). The problem with unprepared statements is that the driver doesn't have a good way to determine column names and types of the bind markers - the `QUERY` request only contains the query string and serialized, untypes values for the bind markers. There is no good way to obtain the types for the bind markers or let the database validate the types when it receives the request.

In the case of prepared statements, information about the bind marker types is available because it is computed by the DB and returned in the response to the prepare request.

I see two ways out:

1. Before sending an unprepared statement, quietly prepare it first to obtain the information about the bind markers (it's only necessary to prepare on one node). This increases latency of unprepared queries twofold, but using unprepared queries is a performance anti-pattern anyway. We could provide an optimization: if the list of values to be bound is empty, we just send it as an unprepared query. This API will also introduce less breakage in the user code.
2. Remove the `values` argument from the `Session::query` and only allow for queries without bind markers. This is simpler to implement than 1. If the user wanted to send an unprepared query with some bind markers, now they will have to prepare it beforehand - latency-wise it will be the same as for 1., though 1. does not have to prepare the statement on every node which would be a win if somebody wants to execute the query only once.

The task doesn't make the API safer by itself, but it is a necessary element of #463 that can be done in parallel from the others sub-tasks.

piodul:
IMO the first option looks better to me at the moment. The empty list optimization can be done as a follow-up.

PR #806, opened 2023-09-01
Quietly prepare unprepared query

sylwiaszunejko:
## Motivation
See the referenced issue (#800)

## What's done
Information about the types of the bind markers is necessary to make the serialization API safe (i.e. so that the user data won't be misinterpreted when sending it to the database). The problem with unprepared statements is that the driver doesn't have a good way to determine column names and types of the bind markers.

Now before sending an unprepared statement, it is quietly prepared on one connection to obtain the information about the bind markers. If the list of values to be bound is empty, we just send it as an unprepared query.

## Evaluation
Integration tests are added.

Refs: #800 

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -657,6 +657,20 @@ impl Connection {
     ) -> Result<QueryResponse, QueryError> {
         let serialized_values = values.serialized()?;
 
+        let values_size = serialized_values.size();
+        if values_size != 0 {
+            let prepared = self.prepare(query).await?;
+            return self
+                .execute_with_consistency(
+                    &prepared,
+                    values,
+                    consistency,
+                    serial_consistency,
+                    paging_state,
+                )
+                .await;
+        }
+
         let query_frame = query::Query {
             contents: Cow::Borrowed(&query.contents),
             parameters: query::QueryParameters {
--- a/scylla/tests/integration/main.rs
+++ b/scylla/tests/integration/main.rs
@@ -5,4 +5,5 @@ mod lwt_optimisation;
 mod new_session;
 mod retries;
 mod shards;
+mod silent_prepare_query;
 pub(crate) mod utils;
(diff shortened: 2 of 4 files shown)


scylla/src/transport/session.rs
```diff
@@ -650,15 +650,34 @@ impl Session {
...
                     let serial_consistency = query
                         .config
                         .serial_consistency
-                        .unwrap_or(execution_profile.serial_consistency);
+                        .unwrap_or(default_consistency);
                     // Needed to avoid moving query and values into async move block
                     let query_ref = &query;
                     let values_ref = &serialized_values;
                     let paging_state_ref = &paging_state;
+                    let values_size = values.serialized().unwrap().into_owned().size();
```
piodul, line 662: `.serialized()` is an operation that can be potentially costly. It is already computed on line 624, so I suggest using the already computed `serialized_values` instead.

scylla/src/transport/session.rs
```diff
@@ -650,15 +650,34 @@ impl Session {
...
                     async move {
+                        if values_size != 0 {
+                            let prepared = connection.prepare(query_ref).await?;
+                            let serial_consistency = prepared
+                                .config
+                                .serial_consistency
+                                .unwrap_or(default_consistency);
```
piodul, line 669: The prepared statement will have exactly the same config as the original query object, so there is no need to calculate serial consistency again.

piodul (changes requested):

piodul (commented):
Apart from moving the preparation into `Connection::query`, we will also have to handle batches... Batches can consist of multiple statements, both prepared and unprepared. I guess we have to update `Connection::batch` as well to handle unprepared statements with arguments. We can start with preparing them sequentially in that function, or with some low (fixed?) concurrency.

scylla/src/transport/session.rs
```diff
@@ -658,7 +658,21 @@ impl Session {
...
+                                .execute_with_consistency(
+                                    &prepared,
+                                    values_ref,
+                                    consistency,
+                                    serial_consistency,
+                                    paging_state_ref.clone(),
+                                )
+                                .await
+                                .and_then(QueryResponse::into_non_error_query_response);
+                        }
```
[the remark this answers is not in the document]
sylwiaszunejko: So maybe we should call `prepare` in `Connection::query_with_consistency`, because both `Session::query` and `Connection::query` use that.

piodul (approved):

PR #807, opened 2023-09-07
Bump verrsion of chrono dependency

Lorak-mmk:
In scylla-cql/src/frame/response/cql_to_rust.rs we use `DateTime::<Utc>::from_naive_utc_and_offset` but it was introduced in chrono 0.4.27. This PR bumps this dependency to correct version.

## Pre-review checklist

### DIFF
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -21,7 +21,7 @@ uuid = "1.0"
 thiserror = "1.0"
 bigdecimal = "0.2.0"
 num-bigint = "0.3"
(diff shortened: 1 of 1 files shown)


piodul (approved):
LGTM

Apparently, there is a flag in `cargo` that allows to resolve minimal versions of the dependencies (https://doc.rust-lang.org/cargo/reference/unstable.html?highlight=minimal-versions#minimal-versions) but, due to the fact that many crates in the ecosystem don't specify correct minimal versions, it's not available on stable. Maybe if the situation improves we can consider this as a CI check in the future.

PR #808, opened 2023-09-09
Added custom hashers support for HashSet and HashMap.

s3rius:
Hello. The library is super great, but I noticed one problem. Custom hashers are not supported. 

I introduced some new generics that allow you to serialize HashMaps and HashSets with custom hashers.

I wanted to speedup my code and have chosen [ahash](https://docs.rs/ahash/latest/ahash/) as custom hasher for my hashmaps. But found out that I need to convert hashmap back to default hasher in order to serialize it to valueslist. 

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -8,6 +8,7 @@ use num_bigint::BigInt;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::convert::TryInto;
+use std::hash::BuildHasher;
 use std::net::IpAddr;
 use thiserror::Error;
 use uuid::Uuid;
@@ -662,13 +663,13 @@ fn serialize_list_or_set<'a, V: 'a + Value>(
     Ok(())
 }
 
-impl<V: Value> Value for HashSet<V> {
+impl<V: Value, S: BuildHasher + Default> Value for HashSet<V, S> {
     fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
         serialize_list_or_set(self.iter(), self.len(), buf)
     }
 }
 
-impl<K: Value, V: Value> Value for HashMap<K, V> {
+impl<K: Value, V: Value, S: BuildHasher> Value for HashMap<K, V, S> {
     fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
         serialize_map(self.iter(), self.len(), buf)
     }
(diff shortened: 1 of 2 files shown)


scylla-cql/src/frame/value.rs
```diff
@@ -866,10 +867,26 @@ macro_rules! impl_value_list_for_map {
...
 }
 
-impl_value_list_for_map!(HashMap, String);
-impl_value_list_for_map!(HashMap, &str);
-impl_value_list_for_map!(BTreeMap, String);
-impl_value_list_for_map!(BTreeMap, &str);
+// Implement ValueList for maps, which serializes named values
+macro_rules! impl_value_list_for_hash_map {
+    ($map_type:ident, $key_type:ty) => {
+        impl<T: Value, S: BuildHasher> ValueList for $map_type<$key_type, T, S> {
```
piodul, line 873: Nit: the `map_type` argument is now not needed, right? It should always be `HashMap`.

scylla-cql/src/frame/value.rs
```diff
@@ -851,7 +852,7 @@ impl<T: Value> ValueList for Vec<T> {
 }
 
 // Implement ValueList for maps, which serializes named values
-macro_rules! impl_value_list_for_map {
+macro_rules! impl_value_list_for_btree_map {
```
piodul, line 855: Similar issue here, it only accepts `BTreeMap` now.

piodul:
The `min_rust` check failed due to one of our indirect dependencies bumping up its MSRV. We'll fix that soon.

piodul (changes requested):
Thanks for the contribution.

It looks like you also need to modify the `impl` on line `cql_to_rust.rs:203` so that deserialization of hashmaps with custom hashers is also possible.

Apart from that, two more nitpicks but LGTM otherwise.

piodul:
The issue with failing `min_rust` check was fixed in https://github.com/scylladb/scylla-rust-driver/pull/810. Rebase on main will be needed.

scylla-cql/src/frame/value.rs
s3rius: Agree.

s3rius: Updated.

piodul:
The test `if_lwt_optimisation_mark_offered_then_negotiatied_and_lwt_routed_optimally` failed twice and only succeeded now in the third run. It's not related to the PR at all so I'll merge, but I'll open an issue about this test being flaky.

s3rius:
Thank you very much. Have a nice day.

PR #809, opened 2023-09-10
Improve calculation for `SerializedValues::with_capacity`

cstyles:
It seems that currently the argument provided to this function is being treated as the _number_ of items that we expect to serialize. Normally that would make sense because the argument is passed along to `Vec::with_capacity` which provisions space for the inner `serialized_values` field. However, that field is a `Vec<u8>` so really the argument corresponds to the total _size_ of the serialized values in bytes.

This commit tries to be smarter about reserving space by taking into account the actual size in memory of the values to be serialized. It's not perfect (e.g., `String`s will always reserve the same amount of space, regardless of how long their actual contents are) but hopefully it will still result in fewer re-allocations.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -830,7 +830,8 @@ impl ValueList for [u8; 0] {
 // Implement ValueList for slices of Value types
 impl<T: Value> ValueList for &[T] {
     fn serialized(&self) -> SerializedResult<'_> {
-        let mut result = SerializedValues::with_capacity(self.len());
+        let size = std::mem::size_of_val(*self);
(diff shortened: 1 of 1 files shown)


piodul (approved):
LGTM. Thanks for the contribution!

In the next release, we are planning to introduce new serialization traits (https://github.com/scylladb/scylla-rust-driver/issues/463) which should make it possible to calculate required size exactly - however, we will keep the old traits in parallel for some time to make it possible to gradually transition to the new traits, so your contribution will stil be useful.

Please rebase on the newest `main` branch, the issue with failing `min_rust` check is fixed there.

piodul (approved):

PR #810, opened 2023-09-11
MSRV: bump to 1.66.0

piodul:
One of our indirect dependencies (`toml_edit`) has updated its MSRV. As we do not keep Cargo.lock committed in the repo and cargo does not honor MSRV during crate resolution, we can just bump our current MSRV by one version forward, from 1.65.0 to 1.66.0.

The version 1.66.0 is over 8 months old, so it's pretty old by the MSRV-breaking standards of the community.

As a bonus, we get rid of `unzip_option` workaround and use the real thing, i.e. `Option::unzip()`.

## Pre-review checklist


muzarski (approved):
LGTM

PR #812, opened 2023-09-18
Quietly prepare unprepared queries in batches

sylwiaszunejko:
## Motivation
See the referenced issue (#800)

## What's done
Information about the types of the bind markers is necessary to make the serialization API safe (i.e. so that the user data won't be misinterpreted when sending it to the database). The problem with unprepared statements in batch is that the driver doesn't have a good way to determine column names and types of the bind markers.

Now before sending a batch, every unprepared statement is quietly prepared on one connection to obtain the information about the bind markers. If the list of values to be bound to the statement is empty, we just send it as an unprepared query.

Fixes: #800

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -26,7 +26,7 @@ pub(crate) use ssl_config::SslConfig;
 
 use crate::authentication::AuthenticatorProvider;
 use scylla_cql::frame::response::authenticate::Authenticate;
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeSet, HashMap, HashSet};
 use std::convert::TryFrom;
 use std::io::ErrorKind;
 use std::net::{IpAddr, SocketAddr};
@@ -52,7 +52,7 @@ use crate::frame::{
     request::{self, batch, execute, query, register, SerializableRequest},
     response::{event::Event, result, NonErrorResponse, Response, ResponseOpcode},
     server_event_type::EventType,
-    value::{BatchValues, ValueList},
+    value::{BatchValues, BatchValuesIterator, ValueList},
     FrameParams, SerializedRequest,
 };
 use crate::query::Query;
@@ -772,11 +772,13 @@ impl Connection {
 
     pub(crate) async fn batch_with_consistency(
         &self,
-        batch: &Batch,
+        init_batch: &Batch,
         values: impl BatchValues,
         consistency: Consistency,
         serial_consistency: Option<SerialConsistency>,
     ) -> Result<QueryResult, QueryError> {
+        let batch = self.prepare_batch(init_batch, &values).await?;
+
         let batch_frame = batch::Batch {
             statements: Cow::Borrowed(&batch.statements),
             values,
@@ -818,6 +820,58 @@ impl Connection {
                 )),
             };
         }
+    }
+
+    async fn prepare_batch<'b>(
+        &self,
+        init_batch: &'b Batch,
+        values: impl BatchValues,
+    ) -> Result<Cow<'b, Batch>, QueryError> {
+        let mut to_prepare = HashSet::<&str>::new();
+
+        {
+            let mut values_iter = values.batch_values_iter();
+            for stmt in &init_batch.statements {
+                if let BatchStatement::Query(query) = stmt {
+                    let value = values_iter.next_serialized().transpose()?;
+                    if let Some(v) = value {
+                        if v.len() > 0 {
+                            to_prepare.insert(&query.contents);
+                        }
+                    }
+                } else {
+                    values_iter.skip_next();
+                }
+            }
+        }
+
+        if to_prepare.is_empty() {
+            return Ok(Cow::Borrowed(init_batch));
+        }
+
+        let mut prepared_queries = HashMap::<&str, PreparedStatement>::new();
+
+        for query in &to_prepare {
+            let prepared = self.prepare(&Query::new(query.to_string())).await?;
+            prepared_queries.insert(query, prepared);
+        }
+
+        let mut batch: Cow<Batch> = Cow::Owned(Default::default());
+        batch.to_mut().config = init_batch.config.clone();
+        for stmt in &init_batch.statements {
+            match stmt {
+                BatchStatement::Query(query) => match prepared_queries.get(query.contents.as_str())
+                {
+                    Some(prepared) => batch.to_mut().append_statement(prepared.clone()),
+                    None => batch.to_mut().append_statement(query.clone()),
+                },
+                BatchStatement::PreparedStatement(prepared) => {
+                    batch.to_mut().append_statement(prepared.clone());
+                }
+            }
+        }
+
+        Ok(batch)
     }
 
     pub(crate) async fn use_keyspace(
--- a/scylla/src/transport/mod.rs
+++ b/scylla/src/transport/mod.rs
@@ -28,6 +28,8 @@ mod authenticate_test;
 mod cql_collections_test;
 #[cfg(test)]
 mod session_test;
+#[cfg(test)]
+mod silent_prepare_batch_test;
 
 #[cfg(test)]
 mod cql_types_test;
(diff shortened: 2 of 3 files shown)


piodul:
Please change the PR title to mention that it's about unprepared queries in batches. Currently, it's too similar to the previous PR's title IMO (https://github.com/scylladb/scylla-rust-driver/pull/806).

scylla/src/transport/connection.rs
```diff
@@ -772,11 +772,50 @@ impl Connection {
...
+        let batch = {
+            let mut batch: Batch = Default::default();
+            batch.config = init_batch.config.clone();
+
+            let mut to_prepare = HashSet::<usize>::new();
+
+            {
+                let mut value_iter = values.batch_values_iter();
+                for (id, stmt) in init_batch.statements.iter().enumerate() {
+                    let value = value_iter.next_serialized().transpose()?;
```
piodul, line 789: Serializing values for a query can be relatively expensive, so it would be best to avoid it. Let's use `skip_next` instead of `next_serialized` in case of `BatchStatement::PreparedStatement`.

piodul (changes requested):

scylla/src/transport/connection.rs
sylwiaszunejko: Done

scylla/src/transport/connection.rs
```diff
@@ -772,11 +772,50 @@ impl Connection {
...
+        init_batch: &Batch,
         values: impl BatchValues,
         consistency: Consistency,
         serial_consistency: Option<SerialConsistency>,
     ) -> Result<QueryResult, QueryError> {
+        let batch = {
+            let mut batch: Batch = Default::default();
+            batch.config = init_batch.config.clone();
+
+            let mut to_prepare = HashSet::<usize>::new();
```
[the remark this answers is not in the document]
piodul: I think that the code looks more complicated than before, actually. The `vec_id` variable and comparing indexes looks tricky...

> I replaced `HashSet` with `Vec`. I tried to store query strings, but I would still need to know the indexes to maintain the right order of the statements in batch

You can first create a hashset with all needed strings. Then, you can prepare the statements and put them in a hashmap, indexed by statement string. Finally, when you re-create the batch, when you iterate and encounter an unprepared statement/query in the original batch, look up the prepared statement in the hashmap.

scylla/src/transport/connection.rs
```diff
@@ -772,11 +772,59 @@ impl Connection {
...
+                        }
+                        BatchStatement::PreparedStatement(prepared) => {
+                            batch.append_statement(prepared.clone());
+                        }
+                    }
+                }
+            }
+
+            batch
+        };
```
piodul, line 826: Finally, let's see if we can extract all of the added logic here to a separate method/function. It's getting quite long.

scylla/src/transport/connection.rs
```diff
@@ -772,11 +772,59 @@ impl Connection {
...
+                    } else {
+                        value_iter.skip_next();
+                    }
+                }
+            }
+
+            let mut batch: Batch = Default::default();
+
+            if to_prepare.is_empty() {
+                batch = init_batch.clone();
```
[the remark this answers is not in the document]
sylwiaszunejko: Done

scylla/src/transport/connection.rs
piodul: I don't think you need to copy or clone `values`, you only need to pass it by reference. As I can see, it's only needed for `.batch_values_iter()` which takes `values` by reference - so you can pass `values` by reference to that function.

I imagine that the final method will look like this (I hope I didn't mess up lifetimes):

```rust
async fn prepare_batch<'b>(
    &self,
    batch: &'b Batch,
    values_iter: BatchValuesIterator<'_>,
) -> Cow<'b, Batch> {
    // ...
}
```

sylwiaszunejko: I have a few errors and I have to change it a little bit to:
```rust
 async fn prepare_batch<'b>(
        &self,
        init_batch: &'b Batch,
        mut values_iter: impl BatchValuesIterator<'_>,
    ) -> Result<Cow<'b, Batch>, QueryError> {
    //...
}
```
but it still does not build, there is a problem with lifetimes in `session_test.rs:408`  (implementation of `std::marker::Send` is not general enough
    |
    = note: `std::marker::Send` would have to be implemented for the type `&'0 str`, for any lifetime `'0`...
    = note: ...but `std::marker::Send` is actually implemented for the type `&'1 str`, for some specific lifetime `'1`).
    I am not sure how to fix it

piodul: I see. Please try passing `&values` instead. For some reason, the `BatchValuesIterator` cannot be held across await points, so I suppose there might be issues when trying to pass it to async functions.

scylla/src/transport/connection.rs
```diff
@@ -820,6 +822,59 @@ impl Connection {
         }
     }
 
+    async fn prepare_batch<'b>(
+        &self,
+        init_batch: &'b Batch,
+        values: impl BatchValues,
+    ) -> Result<Cow<'b, Batch>, QueryError> {
+        let mut to_prepare = HashSet::<String>::new();
```
piodul, line 830: Nit: you can probably use `&str` as the key type. The strings will be borrowed from the `init_batch` object which is alive for the whole duration of this function and we don't modify it, so lifetimes should check out, This will save you an allocation when inserting to the hashset.

scylla/src/transport/connection.rs
```diff
@@ -820,6 +822,59 @@ impl Connection {
...
+                    BatchStatement::Query(query) => match prepared_queries.get(&query.contents) {
+                        Some(prepared) => batch.to_mut().append_statement(prepared.clone()),
+                        None => batch.to_mut().append_statement(query.clone()),
+                    },
+                    BatchStatement::PreparedStatement(prepared) => {
+                        batch.to_mut().append_statement(prepared.clone());
+                    }
+                }
+            }
+        }
```
piodul, line 873: In Rust, `if`s are expressions and evaluate to a value - you can use that to assign to `batch` immediately.

Moreover, in the `else` part, I suggest creating a value of type `Batch`, initializing it and then assigning it to `Cow<Batch>`. This will allow to avoid calling `batch.to_mut()` in multiple places.

```rust
let batch = if to_prepare.is_empty() {
    Cow::Borrowed(init_batch)
} else {
    let mut batch = Batch::default();
    // ...
    Cow::Owned(batch)
}
```

scylla/src/transport/connection.rs
```diff
@@ -820,6 +822,59 @@ impl Connection {
...
+                            to_prepare.insert(query.contents.clone());
+                        }
+                    }
+                } else {
+                    values_iter.skip_next();
+                }
+            }
+        }
+
+        let mut prepared_queries = HashMap::<String, PreparedStatement>::new();
```
piodul, line 848: This one can probably take a `&str`, too.

scylla/src/transport/connection.rs
```diff
@@ -820,6 +822,59 @@ impl Connection {
...
+                        if v.len() > 0 {
+                            to_prepare.insert(query.contents.clone());
+                        }
+                    }
+                } else {
+                    values_iter.skip_next();
+                }
+            }
+        }
+
```
piodul, line 847: How about an early return here?

```rust
if to_prepare.is_empty() {
    return Cow::Borrowed(init_batch);
}
```

By doing that we can skip creating `prepared_queries` and the loop that follows after it altogether on the fast path (i.e. when all statements are prepared).

scylla/src/transport/connection.rs
piodul: Though, if you implement an early return like I suggested before, here you will be able to drop the `if` altogether and only leave the code from the `else` branch.

piodul (changes requested):
Left some final nitpicks with regards to the implementation. What about tests?

I'd like to see some simple test that sends a batch, queries the table and sees whether the batch modified the table as expected. No need to involve the proxy. I'd like to see several test cases that exercise different paths in the `prepare_batch` code:

- Batch only contains prepared statements,
- Batch only contains unprepared statements (queries) without values,
- Batch only contains unprepared statements with values,
- Batch that contains a combination of the above

piodul (approved):

PR #816, opened 2023-09-21
Add additional tracing fields specific to Scylla

karlpvoss:
Introduction of some new struct fields that express the additional fields that are present in the
Scylla schemas for `system_traces.sessions` and `system_traces.events`.

Since I was unable to write a `FromRow` implementation that I was happy with in any form, the
provided implementation will perform conversion for only the fields present in both databases.
This therefore also stands for `Session::get_tracing_info`, where the documentation has been
updated.

## Pre-review checklist

    If there's any feedback about what kind of tests should be added for this type of change let me know and I'll do them.

### DIFF
--- a/docs/source/tracing/tracing.md
+++ b/docs/source/tracing/tracing.md
@@ -17,7 +17,8 @@ Queries that support tracing:
 * [`Session::prepare()`](prepare.md)
 
 After obtaining the tracing id you can use `Session::get_tracing_info()` to query tracing information.\
-`TracingInfo` contains values that are the same in Scylla and Cassandra®, skipping any database-specific ones.\
+`Session::get_tracing_info()` returns a `TracingInfo` which only contains values that are the same
+in both Scylla and Cassandra®, skipping any database-specific ones.\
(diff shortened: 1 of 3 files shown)


piodul:
Hi @karlpvoss, could you tell me about the use case for this PR? As you wrote, this PR adds fields to the tracing-related structs, but never initializes them. Was that the intention, or do you need tips on how to fill these structs in case when connected to Scylla? If it's the former, then I'd rather not add fields that are unused - it's slightly confusing as users who won't carefully read the docstring for `get_tracing_info` or the docs.

Implementing it right now might be a bit convoluted, but doable, as you would need to first detect whether you are connected to Scylla or Cassandra and then use a query that fetches the columns supported by either Scylla or Cassandra. With better deserialization traits (https://github.com/scylladb/scylla-rust-driver/issues/462), you could `SELECT *` from the tables and then fill in the struct in the deserialization trait as you would get information about the names of the columns, though it's difficult for me to tell when the new deserialization traits will be implemented and merged.

wprzytula:
@karlpvoss what's the status of this?

PR #818, opened 2023-09-28
scylla-macros: report errors as compile errors, not panics

piodul:
Our current procedural macros handle errors by panicking. This leads to compilation errors which inform the user about that fact first, and then explain the real issue in the "help" part:

```
error: proc-macro derive panicked
 --> examples/user-defined-type.rs:6:10
  |
6 | #[derive(FromUserType)]
  |          ^^^^^^^^^^^^
  |
  = help: message: derive(FromUserType) works only on structs!
```

This commit gets rid of `.expect()` and `panic()` calls. Instead, errors are now propagated via `Result<T, syn::Error>` and properly converted to a compilation error at the end. This approach results in nicer error messages and allows to attach the message to a particular part of the original code that could cause the problem.

After the changes, error messages look like this:

```
error: derive(FromUserType) works only for structs with named fields
 --> examples/user-defined-type.rs:7:1
  |
7 | enum Foo {
  | ^^^^
```

## Pre-review checklist

### DIFF
--- a/scylla-macros/src/from_row.rs
+++ b/scylla-macros/src/from_row.rs
@@ -3,10 +3,10 @@ use quote::{quote, quote_spanned};
 use syn::{spanned::Spanned, DeriveInput};
 
 /// #[derive(FromRow)] derives FromRow for struct
-pub fn from_row_derive(tokens_input: TokenStream) -> TokenStream {
-    let item = syn::parse::<DeriveInput>(tokens_input).expect("No DeriveInput");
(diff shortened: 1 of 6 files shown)


piodul:
@Lorak-mmk @avelanarius review ping

Lorak-mmk (approved):
Just one question: why is `.into()` in `e.into_compile_error().into()` required? Doesn't `.into_compile_error()` already return `TokenStream`?

PR #819, opened 2023-09-28
Add serialization traits

Lorak-mmk:
This commit adds experimental version of new serialization traits. Blanket implementation are provided to support older traits.
For now there are no comments / other implementations with real type checking as the purpose of this PR is to unblock the work on macros
and other parts of refactor.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types/serialize/mod.rs
+++ b/scylla-cql/src/types/serialize/mod.rs
@@ -1,0 +1,6 @@
+use std::{any::Any, sync::Arc};
+
+pub mod row;
+pub mod value;
+
+type SerializationError = Arc<dyn Any + Send + Sync>;
--- a/scylla-cql/src/types/serialize/row.rs
+++ b/scylla-cql/src/types/serialize/row.rs
@@ -1,0 +1,49 @@
+use std::sync::Arc;
+
+use crate::frame::response::result::ColumnSpec;
+use crate::frame::value::ValueList;
+
+use super::SerializationError;
+
+pub struct RowSerializationContext<'a> {
+    columns: &'a [ColumnSpec],
+}
+
+impl<'a> RowSerializationContext<'a> {
+    #[inline]
+    pub fn columns(&self) -> &'a [ColumnSpec] {
+        self.columns
+    }
+
+    // TODO: change RowSerializationContext to make this faster
+    #[inline]
+    pub fn column_by_name(&self, target: &str) -> Option<&ColumnSpec> {
+        self.columns.iter().find(|&c| c.name == target)
+    }
+}
+
+pub trait SerializeRow {
+    fn preliminary_type_check(ctx: &RowSerializationContext<'_>) -> Result<(), SerializationError>;
+    fn serialize(
+        &self,
+        ctx: &RowSerializationContext<'_>,
+        out: &mut Vec<u8>,
+    ) -> Result<(), SerializationError>;
+}
+
+impl<T: ValueList> SerializeRow for T {
+    fn preliminary_type_check(
+        _ctx: &RowSerializationContext<'_>,
+    ) -> Result<(), SerializationError> {
+        Ok(())
+    }
+
+    fn serialize(
+        &self,
+        _ctx: &RowSerializationContext<'_>,
+        out: &mut Vec<u8>,
+    ) -> Result<(), SerializationError> {
+        self.write_to_request(out)
+            .map_err(|err| Arc::new(err) as SerializationError)
+    }
+}
--- a/scylla-cql/src/lib.rs
+++ b/scylla-cql/src/lib.rs
@@ -3,6 +3,8 @@ pub mod frame;
 #[macro_use]
 pub mod macros;
 
+pub mod types;
+
 pub use crate::frame::response::cql_to_rust;
 pub use crate::frame::response::cql_to_rust::FromRow;
 
--- a/scylla-cql/src/types/mod.rs
+++ b/scylla-cql/src/types/mod.rs
@@ -1,0 +1,1 @@
+pub mod serialize;
--- a/scylla-cql/src/types/serialize/value.rs
+++ b/scylla-cql/src/types/serialize/value.rs
@@ -1,0 +1,22 @@
+use std::sync::Arc;
+
+use crate::frame::response::result::ColumnType;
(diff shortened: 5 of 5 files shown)


scylla-cql/src/types/serialize/serialize_row.rs
```diff
@@ -0,0 +1,44 @@
...
+impl<T: ValueList> SerializeRow for T {
+    fn preliminary_type_check(&self, _ctx: &RowSerializationContext<'_>) -> bool {
+        true
+    }
+
+    fn serialize(
+        &self,
+        _ctx: &RowSerializationContext<'_>,
+        out: &mut Vec<u8>,
+    ) -> Result<(), Arc<dyn Any + Send + Sync>> {
```
piodul, line 38: Please introduce an alias for `Arc<dyn Any + Send + Sync>`, e.g. `SerializationError`.

scylla-cql/src/types/serialize/serialize_value.rs
```diff
@@ -0,0 +1,29 @@
+use std::{any::Any, sync::Arc};
+
+use crate::{_macro_internal::Value, frame::response::result::ColumnType};
+
+pub trait SerializeCql {
+    fn preliminary_type_check(&self, typ: &ColumnType) -> bool;
```
piodul, line 6: I think the preliminary type check should return `Result<(), SerializationError>` so that it can return some information about the error.

scylla-cql/src/types/serialize/serialize_row.rs
```diff
@@ -0,0 +1,44 @@
...
+        self.write_to_request(out).map_err(|err| {
+            let err: Arc<dyn Any + Send + Sync> = Arc::new(err);
+            err
```
piodul, line 41: Nit: this can be written as `Arc::new(err) as Arc<dyn Any + Send + Sync>`, or `Arc::new(err) as SerializationError`

scylla-cql/src/types/serialize/serialize_value.rs
```diff
@@ -0,0 +1,29 @@
+use std::{any::Any, sync::Arc};
+
+use crate::{_macro_internal::Value, frame::response::result::ColumnType};
```
piodul, line 3: Please avoid using the `_macro_internal` paths.

scylla-cql/src/types/serialize/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod serialize_row;
+pub mod serialize_value;
```
piodul, line 2: The `serialize_` part seems unnecessary, these already are in a module called `serialize`.

scylla-cql/src/types/serialize/serialize_row.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::{_macro_internal::ValueList, frame::response::result::ColumnSpec};
+
+pub struct RowSerializationContext<'a> {
+    columns: &'a [ColumnSpec],
+}
+
+impl<'a> RowSerializationContext<'a> {
+    pub fn column_by_idx(&self, i: usize) -> Option<&ColumnSpec> {
+        self.columns.get(i)
+    }
```
piodul, line 12: I'm not exactly a fan of this... This doesn't allow to easily get information about the number of columns, and makes iteration over the columns slightly awkward. How about exposing this field through a getter that returns `&'a [ColumnSpec]`?

piodul (changes requested):

scylla-cql/src/types/serialize/serialize_row.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    // TODO: change RowSerializationContext to make this faster
+    pub fn column_by_name(&self, target: &str) -> Option<&ColumnSpec> {
+        self.columns.iter().find(|&c| c.name == target)
+    }
+}
+
+pub trait SerializeRow {
+    fn preliminary_type_check(&self, ctx: &RowSerializationContext<'_>) -> bool;
```
piodul, line 21: I thought that this was supposed to be an associated function, not a method?

Lorak-mmk: Ooops, yes - that's what I did in previous version but forgot about in this, sorry

Lorak-mmk:
Adressed review comments

scylla-cql/src/types/serialize/row.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct RowSerializationContext<'a> {
+    columns: &'a [ColumnSpec],
+}
+
+impl<'a> RowSerializationContext<'a> {
+    pub fn columns(&self) -> &'a [ColumnSpec] {
+        self.columns
+    }
+
+    // TODO: change RowSerializationContext to make this faster
```
piodul, line 17: Nit: what do you mean by this comment?

scylla-cql/src/types/serialize/row.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::frame::value::ValueList;
+
+use super::SerializationError;
+
+pub struct RowSerializationContext<'a> {
+    columns: &'a [ColumnSpec],
+}
+
+impl<'a> RowSerializationContext<'a> {
+    pub fn columns(&self) -> &'a [ColumnSpec] {
```
piodul, line 13: Nit: this method and `column_by_name` could be made `#[inline]` - they are small and simple, and without it the compiler won't consider inlining them in when used in external crates.

piodul (approved):
Two small issues/comments, otherwise LGTM.

scylla-cql/src/types/serialize/row.rs
Lorak-mmk: Apart from list of clumns we can also keep a hashmap from name to column spec so that name lookups are in constant time instead of linear.

scylla-cql/src/types/serialize/row.rs
Lorak-mmk: Done

scylla-cql/src/types/serialize/row.rs
piodul: I see.

piodul (approved):

Issue #820, opened 2023-10-05
Incorrect error for batching


MathieuDutSik:
I ran some batch with `30000` entries and it works correctly. With `40000` it does not work. I am fine with that however, the errors appears incorrect. The error that occurs is
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidMessage("Frame error: Could not deserialize frame: Integer conversion out of range")', src/main.rs:58:53
```

The code that is triggering it is
```rust
use bcs::serialize_into;
use scylla::query::Query;
use scylla::{IntoTypedRows, Session, SessionBuilder};

fn get_upper_bound_option(key_prefix: &[u8]) -> Option<Vec<u8>> {
    let len = key_prefix.len();
    for i in (0..len).rev() {
        let val = key_prefix[i];
        if val < u8::MAX {
            let mut upper_bound = key_prefix[0..i + 1].to_vec();
            upper_bound[i] += 1;
            return Some(upper_bound);
        }
    }
    None
}

async fn find_keys_by_prefix(session: &Session, key_prefix: Vec<u8>) -> Vec<Vec<u8>> {
    let len = key_prefix.len();
    let rows = match get_upper_bound_option(&key_prefix) {
        None => {
            let values = (key_prefix,);
            let query = "SELECT k FROM kv.pairs WHERE dummy = 0 AND k >= ? ALLOW FILTERING";
            session.query(query, values).await.unwrap()
        }
        Some(upper_bound) => {
            let values = (key_prefix, upper_bound);
            let query =
                "SELECT k FROM kv.pairs WHERE dummy = 0 AND k >= ? AND k < ? ALLOW FILTERING";
            session.query(query, values).await.unwrap()
        }
    };
    let mut keys = Vec::new();
    if let Some(rows) = rows.rows {
        for row in rows.into_typed::<(Vec<u8>,)>() {
            let key = row.unwrap();
            let short_key = key.0[len..].to_vec();
            keys.push(short_key);
        }
    }
    keys
}

async fn write_batch(session: &Session, n: usize) {
    let mut batch_query =
        scylla::statement::batch::Batch::new(scylla::frame::request::batch::BatchType::Logged);
    let mut batch_values = Vec::new();
    for i in 0..n {
        let mut key = vec![0];
        serialize_into(&mut key, &(i as usize)).unwrap();
        let value = key.clone();
[...]

PR #824, opened 2023-10-06
Fix/batch length

samuelorji:
## Pre-review checklist

Modify the type of the number of queries in a batch statement from `i16` to `u16`. As well as add guards to prevent making server calls when the number of queries is over u16::MAX

This is following [this conversation in the issue](https://github.com/scylladb/scylla-rust-driver/issues/820#issuecomment-1751327935)

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/820


scylla-cql/src/frame/types.rs
```diff
@@ -98,6 +98,12 @@ impl From<std::str::Utf8Error> for ParseError {
     }
 }
 
+impl From<Infallible> for ParseError {
+    fn from(_: Infallible) -> Self {
+        ParseError::BadIncomingData("Unexpected Infallible Error".to_string())
+    }
+}
```
cvybhu, line 105: Could you explain what this trait implementation does?
AFAIU `Infallible` is for things that can never happen, so why do we want to convert it to a `ParseError`?

https://doc.rust-lang.org/std/convert/enum.Infallible.html

scylla-cql/src/frame/types.rs
```diff
@@ -174,10 +180,19 @@ pub fn read_short(buf: &mut &[u8]) -> Result<i16, ParseError> {
...
+    Ok(v)
+}
+
 pub fn write_short(v: i16, buf: &mut impl BufMut) {
     buf.put_i16(v);
 }
 
+pub fn write_u16(v: u16, buf: &mut impl BufMut) {
+    buf.put_u16(v);
+}
```
cvybhu, line 194: Maybe we should change `write_short` to accept `u16`.
AFAIU `short` refers to the CQL `short` type, which is a `u16`.
But we can do that in a separate PR as it will be a large change. Let's keep this one focused on the problem with Batches.

scylla-cql/src/errors.rs
```diff
@@ -348,6 +348,10 @@ pub enum BadQuery {
     #[error("Passed invalid keyspace name to use: {0}")]
     BadKeyspaceName(#[from] BadKeyspaceName),
 
+    /// Too many queries in the batch statement
+    #[error("Number of Queries in Batch Statement has exceeded the max value of 65,536")]
+    TooManyQueriesInBatchStatement,
```
cvybhu, line 353: The maximum value of `u16` is 65,535 - one less than 65,536.
https://doc.rust-lang.org/std/primitive.u16.html#associatedconstant.MAX

scylla/src/transport/large_batch_statements_test.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    let err = write_batch(&session, too_many_queries).await;
+
+    assert!(err.is_err());
+}
+
+async fn create_test_session(session: Session, ks: &String) -> Session {
+    session
+        .query(
+            format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{ 'class' : 'SimpleStrategy', 'replication_factor' : 1 }}",ks),
```
cvybhu, line 36: Please change `SimpleStrategy` to `NetworkTopologyStrategy`.
We are currently phasing out `SimpleStrategy`, in the future only `NetworkTopologyStrategy` will be allowed.
The query syntax is the same, only the name has to replaced.

scylla-cql/src/errors.rs
cvybhu: It'd also be nice to display the number of queries that the user has passed. This can be done by adding `(usize)` to the enum variant and displaying it with `{0}`.

scylla/src/transport/large_batch_statements_test.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    let key_prefix = vec![0];
+    let keys = find_keys_by_prefix(&session, key_prefix.clone()).await;
+    assert_eq!(keys.len(), max_number_of_queries);
+
+    let too_many_queries = u16::MAX as usize + 1;
+
+    let err = write_batch(&session, too_many_queries).await;
+
+    assert!(err.is_err());
```
cvybhu, line 30: Please make sure that the error is actually `TooManyQueriesInBatchStatement`.
I think this can be done using `assert_matches!`.

scylla/src/transport/large_batch_statements_test.rs
```diff
@@ -0,0 +1,106 @@
...
+    if let Some(rows) = rows.rows {
+        for row in rows.into_typed::<(Vec<u8>,)>() {
+            let key = row.unwrap();
+            let short_key = key.0[len..].to_vec();
+            keys.push(short_key);
+        }
+    }
+    keys
+}
+
```
cvybhu, line 94: TBH I don't really follow the logic of key prefixes.
Would it be possible to just insert number 1..n and then read them all from the table?
I'm a fan of keeping things as simple as possible.

cvybhu (changes requested):
Thank you for the contribution!
Left some comments, please take a look.

scylla-cql/src/frame/types.rs
piodul: I agree about changing `write_short` to accept `u16`. There is only a small number of uses of this function in the code (I counted only six) so it should be easy to fix it right now.

scylla-cql/src/frame/types.rs
samuelorji: It's for [the conversion ](https://github.com/scylladb/scylla-rust-driver/blob/8dfdcdbc6ea7299b76ee79e625f4a4f8cd009f15/scylla-cql/src/frame/request/batch.rs#L193)of `u16` to a `usize`. I wanted to do a simple `as` but didn't want to modify the existing code as much. I think converting from the previous `i16` to a `usize` would have failed with the `TryFromIntError` error, but with `u16` -> `usize`, it really is in fact infallible

scylla/Cargo.toml
```diff
@@ -60,6 +60,7 @@ criterion = "0.4" # Note: v0.5 needs at least rust 1.70.0
 tracing-subscriber = { version = "0.3.14", features = ["env-filter"] }
 assert_matches = "1.5.0"
 rand_chacha = "0.3.1"
+bcs = "0.1.5"
```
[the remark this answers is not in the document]
samuelorji: Thought so too, was trying to fully replicate the issue. Fixed now

scylla-cql/src/frame/types.rs
samuelorji: The switch wasn't so simple, i had to change the type in some other structs that were linked in a way I don't understand yet. [example](https://github.com/scylladb/scylla-rust-driver/blob/8dfdcdbc6ea7299b76ee79e625f4a4f8cd009f15/scylla-cql/src/frame/value.rs#L66). But as long as the tests pass, it should be fine.

scylla-cql/src/errors.rs
samuelorji: Nice Catch. Fixed

scylla/src/transport/large_batch_statements_test.rs
samuelorji: Thanks, knew there was a way to do this without having to implement `PartialEq` on everything 😢

scylla/src/transport/large_batch_statements_test.rs
samuelorji: modified the tests to be simpler, but still show the intent of batch queries

cvybhu (changes requested):
It looks like the CI has failed.
Please make sure that `cargo check --examples --tests` and `cargo clippy --examples --tests` doesn't produce any errors or warnings.

samuelorji:
@cvybhu. I've made the change

samuelorji:
@cvybhu , any help trying to get the CI to work, it seems to be failing for a myriad of reasons when it works perfectly on my machine.

now, I've added an `#![allow(unused_imports)]` to prevent that failure amongst other things

cvybhu:
> @cvybhu , any help trying to get the CI to work, it seems to be failing for a myriad of reasons when it works perfectly on my machine.
now, I've added an #![allow(unused_imports)] to prevent that failure amongst other things

I saw that the CI is failing on `cargo clippy`. It's possible that you have an older version of rust installed and because of that it doesn't produce some of the warnings that the latest version does. CI always the latest version of stable rust (`1.73.0` at the moment)

Please make sure that `cargo --version` and `cargo clippy --version` print `1.73.0`:
```bash
$ cargo --version
cargo 1.73.0 (9c4383fb5 2023-08-26)
$ cargo clippy --version
clippy 0.1.73 (cc66ad4 2023-10-03)
```

If it doesn't you have to update rust to get the latest warnings. If you installed it using `rustup` you can just run `rustup update`.

Another thing is that you have to pass the flags `--examples --tests` to cargo clippy. Without them it ignores all the code in tests and examples.

Make sure that you are running the same commands that the workflow does: https://github.com/scylladb/scylla-rust-driver/blob/e6d6d3e5bd97ec746e12640a3b861e3cbdffd404/.github/workflows/rust.yml#L29C1-L29C1

Although you can probably skip the --verbose flag to reduce the output.

cvybhu:
> now, I've added an #![allow(unused_imports)] to prevent that failure amongst other things

It'd be better to remove the unused imports that clippy is complaining about.

cvybhu:
> Now, it seems like the test I wrote is a little flakey, although I don't know why. In the previous run, it ran fine, but now, it's failing on this [quite simple assertion](https://github.com/scylladb/scylla-rust-driver/blob/5e731af3bafbe0436874dd0b158417e3c254c584/scylla/src/transport/large_batch_statements_test.rs#L21) that runs fine locally for me

It looks like the test passes on Scylla, but fails on Cassandra. `scylla-rust-driver` supports both databases, so we have to make sure that it passes on both of them.

You can try running this test against Cassandra instead of Scylla.

You can start a single Cassandra node using docker:
```bash
docker run --rm -it -p 9042:9042 -p 19042:19042 cassandra
```
and then run your test on this node and see what goes wrong.

samuelorji:
@cvybhu , I think I found the issue, the batch write result sometimes result in a write timeout error:
```
DbError(WriteTimeout { consistency: LocalQuorum, received: 0, required: 1, write_type: Simple }, "Operation timed out - received only 0 responses.")
```

The only way I could replicate this was by running a cassandra cluster instead of a single instance, and even then, it happened infrequently, but the times when the batch write didn't succeed, it was due to the write timeout, which could be caused by the large size of the data or busy nodes. 

WDYT

cvybhu:
> @cvybhu , I think I found the issue, the batch write result sometimes result in a write timeout error:

Ah that makes sense, Scylla is the faster (better) database after all ;) 60k requests can take a moment to complete.

In that case let's increase the timeout so that Cassandra can complete the work. There used to be `set_request_timeout` function, but AFAIR this was replaced by `ExecutionProfiles`. So you'll have to create an `ExecutionProfile` with larger timeout and make the `Batch` statement use this `ExecutionProfile`.

samuelorji:
> > @cvybhu , I think I found the issue, the batch write result sometimes result in a write timeout error:
> 
> Ah that makes sense, Scylla is the faster (better) database after all ;) 60k requests can take a moment to complete.
> 
> In that case let's increase the timeout so that Cassandra can complete the work. There used to be `set_request_timeout` function, but AFAIR this was replaced by `ExecutionProfiles`. So you'll have to create an `ExecutionProfile` with larger timeout and make the `Batch` statement use this `ExecutionProfile`.

Okay, lemme give that a try

samuelorji:
@cvybhu I've made the change. Added the change as a different commit that should be eventually squashed

samuelorji:
so weird how my cargo and cargo clippy versions are both:

```
$ cargo --version
cargo 1.73.0 (9c4383fb5 2023-08-26)
$ cargo clippy --version
clippy 0.1.73 (cc66ad4 2023-10-03)
```

and I'm running Clippy check an Format Check. and I'm not getting any error, but it's weirdly failing on CI 🤔

cvybhu:
> and I'm running Clippy check an Format Check. and I'm not getting any error, but it's weirdly failing on CI 🤔

Looks like everything is OK in the last version. Clippy check passes in the CI now.

It failed on Cassandra tests again. Maybe you have to increase the timeout even more.

Btw you could replace `assert(batch_result.is_ok())` with `batch_result.unwrap()`. `unwrap()` prints the error message on failure, so we would be able to see the timeout message in CI logs.

samuelorji:
Still a write timeout error even with 240 seconds

```
thread 'transport::large_batch_statements_test::test_large_batch_statements' panicked at 'called `Result::unwrap()` on an `Err` value: DbError(WriteTimeout { consistency: LocalQuorum, received: 0, required: 1, write_type: Simple }, "Operation timed out - received only 0 responses.")', scylla/src/transport/large_batch_statements_test.rs:33:26
```

piodul:
> Still a write timeout error even with 240 seconds
> 
> ```
> thread 'transport::large_batch_statements_test::test_large_batch_statements' panicked at 'called `Result::unwrap()` on an `Err` value: DbError(WriteTimeout { consistency: LocalQuorum, received: 0, required: 1, write_type: Simple }, "Operation timed out - received only 0 responses.")', scylla/src/transport/large_batch_statements_test.rs:33:26
> ```

Let's not introduce unit tests that are very long to execute. I suppose the goal is to just verify that batch serialization succeeds, right? Can we just call the serialization logic directly and test that it returns the expected error or succeeds?

Alternatively, you can see if using prepared statements in the batch helps - but a batch with 60k statements is abnormally large nonetheless, so I doubt the test will end in a reasonable time.

piodul:
One more thing - the timeout that you observed occurs on the DB side, i.e. Cassandra itself returns a timeout error. Increasing the timeout on the driver side will not help - it only affects how long the driver waits for response. You would have to modify the configuration used to run Cassandra in CI.

samuelorji:
> One more thing - the timeout that you observed occurs on the DB side, i.e. Cassandra itself returns a timeout error. Increasing the timeout on the driver side will not help - it only affects how long the driver waits for response. You would have to modify the configuration used to run Cassandra in CI.

Okay, makes sense, because the check for the size of the statements occurs at the `session` level, will a single test that shows the expected error suffice ?

piodul:
> Okay, makes sense, because the check for the size of the statements occurs at the `session` level, will a single test that shows the expected error suffice ?

I think yes. It would be better to also have a test which shows that serializing a large batch that is just under the limit works, but to me it's even better not to have long tests or flaky tests (due to timeouts). I don't suspect that regression here is likely, so if it's hard to write such a test then I won't insist.

cvybhu:
> Let's not introduce unit tests that are very long to execute. I suppose the goal is to just verify that batch serialization succeeds, right? Can we just call the serialization logic directly and test that it returns the expected error or succeeds?

@piodul I think it would be good to have a test that actually runs a batch with ~40k queries. Otherwise how do we know that it actually works? It will take a moment, but IMO it's better to have a few heavy tests rather than no tests at all.
In Pytest there are markers like `@heavy_test` and it's possible to run all tests except for the heavy ones for quick feedback. Maybe Rust has something similar?

Plus inserting 40k rows shouldn't take THAT long. AFAIR Scylla was capable of 100k inserts/s on a single shard, so it shouldn't take more than a few seconds to run this test. Cassandra might be worse, but we could just disable this test for Cassandra.
One more idea would be to insert the same row 40k times - that would all be a single mutation, so AFAIU there wouldn't be any memtable flushes, should be really fast.

cvybhu:
Rust docs mention that there are two main types of tests - unit tests, which are next to the source code and integration tests, which are in the `tests` folder[1]. It's possible to run only unit tests using `cargo test --lib` [2].

I think we could add a test that runs a huge batch as an integration test. It matches the definition of an integration test, because it uses the library in the same way as an outside user would. Then we'd have a split - unit tests are quick, can be ran alone, while integration tests can be heavier, but can be skipped when we need only the quick tests.

[1] https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory
[2] https://github.com/rust-lang/cargo/issues/8396

piodul:
> > Let's not introduce unit tests that are very long to execute. I suppose the goal is to just verify that batch serialization succeeds, right? Can we just call the serialization logic directly and test that it returns the expected error or succeeds?
> 
> @piodul I think it would be good to have a test that actually runs a batch with ~40k queries. Otherwise how do we know that it actually works? It will take a moment, but IMO it's better to have a few heavy tests rather than no tests at all.

I tried running the test locally and it actually doesn't take that long (<1s to insert a batch, both with Scylla and Cassandra). I was concerned about not making the tests too long for the developers to execute them locally, but I don't think this one sticks out too much (it spends most of the time on schema changes, actually - this seems to be a problem with most of our existing tests).

I'm actually more worried about introducing flakiness to the CI.

> In Pytest there are markers like `@heavy_test` and it's possible to run all tests except for the heavy ones for quick feedback. Maybe Rust has something similar?

Rust's default testing framework, sadly, doesn't have too many features, so I doubt it.

> 
> Plus inserting 40k rows shouldn't take THAT long. AFAIR Scylla was capable of 100k inserts/s on a single shard, so it shouldn't take more than a few seconds to run this test. Cassandra might be worse, but we could just disable this test for Cassandra.

Disabling the test for Cassandra is also an option.

> One more idea would be to insert the same row 40k times - that would all be a single mutation, so AFAIU there wouldn't be any memtable flushes, should be really fast.

The batch should already result in a single mutation, it only modifies a single partition. I don't know about Cassandra, but Scylla also seems to struggle with batches this large - I've seen some large stalls in the logs related to applying mutation to the memtable.

samuelorji:
@cvybhu @piodul to move forward, should reinstate the previous tests that does an actual write but then ignore the tests for cassandra as done [here](https://github.com/scylladb/scylla-rust-driver/blob/e6d6d3e5bd97ec746e12640a3b861e3cbdffd404/.github/workflows/cassandra.yml#L32C138-L32C163)

piodul:
> @cvybhu @piodul to move forward, should reinstate the previous tests that does an actual write but then ignore the tests for cassandra as done [here](https://github.com/scylladb/scylla-rust-driver/blob/e6d6d3e5bd97ec746e12640a3b861e3cbdffd404/.github/workflows/cassandra.yml#L32C138-L32C163)

I'm fine with this approach. @cvybhu ?

scylla-cql/src/frame/types.rs
cvybhu: Ah okay, I understand now.
So we have a few pieces of code like this one:
```rust
let a: i16 = 123;
let b: usize = a.try_into()?;
```
And after changing the `i16` to `u16` they no longer compile because `try_into()` has an `Infallible` error type.
We can implement a conversion from `Infallible` to `ParseError` to make it compile, but it's a bit hacky.

I think it would be better to replace the `try_into()?`with `into()`, like this:
```rust
let a: u16 = 123;
let b: usize = a.into();
```

There is a an implentation of `From<u16> for usize`, so we can just use `into()` here.

cvybhu (changes requested):
Thanks, it looks nice.

I left a few more comments, please take a look, and then we can merge it :)

cvybhu (approved):
Looks good to me.

@piodul do you want to take a final look? I'm ready to  merge.

scylla/src/transport/large_batch_statements_test.rs
```diff
@@ -0,0 +1,67 @@
...
+    assert_matches!(
+        batch_insert_result.unwrap_err(),
+        QueryError::BadQuery(BadQuery::TooManyQueriesInBatchStatement(_too_many_queries)) if _too_many_queries == too_many_queries
+    )
+}
+
+async fn create_test_session(session: Session, ks: &String) -> Session {
+    session
+        .query(
+            format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{ 'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1 }}",ks),
```
piodul, line 36: You can use `CREATE KEYSPACE` without `IF EXISTS`, the `ks` name is guaranteed to be unique.

scylla/src/transport/large_batch_statements_test.rs
```diff
@@ -0,0 +1,67 @@
...
+            &[],
+        )
+        .await.unwrap();
+    session
+        .query(format!("DROP TABLE IF EXISTS {}.pairs;", ks), &[])
+        .await
+        .unwrap();
```
piodul, line 43: This is not needed, the test will create a new keyspace because `ks` is guaranteed to be unique.

scylla/src/transport/large_batch_statements_test.rs
```diff
@@ -0,0 +1,67 @@
...
+    let mut batch_query = Batch::new(BatchType::Unlogged);
+    let mut batch_values = Vec::new();
+    for i in 0..n {
+        let mut key = vec![0];
+        key.extend(i.to_be_bytes().as_slice());
+        let value = key.clone();
+        let query = format!("INSERT INTO {}.pairs (dummy, k, v) VALUES (0, ?, ?)", ks);
+        let values = vec![key, value];
+        batch_values.push(values);
+        let query = Query::new(query);
```
piodul, line 63: How about using prepared statements instead? Prepare the statement once before creating the batch and then use it in `append_statement`: `batch_query.append_statement(prepared.clone())`. This should reduce the work needed by DB to process the batch.

piodul (approved):
Some last nitpicks, but after fixing them - LGTM

samuelorji:
@piodul thanks for the review. I've made the change, can you please have another look?

piodul (approved):

cvybhu (approved):
LGTM

Issue #747, opened 2023-06-20
Misleading error when using nonexistent DC


Lorak-mmk:
When using `DefaultPolicy::builder().prefer_datacenter(dc)` with nonexistent datacenter the driver returns a very misleading erro when trying to execute a query: `Error: ProtocolError("Empty query plan - driver bug!")`.

Full reproducer: https://github.com/fee-mendes/rust-driver-example/blob/main/metric-collector
Driver version: 0.8.1, as can be seen in above reproducer.

To reproduce, simply run `cargo run --bin metric-collector <scylla_ip> this_dc_does_not_exist`

samuelorji:
stab: https://github.com/scylladb/scylla-rust-driver/pull/825

PR #825, opened 2023-10-08
Return Query Error for non existent DC

samuelorji:
Return query error when there are no known nodes found during query planning

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/747

## Pre-review checklist

- [-] I have adjusted the documentation in `./docs/source/`.
- [-] I added appropriate `Fixes:` annotations to PR description.

### DIFF
--- a/scylla-cql/src/errors.rs
+++ b/scylla-cql/src/errors.rs
@@ -46,6 +46,10 @@ pub enum QueryError {
     #[error("Request timeout: {0}")]
     RequestTimeout(String),
 
+    /// No known node found to perform query
+    #[error("No known node found: {0}")]
+    NoKnownNodeFoundError(String),
+
     /// Address translation failed
     #[error("Address translation failed: {0}")]
     TranslationError(#[from] TranslationError),
@@ -404,6 +408,10 @@ pub enum NewSessionError {
     #[error("Client timeout: {0}")]
     RequestTimeout(String),
 
+    /// No known node found to perform query
+    #[error("No known node found: {0}")]
+    NoKnownNodeFoundError(String),
+
     /// Address translation failed
     #[error("Address translation failed: {0}")]
     TranslationError(#[from] TranslationError),
@@ -482,6 +490,7 @@ impl From<QueryError> for NewSessionError {
             QueryError::UnableToAllocStreamId => NewSessionError::UnableToAllocStreamId,
             QueryError::RequestTimeout(msg) => NewSessionError::RequestTimeout(msg),
             QueryError::TranslationError(e) => NewSessionError::TranslationError(e),
+            QueryError::NoKnownNodeFoundError(e) => NewSessionError::NoKnownNodeFoundError(e),
         }
     }
 }
--- a/scylla-cql/src/errors.rs
+++ b/scylla-cql/src/errors.rs
@@ -46,9 +46,9 @@ pub enum QueryError {
     #[error("Request timeout: {0}")]
     RequestTimeout(String),
 
-    /// No known node found to perform query
-    #[error("No known node found: {0}")]
-    NoKnownNodeFoundError(String),
+    /// Empty Query Plan
+    #[error("Load balancing policy returned empty query plan. It can happen when the driver is provided with non-existing datacenter name")]
+    EmptyQueryPlan,
 
     /// Address translation failed
     #[error("Address translation failed: {0}")]
@@ -408,9 +408,9 @@ pub enum NewSessionError {
     #[error("Client timeout: {0}")]
     RequestTimeout(String),
 
-    /// No known node found to perform query
-    #[error("No known node found: {0}")]
-    NoKnownNodeFoundError(String),
+    /// Empty Query Plan
+    #[error("Load balancing policy returned empty query plan. It can happen when the driver is provided with non-existing datacenter name")]
+    EmptyQueryPlan,
 
     /// Address translation failed
     #[error("Address translation failed: {0}")]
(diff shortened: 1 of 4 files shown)


scylla/src/transport/session.rs
```diff
@@ -1653,7 +1653,8 @@ impl Session {
         QueryFut: Future<Output = Result<ResT, QueryError>>,
         ResT: AllowedRunQueryResTType,
     {
-        let mut last_error: Option<QueryError> = None;
+        // set default error as no known found as the query plan returns an empty iterator if there are no nodes in the plan
+        let mut last_error: Option<QueryError> = Some(QueryError::NoKnownNodeFoundError("Please confirm the supplied datacenters exists".to_string()));
```
samuelorji, line 1657: The query_plan is an [iterator](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla/src/transport/session.rs#L1661) that provides a Node, but this iterator can be empty due to the inability to [find/pick a node](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla/src/transport/load_balancing/plan.rs#L71) which is mostly [caused here](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla/src/transport/load_balancing/default.rs#L398). 

There could be other reasons why the iterator could be empty, but I can't see any yet

samuelorji:
@Lorak-mmk. I can see that the Cassandra tests are failing on CI, but they work locally for me. Thoughts?

scylla/src/transport/session.rs
```diff
@@ -1653,7 +1653,10 @@ impl Session {
         QueryFut: Future<Output = Result<ResT, QueryError>>,
         ResT: AllowedRunQueryResTType,
     {
-        let mut last_error: Option<QueryError> = None;
+        // set default error as no known found as the query plan returns an empty iterator if there are no nodes in the plan
+        let mut last_error: Option<QueryError> = Some(QueryError::NoKnownNodeFoundError(
+            "Please confirm the supplied datacenters exists".to_string(),
+        ));
         let mut current_consistency: Consistency = context
```
Lorak-mmk, line 1660: If I understand correctly, the problem happens when `query_plan` is empty. The question is, is this the only case when `last_error` can remain empty?
From brief reading of the code it looks like it, but I'm not 100% sure, and it could break in the future.
Could you rewrite this so that the new error is only returned when `query_plan` is empty, and otherwise the current behavior is preserved?

scylla/src/transport/session.rs
```diff
@@ -1653,7 +1653,10 @@ impl Session {
         QueryFut: Future<Output = Result<ResT, QueryError>>,
         ResT: AllowedRunQueryResTType,
     {
-        let mut last_error: Option<QueryError> = None;
+        // set default error as no known found as the query plan returns an empty iterator if there are no nodes in the plan
+        let mut last_error: Option<QueryError> = Some(QueryError::NoKnownNodeFoundError(
+            "Please confirm the supplied datacenters exists".to_string(),
+        ));
```
Lorak-mmk, line 1659: I see we don't have access to provided datacenter name so we can't use it in error message. But the message could be better I think - right now it explains a probable fix, but does not say what went wrong.
Maybe something like `Load balancing policy returned empty query plan. It can happen when driver is provided with non-existing datacenter name`.

scylla-cql/src/errors.rs
```diff
@@ -46,6 +46,10 @@ pub enum QueryError {
     #[error("Request timeout: {0}")]
     RequestTimeout(String),
 
+    /// No known node found to perform query
+    #[error("No known node found: {0}")]
+    NoKnownNodeFoundError(String),
```
Lorak-mmk, line 51: I'm not convinced by this name. What is `KnownNode` in this context? Maybe something like `EmptyQueryPlan`?

Lorak-mmk (changes requested):

scylla/src/transport/session.rs
```diff
@@ -1666,7 +1666,10 @@ impl Session {
             .consistency_set_on_statement
             .unwrap_or(execution_profile.consistency);
 
+        let mut query_plan_is_empty = true;
```
samuelorji, line 1669: Wanted to do a peekable check to ensure the iterator is not empty like:
```rust
query_plan.peekable()
```

but it caused this lifetime error:
```
error: higher-ranked lifetime error
   --> scylla/src/transport/session_test.rs:409:5
    |
409 | /     tokio::spawn(async move {
410 | |         let values = (
411 | |             (1_i32, 2_i32, "abc"),
412 | |             (),
...   |
415 | |         session_clone.batch(&batch, values).await.unwrap();
416 | |     })
    | |______^
    |
    = note: could not prove `[async block@scylla/src/transport/session_test.rs:409:18: 416:6]: std::marker::Send`
    ```
    So, I just defaulted to doing this more simple check

File: scylla/src/transport/session_test.rs [hunk 6]
```diff
@@ -2857,3 +2858,33 @@ async fn test_manual_primary_key_computation() {
         .await;
     }
 }
+
+#[tokio::test]
+async fn test_non_existent_dc_return_correct_error() {
+    let ks = "iot";
+
+    let host = "127.0.0.1";
```
Lorak-mmk, line 2866: The tests fail because they run Scylla on different IP. Please take a look at other tests in this file - they use `create_new_session_builder()` to get a session builder with correct IP. You can then call 
```
        .default_execution_profile_handle(handle)
        .build()
        .await
        .expect("cannot create session");
```
on resulting builder.

samuelorji: Thanks

piodul (changes requested):
Notice that `execute_query` returns `None` in case there were no nodes tried. The `None` is converted into `QueryError::ProtocolError("Empty query plan - driver bug!")` at some point - look those places up by grepping on the error message. I don't like the modifications made to `execute_query` because now

- The function never returns `None`, but it has `Option` in its return type,
- The new behavior will break speculative execution (look at `speculative_execution.rs`).

I don't think that `execute_query` needs to be touched at all, it's just that the error message needs to be changed. You should go over the places where `Empty query plan - driver bug!` is returned, change it to the new error variant you introduced and that should be it.

wprzytula:
@samuelorji what's the status of this?

Lorak-mmk:
Closing in favor of https://github.com/scylladb/scylla-rust-driver/pull/1371

PR #830, opened 2023-10-11
Fix crash when Instant::now() returns the same value twice

Lorak-mmk:
When Instant::now() returns the same value during 2 calls of
TimestampedAverage::compute_next, division by 0 occurs,
and Duration::from_secs_f64 panics because of NaN input.

This PR fixes the issue, adds additional logs in case another conversion problem
occurs in this code and adds regression test for the issue.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -2132,7 +2132,7 @@ mod latency_awareness {
     use futures::{future::RemoteHandle, FutureExt};
     use itertools::Either;
     use scylla_cql::errors::{DbError, QueryError};
-    use tracing::{instrument::WithSubscriber, trace};
+    use tracing::{instrument::WithSubscriber, trace, warn};
     use uuid::Uuid;
 
     use crate::{load_balancing::NodeRef, transport::node::Node};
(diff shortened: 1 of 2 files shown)


piodul (commented):
Looks good, thanks. Only one thing - please swap the commits together. I know it could make some sense to introduce a failing test first and then fix it in order to better show that the fix works (e.g. I checked out the branch and saw how it behaves with and without the fix as a part of the review) - but this breaks bisectability.

Lorak-mmk:
Done

scylla/src/transport/load_balancing/default.rs
```diff
@@ -2192,13 +2192,37 @@ mod latency_awareness {
                 Some(prev_avg) => Some({
                     let delay = (now - prev_avg.timestamp).as_secs_f64();
```
piodul, line 2193: https://doc.rust-lang.org/std/time/struct.Instant.html#monotonicity

Assuming that `compute_next` could be called with timestamp that is smaller than `prev_avg.timestamp`, this code will... work correctly as of now, but may start to panic in the future. Let's use [checked_duration_since](https://doc.rust-lang.org/std/time/struct.Instant.html#method.checked_duration_since) here to be sure.

Lorak-mmk: Is the current version ok?

piodul: Yes, it's OK. TBF I realized that [saturating_duration_since](https://doc.rust-lang.org/std/time/struct.Instant.html#method.saturating_duration_since) would be nicer and shorter, but I won't insist on changing that (unless clippy complains)

Lorak-mmk: Changed it, just to be safe in the future

piodul (approved):

Issue #671, opened 2023-03-22
Provide a way to configure which keyspaces need replica set precomputation


havaker:
With the merge of https://github.com/scylladb/scylla-rust-driver/pull/612, a new module called `scylla::transport::locator` was added to the `scylla-rust-driver`. This module offers a means to compute replica sets for a given `strategy, token` pair, which is commonly used in load balancing. To speed up these computations, replica set precomputation was included as a part of #612. Unfortunately, precomputing replica sets can be a bit costly in terms of memory and compute resources.

Currently, replica set precomputation is enabled by default for all keyspaces. While this can greatly improve the speed of computing replica sets, it may be unnecessary to use memory and compute resources for precomputation for keyspaces that are not in use by the driver.

Adding an option to configure replica set precomputation for specific keyspaces would solve this problem. By default, precomputation could be enabled for all keyspaces to provide fast access to replica sets. Users could selectively disable it for specific keyspaces (or enable only for specific ones) to better adjust the driver's resource consumption to their use cases.

PR #831, opened 2023-10-11
Add ReplicationOptions to SessionConfig for excluding keyspaces from replica precompute

michaelhly:
We add ReplicationOptions which includes a vector of keyspace names to SessionConfig. When instantiating ClusterData, we exclude keyspaces from provided by ReplicationOptions when computing PrecomputedReplicas to conserve memory/compute resources.

Fixes: #671 

## Pre-review checklist


michaelhly:
@havaker would you mind taking a quick look? Happy to write tests and finish up docs if this seems correct.

piodul:
@Lorak-mmk please review

Lorak-mmk:
You can already choose which keyspaces to precompute (see `keyspaces_to_fetch`) in `SessionConfig`, but:
- it is an allowlist, while your approach is denylist. I'm wondering if maybe the best approach is to allow both (make enum Allowlist(Vec) and Denylist(Vec) instead of just Vec).
- It prevents info about other keyspaces to be fetched at all (so it's not available in `ClusterData`), while your solution only avoids precomputing replication strategies.

So I think this change may be valuable. I'll try to post initial review soon.

wprzytula:
@michaelhly Sorry for so delayed response.
I really like the idea. The taken approach is correct: `ReplicationConfig` (a small name change to be in line with other `Config`s) should be passed down to `ClusterData::new()`.
As @Lorak-mmk noted, there should be both an Allowlist and Denylist variant for specyfing keyspaces to precompute.

scylla/src/transport/cluster.rs
```diff
@@ -274,6 +278,7 @@ impl ClusterData {
         known_peers: &HashMap<Uuid, Arc<Node>>,
         used_keyspace: &Option<VerifiedKeyspaceName>,
         host_filter: Option<&dyn HostFilter>,
+        replication_opts: Option<&ReplicationConfigs>,
     ) -> Self {
```
wprzytula, line 282: Let's not make it an `Option`; `ReplicationConfig` should be always passed to where `ReplicaLocator` is constructed.

scylla/src/transport/cluster.rs
```diff
@@ -339,8 +344,16 @@ impl ClusterData {
         Self::update_rack_count(&mut datacenters);
 
         let keyspaces = metadata.keyspaces;
+        // Exclude these keyspaces from replica set precomputation
+        let keyspaces_to_exclude = match replication_opts {
+            Some(opts) => opts.keyspaces_to_exclude().to_vec(),
+            None => vec![],
+        };
```
wprzytula, line 351: Better not to allocate here. Let's just use the borrowed `Vec` there.

```suggestion
        // Exclude these keyspaces from replica set precomputation
        let keyspaces_to_exclude = replication_config.keyspaces_to_exclude();
```

wprzytula: Even better, let's allow for both Allowlist and Denylist variant:

```suggestion
        let replicas_precomputation_config = replication_config.replicas_precomputation_config();
```

scylla/src/transport/cluster.rs
```diff
@@ -339,8 +344,16 @@ impl ClusterData {
...
         let (locator, keyspaces) = tokio::task::spawn_blocking(move || {
-            let keyspace_strategies = keyspaces.values().map(|ks| &ks.strategy);
+            let keyspace_strategies = keyspaces
+                .iter()
+                .filter(|(k, _)| !keyspaces_to_exclude.contains(k))
+                .map(|(_, ks)| &ks.strategy);
```
wprzytula, line 356:
```suggestion
            let keyspace_strategies = keyspaces
                .iter()
                .filter(|(k, _)| match replicas_precomputation_config {
                    Denylist(disallowed_keyspaces) => !disallowed_keyspaces.contains(k),
                    AllowList(allowed_keyspaces) => allowed_keyspaces.contains(k), 
                )
                .map(|(_, ks)| &ks.strategy);
```

scylla/src/transport/locator/mod.rs
```diff
@@ -21,6 +21,23 @@ use std::{
 };
 use tracing::debug;
 
+/// TODO(michael): Docs
+#[derive(Debug, Default, Clone)]
+pub struct ReplicationConfigs {
```
wprzytula, line 26:
```suggestion
pub struct ReplicationConfig {
```

scylla/src/transport/locator/mod.rs
```diff
@@ -21,6 +21,23 @@ use std::{
...
+    // Keyspaces to exclude from precomuted replica sets
+    keyspaces_to_exclude: Vec<String>,
+}
+
+impl ReplicationConfigs {
+    pub fn exclude_keyspace(&mut self, keyspace: String) {
```
wprzytula, line 32:
```suggestion
    pub fn exclude_keyspace(&mut self, keyspace: impl Into<String>) {
```

scylla/src/transport/locator/mod.rs
```diff
@@ -21,6 +21,23 @@ use std::{
...
+        self.keyspaces_to_exclude.push(keyspace)
+    }
+
+    pub fn keyspaces_to_exclude(&self) -> &Vec<String> {
+        &self.keyspaces_to_exclude
+    }
+}
+
```
wprzytula, line 40: Let's create a `ReplicasPrecomputationList` enum:
```rust
enum ReplicasPrecomputationList {
    Allowlist { allowed_keyspaces: Vec<String> },
    Denylist { disallowed_keyspaces: Vec<String> },
}
```
and store it `ReplicationConfig` as `Option<ReplicaPrecomputationList>`.

scylla/src/transport/session.rs
```diff
@@ -285,6 +286,9 @@ pub struct SessionConfig {
     /// for e.g: if they do not want unexpected traffic
     /// or they expect the topology to change frequently.
     pub cluster_metadata_refresh_interval: Duration,
+
+    /// TODO(michael): Docs ...
+    pub replication_configs: ReplicationConfigs,
```
wprzytula, line 291: `SessionBuilder` needs the corresponding setters as well, as it's the intended way to construct a `Session`.

wprzytula (changes requested):

wprzytula:
Thank you for contributing the idea @michaelhly!

Issue #828, opened 2023-10-10
test_coalescing might be flaky


cvybhu:
`test_coalescing` failed in run https://github.com/scylladb/scylla-rust-driver/actions/runs/6473573762/job/17576595276.
It might be a flaky test.

Here's the failure log:
```
---- transport::connection::tests::test_coalescing stdout ----
Unique name: test_rust_1696964594_11
thread 'transport::connection::tests::test_coalescing' panicked at 'assertion failed: `(left == right)`
[...]

PR #834, opened 2023-10-13
test_coalescing: improve error handling

cvybhu:
`test_coalescing` spawns a lot of tokio tasks, each of which runs a query on the database.
If a query fails, the task will panic because of the `unwrap()`.

The problem is that all of the tasks are joined using `join_all`, which doesn't check for errors in the joined tasks.
For example, this piece of code prints "Ok!" and the program finishes successfuly:
```rust
let mut tasks = Vec::new();
for _i in 0..4 {
    tasks.push(tokio::spawn(async { panic!("Panic!") }));
}

futures::future::join_all(tasks).await;

println!("Ok!");
```
There are some error messages in the output, but the important thig is that they don't stop the control flow.

To fix it, let's use `try_join_all` instead of `join_all`. `try_join_all` will check whether the task panicked or not, and return an error if there was a panic.

I manually tested that it works by inserting a panic!() after `conn.query()`.

Let's also make sure that the result of join_all doesn't contain any `Result<>`, as it could hide the errors. The result of `join_all` should be just a `Vec<>` of unit values.
    
The PR also adds a check that `Connection::query` didn't return `QueryResponse::Error`. It is required to check it in order to ensure that the query didn't fail.

Refs: https://github.com/scylladb/scylla-rust-driver/issues/828

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -1833,6 +1833,7 @@ mod tests {
     use super::ConnectionConfig;
     use crate::query::Query;
     use crate::transport::connection::open_connection;
+    use crate::transport::connection::QueryResponse;
     use crate::transport::node::ResolvedContactPoint;
     use crate::transport::topology::UntranslatedEndpoint;
     use crate::utils::test_utils::unique_keyspace_name;
@@ -2024,9 +2025,13 @@ mod tests {
                         let q = Query::new("INSERT INTO t (p, v) VALUES (?, ?)");
                         let conn = conn.clone();
                         async move {
-                            conn.query(&q, (j, vec![j as u8; j as usize]), None)
+                            let response: QueryResponse = conn
+                                .query(&q, (j, vec![j as u8; j as usize]), None)
                                 .await
                                 .unwrap();
+                            // QueryResponse might contain an error - make sure that there were no errors
+                            let _nonerror_response =
+                                response.into_non_error_query_response().unwrap();
                         }
                     });
                     let _joined: Vec<()> = futures::future::join_all(futs).await;
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -2026,16 +2026,16 @@ mod tests {
                         async move {
                             conn.query(&q, (j, vec![j as u8; j as usize]), None)
                                 .await
-                                .unwrap()
+                                .unwrap();
                         }
                     });
-                    futures::future::join_all(futs).await;
+                    let _joined: Vec<()> = futures::future::join_all(futs).await;
                 }));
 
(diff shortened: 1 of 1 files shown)


Lorak-mmk (approved):

scylla/src/transport/connection.rs
```diff
@@ -2023,19 +2023,15 @@ mod tests {
...
                         let conn = conn.clone();
-                        async move {
-                            conn.query(&q, (j, vec![j as u8; j as usize]), None)
-                                .await
-                                .unwrap()
-                        }
+                        async move { conn.query(&q, (j, vec![j as u8; j as usize]), None).await }
                     });
-                    futures::future::join_all(futs).await;
+                    futures::future::try_join_all(futs).await
```
piodul, line 2028: I think this part is OK, actually. If you have a compound future (i.e. a future that polls other futures, like the one returned by `join_all`), panics from the inner futures will be propagated properly. I'm not sure that changes are needed here.

scylla/src/transport/connection.rs
```diff
@@ -2023,19 +2023,15 @@ mod tests {
...
                 }));
 
                 tokio::task::yield_now().await;
             }
 
-            futures::future::join_all(futs).await;
+            futures::future::try_join_all(futs).await.unwrap();
```
piodul, line 2034: On the other hand, I'm not sure this line is correct now after your changes. If you look at the definition of `tokio::task::spawn`, you will see that it returns a handle that is a future which returns its own `Result` - if the future polled by the tokio task panics, the handle will contain `Err`. So, because you changed the futures to return `Result<>`s, now `try_join_all` returns a `Result<Result<_>>`.

Assuming that you won't change the `futs.push` part of the code, you will need to do two unwraps here. On the other hand, if you revert changes you did to the `futs.push` part, you should keep the line directly above as it is and that's it.

To sum up, the whole change should be an one-liner.

piodul (changes requested):

scylla/src/transport/connection.rs
cvybhu: Ahh you're right. It's a shame that rust didn't complain about unused Results here.
I uploaded new version which also checks that the result of `join_all` is always a `Vec<()>`

scylla/src/transport/connection.rs
cvybhu: Yeah, I guess we can just panic there.
I was scared of panics because I don't fully understand how tokio handles panics, but it looks like it would be ok.

Although I found one more issue here. `Connection::query` returns a `QueryResponse`, which can be `QueryResponse::Error`, so we have to also check that. Fixed in the new version.

piodul (approved):

PR #838, opened 2023-10-16
Dereference `datacenter` of type `&&str` when comparing with `&str`

oeb25:
I encountered a compile error when updating from `0.7.0` to `0.10.1`:

```rs
error[E0277]: can't compare `Option<&str>` with `Option<&&str>`
   --> /Users/oeb25/.cargo/registry/src/index.crates.io-6f17d22bba15001f/scylla-0.10.1/src/transport/locator/mod.rs:278:59
    |
278 |                 .filter(|node| node.datacenter.as_deref() == Some(datacenter))
    |                                                           ^^ no implementation for `Option<&str> == Option<&&str>`
    |
    = help: the trait `PartialEq<Option<&&str>>` is not implemented for `Option<&str>`
    = help: the following other types implement trait `PartialEq<Rhs>`:
              <Option<Box<U>> as PartialEq<rkyv::niche::option_box::ArchivedOptionBox<T>>>
              <Option<T> as PartialEq>
              <Option<U> as PartialEq<rkyv::option::ArchivedOption<T>>>
```

<details>
<summary><i>Two more similar errors</i></summary>

```rs
error[E0277]: can't compare `Option<&str>` with `Option<&&str>`
   --> /Users/oeb25/.cargo/registry/src/index.crates.io-6f17d22bba15001f/scylla-0.10.1/src/transport/locator/mod.rs:320:63
    |
320 |                     .filter(|node| node.datacenter.as_deref() == Some(datacenter))
    |                                                               ^^ no implementation for `Option<&str> == Option<&&str>`
    |
    = help: the trait `PartialEq<Option<&&str>>` is not implemented for `Option<&str>`
    = help: the following other types implement trait `PartialEq<Rhs>`:
              <Option<Box<U>> as PartialEq<rkyv::niche::option_box::ArchivedOptionBox<T>>>
              <Option<T> as PartialEq>
              <Option<U> as PartialEq<rkyv::option::ArchivedOption<T>>>

error[E0277]: can't compare `Option<&str>` with `Option<&mut &str>`
   --> /Users/oeb25/.cargo/registry/src/index.crates.io-6f17d22bba15001f/scylla-0.10.1/src/transport/locator/mod.rs:469:54
    |
469 |                     if replica.datacenter.as_deref() == Some(datacenter) {
    |                                                      ^^ no implementation for `Option<&str> == Option<&mut &str>`
    |
    = help: the trait `PartialEq<Option<&mut &str>>` is not implemented for `Option<&str>`
    = help: the following other types implement trait `PartialEq<Rhs>`:
              <Option<Box<U>> as PartialEq<rkyv::niche::option_box::ArchivedOptionBox<T>>>
              <Option<T> as PartialEq>
              <Option<U> as PartialEq<rkyv::option::ArchivedOption<T>>>
```

</details>

When building this repo by itself the error does not occur, which I do not understand, and I'd love to know why if anybody can figure it out. Here's [the commit to reproduce](https://github.com/oeb25/stract/commit/70acab5dad6bbc1f182a1094d58f4d566c13db14).

Introducing these dereferences should not change anything for already working code, but should hopefully fix the issue encountered above.

---

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/locator/mod.rs
+++ b/scylla/src/transport/locator/mod.rs
@@ -275,7 +275,7 @@ impl<'a> ReplicaSet<'a> {
                 datacenter,
             } => replicas
                 .iter()
-                .filter(|node| node.datacenter.as_deref() == Some(datacenter))
+                .filter(|node| node.datacenter.as_deref() == Some(*datacenter))
                 .count(),
             ReplicaSetInner::ChainedNTS {
                 datacenter_repfactors,
@@ -317,7 +317,7 @@ impl<'a> ReplicaSet<'a> {
                     datacenter,
                 } => replicas
                     .iter()
-                    .filter(|node| node.datacenter.as_deref() == Some(datacenter))
+                    .filter(|node| node.datacenter.as_deref() == Some(*datacenter))
                     .nth(index),
                 ReplicaSetInner::ChainedNTS {
                     datacenter_repfactors,
@@ -466,7 +466,7 @@ impl<'a> Iterator for ReplicaSetIterator<'a> {
             } => {
                 while let Some(replica) = replicas.get(*idx) {
                     *idx += 1;
-                    if replica.datacenter.as_deref() == Some(datacenter) {
+                    if replica.datacenter.as_deref() == Some(*datacenter) {
                         return Some(replica);
                     }
                 }


piodul:
AFAIK the standard library uses the `hashbrown` crate internally to implement the standard hashmap and hashset. Perhaps enabling the `rkyv` feature affects the std crate? The code that doesn't compile doesn't seem to directly interact with `std::collections::HashMap` or `std::collections::HashSet`, though.

oeb25:
_Apologizes for a lot of details, feel free to skip to the end._

Did some more digging and tried to replicate the impl's resulting from enabling the `rkyv` feature in `hashbrown` but could not reproduce the error. ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f50093de7bffb6b21d0eda4d8eb2aa3))

Then tried multiple permutations of comparing `Option<&str>` with `Option<&&str>`, and found something interesting. ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=76a76fc49735ad8ba68a4485850e7332))

```rs
// Ok
fn test_1(a: Option<String>, b: &&str) -> bool {
    a.as_deref() == Some(b)
}
// Errors
fn test_2(a: Option<String>, b: &&str) -> bool {
    let c = Some(b);
    a.as_deref() == c
}
// Errors
fn test_3(a: Option<String>, b: &&str) -> bool {
    let c: Option<&&str> = Some(b);
    a.as_deref() == c
}
// Ok
fn test_4(a: Option<String>, b: &&str) -> bool {
    let c: Option<&str> = Some(b);
    a.as_deref() == c
}
```

This seems to indicate that the inference for performing the dereference in `Some(b)` requires the context of `PartialEq<Option<&str>>`. `test_2` shows moving the construction of `Some(b)` out of the `==` loses inference of `&str`. `test_4` shows we can introduce the dereference again by explicitly stating the target type.

I then tried adding `hashbrown` with the `rkyv` feature, but that doesn't error by itself, only when I add `scylla` do I get the error inside of `scylla` again. However, if I set `scylla = { git = "https://github.com/oeb25/scylla-rust-driver", branch = "double-ref-str-deref" }`, then I get none of the original errors, and surprisingly `test_1` doesn't error either.

Turns out you have to trigger compilation of `hashbrown` for the error to appear, so adding `let _ = hashbrown::HashSet::<()>::default();` anywhere to the code makes `test_1` error.

I unsuccessfully tried to reproduce it in a single crate without deps, but failed. ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bc176270af107cbda93e093ddbeff82d))

However, I did manage to reduce the error to just having `rkyv` as a dependency with the following code:
```rs
use rkyv::AlignedVec;
fn does_not_compile() {
    assert!(Some("") == Some(&""));
}
```
I opened [a repo](https://github.com/oeb25/rkyv-partial-eq-error-reproduction) for this and found https://github.com/rkyv/rkyv/issues/328 which might be related.

---

In any case, I think it is clear that this is not an issue stemming from `scylla`, but it would be nice to have the "workaround" just to prevent such issues from occurring for users of `rkyv`.

Lorak-mmk:
Thanks for the effort. I think we can merge this workaround, but I'd like an issue to be opened in `rkyv` first, and I'm not sure the one you linked is the same problem. Do you think you could open a new one?
Also, I'm surprised that it's possible for rkyv to affect compilation of other crates this way. Maybe I'm just not seeing something, but to me it looks like a possible bug in rustc?

oeb25:
Yeah I opened an issue on rkyv! https://github.com/rkyv/rkyv/issues/434

And also managed to track down which `impl PartialEq` caused the error. This is a minimized example:

```rs
struct ArchivedOption<T>(T);

impl<T: PartialEq<U>, U> PartialEq<ArchivedOption<T>> for Option<U> {
    fn eq(&self, _: &ArchivedOption<T>) -> bool {
        todo!()
    }
}

fn test() {
    Some("") == Some(&"");
}
```

You can even skip one of the generics:

```rs
impl<T: PartialEq> PartialEq<ArchivedOption<T>> for Option<T> {
    fn eq(&self, _other: &ArchivedOption<T>) -> bool {
        todo!()
    }
}
```

Lorak-mmk (approved):

Issue #844, opened 2023-10-17
Error when running examples/compare-tokens.rs and examples/custom_deserialization.rs


nsipplswezey:
When running the `compare-tokens` and `custom_deserialization` examples locally using the following:

`SCYLLA_URI="127.0.0.1:9042" cargo run --example compare-tokens` 
`SCYLLA_URI="127.0.0.1:9042" cargo run --example custom_deserialization` 

The scylla-rust-driver reports table-related errors such as:

```
scylla-rust-driver % cargo run --example custom_deserialization
    Finished dev [unoptimized + debuginfo] target(s) in 0.38s
     Running `target/debug/examples/custom_deserialization`
Connecting to 127.0.0.1:9042 ...
Error: Database returned an error: The query is syntactically correct but invalid, Error message: Unknown identifier pk
```

```
scylla-rust-driver % cargo run --example compare-tokens        
    Finished dev [unoptimized + debuginfo] target(s) in 0.29s
     Running `target/debug/examples/compare-tokens`
Connecting to 127.0.0.1:9042 ...
Error: Database returned an error: The query is syntactically correct but invalid, Error message: Exception while binding column pk: marshaling error: Validation failed for type org.apache.cassandra.db.marshal.Int32Type: got 8 bytes

```

After taking a look it seems that these tests use the same simple `ks` for keyspace and `ks.t` for table naming convention for keyspace and table creation. 

```
"CREATE TABLE IF NOT EXISTS ks.t (pk bigint primary key)",
"INSERT INTO ks.t (pk) VALUES (?)"
"INSERT INTO ks.t (pk) VALUES (?)"
"SELECT token(pk) FROM ks.t where pk = {}"
"CREATE TABLE IF NOT EXISTS ks.t (pk int PRIMARY KEY, v text)"
"INSERT INTO ks.t (pk, v) VALUES (1, 'asdf')"
```

When running against a local Scylla cluster, `ks.t` persists between example runs causing examples to run inconsistently. These two examples would error in different ways depending on when they were run... and dropping `ks.t` as needed resolved the issue. This issue also persisted when running examples in my own CI environment on my own `scylla-rust-driver` fork.

nsipplswezey:
Proposed fix/improvement: https://github.com/scylladb/scylla-rust-driver/pull/846

Issue #839, opened 2023-10-16
Run examples in the CI


cvybhu:
The CI checks that all of the code examples compile, but it doesn't check whether they actually work.
The code is constantly evolving, and things that used to work in the past might stop working at some point, without anyone noticing.

In https://github.com/scylladb/scylla-rust-driver/issues/835 a user reported that one of the example doesn't works. It fails because the specified quorum is invalid for this type of request. This should've been caught in the PR that introduced this example, but wasn't.

It would be good to run all of the examples in the CI. This will help us catch all breakages.

cvybhu:
I don't know any simple way to run all of the examples with a single command.
There's an issue about it: https://github.com/rust-lang/cargo/issues/8356

We can just list every example as a separate step in the workflow, there aren't that many of them.

piodul:
Some of the examples cannot be just run like you would run a test - some of them require some CLI options (`allocations.rs`), user input (`cqlsh-rs.rs`) or specific environment (`serverless.rs`).

IMO those tests that don't need any user input should just be integration tests or maybe examples in docstrings (and `cargo test` runs those). Most of the existing examples could be treated in that way. Not sure about the others, but we could consider what to do with them on a per-case basis.

nsipplswezey:
Since I'm the user who reported an error in one of the examples, I figured I'd open a PR https://github.com/scylladb/scylla-rust-driver/pull/845 taking a look at running examples in CI. After running the examples in CI, I agree with @piodul that the examples aren't all consistent enough to run like you would run a test. Along with the examples @piodul highlighted, I just opened this issue https://github.com/scylladb/scylla-rust-driver/issues/844 that I think is caused by two examples that use the same table naming conventions that cause the other to error when run against the same scylladb cluster. Also the `tls` example is already in CI. And the `cloud` example might need a review of its config credentials. 

I think a per-case basis certainly makes sense. Hope this little bit of work is helpful.

PR #845, opened 2023-10-17
add workflow to run examples in ci

nsipplswezey:
https://github.com/scylladb/scylla-rust-driver/issues/839 suggested putting examples into into CI. I took a look into doing so on my own fork. As @piodul pointed out, these examples are not all fundamentally similar. This PR excludes the `cloud` and `tls` examples since it seems like the `cloud` credentials may need to be reviewed and the `tls` example already runs in CI against a specific `tls` image.

Furthermore, when running all examples in series, tables persist between example runs causing errors when tables share names. See https://github.com/scylladb/scylla-rust-driver/issues/844

Possible fix for https://github.com/scylladb/scylla-rust-driver/issues/839

## Pre-review checklist

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,0 +1,87 @@
+# This workflow runs all driver examples to ensure that they run without errors
+name: Examples
+
+on:
+  push:
+    branches:
+    - main
+    - 'branch-*'
+  pull_request:
+    branches:
+    - main
+    - 'branch-*'
+
+env:
+  CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


Lorak-mmk:
Sorry for the long wait.
Generally looks good to me, assuming that CI passes. I see that it didn't, but the logs already expired.
Could you rebase on current main, drop the merge commit and squash the remaining 2?
Then CI will run again and we'll see if everything works.

wprzytula:
@nsipplswezey What's the status of this?

PR #846, opened 2023-10-17
Explicit table naming in examples to prevent errors from shared tables across examples

nsipplswezey:
The following examples use/reuse the following table name and schema:

`"CREATE TABLE IF NOT EXISTS ks.t (a int, b int, c text, primary key (a, b))",`

`basic`
`execution_profile`
`query_history`
`schema-agreement`
`select-paging`
`speculative-execution`
`tls`

The following examples currently use/reuse the `ks.t` table, with a different schema:
`custom_deserializion` uses `"CREATE TABLE IF NOT EXISTS ks.t (pk int primary key, v text)"`
`compare-tokens` uses `"CREATE TABLE IF NOT EXISTS ks.compare_tokens_example (pk bigint primary key)"`

Depending on what order the examples are run/rerun locally, the examples can create inconsistent table schemas for `ks.t` that cause errors in the examples. 

This PR changes the keyspace name in all examples to `examples_ks` and uses the example file name as the table name for all example tables. This should address any errors caused by tables that are shared between examples.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/844
## Pre-review checklist

- [-] I have adjusted the documentation in `./docs/source/`.

### DIFF
--- a/examples/auth.rs
+++ b/examples/auth.rs
@@ -14,9 +14,9 @@ async fn main() -> Result<()> {
         .await
         .unwrap();
 
-    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await.unwrap();
+    session.query("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await.unwrap();
     session
-        .query("DROP TABLE IF EXISTS ks.t;", &[])
+        .query("DROP TABLE IF EXISTS examples_ks.auth;", &[])
         .await
         .unwrap();
 
--- a/examples/cloud.rs
+++ b/examples/cloud.rs
@@ -16,10 +16,10 @@ async fn main() -> Result<()> {
         .await
         .unwrap();
 
-    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}",
+    session.query("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}",
                     &[]).await.unwrap();
     session
-        .query("DROP TABLE IF EXISTS ks.t;", &[])
+        .query("DROP TABLE IF EXISTS examples_ks.cloud;", &[])
         .await
         .unwrap();
 
--- a/examples/allocations.rs
+++ b/examples/allocations.rs
@@ -131,12 +131,12 @@ async fn main() -> Result<()> {
     let session: Session = SessionBuilder::new().known_node(args.node).build().await?;
     let session = Arc::new(session);
 
-    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await?;
+    session.query("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await?;
     session.await_schema_agreement().await.unwrap();
 
     session
         .query(
-            "CREATE TABLE IF NOT EXISTS ks.alloc_test (a int, b int, c text, primary key (a, b))",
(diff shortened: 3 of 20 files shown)


Lorak-mmk:
This is going to be a problem again when someone adds new example with conflicting schema - even running examples in CI is not guaranteed to catch that.
Maybe a better option than changing `custom_deserializion` and `compare-tokens`  would be to change all examples to use table name corresponding to their name, and change keyspace name to e.g. `examples`?

nsipplswezey:
Makes sense to me. Updated PR to match the suggestion.
I used the name `examples_ks` for key spaces because it seemed clearer to me in the examples. I couldn't find any strong suggestions for naming conventions, so if `examples` or some other name is better, it can be changed.

All examples except `cloud` run without error both on my local and in CI on my fork https://github.com/nsipplswezey/scylla-rust-driver/actions/runs/6551696037/job/17793439625.

Lorak-mmk:
Code changes look fine to me. Please update PR description (because it no longer changes just `custom_deserializion` and `compare-tokens`) and put more information in the commit message so that future developers don't need to open PR to see motivation for the change ( https://github.com/scylladb/scylla-rust-driver/blob/main/CONTRIBUTING.md )

Lorak-mmk:
Yep, could you rebase on main, as #847  is fixed now?

Lorak-mmk (approved):
Looks fine to me now, I think we can merge after rebasing on main.

examples/auth.rs
```diff
@@ -14,9 +14,9 @@ async fn main() -> Result<()> {
         .await
         .unwrap();
 
-    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await.unwrap();
+    session.query("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await.unwrap();
     session
-        .query("DROP TABLE IF EXISTS ks.t;", &[])
+        .query("DROP TABLE IF EXISTS examples_ks.auth;", &[])
```
Lorak-mmk, line 19: Hmm... Previously thix example created (if not exists) the same keyspace as other tests, and then dropped (again, conditionally) shared table.
Now the table is not shared, so no other example can create it, dropping it makes not much sense.
Maybe to verify that session really works, let's remove keyspace / table creation / deletion and just perform some query like `SELECT * FROM system.local`?
wdyt @piodul

examples/cloud.rs
```diff
@@ -16,10 +16,10 @@ async fn main() -> Result<()> {
         .await
         .unwrap();
 
-    session.query("CREATE KEYSPACE IF NOT EXISTS ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}",
+    session.query("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}",
                     &[]).await.unwrap();
     session
-        .query("DROP TABLE IF EXISTS ks.t;", &[])
+        .query("DROP TABLE IF EXISTS examples_ks.cloud;", &[])
         .await
```
Lorak-mmk, line 23: The same comment as with auth example

examples/auth.rs
Lorak-mmk: Alternatively, create a table and insert some data into it

piodul: An example can be run more than once and might expect the table to be in an empty state. I think it still makes sense to drop the table at the beginning.

Lorak-mmk: What I meant is that this example does 2 queries: CREATE IF NOT EXISTS and then DROP IF EXISTS - which doesn't really showcase that session actually works, like inserting / selecting / printing data would.
But it's hard for me to imagine scenario where those queries don't return error in a non-functioning session, so I guess it's ok.

piodul: I agree that the statements that this sample executes are weird, but IIUC they do test that the session works - if the credentials were wrong, then we wouldn't be able to establish any connection to the cluster and wouldn't be able to send anything. We are using superuser credentials, so this example doesn't test authorization but rather authentication - we might consider extending it so that it creates some users and verifies that permissions work.

In any case, those issues are out of scope of the PR.

piodul (approved):
I've rebased the pull request - there were conflicts in `cql-time-types.rs` which I had to resolve. Let's wait for the CI to complete.

LGTM from my side - @Lorak-mmk, please approve and merge at your convenience.

PR #853, opened 2023-10-20
benchmarks/CI: make benchmarks compile and build them in CI

piodul:
The recent fix that changed the `types::write_short` function to use u16 instead of i16 broke one of our benchmarks - it tried to call that function with a negative number. The benchmarks aren't compiled in our CI, so it didn't catch it.

In order to prevent it from happening in the future, the `rust` workflow now uses the `--all-targets` flag to build and clippy-check the benchmarks as well.

## Pre-review checklist


piodul:
@Lorak-mmk review ping

Lorak-mmk (approved):

Issue #801, opened 2023-08-25
Serialization refactor: add new serialization traits


piodul:
Sub-task of #463.

Let's use the convention introduced in the not-yet-done deserialization refactor and let's name the new traits `SerializeCql` and `SerializeRow`:

```rust
pub trait SerializeRow {
    fn serialize(&self, ctx: &RowSerializationContext<'_>, out: &mut impl std::io::Write) -> Result<(), std::io::Error>;
}

pub trait SerializeCql {
    fn serialize(&self, typ: &CqlType, out: &mut impl std::io::Write) -> Result<(), std::io::Error>;
}
```

Instead of `std::io::Write`, perhaps we could use a different trait that abstracts away appending data to a buffer (`bytes::BufMut`?). Using a trait instead of e.g. `&mut Vec<u8>` will let us to very easily implement size hints like here: https://github.com/scylladb/scylla-rust-driver/pull/582 based on the existing `serialize()` method.

The `RowSerializationContext` should contain information about the bind markers: column names and their types. Later, we may add more stuff there.

As this change will affect all `Session::query` and `Session::execute` calls which should be ubiquitous in the user codebase, we should make it as easy as possible to migrate to the new API, preferably step by step.

Current serialization API is heavily based on `SerializedValues` type which is an untyped container for serialized values. `Session::{execute,query}` receive the argument list as an `impl ValueList` type - which is just a trait that allows to convert the type to `SerializedValues`. In general, we should implement `SerializeRow` for each type that implements `ValueList`, which also includes `SerializedValues`.

- Calls like `session.execute(&prepared, (1, 2, 3))` should work automatically out of the box as we will implement `SerializeRow` for types that implement `SerializedValues`, and `SerializeCql` for types that implement `Value`. This should be the most common case.
- In a generic context like `session.execute(&prepared, generic_list)` where `generic_list: impl ValueList`, the piece of code should be adjusted so that `generic_list.serialized()?` is called instead - which will convert to `SerializedValues` which _does_ implement the new `SerializedRow`.
- For the case of user impls of `ValueList` and `Value`, we can provide macros (not necessarily procedural) that generate a `SerializedRow`/`SerializedCql` implementation based on the existing `ValueList`/`Value`.

It will be necessary to update some examples and documentation, but I don't expect it to be much work compared to the deserialization refactor.
```[tasklist]
### Tasks
- [ ] https://github.com/scylladb/scylla-rust-driver/issues/821
[...]

PR #855, opened 2023-10-26
types: serialize: constrain the new serialization traits to make them easier and safer to use

piodul:
Currently, `SerializeRow` and `SerializeCql` traits are just given a mutable reference to a Vec<u8> and asked to append their CQL representation to the end. While simple, there are some issues with the interface:

- The serialize method has access to the serialized representation of the values that were appended before it. It's not necessary for a correct implementation to have access to it.
- Implementors technically can append any byte sequence to the end, but actually are expected to produce a CQL [value] containing the serialized value.

While the `SerializeRow` and `SerializeCql` traits are not generally meant to be manually implemented by the users, we can make the interface easier to use and harder to misuse by making it append-only, restricting what the users are allowed to append and requiring the users to append anything by using a dash of type-level magic.

Introduce `RowWriter` and `CellWriter` traits which satisfy the above wishes and constraints, and pass them instead of Vec<u8> in `SerializeRow` and `SerializeCql`.

The new traits have two implementations - a Vec<u8> backed one that actually appends the bytes given to it, and a usize-backed one which just measures the length of the output without writing anything. Passing the latter before doing the actual serialization will allow to preallocate the right amount of bytes and then serialize without reallocations. It should be measured whether the reallocation cost always outweighs the calculation cost before implementing this optimization.

Refs: https://github.com/scylladb/scylla-rust-driver/issues/801

### DIFF
--- a/scylla-cql/src/frame/types.rs
+++ b/scylla-cql/src/frame/types.rs
@@ -104,6 +104,23 @@ impl From<std::array::TryFromSliceError> for ParseError {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum RawValue<'a> {
+    Null,
+    Unset,
+    Value(&'a [u8]),
+}
+
+impl<'a> RawValue<'a> {
+    #[inline]
+    pub fn as_value(&self) -> Option<&'a [u8]> {
+        match self {
+            RawValue::Value(v) => Some(v),
+            RawValue::Null | RawValue::Unset => None,
+        }
+    }
+}
+
 fn read_raw_bytes<'a>(count: usize, buf: &mut &'a [u8]) -> Result<&'a [u8], ParseError> {
     if buf.len() < count {
         return Err(ParseError::BadIncomingData(format!(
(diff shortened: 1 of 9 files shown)


piodul:
@Lorak-mmk review ping

piodul:
v2:
- More docstrings
- RowWriter now does not prepend what it wrote with a u16 value indicating the count of the values written
- Some items that were accidentally imported via `_macro_internal` path in tests are now imported through a proper path

piodul:
v2.1:
- Rebased
- Addressed clippy's complaints about needless `std::mem::drop`

Lorak-mmk (approved):

Lorak-mmk:
Actually I have one question: is mod.rs the best place for those new traits and structs? I don't really know Rust conventions in this matter, but it is quite a lot of code, wouldn't it be better to make a new file for it?

piodul:
> Actually I have one question: is mod.rs the best place for those new traits and structs? I don't really know Rust conventions in this matter, but it is quite a lot of code, wouldn't it be better to make a new file for it?

Makes sense, I can move the `CellWriter` and friends to a separate module.

piodul:
v3.1: fixed links in the docs

Lorak-mmk (approved):
I think there is nothing blocking this, do you think we can merge it?

piodul:
> I think there is nothing blocking this, do you think we can merge it?

Yes, I'll go ahead and merge it.

PR #856, opened 2023-10-26
Commit renamed Cargo.lock file

Lorak-mmk:
It will be used by min_rust workflow to build and check with our MSRV to prevent errors from packages updating their MSRV.

## Pre-review checklist


mykaul:
Do we need something similar for Scylla itself?

piodul:
> Do we need something similar for Scylla itself?

No, Scylla doesn't have this concern because we use a frozen toolchain and already commit `Cargo.lock` (i.e. we pin particular versions of the dependencies). We need this for the rust driver because we support a range of versions of the rust toolchain.

piodul (approved):

.github/workflows/rust.yml
```diff
@@ -60,6 +60,8 @@ jobs:
         rustup override set ${{ env.rust_min }}
     - name: Print Rust version
       run: rustc --version
+    - name: Use MSRV Cargo.lock
+      run: mv Cargo.lock.msrv Cargo.lock
```
piodul, line 64: One more request: could you add the [`--frozen`](https://doc.rust-lang.org/cargo/commands/cargo-check.html#manifest-options) flag to the invocations of `cargo check` in the `min_rust` job? This should assert that cargo doesn't modify the lockfile due to some reason not accounted by us.

piodul (approved):

PR #857, opened 2023-10-31
Bump `num_enum` to `0.6..=0.7`

oeb25:
This PR bumps `num_enum` from `0.5` and `0.6` to `0.6..=0.7`. The motivation for this is that `num_enum@0.7` transitively allows for `toml_edit@0.20`.

Unfortunately another dependency of `scylla`, `ntest`, still depends on `toml_edit@0.19` in its current release, but has bumped to `0.20` on master. `ntest` is however only a `dev-dependency`, so I still think this will have upside for downstream users of scylla. 

---

## Pre-review checklist

### DIFF
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -916,32 +916,11 @@ dependencies = [
 
 [[package]]
 name = "num_enum"
-version = "0.5.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1f646caf906c20226733ed5b1374287eb97e3c2a5c227ce668c1f2ce20ae57c9"
-dependencies = [
- "num_enum_derive 0.5.11",
-]
-
-[[package]]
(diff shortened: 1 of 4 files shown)


piodul:
I'm not sure it is a good idea to allow for two major versions. This will interact weirdly with the cargo resolver and could cause issues when used in a project where some dependencies (or the project itself) specifies `num_enum = 0.6` and others `num_enum = 0.7`. See the recommendation here: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-requirements.

I'd rather stick to only one version (bumping to just 0.6 sounds like less of a hassle).

In the future, I'd like to get rid of the dependency altogether as it is not stable (<1.0, problematic for making this crate's API stable) but appears in the public interface (for Consistency and SerialConsistency) and doesn't offer too much functionality (we can implement TryFrom conversions manually).

PR #858, opened 2023-11-02
Switch Session to new serialization traits

Lorak-mmk:
This PR is based on https://github.com/scylladb/scylla-rust-driver/pull/855 and is the main part of serialization refactor - switching public API to new trait.
The last commit makes the actual change. Previous ones are different changes that were necessary (or helpful) to make the switch.

There are still some things to do:
- [x] Comments on new traits / structs
- [ ] Documentation
- [ ] Rename SerializeCQL / SerializeRow / NewSerializedValues to some sensible names
- [ ] (Maybe) Check who calls query_filter_keyspace_name and use prepared statements in those places to avoid unnecessary prepares.

But the code changes are pretty much ready for review.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types/serialize/row.rs
+++ b/scylla-cql/src/types/serialize/row.rs
@@ -680,7 +680,6 @@ impl SerializedValues {
     }
 
     /// Creates value list from the request frame
-    #[allow(dead_code)]
     pub(crate) fn new_from_frame(buf: &mut &[u8]) -> Result<Self, ParseError> {
         let values_num = types::read_short(buf)?;
         let values_beg = *buf;
@@ -697,7 +696,6 @@ impl SerializedValues {
     }
 
     // Temporary function, to be removed when we implement new batching API (right now it is needed in frame::request::mod.rs tests)
-    #[allow(dead_code)]
     pub fn to_old_serialized_values(&self) -> LegacySerializedValues {
         let mut frame = Vec::new();
         self.write_to_request(&mut frame);
--- a/scylla-cql/src/types/serialize/row.rs
+++ b/scylla-cql/src/types/serialize/row.rs
@@ -4,14 +4,19 @@ use std::fmt::Display;
 use std::hash::BuildHasher;
 use std::{collections::HashMap, sync::Arc};
 
+use bytes::BufMut;
 use thiserror::Error;
 
+use crate::frame::frame_errors::ParseError;
+use crate::frame::response::result::ColumnType;
 use crate::frame::response::result::PreparedMetadata;
+use crate::frame::types;
+use crate::frame::value::SerializeValuesError;
 use crate::frame::value::{LegacySerializedValues, ValueList};
 use crate::frame::{response::result::ColumnSpec, types::RawValue};
 
 use super::value::SerializeCql;
-use super::{RowWriter, SerializationError};
+use super::{CellWriter, RowWriter, SerializationError};
 
 /// Contains information needed to serialize a row.
 pub struct RowSerializationContext<'a> {
@@ -573,9 +578,160 @@ pub enum ValueListToSerializeRowAdapterError {
     NoBindMarkerWithName { name: String },
 }
 
+/// A buffer containing already serialized values.
+///
+/// It is not aware of the types of contained values,
+/// it is basically a byte buffer in the format expected by the CQL protocol.
+/// Usually there is no need for a user of a driver to use this struct, it is mostly internal.
+/// The exception are APIs like `ClusterData::compute_token` / `ClusterData::get_endpoints`.
+/// Allows adding new values to the buffer and iterating over the content.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SerializedValues {
+    serialized_values: Vec<u8>,
+    element_count: u16,
+}
+
+impl SerializedValues {
+    pub const fn new() -> Self {
+        SerializedValues {
+            serialized_values: Vec::new(),
+            element_count: 0,
+        }
+    }
+
+    /// A const empty instance, useful for taking references
+    pub const EMPTY: &'static SerializedValues = &SerializedValues::new();
+
+    pub fn from_serializable<T: SerializeRow>(
+        ctx: &RowSerializationContext,
+        row: &T,
+    ) -> Result<Self, SerializationError> {
+        let mut data = Vec::new();
+        let element_count = {
+            let mut writer = RowWriter::new(&mut data);
+            row.serialize(ctx, &mut writer)?;
+            match writer.value_count().try_into() {
+                Ok(n) => n,
+                Err(_) => {
+                    return Err(SerializationError(Arc::new(
+                        SerializeValuesError::TooManyValues,
+                    )))
+                }
+            }
+        };
+
+        Ok(SerializedValues {
+            serialized_values: data,
+            element_count,
+        })
+    }
+
+    #[inline]
+    pub fn is_empty(&self) -> bool {
+        self.element_count() == 0
+    }
+
+    #[inline]
+    pub fn iter(&self) -> impl Iterator<Item = RawValue> {
+        SerializedValuesIterator {
+            serialized_values: &self.serialized_values,
+        }
+    }
+
+    #[inline]
+    pub fn element_count(&self) -> u16 {
+        // We initialize first two bytes in new() and BufBackedRowWriter does too,
+        // so this unwrap is safe
+        self.element_count
+    }
+
+    #[inline]
+    pub fn buffer_size(&self) -> usize {
+        self.serialized_values.len()
+    }
+
+    pub(crate) fn write_to_request(&self, buf: &mut impl BufMut) {
+        buf.put_u16(self.element_count);
+        buf.put(self.serialized_values.as_slice())
+    }
+
+    /// Serializes value and appends it to the list
+    pub fn add_value<T: SerializeCql>(
+        &mut self,
+        val: &T,
+        typ: &ColumnType,
+    ) -> Result<(), SerializationError> {
+        if self.element_count() == u16::MAX {
+            return Err(SerializationError(Arc::new(
+                SerializeValuesError::TooManyValues,
+            )));
+        }
+
+        let len_before_serialize: usize = self.serialized_values.len();
+
+        let writer = CellWriter::new(&mut self.serialized_values);
+        if let Err(e) = val.serialize(typ, writer) {
+            self.serialized_values.resize(len_before_serialize, 0);
+            Err(e)
+        } else {
+            self.element_count += 1;
+            Ok(())
+        }
+    }
+
+    /// Creates value list from the request frame
+    #[allow(dead_code)]
+    pub(crate) fn new_from_frame(buf: &mut &[u8]) -> Result<Self, ParseError> {
+        let values_num = types::read_short(buf)?;
+        let values_beg = *buf;
+        for _ in 0..values_num {
+            let _serialized = types::read_value(buf)?;
+        }
+
+        let values_len_in_buf = values_beg.len() - buf.len();
+        let values_in_frame = &values_beg[0..values_len_in_buf];
+        Ok(SerializedValues {
+            serialized_values: values_in_frame.to_vec(),
+            element_count: values_num,
+        })
+    }
+
+    // Temporary function, to be removed when we implement new batching API (right now it is needed in frame::request::mod.rs tests)
+    #[allow(dead_code)]
+    pub fn to_old_serialized_values(&self) -> LegacySerializedValues {
+        let mut frame = Vec::new();
+        self.write_to_request(&mut frame);
+        LegacySerializedValues::new_from_frame(&mut frame.as_slice(), false).unwrap()
+    }
+}
+
+impl Default for SerializedValues {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Clone, Copy)]
+pub struct SerializedValuesIterator<'a> {
+    serialized_values: &'a [u8],
+}
+
+impl<'a> Iterator for SerializedValuesIterator<'a> {
+    type Item = RawValue<'a>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.serialized_values.is_empty() {
+            return None;
+        }
+
+        Some(types::read_value(&mut self.serialized_values).expect("badly encoded value"))
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use crate::frame::response::result::{ColumnSpec, ColumnType, TableSpec};
+    use crate::frame::types::RawValue;
     use crate::frame::value::{LegacySerializedValues, MaybeUnset, ValueList};
     use crate::types::serialize::RowWriter;
 
@@ -584,6 +740,7 @@ mod tests {
         BuiltinTypeCheckErrorKind, RowSerializationContext, SerializeCql, SerializeRow,
     };
 
+    use super::SerializedValues;
     use scylla_macros::SerializeRow;
 
     fn col_spec(name: &str, typ: ColumnType) -> ColumnSpec {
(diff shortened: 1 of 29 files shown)


scylla-cql/src/types/serialize/row.rs
```diff
@@ -200,3 +206,127 @@ mod tests {
...
+/// Allows adding new Values and iterating over serialized ones
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct NewSerializedValues {
+    serialized_values: Vec<u8>,
+}
+
+impl NewSerializedValues {
+    pub fn new() -> Self {
+        let mut buf = Vec::new();
+        buf.extend_from_slice(&0u16.to_be_bytes());
```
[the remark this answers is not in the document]
piodul: It's a little sad that `NewSerializedValues::new()` must allocate - this prevents us from making the method `const` and forces us to allocate in cases where we need to pass an empty list of values.

Perhaps we could rework how `BufBackedRowWriter` works and let it count written values in a variable and not in the buffer.

No need to act upon it now, this can be left for a later improvement.

scylla/benches/benchmark.rs
```diff
@@ -40,23 +40,49 @@ fn types_benchmark(c: &mut Criterion) {
...
+        .unwrap();
+    serialized_simple_pk_long_column
+        .add_value(&16_i32, &ColumnType::Int)
+        .unwrap();
+    serialized_simple_pk_long_column
+        .add_value(
+            &String::from_iter(std::iter::repeat('.').take(2000)),
+            &ColumnType::Text,
+        )
+        .unwrap();
```
piodul, line 60: Instead of adding values one by one, why not use `NewSerializedValues::from_serializable()` and pass a tuple to it, similarly to how it is done now? It should make it shorter.

piodul (commented):
The last commit is quite large and hard to review. I think the changes there should be done in a more gradual fashion, utilizing the `into_old_serialized_values` method more.

scylla/benches/benchmark.rs
Lorak-mmk: Because then we need serialization context, which is created from prepared statement - and this benchmark doesn't even connect to Scylla as far as I can tell, so I have no way to create a prepared statement.

Lorak-mmk: I could add a way to craft context manually, but I doubt the resulting benchmark code would be simpler / shorter.

scylla-cql/src/types/serialize/row.rs
```diff
@@ -200,3 +206,127 @@ mod tests {
...
+        T::preliminary_type_check(ctx)?;
+        row.serialize(ctx, &mut writer)?;
+        drop(writer);
+        Ok(NewSerializedValues {
+            serialized_values: data,
+        })
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
```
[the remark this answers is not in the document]
Lorak-mmk: Actually no, this looks fine - len() returns the element count and size() returns buffer size. I'm still going to rename the methods because it's confusing.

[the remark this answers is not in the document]
piodul: I agree that the names are confusing and it makes sense to rename.

scylla/benches/benchmark.rs
piodul: Ok, I see, in that case I agree it's not worth it. Creating a context manually would require adding more irrelevant details compared to what we have now (e.g. column names).

Lorak-mmk:
> Also, the improvements in the proxy related to the detection of control connection could be sent in a separate PR in order to merge them more quickly.

Opened https://github.com/scylladb/scylla-rust-driver/pull/863

Lorak-mmk:
Ooops, misslick

scylla/src/transport/connection.rs
```diff
@@ -4,6 +4,7 @@ use scylla_cql::errors::TranslationError;
 use scylla_cql::frame::request::options::Options;
 use scylla_cql::frame::response::Error;
 use scylla_cql::frame::types::SerialConsistency;
+use scylla_cql::frame::value::SerializedValues;
```
piodul: > connection.rs: Don't allow queries with values
>
> After serialization refactor it will be impossible to perform unprepared
> query with values, because serializing values will require knowing
> column types.
> 
> In order to make refactor easier and better split responsibility, this
> commit removes `values` arguments from `Connection` methods, so that
> it is callers responsibility to prepare the query if necessary.

I'm not sure if I see why it makes the remaining refactor easier. It looks like you are simplifying the API of `Connection` but this is just pushing this concern to the caller. This requires modifying lots of places in the code, including tests and `query_filter_keyspace_name` which manually checks whether values is empty and dispatches to `query` or `execute` anyway.

I wonder whether it wouldn't be just simpler to drop this commit and then rework the last one, but I may be missing something.

scylla-cql/src/types/serialize/row.rs
```diff
@@ -125,13 +131,150 @@ pub enum ValueListToSerializeRowAdapterError {
     NoBindMarkerWithName { name: String },
 }
 
+/// A container for already serialized values. It is not aware of the types of contained values,
+/// it is basically a byte buffer in the format expected by the CQL protocol.
+/// Usually there is no need for a user of a driver to use this struct, it is mostly internal.
```
piodul, line 136: Nit: conventionally, a rust docstring starts with a single summary sentence, followed by a blank line and a more detailed description.

scylla-cql/src/types/serialize/row.rs
```diff
@@ -125,13 +131,150 @@ pub enum ValueListToSerializeRowAdapterError {
...
+            writer.value_count()
+        };
+
+        Ok(NewSerializedValues {
+            serialized_values: data,
+            element_count,
+        })
+    }
+
+    pub fn is_empty(&self) -> bool {
```
piodul, line 171: `#[inline]` pls

scylla-cql/src/types/serialize/row.rs
```diff
@@ -125,13 +131,150 @@ pub enum ValueListToSerializeRowAdapterError {
...
+
+    pub fn element_count(&self) -> u16 {
+        // We initialize first two bytes in new() and BufBackedRowWriter does too,
+        // so this unwrap is safe
+        self.element_count
+    }
+
+    pub fn buffer_size(&self) -> usize {
+        self.serialized_values.len()
+    }
```
piodul, line 189: Those can be inline, too. Remember that non-generic functions are not automatically inlined across crates without the annotation. The functions here are quite trivial, so they would benefit from inlining. The drawback is longer compilation time, but I don't think it should matter as much for such simple functions.

scylla/src/statement/prepared_statement.rs
```diff
@@ -134,9 +137,10 @@ impl PreparedStatement {
     /// [Self::calculate_token()].
     pub fn compute_partition_key(
         &self,
-        bound_values: &SerializedValues,
+        bound_values: &impl SerializeRow,
```
piodul, line 140: This used to take `SerializedValues` as an argument, why does it take `impl SerializeRow` now? Shouldn't it rather take `NewSerializedValues` instead? This should remove the need to add `PartitionKeyError::Serialization` variant.

Lorak-mmk: This is a method on `PreparedStatement` and I wanted it to be type safe - it couldn't with `NewSerializedValues`.
Also, `NewSerializedValues` are not convenient for a user, it's much easier to call this method with `SerializeRow`.

scylla/src/transport/connection.rs
Lorak-mmk: My thinking is that we don't want to send values in unprepared queries, and removing such capability from `Connection` is a way to guarantee that. If the only users of this functionality are tests and `query_filter_keyspace_name`, then I think it's not worth it to have this functionality.

I'll look into this matter again - I remember that some changes were less pleasant than I hoped, maybe reverting those changes in connections would help with that.

Lorak-mmk:
- Rebased on current `adjust_interface` branch
- Addressed most of review comments (I marked those as resolved)

scylla/src/statement/prepared_statement.rs
piodul: OK, makes sense.

scylla/src/transport/connection.rs
piodul: > My thinking is that we don't want to send values in unprepared queries, and removing such capability from `Connection` is a way to guarantee that.

What do you mean exactly by "send values in unprepared queries"? We are no longer doing that on the protocol level after https://github.com/scylladb/scylla-rust-driver/pull/806 and https://github.com/scylladb/scylla-rust-driver/pull/812 . If you mean to disallow calling `.query()` with values, then I'm not sure if it makes sense to disallow it for `Connection` but not for `Session` (and AFAIK we didn't decide to do it for `Session` at this point).

Lorak-mmk:
> > [scylla-cql: make SerializationError a newtype instead of alias](https://github.com/scylladb/scylla-rust-driver/pull/858/commits/56930fe245b4462690c0de66755bbd45fb13af66)
> 
> Please send this commit as a separate PR. It will introduce lots of conflicts with my other PRs, so let's merge this first so that I can rebase on top of it.

Created https://github.com/scylladb/scylla-rust-driver/pull/865

scylla/src/transport/connection.rs
Lorak-mmk: Removing this commit opens us up to the problems like following:
`topology::query_filter_keyspace_name` calls `Connection::query_iter`, potentially with values.
`Connection::query_iter` will call `RowIterator::new_for_connection_query_iter`, which will,
for each page, call `Connection::query_with_consistency` - which, if there are values,
prepares the query underneath to serialize them.
Such problems seem pretty hard to notice normally but removing `values` from Connection pretty much prevents them.

Currently, I tried to revert removal of `values` from `Connection` and from `RowIterator::new_for_query`, as it forced me to prepare a query globally in `Session::query_iter` if there are values - now it is again prepared for each sub query.
Doing this created another problem I'm not yet sure how to solve: - if `RowIterator::new_for_query` accepts `values`, how do we get a size of it
for request span? In either variant of `Connection` API, `new_for_query` won't serialize the values, `Connection` will do it. There is no place where `RowIterator` sees serialized values, so it can't record it's size for span.

For a user of `Connection` struct (e.g. struct `Session`), API without `values` args covers all the use cases that the previous one did - perhaps with a bit more code sometimes, as the caller needs to prepare the query explicitly if they want to send values.

One advantage I see of removing `values` from `Connection` is that it makes the API more explicit - additional overhead of preparing queries is not hidden underneath `Connection::query*` methods, so we are more aware of it while writing other code.

piodul: > Removing this commit opens us up to the problems like following: `topology::query_filter_keyspace_name` calls `Connection::query_iter`, potentially with values. `Connection::query_iter` will call `RowIterator::new_for_connection_query_iter`, which will, for each page, call `Connection::query_with_consistency` - which, if there are values, prepares the query underneath to serialize them. Such problems seem pretty hard to notice normally but removing `values` from Connection pretty much prevents them.

This issue is easy to fix - prepare the query if needed in `Connection::query_iter` and then change `RowIterator::new_connection_query_iter` to be `RowIterator::new_connection_execute_iter`.

> Currently, I tried to revert removal of `values` from `Connection` and from `RowIterator::new_for_query`, as it forced me to prepare a query globally in `Session::query_iter` if there are values - now it is again prepared for each sub query.

This sounds like a harder problem because, unlike `Connection::query_iter`, the iterator might switch to another connection when performing a query fails. Perhaps we could be smarter here and prepare the query when `RowIteratorWorker::work` chooses a connection, but doing that sounds like too much work to be worth it, so your current approach looks OK for this one.

> Doing this created another problem I'm not yet sure how to solve: - if `RowIterator::new_for_query` accepts `values`, how do we get a size of it for request span? In either variant of `Connection` API, `new_for_query` won't serialize the values, `Connection` will do it. There is no place where `RowIterator` sees serialized values, so it can't record it's size for span.

This problem should go away if you apply my first suggestion.

> For a user of `Connection` struct (e.g. struct `Session`), API without `values` args covers all the use cases that the previous one did - perhaps with a bit more code sometimes, as the caller needs to prepare the query explicitly if they want to send values.
> 
> One advantage I see of removing `values` from `Connection` is that it makes the API more explicit - additional overhead of preparing queries is not hidden underneath `Connection::query*` methods, so we are more aware of it while writing other code.

You can look at it from another angle - having to do manual prepare in metadata fetch code sounds like polluting it with unnecessary details which should be handled on a lower level.

----

The examples with `query_iter` convince me that it might be a good idea after all to automatically prepare on the highest layer instead on the lowest, because otherwise unnecessary preparing may occur. Though I'd prefer that `query_iter` API did prepare automatically like I described (I see it as a user API in some sense, where the control connection is the user), there is an issue about changing internal CQL statements to use prepared statements (https://github.com/scylladb/scylla-rust-driver/issues/417) so I suppose having `Connection::execute_iter` right now is acceptable.

To sum up - let's not pursue my original suggestion from my first comment in this thread, I don't think it's worth it anymore.

scylla-cql/src/types/serialize/row.rs
```diff
@@ -126,13 +132,156 @@ pub enum ValueListToSerializeRowAdapterError {
...
 
+/// A buffer containing already serialized values.
+///
+/// It is not aware of the types of contained values,
+/// it is basically a byte buffer in the format expected by the CQL protocol.
+/// Usually there is no need for a user of a driver to use this struct, it is mostly internal.
+/// The exception are APIs like `ClusterData::compute_token` / `ClusterData::get_endpoints`.
+/// Allows adding new values to the buffer and iterating over the content.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct NewSerializedValues {
```
piodul, line 143: `NewSerializedValues` will not be "new" forever, I think it makes sense to just use `SerializedValues` here and rename the old one to `LegacySerializedValues` and eventually remove it.

piodul:
Also, needs rebase after https://github.com/scylladb/scylla-rust-driver/pull/865 was merged.

Lorak-mmk:
I pushed a new version - rebased on main. It also contains commits from https://github.com/scylladb/scylla-rust-driver/pull/851 (rebased on  top of main and my commit that added `is_empty` method - so you can use this to update your PR).
This is not ready - there are still some improvements I want to make.

Problems:
- New macros seem to not work for generic structs. See error messages in CI
- For now I commented out part of session_tests that uses `dyn Value` - I don't like it that SerializeCQL / SerializeRow are not object safe and would like to think / discuss a bit more to see if we can make it object safe.

piodul:
>     * New macros seem to not work for generic structs. See error messages in CI

Right... I'll add it to the PR.

>     * For now I commented out part of session_tests that uses `dyn Value` - I don't like it that SerializeCQL / SerializeRow are not object safe and would like to think / discuss a bit more to see if we can make it object safe.

I have an idea how to make it object safe, I started working on a rework of the API already. I'll try to send something tomorrow.

piodul:
Looks like we are hitting https://github.com/scylladb/scylladb/issues/16329 in CI

Lorak-mmk:
New version:
- Rebased on top of current main, which includes new version of API and new macros
- Split the big commit (the one changing all the APIs to use new serialization methods) into a lot of small ones, to make review, changes and future reading / bisecting easier.
- Renamed `into_old_serialized_values` to `to_old_serialized_values` in `SerializedValues`.
- Removed usage of old traits / macros from examples and many tests

Each commit in the series passes all static checks. I'll run `make ci && cargo build --benches` on each too, and report here when it finishes.

Lorak-mmk:
Hmmm... Book tests are failing, how should I fixed it? I know that you are currently editing the book, so maybe I'll just make a minimal change to fix the error (adding new derive macros)?

piodul:
> @piodul, so maybe I'll just make a minimal change to fix the error (adding new derive macros)?

Yes, please.

piodul:
Thanks for splitting it so thoroughly, it was much easier to see what was happening. I left two nitpicks that should be easily fixable, but otherwise I didn't see any issues. Please update the docs so that we make the CI green and I think it will be alright to merge.

Lorak-mmk:
I think you may have forgotten to send the review - I don't see any comments

piodul (commented):
Yep, I forgot to send the comments, sorry.

scylla/src/transport/connection.rs
```diff
@@ -652,7 +651,7 @@ impl Connection {
             parameters: query::QueryParameters {
                 consistency,
                 serial_consistency,
-                values: Cow::Borrowed(LegacySerializedValues::EMPTY),
+                values: Cow::Owned(SerializedValues::new()),
```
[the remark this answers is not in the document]
Lorak-mmk: I can do this, but I'm curious: why is it better?

scylla/src/transport/session.rs
```diff
@@ -2043,7 +2043,6 @@ impl RequestSpan {
             .record("replicas", tracing::field::display(&ReplicaIps(replicas)));
     }
 
-    #[allow(dead_code)]
```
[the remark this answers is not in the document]
Lorak-mmk: Of course. I thought I did that, must have clicked something wrong...

scylla/src/transport/connection.rs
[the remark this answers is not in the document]
Lorak-mmk: My first thought was performance, but Borrowed has an indirection, while Owned doesnt. So is this because of readability?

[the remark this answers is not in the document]
piodul: I don't have a good argument. It's just more similar to what was there before.

I also think that the destructor of the Cow has less to do in that case (it's dropping a reference which is a no-op, whereas in the other case you have a Vec), however it's pure speculation on my part and without benchmarks it's hard to tell whether it's worth it.

Feel free to disregard the nit if you feel it's not worth the effort.

Lorak-mmk:
Addressed review comments, fixed a book build errors (in a minimal way - I only introduced changes that were strictly necessary, I don't want to expand scope of this PR)

piodul (approved):

Issue #821, opened 2023-10-06
Serialization refactor: implement `SerializeRow` and `SerializeCql` for current implementors of `ValueList` and `Value`


piodul:
Sub-task of https://github.com/scylladb/scylla-rust-driver/issues/801 - see the description for more details.

Issue #813, opened 2023-09-20
`if_lwt_optimisation_mark_offered_then_negotiatied_and_lwt_routed_optimally` may be flaky with Cassandra


piodul:
It failed twice in an unrelated pull request (https://github.com/scylladb/scylla-rust-driver/pull/808).

Link to logs (attempt 2 and 3 failed): https://github.com/scylladb/scylla-rust-driver/actions/runs/6164271508/job/16967689748

Example:

```
---- lwt_optimisation::if_lwt_optimisation_mark_offered_then_negotiatied_and_lwt_routed_optimally stdout ----
Unique name: test_rust_1694796250_3
thread 'lwt_optimisation::if_lwt_optimisation_mark_offered_then_negotiatied_and_lwt_routed_optimally' panicked at 'timeout: the function call took 20000 ms. Max time 20000 ms', scylla/tests/integration/lwt_optimisation.rs:16:1
stack backtrace:
   0:     0x558539fceb41 - std::backtrace_rs::backtrace::libunwind::trace::he648b5c8dd376705
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x558539fceb41 - std::backtrace_rs::backtrace::trace_unsynchronized::h5da3e203eef39e9f
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x558539fceb41 - std::sys_common::backtrace::_print_fmt::h8d28d3f20588ae4c
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x558539fceb41 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hd9a5b0c9c6b058c0
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x558539fff11f - core::fmt::rt::Argument::fmt::h0afc04119f252b53
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/fmt/rt.rs:138:9
   5:     0x558539fff11f - core::fmt::write::h50b1b3e73851a6fe
[...]

PR #862, opened 2023-11-20
serialization: implement new serialization traits for the currently supported types

piodul:
The new `SerializeCql` and `SerializeRow` traits are supposed to eventually replace `Value` and `ValueList`. Currently, the new traits are implemented with a blanket implementation based on the old traits - this was just done in order to parallelize work on adapting the whole codebase to the new traits; the blanket implementations are inefficient and not type safe. This PR gets rid of them and, for each type for which we implement `Value` or `ValueList`, introduces proper implementations of `SerializeCql` and `SerializeRow`.

All types now check whether the CQL type serialized to matches, and most of them serialize exactly the same as they used to with the old traits. However, some slight differences were made in order to utilize the information available to the new traits and make it more robust:

- `CqlValue::UserDefinedType` now doesn't assume that the order of the fields is the same as in the CQL type - it sorts the fields according to the CQL type.
- `impl SerializeRow for BTreeMap/HashMap` doesn't serialize in the "named values" format anymore as it is no longer supported by the new traits. Instead, it uses provided information about the column names to properly order the values before sending them to the database.

Existing tests in `value_tests.rs` are adapted so that they perform serialization using both old and new traits and compare the results (with a small number of exceptions where serialization behavior was changed).

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/821

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types/serialize/row.rs
+++ b/scylla-cql/src/types/serialize/row.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::collections::BTreeMap;
+use std::fmt::Display;
 use std::hash::BuildHasher;
 use std::{collections::HashMap, sync::Arc};
 
@@ -70,12 +71,39 @@ macro_rules! fallback_impl_contents {
     };
 }
 
+macro_rules! impl_serialize_row_for_unit {
+    () => {
+        fn preliminary_type_check(
+            ctx: &RowSerializationContext<'_>,
+        ) -> Result<(), SerializationError> {
+            if !ctx.columns().is_empty() {
+                return Err(mk_typck_err::<Self>(
+                    BuiltinTypeCheckErrorKind::WrongColumnCount {
+                        actual: 0,
+                        asked_for: ctx.columns().len(),
+                    },
+                ));
+            }
+            Ok(())
+        }
+
+        fn serialize<W: RowWriter>(
+            &self,
+            _ctx: &RowSerializationContext<'_>,
+            _writer: &mut W,
+        ) -> Result<(), SerializationError> {
+            // Row is empty - do nothing
+            Ok(())
+        }
+    };
+}
+
 impl SerializeRow for () {
-    fallback_impl_contents!();
+    impl_serialize_row_for_unit!();
 }
 
 impl SerializeRow for [u8; 0] {
-    fallback_impl_contents!();
+    impl_serialize_row_for_unit!();
 }
 
 impl<T: Value> SerializeRow for &[T] {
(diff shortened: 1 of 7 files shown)


piodul:
v1.1: more complete docstrings for the error types

Lorak-mmk:
> 
> Depends on #851

Isn't this the other way around?

piodul:
> > Depends on #851
> 
> Isn't this the other way around?

Yeah, my mistake. Updated the description.

piodul:
v2:
- rebased
- addressed most of the comments that were placed in https://github.com/scylladb/scylla-rust-driver/pull/851 (apart from the one about panicking on overflow)

scylla-cql/src/types/serialize/value.rs
```diff
@@ -29,19 +48,883 @@ pub trait SerializeCql {
...
+            if let Some(err) = err.downcast_ref::<BuiltinSerializationError>() {
+                if err.rust_name != rust_name {
+                    return Arc::new(BuiltinSerializationError {
+                        rust_name,
+                        ..err.clone()
+                    });
+                }
+            }
+        }
+    };
```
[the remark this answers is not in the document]
Lorak-mmk: Oh wait, I didn't read the code properly, sorry.

piodul:
v3:
- adjusted the builder interface to make it safer/easier to handle CQL value size overflows
- adjusted the `impl SerializeCql` blocks to the new interface - new errors are handled by unwraps for fixed sized types, and by properly returning an error for dynamically sized ones
- got rid of the `.expect` calls in map/list/set serialization code

piodul:
Looks like the CI failed due to a known issue: https://github.com/scylladb/scylla-rust-driver/issues/813

scylla-cql/src/types/serialize/row.rs
```diff
@@ -50,22 +55,310 @@ pub trait SerializeRow {
...
     }
 }
 
+impl SerializeRow for SerializedValues {
+    fallback_impl_contents!();
+}
+
+impl<'b> SerializeRow for Cow<'b, SerializedValues> {
+    fallback_impl_contents!();
+}
```
Lorak-mmk, line 269: Is this a implementation created to make transition to new API easier? If so, let's add a TODO / create and issue so that we don't forget to remove it later.

scylla-cql/src/types/serialize/value.rs
```diff
@@ -29,19 +48,940 @@ pub trait SerializeCql {
...
+            )
+        })?;
+    }
+
+    builder
+        .finish()
+        .map_err(|_| mk_ser_err_named(rust_name, typ, BuiltinSerializationErrorKind::SizeOverflow))
+}
 
 pub fn serialize_legacy_value<T: Value, W: CellWriter>(
```
Lorak-mmk, line 986: Is this used anywhere after the changes in this PR? I can't find any references to it. If not, it can be removed

Lorak-mmk (changes requested):

piodul:
v4:
- Rebased
- Implemented two declarative macros that allows to generate a quick and dirty implementation of SerializeRow/SerializeCql based on ValueList/Value

Lorak-mmk (approved):

PR #863, opened 2023-11-20
scylla-proxy: Add option to filter out control connection messages

Lorak-mmk:
Proxy does not allow filtering of control connection messages. Right now this is not a problem, as all queries performed by driver are not prepared, so tests can filter them out by checking query content.

The ongoing serialization refactor changes that and prepares some queries performed automatically by driver, which breaks some tests, as they are not able to differentiate test's queries from internal driver queries.

This PR adds a mechanism to filter out control connection messages by ignoring messages on connections that had any event registered. Control connection is the only one in rust driver that does it, so this should by a reliable heuristic.

## Pre-review checklist

### DIFF
--- a/scylla-proxy/Cargo.toml
+++ b/scylla-proxy/Cargo.toml
@@ -31,3 +31,5 @@ rand = "0.8.5"
 assert_matches = "1.5.0"
 ntest = "0.9.0"
 tracing-subscriber = { version = "0.3.14", features = ["env-filter"] }
+tokio = { version = "1.12", features = ["signal"] }
+
--- a/scylla-proxy/src/actions.rs
+++ b/scylla-proxy/src/actions.rs
@@ -48,11 +48,15 @@ pub enum Condition {
 
     /// True for predefined number of evaluations, then always false.
     TrueForLimitedTimes(usize),
+
+    // True if any REGISTER was sent on this connection. Useful to filter out control connection messages.
+    ConnectionRegisteredAnyEvent,
 }
 
 /// The context in which [`Conditions`](Condition) are evaluated.
 pub(crate) struct EvaluationContext {
     pub(crate) connection_seq_no: usize,
+    pub(crate) connection_has_events: bool,
     pub(crate) opcode: FrameOpcode,
     pub(crate) frame_body: Bytes,
 }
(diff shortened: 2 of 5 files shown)


Lorak-mmk:
I'll try to add some test of this feature

Lorak-mmk:
I added a test.
I also added tokio's signal feature to dev dependencies in scylla-proxy - without it it's not possible to run `cargo test` inside `scylla-proxy` directory. I hope it's small enough change to not require another PR.

scylla-proxy/Cargo.toml
```diff
@@ -31,3 +31,4 @@ rand = "0.8.5"
 assert_matches = "1.5.0"
 ntest = "0.9.0"
 tracing-subscriber = { version = "0.3.14", features = ["env-filter"] }
+
```
piodul, line 34: The last commit adds nothing useful to `scylla-proxy`, only an empty line.

piodul:
```
failures:

---- transport::connection::tests::test_coalescing stdout ----
Unique name: test_rust_1700493980_11
thread 'transport::connection::tests::test_coalescing' panicked at scylla/src/transport/connection.rs:2034:74:
called `Result::unwrap()` on an `Err` value: DbError(Unprepared { statement_id: b"i\xb60\x01/\xdc\x02\xb6\xec;\x18\x8e\xaa\xf38\xcf" }, "Prepared query with ID 69b630012fdc02b6ec3b188eaaf338cf not found (either the query was not prepared on this host (maybe the host has been restarted?) or you have prepared too many queries and it has been evicted from the internal cache)")
```

Strange; I know that we now sometimes prepare statements in case of `query` as well, but it falls back to `execute` which automatically reprepares the statement (but just once).

Lorak-mmk:
This failing test can't be result of my changes, can it? It doesn't use proxy and I didn't touch scylla crate apart from 2 other tests.

scylla-proxy/Cargo.toml
piodul: As for the reason why it used to work: I suppose that Cargo resolves dependencies for all crates in the workspace at once, and some other crate than scylla-proxy depends on tokio and enables the "signal" feature.

piodul (approved):

PR #869, opened 2023-12-05
SerializeRow: Add is_empty method

Lorak-mmk:
Currently ongoing serialization refactor requires to only perform queries with values using prepared statements. Session interface still allows to pass values in query* methods, so we need a way to know if we can perform unprepared query or not.

One way is to always perform prepared queries, but that is not optimal. Before ongoing refactor we could serialize values and check if they are empty. This will no longer be possible soon, because serializing will require knowing column types, which in this case means having a prepared statement.

This commit implements third way - additional method on SerializeRow trait that allows us to check if values are empty (in which case we can skip the prepare).

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types/serialize/row.rs
+++ b/scylla-cql/src/types/serialize/row.rs
@@ -53,6 +53,8 @@ pub trait SerializeRow {
         ctx: &RowSerializationContext<'_>,
         writer: &mut W,
     ) -> Result<(), SerializationError>;
+
+    fn is_empty(&self) -> bool;
 }
 
 macro_rules! fallback_impl_contents {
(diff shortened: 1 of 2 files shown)


scylla-cql/src/types/serialize/row.rs
```diff
@@ -96,6 +101,10 @@ macro_rules! impl_serialize_row_for_unit {
             // Row is empty - do nothing
             Ok(())
         }
+
+        fn is_empty(&self) -> bool {
```
piodul, line 105: Add `#[inline]` here, and to other `is_empty` impls that you add. It's very simple but is not generic so it won't get inlined across crates.

piodul (changes requested):
In order to test the new method, I suggest modifying the `serialize_values`/`serialize_values_only_new` function in `scylla-cql/src/frame/value_tests.rs` to call `SerializeRow::is_empty` and compare its result with what `<T as SerializeRow>::serialize` has produced.

Lorak-mmk:
Resolved comments, added asserts to serialize_values/serialize_values_only_new

PR #874, opened 2023-12-08
docs: add myst parser

dgarcia360:
Related PR: https://github.com/scylladb/sphinx-scylladb-theme/pull/803

This PR replaces Sphinx's deprecated extension recommonmark with myst-parser, as per https://sphinx-theme.scylladb.com/stable/configuration/markdown.html#recommonmark-parser.

# How to test

Check the docs build and display without errors.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -3,7 +3,6 @@ import os
 import sys
 from datetime import date
 
-from recommonmark.transform import AutoStructify
 from pygments.lexers.configs import TOMLLexer
 from pygments.lexers.rust import RustLexer
 from sphinx.highlighting import lexers
@@ -36,14 +35,14 @@ extensions = [
     'sphinx.ext.extlinks',
     'sphinx_sitemap',
     'sphinx_scylladb_theme',
-    'sphinx_multiversion',  # optional
-    'recommonmark',  # optional
+    'sphinx_multiversion',
+    'sphinx_scylladb_markdown'
 ]
 
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
 #
-source_suffix = ['.rst', '.md']
+source_suffix = ['.rst']
 autosectionlabel_prefix_document = True
 
 # The master toctree document.
@@ -64,14 +63,8 @@ pygments_style = 'sphinx'
 
 # Setup Sphinx
 def setup(sphinx):
-    sphinx.add_config_value('recommonmark_config', {
-        'enable_eval_rst': True,
-        'enable_auto_toc_tree': False,
-    }, True)
-    sphinx.add_transform(AutoStructify)
     lexers['rust'] = RustLexer()
     lexers['toml'] = TOMLLexer()
-
 
 # -- Options for not found extension
 
@@ -80,6 +73,11 @@ notfound_template =  '404.html'
 
 # Prefix added to all the URLs generated in the 404 page.
 notfound_urls_prefix = ''
+
+# -- Options for markdown extension
+scylladb_markdown_enable = True
+scylladb_markdown_recommonmark_versions = ['v0.10.1', 'v0.11.1']
+suppress_warnings = ["myst.header","myst.xref_missing"]
 
 # -- Options for multiversion extension
 
--- a/docs/source/load-balancing/default-policy.md
+++ b/docs/source/load-balancing/default-policy.md
@@ -168,11 +168,9 @@ And only if latency awareness is enabled:
 
 If no preferred datacenter is specified, all nodes are treated as local ones.
 
-Replicas in the same priority groups are shuffled[^*]. Non-replicas are randomly
+Replicas in the same priority groups are shuffled[^1]. Non-replicas are randomly
 rotated (similarly to a round robin with a random index).
 
-[^*]: There is an optimisation implemented for LWT requests[^**] that routes them
-to the replicas in the ring order (as it prevents contention due to Paxos conflicts),
-so replicas in that case are not shuffled in groups at all.
-
-[^**]: In order for the optimisation to be applied, LWT statements must be prepared before.
+[^1]: There is an optimisation implemented for LWT requests that routes them
+to the replicas in the ring order (as it prevents contention due to Paxos conflicts), so replicas in that case are not shuffled in groups at all.
+In order for the optimisation to be applied, LWT statements must be prepared before.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ The documentation book is written using [mdbook](https://github.com/rust-lang/md
 Book source is in `docs/source`\
 This source has to be compatible with `Sphinx` so it might sometimes contain chunks like:
 ````
-```eval_rst
+```{eval-rst}
 something
 ```
 ````
--- a/docs/pyproject.toml
+++ b/docs/pyproject.toml
@@ -8,13 +8,13 @@ authors = ["ScyllaDB Documentation Contributors"]
 python = "^3.9"
 pyyaml = "6.0.1"
 pygments = "2.15.1"
-recommonmark = "0.7.1"
 redirects_cli ="~0.1.2"
 sphinx-scylladb-theme = "~1.6.1"
 sphinx-sitemap = "2.5.1"
 sphinx-autobuild = "2021.3.14"
 Sphinx = "7.2.6"
 sphinx-multiversion-scylla = "~0.3.1"
+sphinx-scylladb-markdown = "^0.1.2"
 
 [build-system]
 requires = ["poetry>=0.12"]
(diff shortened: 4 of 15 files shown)


docs/source/load-balancing/default-policy.md
```diff
@@ -168,11 +168,10 @@ And only if latency awareness is enabled:
...
 rotated (similarly to a round robin with a random index).
 
-[^*]: There is an optimisation implemented for LWT requests[^**] that routes them
+[^1]: There is an optimisation implemented for LWT requests that routes them
 to the replicas in the ring order (as it prevents contention due to Paxos conflicts),
-so replicas in that case are not shuffled in groups at all.
-
-[^**]: In order for the optimisation to be applied, LWT statements must be prepared before.
+so replicas in that case are not shuffled in groups at all. 
+In order for the optimisation to be applied, LWT statements must be prepared before.
```
dgarcia360, line 177: **Reason:** MystParser raises an error when trying to include a footnote reference within a footnote.

piodul:
I checked out the branch and generated the documentation with:

```
make -C docs clean && make -C docs multiversionpreview
```

Unfortunately, it looks like only top-level pages are generated, and their TOCs are not rendered properly:

image

I also get a ton of warnings like these:

```
/tmp/tmpjo_qlxb7/1a66cb61e3ccfe42646c8fc75dd310a28f93ea25/docs/source/connecting/authentication.md:4: WARNING: Non-consecutive header level increase; H1 to H6 [myst.header]
/tmp/tmpjo_qlxb7/1a66cb61e3ccfe42646c8fc75dd310a28f93ea25/docs/source/connecting/authentication.md:18: WARNING: Non-consecutive header level increase; H1 to H3 [myst.header]
/tmp/tmpjo_qlxb7/1a66cb61e3ccfe42646c8fc75dd310a28f93ea25/docs/source/connecting/tls.md:8: WARNING: Non-consecutive header level increase; H1 to H3 [myst.header]
/tmp/tmpjo_qlxb7/1a66cb61e3ccfe42646c8fc75dd310a28f93ea25/docs/source/connecting/tls.md:40: WARNING: Non-consecutive header level increase; H1 to H3 [myst.header]
/tmp/tmpjo_qlxb7/1a66cb61e3ccfe42646c8fc75dd310a28f93ea25/docs/source/data-types/primitive.md:3: WARNING: Non-consecutive header level increase; H1 to H3 [myst.header]
/tmp/tmpjo_qlxb7/1a66cb61e3ccfe42646c8fc75dd310a28f93ea25/docs/source/data-types/primitive.md:23: WARNING: Non-consecutive header level increase; H1 to H3 [myst.header]
```

piodul (changes requested):
The documentation does not generate properly (with the method I tried, at least - but given that it's just a makefile command, it should work). Please fix it.

piodul:
Maybe the page I showed was not the best example because it contains links to subpages - and those pages are indeed generated unlike what I wrote earlier - but the `toctree` part shouldn't be there and it looks like menu on the left should be expanded to show the path to the current document.

This is the same page, but on the current https://rust-driver.docs.scylladb.com/:

image

Note the difference in the menu on the left.

piodul:
@annastuchlik I tried with `make preview` and it works. However, I also tried again with `multiversionpreview`, this time using the procedure described in https://sphinx-theme.scylladb.com/stable/configuration/multiversion.html#previewing-production-builds:

- If I try doing this and build the documentation on the current `main`, I don't get the issues I described here and the documentation looks ok. 
- If I add the fork as a remote, pull the PR's branch and then check it out (`git remote add dgarcia360 https://github.com/dgarcia360/scylla-rust-driver/; git fetch dgarcia360; git checkout dgarcia/docs-myst-parser`) and then try to build `mutliversionpreview`, I get the same issues again. I also tried merging the branch to `main` and then building, but I get the same issues again.

I'm a bit hesitant to merge this as I'm not sure whether the documentation will look alright after this PR is merged. There seems to be a difference in output between `preview` and `multiversionpreview` - which is not present as of now - so I'm not convinced after seeing the output of `preview` that it's ok to merge.

docs/source/queries/paged.md
```diff
@@ -5,6 +5,7 @@ allow to receive the whole result page by page.
 `Session::query_iter` and `Session::execute_iter` take a [simple query](simple.md) or a [prepared query](prepared.md)
 and return an `async` iterator over result `Rows`.
 
+
```
Lorak-mmk, line 8: why?

docs/source/queries/result.md
```diff
@@ -51,7 +51,7 @@ if let Some(rows) = session.query("SELECT a, b from ks.tab", &[]).await?.rows {
 # }
 ```
 
-## Parsing using convenience methods
+# Parsing using convenience methods
```
Lorak-mmk, line 54: It doesn't look like this should be there, the whole chapter is about query result, but now this is a sibling to `Query Result` chapter.
After undoing this, there will probably be some other changes in this file required.

Lorak-mmk (changes requested):
I left a few comments. One other issue, not strictly related to this PR, but I just noticed it and you are touching this region:
When rendering docs using `mdbook` (`mdbook build docs --open`), `toctree` sections are not ignored, and it looks like this:
image

Those should be somehow marked so that mdbook ignores them.

Lorak-mmk:
> @Lorak-mmk It seems you're reviewing the docs on GitHub. When it comes to the docs, GitHub is just a fallback. The primary documentation is the published online docs - you can review them with the make `preview command`. For this reason, I have to skip your feedback regarding rendering if it's connected with the GitHub look and feel.

mdbook is a standard tool for generating Rust book docs, afaik it's not related to GitHub in any way. I strongly feel, that book docs in this repo should be compatible with both mdbook and official docs. Why?
1. We use mdbook in CI, to run examples from the book to make sure they work correctly and prevent code rot.
2. Most Rust developers are familiar with the format and the tool, and are used to reading generated content.
3. It is also helpful for me as a developer, as it's much faster that `make preview` and doesn't require me to fight with the Python mess for an hour each time I use it, so I get quick feedback loop thanks to it.
4. For me - and I know I'm not the only one - the look of official docs is much worse and harder to read. If I need to look something up, it's more pleasant to use mdbook output.

By the way, the only thing in my review that is related to mdbook is the `toctree` thing, everything else is about the official docs. If you still don't want to touch mdbook, feel free to ignore this specific comment, I'll try to fix the issue myself later.

Lorak-mmk:
Oh, nevermind about the toc_tree issue @annastuchlik , I started to look into why it doesn't work and saw we already have a solution for that - there is docs/build_book.py script, which copies the source to new dir, removes eval_rst blocks and build the book using mdbook.
So I guess the only think for this PR is to update this script, as it looks for 
```
```eval_rst
```

and this PR change the format to
```
```{eval_rst}
```

docs/source/connecting/authentication.md
```diff
@@ -1,7 +1,8 @@
 # Authentication
 
 Driver supports both authentication by username and password and custom authentication defined by a user.
-###### Important: The default authentication credentials are sent in plain text to the server. For this reason, it is highly recommended that this be used in conjunction with client-to-node encryption (SSL), or in a trusted network environment.
+
+## Important: The default authentication credentials are sent in plain text to the server. For this reason, it is highly recommended that this be used in conjunction with client-to-node encryption (SSL), or in a trusted network environment.
```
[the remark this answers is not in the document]
dgarcia360: Here's the syntax to define admonitions with this new parser: https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#admonition-types

And, the subset of admonitions supported  by the theme: https://sphinx-theme.scylladb.com/stable/examples/admonitions.html

dgarcia360:
I've rebased the branch and applied the suggested changes. To maintain simplicity in this pull request, I chose not to modify the H1-H6 headings order. Instead, I've suppressed the related warning through the `conf.py` file by using the `suppress_warnings` option.

@piodul 

> Unfortunately, it looks like only top-level pages are generated, and their TOCs are not rendered properly:

I believe v0.11.0 was not released at the time of creating the branch for this pull request. I've fixed the issue by adding it to the versions that should use recommonmark instead of Myst.

docs/source/conf.py
```diff
@@ -81,6 +74,11 @@ def setup(sphinx):
 # Prefix added to all the URLs generated in the 404 page.
 notfound_urls_prefix = ''
 
+# -- Options for markdown extension
+scylladb_markdown_enable = True
+scylladb_markdown_recommonmark_versions = ['v0.9.1', 'v0.10.1']
```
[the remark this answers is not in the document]
Lorak-mmk: Thanks for explanation!

piodul (approved):
I no longer see the issues I reported about `make multiversionpreview`, so looks good from my point of view.

Lorak-mmk (approved):

PR #878, opened 2023-12-13
serialize: add more tests and docstrings

piodul:
This PR aims to improve the quality of the serialize module:

- Adds a couple of tests which check that proper errors are returned in case when serialization fails,
- Fixes a couple of bugs in errors reported from serialization logic - thanks to the aforementioned tests,
- Documents remaining public items in the `types/serialize` module and adds a compile-time warning about missing public items in order to prevent regressions.

## Pre-review checklist


scylla-cql/src/types/serialize/row.rs
```diff
@@ -892,6 +901,123 @@ mod tests {
...
+            }
+        ));
+
+        // Serialization of one of the element fails
+        let v = vec!["Ala ma kota", "Kot ma pchły"];
+        let spec = [col("a", ColumnType::Text), col("b", ColumnType::Int)];
+        let err = do_serialize_err(v, &spec);
+        let err = get_ser_err(&err);
+        assert_eq!(err.rust_name, std::any::type_name::<Vec<&str>>());
+        let BuiltinSerializationErrorKind::ColumnSerializationFailed { name, err: _ } = &err.kind;
```
Lorak-mmk, line 1042: I didn't know this code is possible - what happends if the variant is different than `ColumnSerializationFailed`? `BuiltinSerializationErrorKind` is `#[non_exhaustive]` so adding a new variant is backwards compatible afaiu

scylla-cql/src/types/serialize/mod.rs
```diff
@@ -7,6 +11,24 @@ pub mod value;
...
+///   one of types with an impl built into the driver fails. It is also returned
+///   from impls generated by the `SerializeRow` macro.
+/// - [`value::BuiltinSerializationError`] is analogous to the above but is
+///   returned from [`SerializeCql::serialize`](value::SerializeCql::serialize)
+///   instead both in the case of builtin impls and impls generated by the
+///   `SerializeCql` macro. It won't be returned by the `Session` directly,
+///   but it might be nested in the [`row::BuiltinSerializationError`].
+/// - [`row::ValueListToSerializeRowAdapterError`] is returned in case when
+///   a list of values encoded with the legacy `ValueList` trait is passed
+///   as an argument to the statement.
```
Lorak-mmk, line 31: Is this always returned or only when serialization fails?

Lorak-mmk (changes requested):

scylla-cql/src/types/serialize/row.rs
piodul: `ColumnSerializationFailed` is the only variant of `BuiltinSerializationErrorKind`, so this pattern is irrefutable and there is no need for `else`. In fact, I tried putting an `else {}` to future-proof the expression, and while `rustc` ate it, `clippy` complained that it is not needed.

As far as `#[non_exhaustive]` goes, it only affects other crates than the current one - which makes sense; if you add an enum variant, you can immediately fix all the uses of the enum in the same crate.

Lorak-mmk: Wow, I didn't expect Rust to be that smart in this regard, nice.

scylla-cql/src/types/serialize/mod.rs
piodul: Right, the comment doesn't make sense. The error is only returned if "rewriting" the ValueList via SerializeRow interface fails. I'll update.

scylla-cql/src/types/serialize/row.rs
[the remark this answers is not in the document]
piodul: I'm not sure where would be the best way to put it and how to formulate it... Let's leave it for a follow up for now.

scylla-cql/src/types/serialize/mod.rs
piodul: Fixed

Lorak-mmk (approved):

PR #880, opened 2023-12-14
Serialization documentation

Lorak-mmk:
This PR adds warnings about changes introduced in #806 and #812 and updates documentation about new serialization API.

## Pre-review checklist

### DIFF
--- a/docs/source/data-types/udt.md
+++ b/docs/source/data-types/udt.md
@@ -8,17 +8,25 @@ For example let's say `my_type` was created using this query:
 CREATE TYPE ks.my_type (int_val int, text_val text)
 ```
 
-To use this type in the driver, create a matching struct and derive `IntoUserType` and `FromUserType`:
+To use this type in the driver, create a matching struct and derive:
+- `SerializeCql`: in order to be able to use this struct in query parameters. \
+    This macro requires fields of UDT and struct to have matching names, but the order
+    of the fields is not required to be the same. \
+    Note: you can use different name using `rename` attribute - see `SerializeCql` macro documentation.
+- `FromUserType`:  in order to be able to use this struct in query results. \
+    This macro requires fields of UDT and struct to be in the same *ORDER*. \
+    This mismatch between `SerializeCql` and `FromUserType` requirements is a temporary situation - in the future `FromUserType` (or  the macro that replaces it) will also require matching names.
 
 ```rust
 # extern crate scylla;
 # async fn check_only_compiles() {
-use scylla::macros::{FromUserType, IntoUserType};
+use scylla::macros::{FromUserType, SerializeCql};
 
 // Define a custom struct that matches the User Defined Type created earlier.
-// Fields must be in the same order as they are in the database.
+// Fields must be in the same order as they are in the database and also
+// have the same names.
 // Wrapping a field in Option will gracefully handle null field values.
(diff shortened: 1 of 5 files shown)


docs/source/data-types/udt.md
```diff
@@ -8,17 +8,25 @@ For example let's say `my_type` was created using this query:
 CREATE TYPE ks.my_type (int_val int, text_val text)
 ```
 
-To use this type in the driver, create a matching struct and derive `IntoUserType` and `FromUserType`:
+To use this type in the driver, create a matching struct and derive:
+- `SerializeCql`: in order to be able to use this struct in query parameters. \
+    This macro requires fields of UDT and struct to have matching names.
```
piodul, line 13: Maybe worth highlighting here that the order does not have to match with the UDT definition in the database.

docs/source/data-types/udt.md
```diff
@@ -27,8 +35,9 @@ struct MyType {
 ```
 
 > ***Important***\
-> Fields in the Rust struct must be defined in the same order as they are in the database.
-> When sending and receiving values, the driver will (de)serialize fields one after another, without looking at field names.
+> Fields in the Rust struct must be defined in the same order and with the same names as they are in the database.
```
piodul, line 38: I think this should be split into two remarks, one about deserialization (assumption about the order) and serialization (assumption that the names match). This sentence looks like a general recommendation while only parts of it apply to serialization, and others to deserialization.

docs/source/queries/paged.md
```diff
@@ -119,6 +127,11 @@ let res2 = session
 # }
 ```
 
+> ***Warning***\
+> If the values are not empty, driver first needs to send a `PREPARE` request \
+> in order to fetch information required to serialize values. This will affect \
+> performance because 2 requests will be sent instead of 1.
```
piodul, line 133: Suggestion: `s/because 2 requests will be sent/because 2 round trips will be required/`, to emphasize that those requests are executed serially (though technically speaking this is implied by the previous sentence).

docs/source/queries/simple.md
```diff
@@ -22,6 +22,11 @@ session
 > 
 > When page size is set, `query` will return only the first page of results.
 
+> ***Warning***\
+> If the values are not empty, driver first needs to send a `PREPARE` request \
+> in order to fetch information required to serialize values. This will affect \
+> performance because 2 requests will be sent instead of 1.
```
piodul, line 28: Ditto

scylla/src/transport/session.rs
```diff
@@ -558,6 +558,10 @@ impl Session {
     ///
     /// This is the easiest way to make a query, but performance is worse than that of prepared queries.
     ///
+    /// It is discouraged to use this method with non-empty values argument (`is_empty()` method from `SerializeRow`
+    /// trait returns false). In such case, query first needs to be prepared (on a single connection), so
+    /// driver will issue 2 requests instead of 1. Please use [`Session::execute()`] instead.
```
piodul, line 563: Ditto, and in other places in this file.

Lorak-mmk:
v2:
- Addressed review comments
- Added note about `rename` attribute
- Fixed formatting by removing unnecessary backslashes

piodul (approved):

PR #881, opened 2023-12-14
serialization: implement new, type-safe variants of BatchValues

piodul:
This PR replaces the existing BatchValues API with one that leverages the recently introduced `SerializeCql`/`SerializeRow` infrastructure and is type-safe.

Refs: #801

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/request/mod.rs
+++ b/scylla-cql/src/frame/request/mod.rs
@@ -7,6 +7,7 @@ pub mod query;
 pub mod register;
 pub mod startup;
 
+use crate::types::serialize::row::SerializedValues;
 use crate::{frame::frame_errors::ParseError, Consistency};
 use bytes::Bytes;
 use num_enum::TryFromPrimitive;
@@ -22,7 +23,6 @@ pub use startup::Startup;
 use self::batch::BatchStatement;
 
 use super::types::SerialConsistency;
-use super::value::LegacySerializedValues;
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, TryFromPrimitive)]
 #[repr(u8)]
@@ -59,7 +59,7 @@ pub trait DeserializableRequest: SerializableRequest + Sized {
 pub enum Request<'r> {
     Query(Query<'r>),
     Execute(Execute<'r>),
-    Batch(Batch<'r, BatchStatement<'r>, Vec<LegacySerializedValues>>),
+    Batch(Batch<'r, BatchStatement<'r>, Vec<SerializedValues>>),
 }
 
 impl<'r> Request<'r> {
@@ -190,8 +190,8 @@ mod tests {
 
             // Not execute's values, because named values are not supported in batches.
             values: vec![
-                query.parameters.values.deref().to_old_serialized_values(),
-                query.parameters.values.deref().to_old_serialized_values(),
+                query.parameters.values.deref().clone(),
+                query.parameters.values.deref().clone(),
             ],
         };
         {
@@ -262,7 +262,7 @@ mod tests {
             serial_consistency: None,
             timestamp: None,
 
-            values: vec![query.parameters.values.deref().to_old_serialized_values()],
+            values: vec![query.parameters.values.deref().clone()],
         };
         {
             let mut buf = Vec::new();
--- a/scylla-cql/src/frame/request/mod.rs
+++ b/scylla-cql/src/frame/request/mod.rs
@@ -8,7 +8,7 @@ pub mod register;
 pub mod startup;
 
 use crate::{frame::frame_errors::ParseError, Consistency};
-use bytes::{BufMut, Bytes};
+use bytes::Bytes;
 use num_enum::TryFromPrimitive;
 
 pub use auth_response::AuthResponse;
@@ -40,7 +40,7 @@ pub enum RequestOpcode {
 pub trait SerializableRequest {
     const OPCODE: RequestOpcode;
 
-    fn serialize(&self, buf: &mut impl BufMut) -> Result<(), ParseError>;
+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ParseError>;
 
     fn to_bytes(&self) -> Result<Bytes, ParseError> {
         let mut v = Vec::new();
--- a/scylla-cql/src/types/serialize/batch.rs
+++ b/scylla-cql/src/types/serialize/batch.rs
@@ -1,0 +1,307 @@
+//! Contains the [`BatchValues`] and [`BatchValuesIterator`] trait and their
+//! implementations.
+
+use super::row::{RowSerializationContext, SerializeRow};
+use super::{RowWriter, SerializationError};
+
+/// Represents a list of sets of values for a batch statement.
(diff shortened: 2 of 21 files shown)


scylla/src/transport/session.rs
```diff
@@ -1200,13 +1200,15 @@ impl Session {
...
+            // The temporary "p" is necessary because lifetimes
+            let p = match (first_serialized_value, batch.statements.first()) {
                 (Some(first_serialized_value), Some(BatchStatement::PreparedStatement(ps))) => {
                     let token = ps.calculate_token(&first_serialized_value)?;
                     (Some(first_serialized_value), token, ps.get_keyspace_name())
                 }
                 _ => (None, None, None),
-            }
+            };
+            p
```
Lorak-mmk: It's interesting that this would cause issues, I thought both versions should have identical semantics.

Lorak-mmk: Also, if the problem happens with the new API but not with old, doesn't it mean there is possible some issue within new API definitions?

Lorak-mmk (commented):
After seeing new definitions of `BatchValuesIterator` / `RawBatchValuesIterator` I wonder - isn't it basically a re-implementation of `LendingIterator`? Maybe we could use it?

scylla-cql/src/frame/request/mod.rs
```diff
@@ -40,7 +40,7 @@ pub enum RequestOpcode {
 pub trait SerializableRequest {
     const OPCODE: RequestOpcode;
 
-    fn serialize(&self, buf: &mut impl BufMut) -> Result<(), ParseError>;
+    fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ParseError>;
```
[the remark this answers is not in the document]
piodul: I don't think `scylla` does, or at least should, reexport this API. Users aren't supposed to serialize frames themselves, and the driver is responsible for allocating the buffer for the frame.

piodul:
> After seeing new definitions of `BatchValuesIterator` / `RawBatchValuesIterator` I wonder - isn't it basically a re-implementation of `LendingIterator`? Maybe we could use it?

AFAIK there is not standard library definition of a lending iterator. Not sure whether there are "industry standard" crates that would define such a trait. Besides, this issue is irrelevant in the newest version of the PR.

scylla-cql/src/types/serialize/batch.rs
```diff
@@ -0,0 +1,273 @@
...
+/// `session.batch(&batch, BatchValuesFromIter::from(lines_to_insert.iter().map(|l| &l.value_list)))`
+/// where `lines_to_insert` may also contain e.g. data to pick the statement...
+///
+/// The underlying iterator will always be cloned at least once, once to compute the length if it can't be known
+/// in advance, and be re-cloned at every retry.
+/// It is consequently expected that the provided iterator is cheap to clone (e.g. `slice.iter().map(...)`).
+pub struct BatchValuesFromIterator<'sr, IT> {
+    it: IT,
+
+    // Why is this needed??? Without it impl BatchValues for BatchValuesFromIter doesn't work
```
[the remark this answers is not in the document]
piodul: Done, also changed the tone of the comment

scylla-cql/src/types/serialize/raw_batch.rs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        let mut count = 0;
+        while self.skip_next().is_some() {
+            count += 1;
+        }
+        count
+    }
+}
+
+/// A proxy for a `SerializeValue` which knows how to serialize itself into
```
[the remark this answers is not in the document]
piodul: Ugh, those names are too similar, and now I am mixing them... I think this was supposed to be `SerializeRow`. Should not be relevant in the current version of the PR.

scylla/src/transport/session.rs
piodul: > It's interesting that this would cause issues, I thought both versions should have identical semantics.

In the version of the PR that you reviewed, `BatchValueIterator` would lend its value. I suspect that could have an effect here. I'm not sure it is an issue anymore.

> Also, if the problem happens with the new API but not with old, doesn't it mean there is possible some issue within new API definitions?

Well, I saw that there is an existing test for batches:

```rust
#[test]
fn check_batch_values_iterator_is_not_lending() {
    // This is an interesting property if we want to improve the batch shard selection heuristic
    fn f(bv: impl LegacyBatchValues) {
        let mut it = bv.batch_values_iter();
        let mut it2 = bv.batch_values_iter();
        // Make sure we can hold all these at the same time
        let v = vec![
            it.next_serialized().unwrap().unwrap(),
            it2.next_serialized().unwrap().unwrap(),
            it.next_serialized().unwrap().unwrap(),
            it2.next_serialized().unwrap().unwrap(),
        ];
        let _ = v;
    }
    f(((10,), (11,)))
}
```

On the other hand, I don't see _why_ it was important for the iterator not to lend (I couldn't infer it from the comments, GitHub discussions, commit descriptions etc.). Nevertheless, In the latest version of the PR the `BatchValuesIterator` does not lend (it's not really a true iterator anymore), so the issue is not there and I adapter the test successfully.

piodul:
Marking as "ready", I'll update the docs in a separate PR.

scylla-cql/src/types/serialize/raw_batch.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// An iterator-like for `ValueList`
+///
+/// An instance of this can be easily obtained from `IT: Iterator<Item: ValueList>`: that would be
+/// `BatchValuesIteratorFromIterator<IT>`
+///
+/// It's just essentially making methods from `ValueList` accessible instead of being an actual iterator because of
+/// compiler limitations that would otherwise be very complex to overcome.
+/// (specifically, types being different would require yielding enums for tuple impls)
```
Lorak-mmk, line 31: In this comment you probably meant `SerializedValues`, not `ValueList`, right?

Lorak-mmk (changes requested):
Just one small comment

scylla-cql/src/types/serialize/raw_batch.rs
piodul: Right, this comment was copied over from the legacy `BatchValuesIterator` but not updated. I'll update in a second.

piodul: Done

scylla-cql/src/types/serialize/raw_batch.rs
```diff
@@ -0,0 +1,162 @@
...
+    // See <https://github.com/scylladb/scylla-rust-driver/issues/599> for more details
+    type RawBatchValuesIter<'r>: RawBatchValuesIterator<'r>
+    where
+        Self: 'r;
+
+    /// Returns an iterator over the data contained in this object.
+    fn batch_values_iter(&self) -> Self::RawBatchValuesIter<'_>;
+}
+
+/// An iterator-like for `ValueList`
```
Lorak-mmk, line 31: One more value list hides here @piodul

Lorak-mmk (approved):

Issue #802, opened 2023-08-25
Serialization refactor: macros for the new traits


piodul:
Sub-task of https://github.com/scylladb/scylla-rust-driver/issues/463.

The new `SerializeRow` and `SerializeCql` traits will need to have their corresponding procedural macros implemented. Those macros should match struct fields to columns/UDT fields by name. This should be done by default, but there should be an attribute that allows disabling it.

piodul:
This depends on the definition of the traits in #801.

Lorak-mmk:
This was implemented in https://github.com/scylladb/scylla-rust-driver/pull/851 with some smaller follow ups.

PR #882, opened 2023-12-15
scylla-macros: attributes for better control over name checks

piodul:
The PR introduces two attributes related to matching Rust struct fields by name in the new `SerializeRow`/`SerializeCql` macros:

- `#[scylla(rename = "xyz")]` - instead of using the name of the field verbatim, the name provided in the attribute is used when matching to a column/bind marker/UDT field. This can be useful e.g. if the user has different naming convention in Rust code and in the schema definitions.
- `#[scylla(skip_name_checks)`] - in `enforce_order` flavor, skips checking the names of the fields as they are being serialized. With this attribute, the names of the corresponding fields in Rust struct and UDT/columns/bind markers don't have to match at all, but their _types_ still need to. This attribute is supposed to help migrate from the old `IntoUserType`/`ValueList` macros.

Refs: #802

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/macros.rs
+++ b/scylla-cql/src/macros.rs
@@ -87,6 +87,15 @@ pub use scylla_macros::ValueList;
 /// macro itself, so in those cases the user must provide an alternative path
 /// to either the `scylla` or `scylla-cql` crate.
 ///
+/// `#[scylla(skip_name_checks)]
+///
+/// _Specific only to the `enforce_order` flavor._
+///
+/// Skips checking Rust field names against names of the UDT fields. With this
+/// annotation, the generated implementation will allow mismatch between Rust
+/// struct field names and UDT field names, i.e. it's OK if i-th field has a
+/// different name in Rust and in the UDT. Fields are still being type-checked.
+///
 /// # Field attributes
 ///
 /// `#[scylla(rename = "name_in_the_udt")]`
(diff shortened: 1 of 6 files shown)


scylla-cql/src/macros.rs
```diff
@@ -87,6 +87,15 @@ pub use scylla_macros::ValueList;
...
 ///
+/// `#[scylla(skip_name_checks)]
+///
+/// _Specific only to the `enforce_order` flavor._
+///
+/// Skips checking Rust field names against names of the UDT fields. With this
+/// annotation, the generated implementation will allow mismatch between Rust
+/// struct field names and UDT field names, i.e. it's OK if i-th field has a
+/// different name in Rust and in the UDT. Fields are still being type-checked.
+///
```
Lorak-mmk, line 98: I'm wondering if it wouldn't be better to make this a new flavor instead of an attribute - one less place to make a mistake for a user. But seeing that it's a compile time error, maybe it doesn't matter?

Lorak-mmk (approved):

scylla-cql/src/macros.rs
piodul: I'm not entirely sure either; implementation-wise it was more natural to add an attribute as it introduces very slight changes to how the `enforce_order` flavor generates code.

One argument that I see for leaving it as a separate attribute is that it disables some safety checks, while both current flavors don't compromise on safety. It's good to make the less safe option more verbose IMO.

In the future we should consider adding support for tuple structs which very natually fit into the `enforce_order` + `skip_name_checks` pattern, we could probably remove the current `skip_name_checks` attribute when doing that.

PR #885, opened 2023-12-18
Don't return error on unexpected UDT field in SerializeCql macro

Lorak-mmk:
In order to properly support ALTERing UDTs in database, code that serializes UDT shouldn't return error when there is column in UDT that is not present in Rust struct.
This PR changes the default behavior like this, and adds an `force_exact_match` attribute to opt-in to previous behavior.

CQL specification says:
```
  A UDT value will generally have one value
  for each field of the type it represents, but it is allowed to have less values than
  the type has fields.
```
so I decided to not send anything in place of missing fields rather than send e.g. NULLs.

## Pre-review checklist


scylla-cql/src/macros.rs
```diff
@@ -42,7 +46,6 @@ pub use scylla_macros::ValueList;
 /// struct MyUdt {
 ///     a: i32,
 ///     b: Option<String>,
-///     c: Vec<u8>,
```
piodul, line 45: Hmm, I assume that you remove this field to show the new behavior, but I wonder whether it won't be confusing - especially for those readers which skim through the description right to the example. Perhaps a short comment in place of the removed field would be in order?

scylla-cql/src/macros.rs
```diff
@@ -18,8 +18,12 @@ pub use scylla_macros::ValueList;
 ///
```
piodul, line 18: > New attribut, called

Nit: typo in the commit message of the first commit: "attribut"

scylla-cql/src/types/serialize/value.rs
```diff
@@ -2069,6 +2069,44 @@ mod tests {
         assert_eq!(reference, udt);
     }
 
+    #[test]
```
piodul, line 2072: > This commit adds tests for new behavior of SerializeCql - ignoring
unknown UDT fields.\

Nit: second commit message: unnecessary backslash at the end of the line

scylla-cql/src/types/serialize/value.rs
```diff
@@ -2069,6 +2069,44 @@ mod tests {
...
+                // Unexpected field
+                ("d".to_string(), ColumnType::Counter),
+            ],
+        };
+
+        let result_normal = do_serialize(&udt, &typ_normal);
+        let result_additional_field = do_serialize(&udt, &typ_unexpected_field);
+
+        assert_eq!(result_normal, result_additional_field);
+    }
```
piodul, line 2108: In addition to those tests, I'd like to see the new behavior tested in `cql_types_test.rs` with a real Scylla cluster, to see whether sending one field less does actually work.

piodul (approved):

PR #889, opened 2023-12-19
Migration guide fix

Lorak-mmk:
This PR fixes 2 problems:
- One broken link to new migration guide.
- Table not rendering correctly in migration guide. It looks like Scylla's documentation system doesn't support tables in Markdown.

## Pre-review checklist


Lorak-mmk:
@dgarcia360 Regarding second point, are we doing something wrong or tables are really not supported? It is render correctly by both VSCode and mdbook

piodul:
Sorry for the sloppiness on my side. The table looked correct in vscode/github, but AFAIK tables are a markdown extension, so I shouldn't take it for granted that sphinx supports it.

As for the incorrect file name in the link, it looks like something that should be caught by CI. I am surprised that mdbook (?) doesn't produce an error. Maybe it's possible to adjust the CLI flags to catch issues like that in the future.

piodul (approved):

Issue #749, opened 2023-06-23
Add a `TimeUuid` struct


cvybhu:
`timeuuid` is a UUID v1 - it contains a timestamp and some random bits that together form a unique identifier.
Their comparison operator should compare them primarily based on this timestamp.

`timeuuid` values are currently represented using `uuid::Uuid`, but the comparison operators of `uuid::Uuid` don't compare `timeuuid` values in the same way that Cassandra/Scylla does.

For example, with values `00000257-0efc-11ee-9547-00006490e9a6` and `fed35080-0efb-11ee-a1ca-00006490e9a4`:

Cassandra believes that `fed35080-0efb-11ee-a1ca-00006490e9a4` is smaller:
```cql
CREATE KEYSPACE ks WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1};
CREATE TABLE ks.tab (p int, c timeuuid, PRIMARY KEY (p, c));
INSERT INTO ks.tab (p, c) VALUES (0, 00000257-0efc-11ee-9547-00006490e9a6);
INSERT INTO ks.tab (p, c) VALUES (0, fed35080-0efb-11ee-a1ca-00006490e9a4);
SELECT p, c FROM ks.tab;
 p | c
---+--------------------------------------
 0 | fed35080-0efb-11ee-a1ca-00006490e9a4
 0 | 00000257-0efc-11ee-9547-00006490e9a6

(2 rows)
```
Because it has a lower timestamp value:
```cql
 timeuuid_col                         | system.totimestamp(timeuuid_col)
--------------------------------------+----------------------------------
 fed35080-0efb-11ee-a1ca-00006490e9a4 |  2023-06-19 23:49:56.990000+0000
 00000257-0efc-11ee-9547-00006490e9a6 |  2023-06-19 23:49:58.961000+0000
```

But `Uuid` comparison says the opposite, as it compares the bytes in lexicographical order:
```rust
let mut vals = [
    uuid::uuid!("fed35080-0efb-11ee-a1ca-00006490e9a4"),
    uuid::uuid!("00000257-0efc-11ee-9547-00006490e9a6"),
];
vals.sort();
dbg!(vals);

// Prints:
vals = [
    000002570efc11ee954700006490e9a6,
    fed350800efb11eea1ca00006490e9a4,
]
```

It would be useful to have a struct that represents `timeuuid` values and has the same semantics as the values in Cassandra.

[...]

PR #894, opened 2023-12-22
Introduce `CqlTimeuuid` type

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/749

## Changes
- Introduced `CqlTimeuuid` type as a wrapper for version 1 uuids
- Implemented `uuid::Uuid` delegate methods for `CqlTimeuuid`
- Implemented strong ordering on `CqlTimeuuid` type so it has the same semantics as Scylla/Cassandra
- `CqlValue::Timeuuid` now wraps `CqlTimeuuid`
- Implemented serialization and deserialization logic for `CqlTimeuuid`
- Got rid of `timeuuid` <-> `uuid::Uuid` mapping

## Pre-review checklist


docs/source/data-types/data-types.md
```diff
@@ -20,7 +20,8 @@ Database types and their Rust equivalents:
 * `Counter` <----> `value::Counter`
 * `Blob` <----> `Vec<u8>`
 * `Inet` <----> `std::net::IpAddr`
-* `Uuid`, `Timeuuid` <----> `uuid::Uuid`
+* `Uuid` <----> `uuid::Uuid`
+* `Timeuuid` <----> `value::CqlTimeuuid`
```
piodul, line 24: Please add a page for the new `CqlTimeuuid` type, like we already have for `uuid::Uuid` in `uuid.md`

scylla-cql/src/frame/response/result.rs
```diff
@@ -769,8 +769,15 @@ pub fn deser_cql_value(typ: &ColumnType, buf: &mut &[u8]) -> StdResult<CqlValue,
...
-            CqlValue::Timeuuid(uuid)
+            let timeuuid = match CqlTimeuuid::from_slice(buf) {
+                Ok(timeuuid) => timeuuid,
+                Err(e) => {
+                    return Err(ParseError::BadIncomingData(format!(
+                        "Timeuuid parsing failed: {e}"
+                    )))
+                }
+            };
+            CqlValue::Timeuuid(timeuuid)
```
piodul, line 779: Please verify whether Scylla (and Cassandra) validates timeuuids like this, i.e. whether it allows inserting uuids that are not necessarily timeuuids into timeuuid column. If they are then perhaps we can do this validation here as well, otherwise I would abstain from it so that users don't run into a situation where they cannot read data they inserted earlier or via another driver.

Perhaps it is safer not to validate here anyway - even if the DBs properly validate right now, just in case if validation breaks in Scylla/Cassandra at some future release.

piodul (changes requested):

scylla-cql/src/frame/response/result.rs
muzarski: > Please verify whether Scylla (and Cassandra) validates timeuuids like this, i.e. whether it allows inserting uuids that are not necessarily timeuuids into timeuuid column.

A simple test case shows that Scylla validates whether uuid inserted to timeuuid column is a valid timeuuid.
```rust
#[tokio::test]
async fn test_invalid_timeuuid() {
    let session: Session = create_new_session_builder().build().await.unwrap();
    let ks = unique_keyspace_name();

    session
        .query(
            format!(
                "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = \
            {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}",
                ks
            ),
            &[],
        )
        .await
        .unwrap();
    session.use_keyspace(ks, false).await.unwrap();

    session
        .query(
            "CREATE TABLE tab (p int, t timeuuid, PRIMARY KEY (p, t))",
            (),
        )
        .await
        .unwrap();

    // Invalid timeuuid (version 2 uuid)
    let invalid_timeuuid = uuid::uuid!("00000000-0000-2000-8080-808080808080");

    session
        .query("INSERT INTO tab (p, t) VALUES (0, ?)", (invalid_timeuuid,))
        .await
        .unwrap();
}
```

The result:
```
thread 'transport::cql_types_test::test_invalid_timeuuid' panicked at scylla/src/transport/cql_types_test.rs:1731:10:
called `Result::unwrap()` on an `Err` value: DbError(Invalid, "Exception while binding column t: marshaling error: Unsupported UUID version (2)")
```



> Perhaps it is safer not to validate here anyway - even if the DBs properly validate right now, just in case if validation breaks in Scylla/Cassandra at some future release.

But I agree, I'll get rid of the validation part in `CqlTimeuuid` constructors i.e. replace
```rust
impl TryFrom<Uuid> for CqlTimeuuid {
    type Error = InvalidTimeuuidVersion;

    fn try_from(value: Uuid) -> Result<Self, Self::Error> {
        let version = value.get_version_num();
        if version != 1 {
            return Err(InvalidTimeuuidVersion(version));
        }

        Ok(Self(value))
    }
}
```
with
```rust
impl From<Uuid> for CqlTimeuuid { ... }
```

muzarski:
v2:
- rebased on top of master
- fixed review comments

scylla-cql/src/frame/value_tests.rs
```diff
@@ -41,6 +42,26 @@ fn serialized_only_new<T: SerializeCql>(val: T, typ: ColumnType) -> Vec<u8> {
...
+        CqlTimeuuid::from_str("ffffffff-ffff-1fff-ffff-ffffffffffff").unwrap(),
+        CqlTimeuuid::from_str("ffffffff-ffff-1fff-0000-000000000000").unwrap(),
+        CqlTimeuuid::from_str("ffffffff-ffff-1fff-7f7f-7f7f7f7f7f7f").unwrap(),
+    ];
+
+    let mut driver_sorted_timeuuid_vals = sorted_timeuuid_vals.clone();
+    driver_sorted_timeuuid_vals.sort();
+
+    assert_eq!(sorted_timeuuid_vals, driver_sorted_timeuuid_vals);
+}
```
Lorak-mmk, line 63: I'd like to see an example where values are not already sorted

scylla/src/transport/cql_types_test.rs
```diff
@@ -1149,6 +1149,72 @@ async fn test_timeuuid() {
...
+        CqlTimeuuid::from_str("00000000-0000-1000-8080-808080808080").unwrap(),
+        CqlTimeuuid::from_str("00000000-0000-1000-ffff-ffffffffffff").unwrap(),
+        CqlTimeuuid::from_str("00000000-0000-1000-0000-000000000000").unwrap(),
+        CqlTimeuuid::from_str("fed35080-0efb-11ee-a1ca-00006490e9a4").unwrap(),
+        CqlTimeuuid::from_str("00000257-0efc-11ee-9547-00006490e9a6").unwrap(),
+        CqlTimeuuid::from_str("ffffffff-ffff-1fff-ffff-ffffffffffef").unwrap(),
+        CqlTimeuuid::from_str("ffffffff-ffff-1fff-ffff-ffffffffffff").unwrap(),
+        CqlTimeuuid::from_str("ffffffff-ffff-1fff-0000-000000000000").unwrap(),
+        CqlTimeuuid::from_str("ffffffff-ffff-1fff-7f7f-7f7f7f7f7f7f").unwrap(),
+    ];
```
Lorak-mmk, line 1189: Again, if we test sorting, it would be good to have unsorted data. You can use https://docs.rs/rand/latest/rand/seq/trait.SliceRandom.html#tymethod.shuffle to avoid having 2 literals in the code.

scylla-cql/src/frame/value.rs
```diff
@@ -51,6 +51,150 @@ pub enum MaybeUnset<V> {
     Set(V),
 }
 
+/// Represents timeuuid (uuid V1) value
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
```
Lorak-mmk, line 55: I'm wondering if it would be beneficial to users to derive `Hash` to, so this type is easier to use in hash maps. wdyt?

Lorak-mmk (changes requested):

scylla-cql/src/frame/value.rs
muzarski: It's good that you mentioned that.

Answering your question - we shouldn't derive `Hash` as our `PartialOrd`/`Ord` custom logic masks off 7th bit msbyte. Scylla/Cassandra both behave in the same way.

The same applies to deriving `Eq`/`PartialEq`. We should implement them in a way so the following property holds: `x.cmp(&y) == Ordering::Equal <=> x == y`.

Here is an exemplary test case showing that, as of now, this property doesn't hold:
```rust
#[test]
fn timeuuid_hash() {
    let x = CqlTimeuuid::from_str("00000000-0000-1000-8080-808080808080").unwrap();
    let y = CqlTimeuuid::from_str("00000000-0000-2000-8080-808080808080").unwrap();

    let cmp_res = x.cmp(&y);
    assert_eq!(Ordering::Equal, cmp_res);

    assert_eq!(x, y); // this panics

    let compute_hash = |x: &CqlTimeuuid| {
        let mut hasher = DefaultHasher::new();
        x.hash(&mut hasher);
        hasher.finish()
    };
    assert_eq!(compute_hash(&x), compute_hash(&y)); // this panics
}
```

I'll implement the `Hash` trait as well.

scylla-cql/src/frame/value_tests.rs
muzarski: Actually this test is reduntant, since the test contained in `scylla` crate already tests it.

scylla-cql/src/frame/response/result.rs
```diff
@@ -993,10 +992,6 @@ mod tests {
         let uuid_serialize = super::deser_cql_value(&ColumnType::Uuid, uuid_slice).unwrap();
         assert_eq!(uuid_serialize, CqlValue::Uuid(my_uuid));
 
-        let time_uuid_serialize =
-            super::deser_cql_value(&ColumnType::Timeuuid, uuid_slice).unwrap();
-        assert_eq!(time_uuid_serialize, CqlValue::Timeuuid(my_uuid));
-
```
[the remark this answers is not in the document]
muzarski: I don't know what happened here - restored the test case.

scylla-cql/src/types/serialize/value.rs
```diff
@@ -223,7 +223,13 @@ impl SerializeCql for f64 {
 }
 impl SerializeCql for Uuid {
     impl_serialize_via_writer!(|me, typ, writer| {
-        exact_type_check!(typ, Uuid, Timeuuid);
+        exact_type_check!(typ, Uuid);
```
Lorak-mmk, line 226: Just a note: this change means that PR is API-breaking I think, we need to mark it as such so that it's properly described in release notes.

Lorak-mmk (approved):

piodul (changes requested):

scylla/src/transport/cql_types_test.rs
piodul: We should also use a random generator with known algorithm - `std::thread_rng` uses an unspecified algorithm and can probably change in the future. We already have tests with fixed pseudorandomness - for example, see `partitioners_output_same_result_no_matter_how_input_is_partitioned`.

piodul: Also, `362880` is not a huge number. Perhaps checking all the permutations would not take much time (let's aim under 100ms)? We can iterate over permutations using [`Itertools::permutations()`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.permutations).

muzarski:
v3:
- rebased on top of master
- addressed review comments

scylla/src/transport/cql_types_test.rs
```diff
@@ -1149,6 +1150,84 @@ async fn test_timeuuid() {
...
+
+    // Generate all permutations.
+    let perms = Itertools::permutations(sorted_timeuuid_vals.iter(), sorted_timeuuid_vals.len())
```
piodul, line 1193: Nit: `Itertools` is an extention trait for iterators. While what you wrote works, you are supposed to use it on the iterators directly:

```rust
let perms = sorted_timeuuid_vals
    .iter()
    .permutations(sorted_timeuuid_vals.len())
    .collect::<Vec<_>>();
```

scylla/src/transport/cql_types_test.rs
```diff
@@ -1149,6 +1150,84 @@ async fn test_timeuuid() {
...
+        .collect::<Vec<_>>();
+    // Ensure that all of the permutations were generated.
+    assert_eq!(362880, perms.len());
```
piodul, line 1196: Nit: the `perms` are going to take a few tens of megabytes of memory: an allocation of `362880 * 3 * 8 = ~8.3MB` of contiguous memory for the direct storage of `perms` alone (but may be up to twice as much due to vec exponential reallocation) and `362880 * 9 * 16 = ~50MB` for the timeuuids alone (spread over `362880` individual allocations).

While this is not very much for today's standards, you need to take into account that many test cases can run in parallel, so this memory usage adds up - and it's a good practice in general to be mindful of those allocations.

In this case, it could have been easily and completely avoided by not collecting the `permutations` iterator but rather consuming it in a loop.

Again, not a huge problem and not blocking the merge in my opinion. Can be fixed in a followup.

piodul (approved):
Left some nitpicks, but they aren't blocking the merge - can be fixed later. LGTM.

Issue #897, opened 2023-12-28
Compile error with `--no-default-features -F chrono`


oeb25:
When the `chrono` features is enabled and the `time` feature is not, the `scylla-cql` crate fails to compile with the following error:

```rs
❯ cargo check --no-default-features -F chrono
    Checking scylla-cql v0.0.11 (scylla-rust-driver/scylla-cql)
error[E0433]: failed to resolve: use of undeclared crate or module `time`
   --> scylla-cql/src/types/serialize/value.rs:176:23
    |
176 | impl SerializeCql for time::Date {
    |                       ^^^^ use of undeclared crate or module `time`

error[E0433]: failed to resolve: use of undeclared crate or module `time`
   --> scylla-cql/src/types/serialize/value.rs:183:23
    |
183 | impl SerializeCql for time::OffsetDateTime {
    |                       ^^^^ use of undeclared crate or module `time`

error[E0433]: failed to resolve: use of undeclared crate or module `time`
   --> scylla-cql/src/types/serialize/value.rs:190:23
    |
190 | impl SerializeCql for time::Time {
    |                       ^^^^ use of undeclared crate or module `time`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `scylla-cql` (lib) due to 3 previous errors
```

These are all annotated with `#[cfg(feature = "chrono")]` which perhaps should have been `#[cfg(feature = "time")]`. Changing these to the latter makes the crate compile; #898 proposes these changes.

PR #898, opened 2023-12-28
scylla-cql: Use `#[cfg(feature = "time")]` for `time` related `impl SerializeCql`'s

oeb25:
Fixes: #897, see the issue for details.

Additionally I went ahead and checked that `--no-default-features -F time` also works.

## Pre-review checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,6 +34,12 @@ jobs:
       run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features ""
     - name: Cargo check with all serialization features
       run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "full-serialization"
+    - name: Cargo check with secret feature
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "secret"
+    - name: Cargo check with chrono feature
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "chrono"
+    - name: Cargo check with time feature
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "time"
     - name: Build scylla-cql
       run: cargo build --verbose --all-targets --manifest-path "scylla-cql/Cargo.toml" --features "full-serialization"
     - name: Build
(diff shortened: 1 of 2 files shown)


piodul (commented):
Thanks.

When I reviewed the changes which introduced the `time` and `chrono` features, I was a bit too optimistic and thought that only checking compilation with and without all features would suffice. Would you mind adding some steps to `.github/workflows.yml` to check that everything compiles when only one of the `time`, `chrono`, `secret` features is enabled? You could base them on the existing `Cargo check with all serialization features` step.

oeb25:
Good idea! I've added three steps to the job, and it seems to work.

As a complete side note, I've found that [`- uses: Swatinem/rust-cache@v2`](https://github.com/Swatinem/rust-cache) can speed-up CI quite a bit! So if that's something that you are interested in I can open up a PR to measure the difference?

.github/workflows/rust.yml
```diff
@@ -34,6 +34,12 @@ jobs:
       run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features ""
     - name: Cargo check with all serialization features
       run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "full-serialization"
+    - name: Cargo check with all secret feature
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "secret"
+    - name: Cargo check with all chrono feature
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "chrono"
+    - name: Cargo check with all time feature
```
Lorak-mmk, line 41: Please remove "all" words from descriptions.

Lorak-mmk:
> Good idea! I've added three steps to the job, and it seems to work.
> 
> As a complete side note, I've found that [`- uses: Swatinem/rust-cache@v2`](https://github.com/Swatinem/rust-cache) can speed-up CI quite a bit! So if that's something that you are interested in I can open up a PR to measure the difference?

Sure, that would be great

.github/workflows/rust.yml
oeb25: Oops!

Lorak-mmk:
Code change looks good to me now. Could you squash 2 commits that modify CI?

oeb25:
Squashed!

Lorak-mmk (approved):
Looks good to me, @piodul ?

piodul (approved):

PR #900, opened 2023-12-30
scylla-cql, examples: fix clippy 1.75 issues

piodul:
With the new rust toolchain version comes a new clippy version, and therefore new complaints of the linter to address. This PR fixes the newly reported issues.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -941,7 +941,7 @@ fn serialize_empty(buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
 impl Value for CqlValue {
     fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
         match self {
-            CqlValue::Map(m) => serialize_map(m.iter().map(|(k, v)| (k, v)), m.len(), buf),
+            CqlValue::Map(m) => serialize_map(m.iter().map(|p| (&p.0, &p.1)), m.len(), buf),
             CqlValue::Tuple(t) => serialize_tuple(t.iter(), buf),
 
             // A UDT value is composed of successive [bytes] values, one for each field of the UDT
--- a/scylla-cql/src/types/serialize/value.rs
+++ b/scylla-cql/src/types/serialize/value.rs
@@ -488,7 +488,7 @@ fn serialize_cql_value<'b>(
         CqlValue::Map(m) => serialize_mapping(
             std::any::type_name::<CqlValue>(),
             m.len(),
-            m.iter().map(|(ref k, ref v)| (k, v)),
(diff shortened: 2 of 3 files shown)


scylla-cql/src/frame/value.rs
```diff
@@ -941,7 +941,7 @@ fn serialize_empty(buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
 impl Value for CqlValue {
     fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ValueTooBig> {
         match self {
-            CqlValue::Map(m) => serialize_map(m.iter().map(|(k, v)| (k, v)), m.len(), buf),
+            CqlValue::Map(m) => serialize_map(m.iter().map(|p| (&p.0, &p.1)), m.len(), buf),
```
Lorak-mmk, line 944: What lint does it fail? Can the change be made differently? Imo the old version is much easier to read - you know that the argument is 2-element tuple, in the new version you don't know that

scylla-cql/src/types/serialize/value.rs
```diff
@@ -488,7 +488,7 @@ fn serialize_cql_value<'b>(
         CqlValue::Map(m) => serialize_mapping(
             std::any::type_name::<CqlValue>(),
             m.len(),
-            m.iter().map(|(ref k, ref v)| (k, v)),
+            m.iter().map(|p| (&p.0, &p.1)),
```
Lorak-mmk, line 491: Same here

Lorak-mmk (changes requested):

scylla-cql/src/frame/value.rs
piodul: > What lint does it fail?

https://rust-lang.github.io/rust-clippy/master/index.html#map_identity

For example, I get this message:

```
warning: unnecessary map of the identity function
   --> scylla-cql/src/frame/value.rs:944:55
    |
944 |             CqlValue::Map(m) => serialize_map(m.iter().map(|(k, v)| (k, v)), m.len(), buf),
    |                                                       ^^^^^^^^^^^^^^^^^^^^^ help: remove the call to `map`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#map_identity
    = note: `#[warn(clippy::map_identity)]` on by default
```

See the commit description for more details.

> Imo the old version is much easier to read - you know that the argument is 2-element tuple, in the new version you don't know that

I think it's quite clear here that `m.iter()` iterates over pairs, so I'm not sure if it's a huge problem. I'd argue that the code is now more clear as it's now more obvious what the lambda in the `.map` does.

> Can the change be made differently?

Perhaps it can, but I'm not sure how would you like me to change it.

Lorak-mmk (approved):

Issue #679, opened 2023-03-26
`num_bigint::BigInt: Value` is not satisfied error when compiling varint example from docs


j-chmielewski:
[Varint](https://rust-driver.docs.scylladb.com/stable/data-types/varint.html) example throws compilation errors.

My system:

```
Linux blade 6.2.2-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 03 Mar 2023 15:58:31 +0000 x86_64 GNU/Linux
```

Code:
```rust
use num_bigint::BigInt;
use scylla::{IntoTypedRows, SessionBuilder};
use std::{str::FromStr, error::Error};

#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>>{
    // Insert a varint into the table
    let to_insert: BigInt = BigInt::from_str("12345")?;
    let session = SessionBuilder::new().known_node("locahost:9042").build().await.expect("Connection failed");
    session
        .query("INSERT INTO keyspace.table (a) VALUES(?)", (to_insert,))
        .await?;

    // Read a varint from the table
    if let Some(rows) = session
        .query("SELECT a FROM keyspace.table", &[])
        .await?
        .rows
    {
        for row in rows.into_typed::<(BigInt,)>() {
            let (varint_value,): (BigInt,) = row?;
        }
    }

    Ok(())
}
```

[Steps to reproduce](https://github.com/j-chmielewski/scylla-varint-repro)

Error:

```
error[E0277]: the trait bound `num_bigint::BigInt: Value` is not satisfied                                                                                                                                         
   --> src/main.rs:11:60                                                                                                                                                                                           
    |                                                                                                                                                                                                              
11  |         .query("INSERT INTO keyspace.table (a) VALUES(?)", (to_insert,))                                                                                                                                     
    |          ----- required by a bound introduced by this call ^^^^^^^^^^^^ the trait `Value` is not implemented for `num_bigint::BigInt`                                                                        
    |                                                                                                                                                                                                              
    = help: the following other types implement trait `Value`:                                                                                                                                                     
              &T                                                                                                                                                                                                   
              &[T]                                                                                                                                                                                                 
              &str                                                                                                                                                                                                 
              (T0, T1)                                                                                                                                                                                             
              (T0, T1, T2)                                                                                                                                                                                         
              (T0, T1, T2, T3)                                                                                                                                                                                     
              (T0, T1, T2, T3, T4)                                                                                                                                                                                 
[...]

Issue #771, opened 2023-07-28
Remove types from pre-1.0 crates from the public API, or hide them behind features


piodul:
Currently, the driver has a number of dependencies on crates that did not commit to the API stability, i.e. their latest releases are pre-1.0. Those crates do sometimes release new major versions and we usually, after some time, update those dependencies as well. However, updating dependencies on crates whose types appear in our public API generates friction for users: they may need to update the dependencies in their projects as well. This can require some effort, depending on how much stuff has been modified in the dependencies. Even if the effort is not large, the fact that we need to do it regularly prevents us from stabilising _our_ API.

We should approach this as follows:

- "Unstable" types should be removed from the signatures of public functions and should not appear as fields of our public structs/enums, in case when no features are enabled.
- Each unstable dependency should be hidden behind a feature. There should be a separate feature for each version of the dependency, e.g. `num-bigint_03` for version 0.3 and `num-bigint_04` for v0.4. This will allow us to introduce support for newer versions without breaking support for the old ones, as we do now. If we feel brave enough, we can consider deprecating support for the old versions and eventually remove them when they are no longer supported by the authors.
  If such a feature is enabled, the API that it brings can, of course, depend on the unstable crate. This also includes trait implementations, e.g. `impl FromCqlVal for num_bigint::BigInt`.

The biggest offender seems to be the `CqlValue`. This PR: https://github.com/scylladb/scylla-rust-driver/pull/745 seems to be a step in the right direction to fixing this type. Of course, there may be more issues and we should vet all current pre-1.0 dependencies.

Lorak-mmk:
Why limit this to pre-1.0 crates?
Crates that already released 1.0 may release 2.0 in the future and we will face the same problem (assuming we are on 1.0 too) - either we break SemVer, release 2.0, or force users to use old version of dependent crate.

piodul:
FWIW, the issue was inspired by this guideline: https://rust-lang.github.io/api-guidelines/necessities.html#public-dependencies-of-a-stable-crate-are-stable-c-stable , and the trick with the features was inspired by what the MongoDB driver did: https://docs.rs/mongodb/2.8.0/mongodb/#all-feature-flags .

[...]

PR #902, opened 2024-01-03
Hide public usages of `num_bigint::BigInt` behind a feature flag

muzarski:
Related issue: https://github.com/scylladb/scylla-rust-driver/issues/771 (not fixed yet, as `bigdecimal::BigDecimal` is an unstable type as well).

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/679

# Changes:
- added `CqlVarint` type which is a simple native representation of cql `varint` type
- wrapped `CqlVarint` with `CqlValue::Varint`
- hidden public usages of unstable types (`num_bigint` - v0.3 and v0.4) behind feature flags. Notice, however, that `num_bigint v0.3` is still used within driver's code even without any feature flags. We make use of some library utilities such as pretty printing, or normalization of value's binary representation (e.g. getting rid of leading zeros).

## Pre-review checklist


scylla-cql/src/frame/response/cql_to_rust.rs
```diff
@@ -125,7 +125,7 @@ impl_from_cql_value_from_method!(i32, as_int); // i32::from_cql<CqlValue>
 impl_from_cql_value_from_method!(i64, as_bigint); // i64::from_cql<CqlValue>
 impl_from_cql_value_from_method!(Counter, as_counter); // Counter::from_cql<CqlValue>
 impl_from_cql_value_from_method!(i16, as_smallint); // i16::from_cql<CqlValue>
-impl_from_cql_value_from_method!(BigInt, into_varint); // BigInt::from_cql<CqlValue>
+impl_from_cql_value_from_method!(CqlVarint, into_cqlvarint); // CqlVarint::from_cql<CqlValue>
```
piodul, line 127: Nit: `into_cql_varint` - other getters separate `cql` from the rest by `_`

scylla/src/utils/pretty.rs
```diff
@@ -51,7 +52,11 @@ where
             CqlValue::Inet(i) => write!(f, "'{}'", i)?,
             CqlValue::SmallInt(si) => write!(f, "{}", si)?,
             CqlValue::TinyInt(ti) => write!(f, "{}", ti)?,
-            CqlValue::Varint(vi) => write!(f, "{}", vi)?,
+            CqlValue::Varint(vi) => write!(
+                f,
+                "{}",
+                BigInt::from_signed_bytes_be(vi.to_signed_bytes_be_slice())
```
piodul, line 58: I'd like to get rid of the dependency on num_bigint v3 here. I'd rather not reimplement printing bigint values in base 10 as the algorithm is not that simple (at least the one in num_bigint itself isn't), but maybe we can just print in hexadecimal here?

Preferably, this should be done in a CQL-compatible syntax (i.e. you can put the value formatted in this way into a query string as a literal and it should work), but it's not mandatory if it needs something complicated like printing in base 10.

scylla-cql/Cargo.toml
```diff
@@ -21,6 +21,7 @@ uuid = "1.0"
 thiserror = "1.0"
 bigdecimal = "0.2.0"
 num-bigint-03 = { package = "num-bigint", version = "0.3" }
```
piodul, line 23: Apart from the `CqlValueDisplayer`, what else in the driver currently requires this crate?

scylla-cql/src/frame/response/result.rs
```diff
@@ -1090,6 +1090,71 @@ mod tests {
...
+            },
+            Test {
+                value: (-128).to_bigint().unwrap(),
+                encoding: &[0x80],
+            },
+            Test {
+                value: (-129).to_bigint().unwrap(),
+                encoding: &[0xFF, 0x7F],
+            },
+        ];
```
piodul: I don't like the amount of copy paste here... Can we extract these test cases from `test_varint03` and `test_varint04` somewhere else, e.g. create a function that returns the test cases? The type of `value` could be changed to `Vec<u8>` probably.

scylla-cql/src/frame/value_tests.rs
```diff
@@ -124,6 +124,32 @@ fn bigint03_serialization() {
...
+    let cases_from_the_spec: &[(i64, Vec<u8>)] = &[
+        (0, vec![0x00]),
+        (1, vec![0x01]),
+        (127, vec![0x7F]),
+        (128, vec![0x00, 0x80]),
+        (129, vec![0x00, 0x81]),
+        (-1, vec![0xFF]),
+        (-128, vec![0x80]),
+        (-129, vec![0xFF, 0x7F]),
+    ];
```
piodul: This is the fourth test case that depends on those cases and the fourth time they appear in the code. Please deduplicate.

scylla/src/transport/cql_types_test.rs
```diff
@@ -122,6 +122,25 @@ async fn test_varint() {
...
+        "1",
+        "127",
+        "128",
+        "129",
+        "-1",
+        "-128",
+        "-129",
+        "123456789012345678901234567890",
+        "-123456789012345678901234567890",
+    ];
```
piodul, line 139: Would be nice to deduplicate those as well

scylla-cql/src/frame/value.rs
```diff
@@ -51,6 +52,52 @@ pub enum MaybeUnset<V> {
...
+    pub fn from_signed_bytes_be(digits: &[u8]) -> Self {
+        let normalized_bigint = BigInt::from_signed_bytes_be(digits);
+        Self(normalized_bigint.to_signed_bytes_be())
+    }
+
+    pub fn to_signed_bytes_be(&self) -> Vec<u8> {
+        self.0.clone()
+    }
+
+    pub fn to_signed_bytes_be_slice(&self) -> &[u8] {
```
piodul, line 74: Please add docstrings to the public methods.

piodul (changes requested):

scylla-cql/Cargo.toml
muzarski: The constructors of `CqlVarint` make use of the crate. The reason is that the `num_bigint` library normalizes the user provided `u8` slice/array.

For example:
```rust
pub fn from_signed_bytes_be(digits: &[u8]) -> Self {
        let normalized_bigint = num_bigint_03::BigInt::from_signed_bytes_be(digits);
        Self(normalized_bigint.to_signed_bytes_be())
}
```

I wasn't sure if this is necessary as Scylla accepts non-normalized bytes (with leading zeros) and stores them as is (result of the query on the inserted data has leading zeros as well).

The only issue is a following client-side scenario:
```rust
let one_with_leading_zeros_be = vec![0x00, 0x00, 0x00, 0x01];
let cql_varint: CqlVarint = CqlVarint::from_signed_bytes_be(&one_with_leading_zeros_be);
let bigint: num_bigint::BigInt = cql_varint.clone().into();
let converted_cql_varint: CqlVarint = bigint.into();

// This panics!
// cql_varint.signed_bytes_be().len() == 4
// converted_cql_varint.to_signed_bytes_be().len() == 1
assert_eq!(cql_varint.to_signed_bytes_be(), converted_cql_varint.to_signed_bytes_be());
```

If we want the dependency to be optional, I see two ways to achieve it:
1. ignore above issue and insert non-normalized data to DB
2. reimplement normalization algorithm in the driver

piodul: I don't think that lack of normalization should be a big issue. Converting the CqlVarint to a useful form (i.e. BigInt) will normalize it. Besides, in order to create a non-normalized value one has to create the CqlVarint from bytes - I think this is rare.

muzarski:
v3: adjusted `Cargo.lock.msrv` file.

muzarski:
v4:
implemented `From<Vec<u8>> for CqlVarint` and `From<CqlVarint> for Vec<u8>`

scylla/Cargo.toml
```diff
@@ -66,6 +67,7 @@ tracing-subscriber = { version = "0.3.14", features = ["env-filter"] }
 assert_matches = "1.5.0"
 rand_chacha = "0.3.1"
 time = "0.3"
+num-bigint-04 = { package = "num-bigint", version = "0.4" }
```
piodul, line 70: Please put it higher, near `num-bigint-03`

piodul (changes requested):

muzarski:
v5: fixes based on review comments

muzarski:
v6: removed `derive(Eq)` from `CqlVarint` type

scylla/src/transport/cql_types_test.rs
```diff
@@ -116,9 +116,103 @@ async fn test_varint() {
         "-129",
         "123456789012345678901234567890",
         "-123456789012345678901234567890",
```
Lorak-mmk, line 118: You could add a test with bigger number, that can't be contained in any primitive type (this number is in range of i128).

scylla/src/transport/cql_types_test.rs
```diff
@@ -116,9 +116,103 @@ async fn test_varint() {
...
+            .unwrap()
+            .rows
+            .unwrap()
+            .into_typed::<(CqlVarint,)>()
+            .map(Result::unwrap)
+            .map(|row| row.0)
+            .collect::<Vec<_>>();
+
+        assert_eq!(read_values, vec![cql_varint])
+    }
```
Lorak-mmk, line 215: Why it can't be expressed as a call to `run_tests`?

scylla/Cargo.toml
```diff
@@ -57,6 +58,8 @@ rand_pcg = "0.3.1"
 socket2 = { version = "0.5.3", features = ["all"] }
 
 [dev-dependencies]
+num-bigint-03 = { package = "num-bigint", version = "0.3" }
+num-bigint-04 = { package = "num-bigint", version = "0.4" }
```
Lorak-mmk, line 62: Why do we need those? I thought it's for cql_types_test, but those previously used `num_bigint` so what changed?

Lorak-mmk (changes requested):

scylla/src/transport/cql_types_test.rs
muzarski: As `run_tests` requires that the type implements `std::str::FromStr`. It's not trivial to implement it without the use of the `num_bigint` crate.

scylla/Cargo.toml
muzarski: I just introduced test cases for both versions of the library. Previously we only had test cases for `num_bigint v0.3`.

scylla/src/transport/cql_types_test.rs
muzarski: Implementing `FromStr` for hexadecimal numbers probably wouldn't be hard, but I'm not sure if it's necessary.

Note that `num_bigint` parses the strings representing the number in decimal format.

muzarski:
v7: added test cases for varints that can't be contained in any primitive type

scylla-cql/src/frame/value.rs
```diff
@@ -51,6 +51,81 @@ pub enum MaybeUnset<V> {
...
+        match digits.last() {
+            // Positive number. Most-significant bit is 1, so we need to append a 0x00 byte.
+            Some(x) if x > &0x7f && has_leading_zero => digits.push(0x00),
+            // Negative number - leave as is.
+            Some(_) => (),
+            // Empty vector (or filled with zeros) - represent as 0.
+            None => digits.push(0x00),
+        }
+
+        digits.reverse();
```
Lorak-mmk, line 97: Using `VecDeque` would allow to skip those `reverse()` calls. Otoh it would require `[make_contiguous](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.make_contiguous)` call in `as_signed_bytes_be_slice`, so there would be some overhead on first call. Wdyt?

scylla/Cargo.toml
Lorak-mmk: Previously there was no `num_bigint` dependency in dev-dependencies, now there are two. If the previous test cases didn't need `num_bigint` there why the new ones do?
Is it added so that in test those features are always enabled? If so, I don't think that's a good approach - we should be able to test with and without any feature, it should be controlled using command line options.

scylla/src/transport/cql_types_test.rs
Lorak-mmk: > As `run_tests` requires that the type implements `std::str::FromStr`. It's not trivial to implement it without the use of the `num_bigint` crate.
> 
> CQL accepts `varint` string literals represented in decimal/octal. Implementing `FromStr` for hexadecimal numbers wouldn't be hard but the hexadecimal strings provided to `run_tests` couldn't be contained as literals in queries.

I don't understand the last sentence. Why couldn't they?

Lorak-mmk (commented):
.

scylla/Cargo.toml
muzarski: > Previously there was no num_bigint dependency in dev-dependencies

You are right, but there was one in `[dependencies]` of scylla crate. The only things the dependency was used for were tests and `CqlValueDisplayer` (`scylla/src/utils/pretty.rs`).

Now, the dependency is not needed for `CqlValueDisplayer`, and so I moved it to `dev-dependencies`, so the tests can still use it (now we have two dependencies, one dependency per crate version).

> Is it added so that in test those features are always enabled?

Note that `dev-dependencies` cannot be specified as optional dependencies (and activated via feature flag). However, the test cases with specific `num-bigint` version are still enabled only when the corresponding feature flag is enabled, e.g.:
```rust
#[cfg(feature = "num-bigint-03")]
#[tokio::test]
async fn test_varint03() {
    let tests = varint_test_cases();
    run_tests::<num_bigint_03::BigInt>(&tests, "varint").await;
}
```

scylla/src/transport/cql_types_test.rs
muzarski: Because you can't put a hexadecimal literal for a value of type varint in CQL query:
```cql
cqlsh> CREATE TABLE ks.t (a int PRIMARY KEY, b varint);
cqlsh> INSERT INTO ks.t (a, b) VALUES (0, 0x1);
InvalidRequest: Error from server: code=2200 [Invalid query] message="Invalid HEX constant (0x1) for "b" of type varint"
```

scylla-cql/src/frame/value.rs
muzarski: It would also require that `as_signed_bytes_be_slice` accepts `&mut self` instead of `&self`. This looks a bit strange to me.

I'm really not sure which approach is better here. Maybe @piodul has a stronger opinion on that.

piodul: No need to move data around so much or involve `VecDeque`. I propose this algorithm:

- Find the first non-zero digit `d` and its position `p`,
- If all digits are zero, resize/truncate the vec to size 1,
- If `d > 0x7f` then cut the first `p - 1` chars,
- Otherwise, cut the first `p` chars.

Popping multiple items from the front can be done via `Vec::drain`.

It might be worth checking whether `Vec::drain` on a zero-sized range is a no-op or not. If it's `O(N)`, then we should have an explicit check at the beginning whether anything needs to be done or not.

scylla/src/transport/cql_types_test.rs
piodul: I just found out that you actually can, with `blobAsVarint` built-in CQL function:

```
cqlsh> create table ks.t (pk int, v varint, primary key (pk));
cqlsh> insert into ks.t (pk, v) values (2, blobasvarint(0x123421934815718293482108421905702184921834021943));
cqlsh> select pk, v from ks.t;

 pk | v
----+-----------------------------------------------------------
  2 | 446351888295288528376725140517290295205369310673010563395

(1 rows)

```

scylla/src/utils/pretty.rs
```diff
@@ -51,7 +51,7 @@ where
             CqlValue::Inet(i) => write!(f, "'{}'", i)?,
             CqlValue::SmallInt(si) => write!(f, "{}", si)?,
             CqlValue::TinyInt(ti) => write!(f, "{}", ti)?,
-            CqlValue::Varint(vi) => write!(f, "{}", vi)?,
+            CqlValue::Varint(vi) => write!(f, "0x{:x}", HexBytes(vi.as_signed_bytes_be_slice()))?,
```
piodul, line 54: Suggestion: change the format string to `"blobAsVarint(0x{:x})"` - it will be actually be a valid cql expression.

scylla-cql/src/frame/value.rs
```diff
@@ -51,6 +51,81 @@ pub enum MaybeUnset<V> {
...
+/// Represented as two's-complement binary in big-endian order.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct CqlVarint(Vec<u8>);
+
+/// Constructors from bytes
+impl CqlVarint {
+    /// Creates a [`CqlVarint`] from an array of bytes in
+    /// two's complement big-endian binary representation.
+    pub fn from_signed_bytes_be(mut digits: Vec<u8>) -> Self {
+        Self::normalize(&mut digits);
```
piodul, line 64: @muzarski could you remind me why we are normalizing? We discussed it offline some time ago, but I don't remember it now. Please include the rationale in a comment in the code.

scylla/Cargo.toml
Lorak-mmk: Hmm... I can't say I love it, but no better solution comes to my mind right now. I think we can revisit if number of such dependencies grows too large.

scylla-cql/src/frame/value.rs
muzarski: We decided to normalize because `CqlValue` enum derives `PartialEq/Eq`.

It means that our new type `CqlVarint` needs to implement/derive `PartialEq/Eq` as well. The issue with this approach: take for example `[0x00, 0x01]` and `[0x01]`. If we derived `PartialEq`, the corresponding CqlVarints wouldn't be equal. However, both of these blob arrays represent the same varint value - simply 1. If we wanted them to be equal in our semantics, we would have to implement normalization anyway.

piodul: Ok, right.

I'm thinking about an alternative: if the only issue is that we need to have `PartialEq`/`Eq` so that it works with `CqlValue`, perhaps we could skip normalization during initialization? We could normalize the values when we compare (i.e. compute a subslice which represents the normalized number and compare them).

I suppose that comparing `CqlValue`s is quite rare, while normalizing when creating the `CqlVarint` introduces some overhead to a more common path. I think this would also be a bit more consistent with the behavior of `CqlTimeuuid` in that you can pass denormalized data (in the case of CqlTimeuuid, non-timeuuid uuid).

What do you think?

piodul: cc: @Lorak-mmk

muzarski: > I suppose that comparing `CqlValue`s is quite rare, while normalizing when creating the `CqlVarint` introduces some overhead to a more common path.

I agree. In addition, during comparison we only need to compute the subslice instead of removing elements from `Vec` - which is much cheaper.

I have an additional question regarding one edge case. What should we do with the empty vectors? Replace them with `vec![0x00]` or leave as is? The empty vector is serialized to `null`, and I'm not sure if this should be expected behaviour.

piodul: > I have an additional question regarding one edge case. What should we do with the empty vectors? Replace them with `vec![0x00]` or leave as is? The empty vector is serialized to `null`, and I'm not sure if this should be expected behaviour.

Are you sure that they are serialized as nulls? Nulls have a different binary representation than zero-sized values: https://github.com/apache/cassandra/blob/f6707e6191a994371704f639323a31f9ecc6bcb5/doc/native_protocol_v4.spec#L229-L233

muzarski: OK, I have reproduced that and you are right. This is not a `null` value. However, the following test case (without normalization) doesn't pass:

```rust
#[tokio::test]
async fn test_cql_varint() {
    let tests = [
        vec![],
    ];

    let table_name = "cql_varint_tests";
    let session: Session = create_new_session_builder().build().await.unwrap();
    let ks = "foo_varint";

    session
        .query(
            format!(
                "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = \
            {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}",
                ks
            ),
            &[],
        )
        .await
        .unwrap();
    session.use_keyspace(ks, false).await.unwrap();

    session
        .query(
            format!(
                "CREATE TABLE IF NOT EXISTS {} (id int PRIMARY KEY, val varint)",
                table_name
            ),
            &[],
        )
        .await
        .unwrap();

    let prepared_insert = session
        .prepare(format!(
            "INSERT INTO {} (id, val) VALUES (0, ?)",
            table_name
        ))
        .await
        .unwrap();
    let prepared_select = session
        .prepare(format!("SELECT val FROM {} WHERE id = 0", table_name))
        .await
        .unwrap();

    for test in tests {
        let cql_varint = CqlVarint::from_signed_bytes_be_slice(&test);
        session
            .execute(&prepared_insert, (&cql_varint,))
            .await
            .unwrap();

        let read_values: Vec<CqlVarint> = session
            .execute(&prepared_select, &[])
            .await
            .unwrap()
            .rows
            .unwrap()
            .into_typed::<(CqlVarint,)>()
            .map(Result::unwrap)
            .map(|row| row.0)
            .collect::<Vec<_>>();

        assert_eq!(read_values, vec![cql_varint])
    }
}
```

The error message:
```
called `Result::unwrap()` on an `Err` value: BadCqlVal { err: BadCqlType, column: 0 }
```

cqlsh output:
```cql
cqlsh> SELECT * FROM foo_varint.cql_varint_tests ;

 id | val
----+-----
  0 |    

```

However, if I change the type of `read_values` to `Vec<Option<CqlVarint>>`, and replace assertion with `assert_eq!(read_values, vec![None])` the test passes.

muzarski: The query result from the test is deserialized as `CqlValue::Empty`: https://github.com/apache/cassandra/blob/f6707e6191a994371704f639323a31f9ecc6bcb5/doc/native_protocol_v4.spec#L861-L863. Should this be expected behaviour for `CqlVarint::from_signed_bytes_be(vec![])`?

piodul: I think this is fine. I'd still keep `CqlVarint` as a simple wrapper over a raw representation and not do any normalization at all. The protocol specification only describes normalized values, it doesn't really say what happens to the denormalized values. Maybe ScyllaDB converts them to nulls for some reason (maybe Cassandra does the same and it's about compatibility).

piodul:
Now that https://github.com/scylladb/scylla-rust-driver/pull/894 was merged, there are conflicts with main. Please rebase.

scylla-cql/src/frame/value.rs
```diff
@@ -51,6 +51,68 @@ pub enum MaybeUnset<V> {
     Set(V),
 }
 
+/// Native CQL `varint` representation.
+///
+/// Represented as two's-complement binary in big-endian order.
```
piodul, line 55: I think it would be useful to tell more about the purpose of this struct:

- It's a "default" implementation, independent of any external crate or crate feature, most prominently used in CqlValue
- Other, more useful types like `num_bigint::BigInt` can be used for serialization/deserialization, but they require enabling crate features.

scylla-cql/src/frame/value.rs
```diff
@@ -64,7 +64,10 @@ pub enum MaybeUnset<V> {
 /// Bytes provided by the user via constructor are passed to DB as is.
 /// As of now, users are guaranteed to fetch the exact same data they inserted
 /// previously (even if it was not normalized).
-#[derive(Clone, Debug)]
+///
+/// The implementation of [`PartialEq`], however, normalizes the underlying bytes
+/// before comparison. For details, check [examples](#impl-PartialEq-for-CqlVarint).
+#[derive(Clone, Eq, Debug)]
```
piodul, line 69: Does `num_bigint::BigInt` implement `Hash`? If so, then I think we should implement it, too.

scylla-cql/src/frame/value.rs
```diff
@@ -101,6 +104,42 @@ impl CqlVarint {
     }
 }
 
+impl CqlVarint {
+    fn as_normalized_slice(&self) -> &[u8] {
+        let digits = self.0.as_slice();
+        if digits.is_empty() {
+            return digits;
+        }
```
piodul, line 111: `num_bigint::BigInt` seems to convert empty slices into a zero, see here: https://docs.rs/num-bigint/latest/src/num_bigint/bigint/convert.rs.html#385. So, 

```
CqlVarint::from_signed_bytes_be(vec![]) != CqlVarint::from_signed_bytes_be(vec![0])
```
but

```
BigInt::from_signed_bytes_be(vec![]) == BigInt::from_signed_bytes_be(vec![0])
```

I think it would make the structure more well-behaved if we exactly preserve properties such as equality or ordering when we map to BigInt, so here I think we should `return &[0]`.

scylla-cql/src/frame/value.rs
```diff
@@ -101,6 +104,42 @@ impl CqlVarint {
...
+
+        let non_zero_position = match digits.iter().position(|b| *b != 0) {
+            Some(pos) => pos,
+            None => {
+                // Vector is filled with zeros. Represent it as 0.
+                // SAFETY: digits.is_empty() check at the beginning ensures
+                // that digits.len() >= 1.
+                return &digits[digits.len() - 1..];
```
piodul, line 119: Nit: you can just `return &[0]` and not have to explain why this is panic-safe.

scylla-cql/src/frame/value_tests.rs
```diff
@@ -72,6 +72,30 @@ fn counter_serialization() {
...
+        (vec![0x00, 0x80], vec![0x00, 0x80]), // 128
+        (vec![0xff], vec![0xff]),             // -1
+        (vec![0x00, 0xff], vec![0x00, 0xff]), // 255
+    ];
+
+    for test in test_cases {
+        let non_normalized = CqlVarint::from_signed_bytes_be(test.0);
+        let normalized = CqlVarint::from_signed_bytes_be(test.1);
+
+        assert_eq!(non_normalized, normalized);
```
piodul, line 95: Would be nice to check here that these properties hold after conversion to BigInt, too.

scylla/src/transport/cql_types_test.rs
piodul: Perhaps later we can adjust `run_tests` to take a list of (Cql?)Values instead of strings, use `CqlValueDisplayer` to print them when putting them into the query and finally compare the values fetched from the database.

piodul (changes requested):
Please add steps to the `rust.yml` workflow which `cargo check` the crate when only `num-bigint-03` or `num-bigint-04` features are enabled.

muzarski:
v9: rebased on top of master so the upcoming diff addressing review comments is clearer

scylla/src/transport/cql_types_test.rs
muzarski: Can we do it in a separate PR? I can create an issue regarding this.

piodul: > Can we do it in a separate PR?

Yes, that's what I meant by "later".

muzarski:
v10: addressed review comments

scylla-cql/src/frame/value_tests.rs
```diff
@@ -110,13 +109,14 @@ fn cql_varint_normalization() {
     }
 }
 
+#[cfg(feature = "num-bigint-03")]
 #[test]
-fn cql_varint_normalization_with_bigint() {
+fn cql_varint_normalization_with_bigint03() {
```
piodul, line 114: Nit: there is no corresponding test for `0.4`. I just checked that the normalization behavior is the same in `0.4`, fortunately. The test case can be added in a follow-up.

piodul (approved):
LGTM

scylla-cql/src/frame/value_tests.rs
muzarski: Yes, I've also checked that normalization works the same in this case. Wasn't sure if such test case is redundant or not.

Lorak-mmk (approved):

Issue #904, opened 2024-01-04
Documentation for proc macros not rendered correctly


Lorak-mmk:
Documentation for `SerializeCql` and `SerializeRow` is not rendered properly.
https://docs.rs/scylla/latest/scylla/derive.SerializeCql.html
https://docs.rs/scylla/latest/scylla/derive.SerializeCql.html

Both docs say `See the documentation for this item in the scylla crate.` - which we are obviously looking at.

I'm not sure why that happens. `scylla` re-exports those items from `scylla-cql` crate, which re-exports them from `scylla-macros` and documents them.
Documentation for `scylla-cql` renders correctly, so I don't get why `scylla` uses docstrings directly from `scylla-macros`. @piodul any ideas?

piodul:
I'm not sure either. Perhaps the behavior of `cargo doc` changed in some version.

The reason why the "proper" documentation for those macros is in `scylla-cql` (the docstrings are actually there, not in `scylla`) is because the code examples need stuff from `scylla-cql`, so putting them in `scylla-macros` does not work out of the box.

We should check if there is a way to put the docstrings into `scylla-macros` and make them compile. If that fails, we should see whether we can control the behavior of `cargo doc` and tell it to use the docstring from `scylla-cql`.

Lorak-mmk:
I tried to get the right documentation to render in `scylla` without moving the comments, without success. No combination of `#[doc(inline)]`, `#[doc(no_inline)]` and changing exports (e.g. importing SerializeRow / SerializeCql directly apart from the wildcard import) did the trick.

Do we need those examples to compile? We have other tests for those macros, so imo the easiest solution is to move the docs to scylla_macros crate and mark examples as `ignore`. This would get rid of another problem that I don't think can be solved differently: even in scylla_cql crate, the placeholder comment from scylla_macros is appended to docs. So the end of SerializeCql docs is currently:
```
#[scylla(rename = "name_in_the_udt")]

Serializes the field to the UDT struct field with given name instead of its Rust name. See the documentation for this item in the scylla crate.
```

Note the last sentence, that should not be there.

piodul:
> Do we need those examples to compile? We have other tests for those macros, so imo the easiest solution is to move the docs to scylla_macros crate and mark examples as ignore.

It's not about testing the implementation, it's about making sure that examples in the documentation are up-to-date. I'd rather avoid marking them as "ignore" so that we don't accidentally suggest incorrect code to the users.

[...]

PR #907, opened 2024-01-09
Move derive macros docs to scylla crate

Lorak-mmk:
This PR moves docs for serialization / deserialization related derive macros from scylla-cql crate to scylla crate.
This is done to work around the issue that this documentation was not rendered correctly on docs.rs for scylla crate.
It's hard to tell why that is exactly - it's probably a bug / limitation in rustdoc.
It looks like when we have item X in crate A, re-export X as Y in crate B and document it, and re-export Y az Z in crate C,
Y's docs won't be rendered in crate C, only X's and Z's will.

Fixes: #904 

## Pre-review checklist


Lorak-mmk:
After this is merged, we should release a new version, so that users are able to find documentation.

scylla/src/macros.rs
```diff
@@ -10,7 +10,7 @@ pub use scylla_cql::macros::FromUserType;
 /// Works only on simple structs without generics etc
 pub use scylla_cql::macros::IntoUserType;
 
-/// Derive macro for the [`SerializeCql`](crate::types::serialize::value::SerializeCql) trait
```
piodul, line 16: Nit (last commit): typo in the commit message: `scyll` -> `scylla`

scylla-macros/src/lib.rs
```diff
@@ -9,7 +9,10 @@ mod value_list;
 
 mod serialize;
 
-/// See the documentation for this item in the `scylla` crate.
+///
+/// Documentation for this macro can only be found
+/// in `scylla` crate - not in scylla-macros nor in scylla-cql.
+/// This is because if rustdocs limitations that are hard to explain here.
```
piodul, line 15: Typo: `if` -> `of`, here and in other docstrings in the same commit

piodul (changes requested):
While the documentation in the `scylla` crate looks much better, it looks like the docstrings from multiple crates are still being squashed. For example, the message about the macro being documented in the `scylla` crate still appears in `scylla::macros::ScyllaCql`:

image

However, I can't find anything on the net about overriding docstrings on reexport, so I can't suggest any way to get rid of them (if it is possible at all). As we discussed on slack, let's at least make the message from `scylla-macros` more separated from the rest by putting a separator at the end of the docstrings in `scylla`.

Lorak-mmk:
v2
- Addresed review comments (fixed typos, added separators in scylla crate instead of empty lines in scylla-macros)
- Added empty lines in documentation of old macros, so that only the first line is shown in the summary in the main page of documentation.
- Fixed formatting of commit messages

Issue #906, opened 2024-01-05
Introduce cargo_semver_checks


Lorak-mmk:
In order to decrease the risk of accidentaly introducing breaking changes in minor version bump, we should use cargo_semver_checks ( https://crates.io/crates/cargo-semver-checks ) in CI.
My proposed usage:
- In each PR it would be ran against PR base, and if there are breaking changes, set a label on a PR / add a comment to PR
- For each pushed tag it would run against the previous version (e.g. for tag v0.11.1 it would compare to v0.11.0, for v0.11.0 it would compare to v0.10.x where x is latest version) to see if the version updates were correct.
- There would be convenience script to run locally before pushing a tag, in order to avoid having to pull a tag if the version updates are deemed incorrect.

PR #909, opened 2024-01-10
Semver checks

Lorak-mmk:
This PR introduces cargo-semver-checks to our CI.
The behavior is very similar to what I described in the issue: https://github.com/scylladb/scylla-rust-driver/issues/906
- On a PR, it sets `semver-checks-breaking` label if the PR breaks API (compared to PR base). Report from the tool, containing info about incompatibilities, is posted as a comment.
- When a tag is pushed, the job will fail if update type (major / minor / patch) was incorrect. This requires new version to NOT be published to crates.io. So we should wait for the workflow to finish before publishing.

Apart from CI integration this PR also modifies `Makefile` and `CONTRIBUTING.MD` to make it easier for contributors to run checks locally.

cargo-semver-checks has some false-negatives and false-positives, so it is NOT a replacement for a reviewer.
Reviewer still needs to check the code to see if it's API-breaking. semver-checks should still be a big help with that, catching many cases that
are very hard to identify manually or that people forget about.

Security:

PR CI uses `pull_request_target` trigger which can receive read-write GITHUB_TOKEN (while `pull_request` can't if a PR is from a fork - which is a vast majority of PRs).
This is required in order to be able to add/remove label to a PR. It makes such jobs a valuable target for attackers.

Workflows using `pull_request_target` are run in the context of base of a PR instead of head of a PR (like `pull_request` workflows do). This prevents the attacker from modifying the workflow file.

PR CI is split into two parts. First, unprivileged, job (called `semver-pull-request-check`) performs actual check, which is an insecure process (for example: attacker could add malicious `build.rs` file).
For this reason, this first job has no permissions (`permissions: {}` attribute), so taking over it should not give anything interesting to the attacker.

The second, privileged, job (called `semver-pull-request-label`) has write permissions for pull requests and its only task is adding or removing a label to/from a PR.
It checks the output from the first job and labels a PR based on that. The only thing done with outputs from unprivileged job is a simple comparison with "0",
so modifying those outputs will not result in takeover of privileged job.

As far as I can tell, from GitHub docs each job is run on a clean VM, so taking over unprivileged job should not allow attacker to take over privileged job.

Fixes: #906 

## Pre-review checklist

### DIFF
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -7,6 +7,9 @@ on:
     branches:
     - main
     - 'branch-*'
+  push:
+    tags:
+      - v*.*.*
 
 env:
   CARGO_TERM_COLOR: always
@@ -79,3 +82,14 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         GH_REPO: ${{ github.repository }}
+
+  semver-push-tag:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    timeout-minutes: 30
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install semver-checks
+      run: cargo install cargo-semver-checks --no-default-features
+    - name: Run semver-checks to see if it agrees with version updates
+      run: make semver-version
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -1,0 +1,81 @@
+# This workflow tests semver compatibilty.
+# For PRs it checks if PR makes any API breaking changes, and assings appropriate label if so.
+name: Semver checks
+
+on:
+  pull_request_target:
+    branches:
+    - main
+    - 'branch-*'
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+  PR_BASE: ${{ github.event.pull_request.base.sha }}
+  PR_HEAD: ${{ github.event.pull_request.head.sha }}
+  PR_ID: ${{ github.event.number }}
+
+jobs:
+  semver-pull-request-check:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request_target'
+    # Disable all permissions
+    # This is important, because this job runs on untrusted input from
+    # the user and it's possible for the user to take over the job,
+    # for example by adding malicious build.rs file. If the job had,
+    # for example, `pull_requests: write` permission, malicous user
+    # could do us a lot of harm. This is also the reason that there are
+    # 2 jobs - it's so that it's not possible to take over a job that
+    # has permissions.
+    permissions: {} 
+    timeout-minutes: 30
+    outputs:
+      exitcode: ${{ steps.semver-pr-check.outputs.exitcode }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: "2"
+        ref: "refs/pull/${{ env.PR_ID }}/merge"
+    # Check if there was another push before this job started.
+    # If there was, wrong commit would be checked out.
+    - name: Sanity check
+      run: |
+        [[ "$(git rev-parse 'HEAD^2')" == "$PR_HEAD" ]]
+    # I don't know any way to do this using checkout action
+    - name: Fetch PR base
+      run: git fetch origin "$PR_BASE"
+    - name: Install semver-checks
+      # Official action uses binary releases fetched from GitHub
+      # If this pipeline becomes too slow, we should do this too
+      run: cargo install cargo-semver-checks --no-default-features
+    - name: Verify the API compatibilty with PR base
+      id: semver-pr-check
+      run: |
+        set +e
+        make semver-rev rev="$PR_BASE"
+        exitcode=$?
+        echo "exitcode=$exitcode" >> $GITHUB_OUTPUT
+        exit "$exitcode"
+      continue-on-error: true
+
+  semver-pull-request-label:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request_target'
+    permissions:
+      pull-requests: write
+    needs: semver-pull-request-check
+    timeout-minutes: 3
+    steps:
+    - name: Remove breaking label on success
+      run: gh pr edit "$PR_ID" --remove-label semver-checks-breaking
+      if: needs.semver-pull-request-check.outputs.exitcode == '0'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_REPO: ${{ github.repository }}
+    - name: Add breaking label on failure
+      run: gh pr edit "$PR_ID" --add-label semver-checks-breaking
+      if: needs.semver-pull-request-check.outputs.exitcode != '0'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_REPO: ${{ github.repository }}
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -32,6 +32,13 @@ jobs:
     # has permissions.
     permissions: {} 
     timeout-minutes: 30
+    # This is to prevent a situation, when job A triggered by push 1 finishes
+    # after job B triggered by push 2. That could result in incorrectly assigning
+    # or removing a PR label.
+    concurrency:
+      # Can't use `env.PR_ID` because concurrency doesn't have access to env context.
+      group: semver-pull-request-check-${{ github.event.number }}
+      cancel-in-progress: true
     outputs:
       exitcode: ${{ steps.semver-pr-check.outputs.exitcode }}
       output: ${{ steps.semver-pr-check.outputs.output }}
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -34,6 +34,7 @@ jobs:
     timeout-minutes: 30
     outputs:
       exitcode: ${{ steps.semver-pr-check.outputs.exitcode }}
+      output: ${{ steps.semver-pr-check.outputs.output }}
     steps:
     - name: Checkout
       uses: actions/checkout@v3
@@ -56,8 +57,11 @@ jobs:
       id: semver-pr-check
       run: |
         set +e
-        make semver-rev rev="$PR_BASE"
-        exitcode=$?
+        echo "output<<SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
+        # Weird sed strip ANSI colors from output
+        make semver-rev rev="$PR_BASE" |& tee /dev/tty | sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g" >> $GITHUB_OUTPUT
+        exitcode=${PIPESTATUS[0]}
+        echo "SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
         echo "exitcode=$exitcode" >> $GITHUB_OUTPUT
         exit "$exitcode"
       continue-on-error: true
@@ -82,6 +86,26 @@ jobs:
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         GH_REPO: ${{ github.repository }}
+    - name: Post report on semver break
+      run: |
+        gh pr comment "$PR_ID" --body "\
+        \`cargo semver-checks\` detected some API incompatibilities in this PR.
+        See the following report for details:
+        <details>
+        <summary>cargo semver-checks output</summary>
+
+        \`\`\`
+        $SEMVER_OUTPUT
+        \`\`\`
+
+        </details>
+        "
+      if: needs.semver-pull-request-check.outputs.exitcode != '0'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_REPO: ${{ github.repository }}
+        SEMVER_OUTPUT: ${{ needs.semver-pull-request-check.outputs.output }}
+
 
   semver-push-tag:
     runs-on: ubuntu-latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,22 @@ There are a few scenarios:
          using command like `cargo update -p toml_datetime --precise 0.6.3` and go back to step 3.
       5. Rename `Cargo.lock` to `Cargo.lock.msrv`.
 
+### Semver checking 
+
+Our CI runs cargo semver-checks and labels PRs that introduce breaking changes.
+If you don't intend to change public API, you can perform the checks locally,
+using command `make semver-rev`. Make sure you have semver-checks installed first,
+you can install it using `cargo install cargo-semver-checks`.
+
+`make semver-rev` will check for API breaking changes using `main` branch as baseline.
+To use different branch / commit call `make semver-rev rev=BRANCH`.
+
+The tool is NOT perfect and only checks some aspect of semver-compatibility.
+It is NOT a replacement for a human reviewer, it is only supposed to help them
+and catch some erros that they might have missed.
+
+Tool that we curently use: https://github.com/obi1kenobi/cargo-semver-checks
+
 ## Contributing to the book
 
 The documentation book is written using [mdbook](https://github.com/rust-lang/mdBook)\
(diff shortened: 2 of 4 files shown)


Lorak-mmk:
v2:
Moved commands to a separate script, called by CI and newly introduced Makefile targets.
This is because invocation of semver-checks may become more complicated, e.g. when https://github.com/scylladb/scylla-rust-driver/pull/911 gets merged, as it introduces mutually-exclusive features.

Lorak-mmk:
v3: Noticed that I didn't install semver-checks in push action... Well, I did say I have no idea how to test it.

Lorak-mmk:
TODO / Considerations:
- We are before 1.0, so we don't have patch versions. After we release 1.0, it may be beneficial to replace `semver-checks-breaking` label with `semver-major-required` / `semver-minor-required` / `semver-patch-required` trio, so we know exactly which type of release to perform.
- I tested that PR job works, both for adding and removing label. I didn't test the push job - I don't see any way to do it. I'll push a testing tag after PR is merged, and if it turns out to not work I'll fix it.
- Official semver-checks action (which is not suitable for our usage) installs semver-checks from a github release binary instead of using `cargo install`. I think we can skip this for now, as we have pipelines that take more time anyway.

Lorak-mmk:
v4:
- Workflow now works for PRs opened for forks and should be secure.
- Included security considerations in PR description. It would be good for someone more experienced with GH Actions to verify if everything is correct or if there is some vulnerability I missed. @fruch ?
- I tested PR jobs in https://github.com/scylladb/scylla-rust-driver/pull/920 . Push job remains untested. I could push a tag to check it, but I'm not sure the workflow will even run if it's not in main - wdyt @piodul ?

fruch:
> v4:
> - Workflow now works for PRs opened for forks and should be secure.
> - Included security considerations in PR description. It would be good for someone more experienced with GH Actions to verify if everything is correct or if there is some vulnerability I missed. @fruch ?

I'm not expert on security, but seems that as long those branches are protected it good enough, you need to keep in mind those step have access to repo secrets, if any. so need to pay attention in reviews that no one is changing those.


> - I tested PR jobs in https://github.com/scylladb/scylla-rust-driver/pull/920 . Push job remains untested. I could push a tag to check it, but I'm not sure the workflow will even run if it's not in main - wdyt @piodul ?
> 
>

avelanarius (approved):
Security-wise this seems correct to me

piodul:
> cargo-semver-checks has some false-negatives and false-positives

It has _both_ false positives and false negatives? Doesn't it make it a bit useless? What kind of help is it if we have to always verify the conclusion of the tool anyway?

piodul:
> Doesn't it make it a bit useless?

Maybe the tool only checks some particular areas of semver-compatibility, and in those areas it only has positives/negatives? In that case I think that the job should post some kind of report from the cargo-semver-checks tool in a comment instead of setting a label. Reviewers should verify whether the PR breaks the API or not and it should be their decision to mark it as such - the report should only serve as help, allow to reduce work.

Lorak-mmk:
> > cargo-semver-checks has some false-negatives and false-positives
> 
> It has _both_ false positives and false negatives? Doesn't it make it a bit useless? What kind of help is it if we have to always verify the conclusion of the tool anyway?

I meant that this software, like any software ever written, has bugs (which means false positives). Afaik any false-positive is considered a bug in the tool.

False-negatives are cases that it can't (yet) check, but it doesn't make it useless, because it still does check things that people usually don't think about (e.g. if some pub struct stopped having some auto-trait).
As I mentioned in the description, it does not replace human reviewer, just aids them.

Lorak-mmk:
> > Doesn't it make it a bit useless?
> 
> Maybe the tool only checks some particular areas of semver-compatibility, and in those areas it only has positives/negatives? In that case I think that the job should post some kind of report from the cargo-semver-checks tool in a comment instead of setting a label. Reviewers should verify whether the PR breaks the API or not and it should be their decision to mark it as such - the report should only serve as help, allow to reduce work.

Posting a report seems like a great idea, I'll try to add that. I think that label is not harmful in any way (as it is a different one from the one we assign manually), so I'd like for it to remain - because comment is only visible when you visit a specific PR, not when you browse trough them.
One problem with reports is that it may become noisy for PRs intentionally changing API, as each push would result in another comment. Not yet sure how to solve that.

piodul:
> > > cargo-semver-checks has some false-negatives and false-positives
> > 
> > 
> > It has _both_ false positives and false negatives? Doesn't it make it a bit useless? What kind of help is it if we have to always verify the conclusion of the tool anyway?
> 
> I meant that this software, like any software ever written, has bugs (which means false positives). Afaik any false-positive is considered a bug in the tool.

I am actually fine with false positives because incorrect ones are easy to dismiss. False negatives are far worse because you don't know which cases were missed and you need to go over the code anyway.

> 
> False-negatives are cases that it can't (yet) check, but it doesn't make it useless, because it still does check things that people usually don't think about (e.g. if some pub struct stopped having some auto-trait). As I mentioned in the description, it does not replace human reviewer, just aids them.

Hm, right, I suppose the tool is useful because when it finds something, we can be sure that the PR is breaking (or we can easily dispute it if the detected issue is incorrect). We need to be as vigilant as we do now when it _doesn't_ detect anything.

I was just wondering how much trust can we put in the tool. If it claimed to have no false-negatives in some areas and it was battle-tested in that regard for some time then I suppose we could relax our manual semver-checking errors there as well. Of course, bugs can always happen, but we can always yank - it would just be nice not to have to do it too frequently.

Ofc, there are non-API breaking changes that can happen and I don't think any tool can check that. I suppose that unit tests serve that purpose, but it's still a matter of how much trust we are willing to put into them.

piodul:
> > > Doesn't it make it a bit useless?
> > 
> > 
> > Maybe the tool only checks some particular areas of semver-compatibility, and in those areas it only has positives/negatives? In that case I think that the job should post some kind of report from the cargo-semver-checks tool in a comment instead of setting a label. Reviewers should verify whether the PR breaks the API or not and it should be their decision to mark it as such - the report should only serve as help, allow to reduce work.
> 
> Posting a report seems like a great idea, I'll try to add that. I think that label is not harmful in any way (as it is a different one from the one we assign manually), so I'd like for it to remain - because comment is only visible when you visit a specific PR, not when you browse trough them. One problem with reports is that it may become noisy for PRs intentionally changing API, as each push would result in another comment. Not yet sure how to solve that.

I'm not sure whether the noisiness is a huge problem. We already have something like this for tests in the scylladb repo and it works fine (random example: https://github.com/scylladb/scylladb/pull/16947#issuecomment-1906321440).

If you are worried about the surface area that the comments take up on screen in total (which I agree is a concern), then maybe the comment could just link to the workflow step which prints the report - something similar to the report like there is in scylladb?

Lorak-mmk:
> If you are worried about the surface area that the comments take up on screen in total (which I agree is a concern), then maybe the comment could just link to the workflow step which prints the report - something similar to the report like there is in scylladb?

I can put the details in spoiler, I'm pretty sure GH supports them, so that's not a problem. And I think it's a better solution than linking to CI outputs that will disappear after a few days.

.github/workflows/semver_checks.yml
```diff
@@ -71,3 +74,14 @@ jobs:
...
+  semver-push-tag:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    timeout-minutes: 30
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install semver-checks
+      run: cargo install cargo-semver-checks --no-default-features
+    - name: Verify that's it's safe to publish the version.
+      run: cargo semver-checks -p scylla -p scylla-cql
```
piodul: I think we should use the script here as well.

.github/workflows/semver_checks.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       id: semver-pr-check
       run: |
         set +e
-        cargo semver-checks -p scylla -p scylla-cql --baseline-rev "$PR_BASE"
+        make semver-rev rev="$PR_BASE"
```
piodul, line 51: Ok, I see that you replace the manual calls to semver-checks in the last commit. Please move the commit that introduces the script to the beginning.

CONTRIBUTING.md
```diff
@@ -86,6 +86,16 @@ There are a few scenarios:
...
 
+### Semver checking 
+
+Our CI runs cargo semver-checks and labels PRs that introduce breaking changes.
+If you don't intend to change public API, you can perform the checks locally,
+using command `make semver-rev`. Make sure you have semver-checks installed first,
+you can install it using `cargo install cargo-semver-checks`.
+
+`make semver-rev` will check for API breaking changes using `main` branch as baseline.
+To use different branch / commit call `make semver-rev rev=BRANCH`.
```
piodul, line 97: We should mention here that the tool is not perfect and does not detect all the possible issues related to semantic versioning.

Perhaps a link to the github repo of the tool would be useful as well.

piodul (changes requested):

.github/workflows/semver_checks.yml
```diff
@@ -0,0 +1,73 @@
...
+    permissions: {} # Disable all permissions
+    timeout-minutes: 30
+    outputs:
+      exitcode: ${{ steps.semver-pr-check.outputs.exitcode }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: "2"
+        ref: "refs/pull/${{ env.PR_ID }}/merge"
```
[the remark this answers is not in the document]
Lorak-mmk: `the ref/SHA that triggered the workflow` is `GITHUB_SHA`.
For `pull_request` workloads that is the merge commit.
For `pull_request_target` it is head of base branch.

.github/workflows/semver_checks.yml
```diff
@@ -0,0 +1,73 @@
...
+      if: needs.semver-pull-request-check.outputs.exitcode == '0'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_REPO: ${{ github.repository }}
+    - name: Add breaking label on failure
+      run: gh pr edit "$PR_ID" --add-label semver-checks-breaking
+      if: needs.semver-pull-request-check.outputs.exitcode != '0'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GH_REPO: ${{ github.repository }}
```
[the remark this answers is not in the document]
Lorak-mmk: Hmm... It seems there is such race condition, but I think it would be hard do trigger. I can see if GHA has a mechanism to abort previous job when a new one is submitted.
This should be less relevant after adding report mechanism, as it will be the primary output from a job

scripts/semver-checks.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -x
+
+cargo semver-checks -p scylla -p scylla-cql $@
```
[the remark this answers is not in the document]
Lorak-mmk: proc-macros crates are not checked by cargo-semver-checks. If you run just `cargo semver-checks` it will check scylla, scylla-cql, scylla-proxy, but not scylla-macros.

.github/workflows/semver_checks.yml
```diff
@@ -0,0 +1,73 @@
...
+        fetch-depth: "2"
+        ref: "refs/pull/${{ env.PR_ID }}/merge"
+    # Check if there was another push before this job started.
+    # If there was, wrong commit would be checked out.
+    - name: Sanity check
+      run: |
+        [[ "$(git rev-parse 'HEAD^2')" == "$PR_HEAD" ]]
+    # I don't know any way to do this using checkout action
+    - name: Fetch PR base
+      run: git fetch origin "$PR_BASE"
```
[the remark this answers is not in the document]
Lorak-mmk: I think commit sha (which I'm using) is better. `$GITHUB_BASE_REF` can be a branch name, so if some PR is merged between last push and this step, wrong commit will be fetched.

.github/workflows/semver_checks.yml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Install semver-checks
+      # Official action uses binary releases fetched from GitHub
+      # If this pipeline becomes too slow, we should do this too
+      run: cargo install cargo-semver-checks --no-default-features
```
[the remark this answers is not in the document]
Lorak-mmk: Other jobs are definitely longer, so this is not a problem now, but I agree we should definitely do this at some point.

Lorak-mmk:
v5:
- Addresed review comments (responded to some, changed the PR accordingly for the rest)
- Added a report mechanism. You can see example comment here: https://github.com/scylladb/scylla-rust-driver/pull/920#issuecomment-1913709910 . Message in this can of course be changed. I'm also not sure if I should add posting a comment that there were no incompatibilities? I think it's reduntant and would increase spam/

.github/workflows/semver_checks.yml
```diff
@@ -56,8 +57,11 @@ jobs:
       id: semver-pr-check
       run: |
         set +e
-        make semver-rev rev="$PR_BASE"
-        exitcode=$?
+        echo "output<<SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
+        # Weird sed strip ANSI colors from output
+        make semver-rev rev="$PR_BASE" |& tee /dev/tty | sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g" >> $GITHUB_OUTPUT
```
piodul, line 62: `CARGO_TERM_COLOR=never` should force the output to be emitted without colors, no need for complicated regexes.

.github/workflows/semver_checks.yml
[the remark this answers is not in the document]
piodul: It looks like there is such a mechanism: https://docs.github.com/en/actions/using-jobs/using-concurrency. It is possible to limit the concurrency to one at a time and let the second run cancel the first one if it's still running.

.github/workflows/semver_checks.yml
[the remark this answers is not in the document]
piodul: Right, I got confused about this one.

CONTRIBUTING.md
```diff
@@ -86,6 +86,22 @@ There are a few scenarios:
...
+
+The tool is NOT perfect and only checks some aspect of semver-compatibility.
+It is NOT a replacement for a human reviewer, it is only supposed to help them
+and catch some erros that they might have missed.
+
+Tool that we curently use: https://github.com/obi1kenobi/cargo-semver-checks-action
```
piodul, line 103: This points to a github action. Shouldn't it be just `https://github.com/obi1kenobi/cargo-semver-checks`?

.github/workflows/semver_checks.yml
```diff
@@ -79,3 +82,14 @@ jobs:
...
+
+  semver-push-tag:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    timeout-minutes: 30
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install semver-checks
+      run: cargo install cargo-semver-checks --no-default-features
+    - name: Verify that's it's safe to publish the version.
```
piodul, line 118: I think this label implies too much, that the tool does really exhaustive checks and its assessment about semver compatibility can be trusted.

piodul:
> I'm also not sure if I should add posting a comment that there were no incompatibilities? I think it's reduntant and would increase spam

If the workflow removes the `semver-checks-breaking` label after declaring that there are no incompatibilities, then I think it is sufficient and there is no need for the additional comment.

.github/workflows/semver_checks.yml
Lorak-mmk: I wanted the colors to be visible in actions output (because it's prettier), but removed in report comment (because there is no way to render them)

piodul: OK. I just wish there were a solution which doesn't require such incantations, but it's fine if it works.

Lorak-mmk: Yeah, I don't like it either, but I didn't find any easier way. If you know any, let me know

.github/workflows/semver_checks.yml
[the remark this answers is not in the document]
Lorak-mmk: Done

Lorak-mmk:
v6: Addressed review comments, added concurrency control

piodul (approved):
Looks good. Please drop the "TMP" commit and I'll merge.

Lorak-mmk:
> Looks good. Please drop the "TMP" commit and I'll merge.

Done

piodul (approved):

PR #910, opened 2024-01-11
Docs: Add a book preprocessor

Lorak-mmk:
This PR adds a preprocessor for the book that removes Sphinx-related sections.
This was previously done by `docs/build_book.py` script, which this PR removes.
There are some advantages of new approach:
- No unnecessary file writing. Book content is passed to preprocessor by stdin. This is both faster and more friendly to disk.
- No way to accidentally build a book without preprocessing. Also now developer doesn't need to know about the script or Makefile target that uses it.
- New script is simpler as it doesn't need to handle files anymore. As a contributor I can certify that it's possible to not know about the script or Makefile target and be frustrated about Sphinx sections.

There is one disadvantage: book must be built either from main directory, or docs directory. If it's built from different location, build will succeed buy Sphinx sections won't be stripped.
This is because of mdbook limitation - PWD of build process is the folder the command is executed in, not the folder in which book.toml is located.
This shouldn't be a problem - I suspect book build is never / very rarely executed from other dir.
If it becomes a problem, it's possible to fix it. The script in bash.sh will need to be modified:
- if executed with `supports` argument - just exit.
- Otherwise, read stdin using jq. From first element extract `root` element, `cd` to it's value.
- Execute `python ./sphinx_preprocessor.py` with previosuly-read stdin.

## Pre-review checklist


Lorak-mmk:
Submitted an issue to mdbook: https://github.com/rust-lang/mdBook/issues/2289

Lorak-mmk:
mdbook issue was closed in favor of https://github.com/rust-lang/mdBook/issues/1424 which is open and not touches for 3 years now, so we shouldn't hope for mdbook fix. Imo the script I wrote should be enough, if it isn't I can write more complicated one that would always work.

docs/sphinx_preprocessor.py
```diff
@@ -0,0 +1,34 @@
+import json
```
piodul, line 1: Nit: how about adding a shebang? We had it in the previous script.

Lorak-mmk: This script is not supposed to be run manually, so I don't see how shebang helps?

Lorak-mmk:
v2
- Changed contributing.md: added a note about limitations
- Removed unnecessary in the script coment.

docs/sphinx_preprocessor.py
piodul: Makes sense not to do it.

piodul (approved):

PR #911, opened 2024-01-11
scylla: Add support for rustls Fixes #293

### LINKED ISSUE #293: rustls support

nemosupremo:
This PR adds support for [rustls](https://github.com/rustls/rustls) behind a feature. Rustls is a pure rust implementation of TLS. With this PR, it obviates the need for clients to install libssl separately, which can be advantageous for users who with to cross compile binaries or just want a simpler building process.

There is one breaking change:
* Since the rustls and ssl features cannot be used together, the `cloud` feature no longer automatically pulls in `ssl`. Instead a compiler error is generated when the `cloud` feature is enabled without either `rustls` or `ssl`. A compiler error is also generated when both `rustls` and `ssl` are enabled.

***This PR is not yet ready, trying to figure out how to test these changes***

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -14,9 +14,10 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = []
+default = ["rustls"]
 ssl = ["dep:tokio-openssl", "dep:openssl"]
-cloud = ["ssl", "scylla-cql/serde", "dep:serde_yaml", "dep:serde", "dep:url", "dep:base64"]
+rustls = ["dep:tokio-rustls"]
+cloud = ["scylla-cql/serde", "dep:serde_yaml", "dep:serde", "dep:url", "dep:base64"]
 secret = ["scylla-cql/secret"]
 chrono = ["scylla-cql/chrono"]
 time = ["scylla-cql/time"]
@@ -42,6 +43,7 @@ tracing = "0.1.36"
 chrono = { version = "0.4.20", default-features = false, features = ["clock"] }
 openssl = { version = "0.10.32", optional = true }
 tokio-openssl = { version = "0.6.1", optional = true }
+tokio-rustls = { version = "0.25", optional = true }
 arc-swap = "1.3.0"
 dashmap = "5.2"
 strum = "0.23"
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -142,3 +142,9 @@ pub use transport::retry_policy;
 pub use transport::speculative_execution;
 
 pub use transport::metrics::Metrics;
+
+#[cfg(all(feature = "ssl", feature = "rustls"))]
+compile_error!("both rustls and ssl should not be enabled together.");
+
+#[cfg(all(feature = "cloud", not(any(feature = "ssl", feature = "rustls"))))]
+compile_error!("cloud feature requires either the rustls or ssl feature.");
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -23,8 +23,10 @@ use std::sync::atomic::AtomicU64;
 use std::time::Duration;
 #[cfg(feature = "ssl")]
 use tokio_openssl::SslStream;
+#[cfg(feature = "rustls")]
+use tokio_rustls::TlsConnector;
 
-#[cfg(feature = "ssl")]
+#[cfg(any(feature = "ssl", feature = "rustls"))]
 pub(crate) use ssl_config::SslConfig;
 
 use crate::authentication::AuthenticatorProvider;
@@ -280,12 +282,19 @@ impl NonErrorQueryResponse {
         })
     }
 }
-#[cfg(feature = "ssl")]
+#[cfg(any(feature = "ssl", feature = "rustls"))]
 mod ssl_config {
+    #[cfg(feature = "ssl")]
     use openssl::{
         error::ErrorStack,
         ssl::{Ssl, SslContext},
     };
+    #[cfg(feature = "rustls")]
+    use std::{net::IpAddr, sync::Arc};
+    #[cfg(feature = "rustls")]
+    use tokio_rustls::rustls::pki_types::ServerName;
+    #[cfg(feature = "rustls")]
+    use tokio_rustls::rustls::ClientConfig;
     #[cfg(feature = "cloud")]
     use uuid::Uuid;
 
@@ -299,6 +308,7 @@ mod ssl_config {
     // NodeConnectionPool::new(). Inside that function, the field is mutated to contain SslConfig specific
     // for the particular node. (The SslConfig must be different, because SNIs differ for different nodes.)
     // Thenceforth, all connections to that node share the same SslConfig.
+    #[cfg(feature = "ssl")]
     #[derive(Clone)]
     pub struct SslConfig {
         context: SslContext,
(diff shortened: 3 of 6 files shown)


nemosupremo:
While implementing this, I noticed that 

1. The actual hostname that is used to connect to the cluster is not preserved. I think this is because Scylla doesn't actually know the hostnames of other nodes in the cluster, just the IP addresses. This presents a problem for Rustls - the default implementation expects to find the hostname or the IP address in either the common name or subject alternate name of the certificate. 

2. Disabling hostname verification isn't easy and requires some boilerplate. There was a discussion in rustls/rustls#578 and rustls/rustls#1197 on making this easier in rustls itself that didn't make it as other database drivers had similar issues. What a driver like github.com/redis-rs/redis-rs opted to do was to manage the configuration internally and handle the implementation of disabling hostname verification. However, `scylla-rust-driver`'s design has a "BYOB" style of configuration and it seems expected for the user to do this themselves. This is what I have done in the rustls example.

nemosupremo:
I'm not sure how to fix the `Rust / min_rust (pull_request)` test.

mykaul:
Do we have some performance numbers to compare between the two implementations?

nemosupremo:
As far as the implementations go, rustls is a bit faster than openssl as benchmarked here:

https://jbp.io/2019/07/01/rustls-vs-openssl-performance.html

> See those posts for details and analysis. To summarise the results, though, we can say approximately:
> * rustls is 15% quicker to send data.
> * rustls is 5% quicker to receive data.
> * rustls is 20-40% quicker to set up a client connection.
> * rustls is 10% quicker to set up a server connection.
> * rustls is 30-70% quicker to resume a client connection.
> * rustls is 10-20% quicker to resume a server connection.
> * rustls uses less than half the memory of OpenSSL.

However I don't have the infrastructure to test the change in performance against a scylla cluster.

wprzytula:
> I started to review this, but I decided against posting comments to specific fragments of code. I really like the idea of adding rustls support and we'd like such support to be part of 1.0. But I think we should take a different approach. In this PR you introduced a new `rustls` feature, that is conflicting with `ssl` feature. This will make documentation confusing (now you have 2 instances of struct with the same name, but enabled for different features). It will prevent use of e.g. `cargo check --all-features` and make the features non-additive. Also the amount of `cfg` directives is quite high and it makes it hard to read the code.

I strongly agree that our features should be additive.

> The approach I think should be taken is for `SslConfig` to be an enum or trait. Enabling `rustls` / `ssl` features would introduce implementation of the trait / variants of the enum. This makes features additive, and users would be able to choose at runtime the SSL implementation they want to use. I'm not sure which is better here: enum or trait. We also need to think about how to make `cloud` work best with this approach. WDYT @wprzytula @piodul ? Trait would enable external implementations. This makes sense if someone is using `BoringSSL` / new Amazon library or something else we don't (yet) support. Enum OTOH might provide simpler API / better performance.

I'm fine with the enum.
The usual tradeoff between enum and trait is that enum is more efficient, whereas a boxed trait enables user's custom implementations.
I think we could have pros of both this way:
```rust
trait SslProvider {
  ...
}

#[cfg(feature = ssl) // consider if putting this under feature is beneficial wrt performance
enum SslConfig {
  #[cfg(feature = "rustls")
  Rustls(RustlsProvider),
  #[cfg(feature = "openssl")
  Openssl(OpensslProvider),
  // maybe a cfg here as well?
  Custom(Arc(dyn CustomSslProvider),
}

impl SslProvider for RustlsProvider { ... }
impl SslProvider for OpensslProvider { ... }
impl SslProvider for SslConfig {
   // match and use trait impl on any of the 3 enum variants
}
```

Lorak-mmk:
> > The approach I think should be taken is for `SslConfig` to be an enum or trait. Enabling `rustls` / `ssl` features would introduce implementation of the trait / variants of the enum. This makes features additive, and users would be able to choose at runtime the SSL implementation they want to use. I'm not sure which is better here: enum or trait. We also need to think about how to make `cloud` work best with this approach. WDYT @wprzytula @piodul ? Trait would enable external implementations. This makes sense if someone is using `BoringSSL` / new Amazon library or something else we don't (yet) support. Enum OTOH might provide simpler API / better performance.
> 
> I'm fine with the enum. The usual tradeoff between enum and trait is that enum is more efficient, whereas a boxed trait enables user's custom implementations. I think we could have pros of both this way:
> 
> ```rust
> trait SslProvider {
>   ...
> }
> 
> #[cfg(feature = ssl) // consider if putting this under feature is beneficial wrt performance
> enum SslConfig {
>   #[cfg(feature = "rustls")
>   Rustls(RustlsProvider),
>   #[cfg(feature = "openssl")
>   Openssl(OpensslProvider),
>   // maybe a cfg here as well?
>   Custom(Arc(dyn CustomSslProvider),
> }
> 
> impl SslProvider for RustlsProvider { ... }
> impl SslProvider for OpensslProvider { ... }
> impl SslProvider for SslConfig {
>    // match and use trait impl on any of the 3 enum variants
> }
> ```

In your example our functions would accept `SslConfig` instead of `Arc<dyn CustomSslProvider>`, right? I like this a lot.

aldanor:
@Lorak-mmk @wprzytula I will take a look at this to move this forward if noone minds.

Correct me if I'm wrong, but one problem with the trait + enum suggestion, if you look closely, is that there's SslContext (which is a different for openssl / rustls) stored in Session. And you can't have that `Custom(Arc<dyn>)` if you have an associated type in the trait... so the only option if you wanted to retain what you've suggested is to store it as `Arc<dyn Any>`, would that be suitable?

Another question I had - in regards to feature names, it gets a bit confusing, should ssl be renamed to openssl? (a bit of a breaking change, of course) Does there need to be a separate ssl feature as well then, or perhaps not?

nemosupremo:
This fell off for me as it wasn't clear if rustls was something scylla wanted to support. I see it has the waiting-on-author label now. w.r.t `ssl_context` could just be `SslConfig` that does handwritten dynamic dispatch to the enum variants. 

That said, I think changing `ssl_context` may be a more invasive change, and my original goal was to minimize the amount of code I touched. @aldanor let me know if you plan on taking this over - I'm not planning on refactoring this PR unless you are no longer interested.

aldanor:
> I don't see why would there by any need for associated type in SslProvider trait.
Regarding the enum idea, we discussed it with @wprzytula and this optimization is not necessary here, we can just use Arc<dyn SslProvider>.

@Lorak-mmk As it's currently written, in Session you have an instance of an sslprovider-dependent type which is not the provider itself. `Arc<dyn SslProvider>` won't help you here unless you agree to store the context as `Arc<dyn Any>` or perhaps `Arc<dyn SslContext>` if the latter even makes sense. Or, do something like what @nemosupremo suggested above (config with dynamic dispatch).

---

@nemosupremo Thanks for answering (and apologies for taking long to get back).

> @aldanor let me know if you plan on taking this over - I'm not planning on refactoring this PR unless you are no longer interested.

I think it makes sense for those present in this thread to jointly get this PR to completion in some way since it's more than halfway there already. To be frank I am barely familiar with either scylla or this particular codebase, so I'd definitely prefer it to be led by someone more knowledgeable than I am :) But if noone else wants or has time to, I could give it a try too, would really like to see it get merged in.

Cheers.

Lorak-mmk:
> 
> > Did you wait a few minutes? Launching Scylla takes some time. It sometimes just hangs for me too - trying again (make down && make up) is usually enough.
> 
> Yep, tried it all possible ways, and waited over 10 minutes, make up/down, nothing helps, _sigh_ (I'm on an m1 mac, if that's any relevant).
> 

Could you share logs from Scylla? Either from the node that hangs at boot or from all 3. `docker logs <container name>`, you can obtain container name from `docker ps` (unless something is very different on macs...).

wprzytula:
Closing in favour of #1190.

PR #914, opened 2024-01-14
Docs: Add recommendations about Session usage.

Lorak-mmk:
#216 showed that best practices of `Session` usage are not obvious and should be stated in documentation.
This PR adds the relevant section that explains how to use it in order to not decrease performance.

Fixes: #216 

## Pre-review checklist

### DIFF
--- a/docs/source/connecting/connecting.md
+++ b/docs/source/connecting/connecting.md
@@ -35,6 +35,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+## Best practices for using Session
+
 The driver maintains its own pool of connections to each node and each connection is capable of handling multiple requests in parallel. Driver will also route requests to nodes / shards that actually own the data (unless the load balancing policy that you use doesn't support it).
 
 For those reasons, we recommend using one instance of `Session` per application.
@@ -44,8 +46,12 @@ Creating many `Session`'s in one application (e.g. `Session` per thread / per To
 
 If you need to share `Session` with different threads / Tokio tasks etc. use `Arc<Session>` - all methods of `Session` take `&self`, so it doesn't hinder the functionality in any way.
 
+## Metadata
+
 The driver refreshes the cluster metadata periodically, which contains information about cluster topology as well as the cluster schema. By default, the driver refreshes the cluster metadata every 60 seconds. 
 However, you can set the `cluster_metadata_refresh_interval` to a non-negative value to periodically refresh the cluster metadata. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
+
+## Scylla Cloud Serverless
 
 Scylla Serverless is an elastic and dynamic deployment model. When creating a `Session` you need to
 specify the secure connection bundle as follows:
--- a/docs/source/connecting/connecting.md
+++ b/docs/source/connecting/connecting.md
@@ -35,6 +35,15 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+The driver maintains its own pool of connections to each node and each connection is capable of handling multiple requests in parallel. Driver will also route requests to nodes / shards that actually own the data (unless the load balancing policy that you use doesn't support it).
+
+For those reasons, we recommend using one instance of `Session` per application.
+
+Creating short-lived `Session`'s (e.g. `Session` per request) is strongly discouraged because it will result in great performance penalties because creating a `Session` is a costly process - it requires estabilishing a lot of TCP connections.
+Creating many `Session`'s in one application (e.g. `Session` per thread / per Tokio task) is also discouraged, because it wastes resources - as mentioned before, `Session` maintains a connection pool itself and can handle parallel queries, so you would be holding a lot of connections unnecessarily.
+
+If you need to share `Session` with different threads / Tokio tasks etc. use `Arc<Session>` - all methods of `Session` take `&self`, so it doesn't hinder the functionality in any way.
+
 The driver refreshes the cluster metadata periodically, which contains information about cluster topology as well as the cluster schema. By default, the driver refreshes the cluster metadata every 60 seconds. 
 However, you can set the `cluster_metadata_refresh_interval` to a non-negative value to periodically refresh the cluster metadata. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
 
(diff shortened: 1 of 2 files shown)


docs/source/connecting/connecting.md
```diff
@@ -35,6 +35,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+Driver maintains this pool of connections and each connection is capable of handling multiple requests in parallel. Driver will also route the request to node / shard that actually owns the data (unless load balancing policy that you use doesn't support it).
+For those reasons, we recommend to use one instance of `Session` per application.
+Creating short-lived `Session`'s (e.g. `Session` per request) is extremely discouraged and will result in great performance penalties because creating a `Session` is a costly process - it requires estabilishing a lot of TCP connections.
+Creating many `Session`'s in one application (e.g. `Session` per thread / per Tokio taks) is also discouraged, because it wastes resources - as mentioned before, `Session` maintains connection pool itself and can handle parallel queries, so you would be holding a lot of connections unnecessarily.
```
piodul, line 41: typo: `per Tokio task`

piodul: `maintains connection pool` -> `maintains a connection pool`

docs/source/connecting/connecting.md
```diff
@@ -35,6 +35,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+If you need to share `Session` to different threads / Tokio tasks etc. use `Arc<Session>` - all methods of `Session` take `&self`, so it doesn't hinder the functionality in any way. In order to avoid cost of `Arc`'s atomic operations, you can estabilish all your `Arc`'s in the beggining as the atomic operations only happend when cloning `Arc`, not when using it's inner object.
```
piodul, line 42: `share 'Session' to different threads` -> `share 'Session' with different threads`

piodul: `avoid cost` -> `avoid the cost`

piodul: typo: `happend` -> `happen`

piodul: `it's inner object` -> `its inner object`

piodul: I'm not sure that the remark about avoiding `Arc` atomic operations is realistic. Many client programs will spawn new tasks (I bet most of the web frameworks based on tokio do this) and then you will have to clone the `Arc`. The only alternative I see is putting the session into a global - but then you would either have to use unsafe to access it, or some wrapper that introduces synchronization overhead (rwlock, once cell, lazy static etc.)

docs/source/connecting/connecting.md
```diff
@@ -35,6 +35,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+Driver maintains this pool of connections and each connection is capable of handling multiple requests in parallel. Driver will also route the request to node / shard that actually owns the data (unless load balancing policy that you use doesn't support it).
```
piodul, line 38: `Driver maintains this pool`? Maybe `a pool`, or `its own pool`?

piodul: Also, `The driver` instead of `Driver` sounds slightly better

docs/source/connecting/connecting.md
```diff
@@ -35,6 +35,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+For those reasons, we recommend to use one instance of `Session` per application.
+Creating short-lived `Session`'s (e.g. `Session` per request) is extremely discouraged and will result in great performance penalties because creating a `Session` is a costly process - it requires estabilishing a lot of TCP connections.
```
piodul, line 40: IMO "extremely" is too strong of a word here. There are cases where short sessions make sense, e.g. an administrative tool that connects only to execute a single statement. Of course, creating multiple sessions over the course of the application is wasteful.

docs/source/connecting/connecting.md
```diff
@@ -35,6 +35,12 @@ async fn main() -> Result<(), Box<dyn Error>> {
 After successfully connecting to some specified node the driver will fetch topology information about
 other nodes in this cluster and connect to them as well.
 
+Driver maintains this pool of connections and each connection is capable of handling multiple requests in parallel. Driver will also route the request to node / shard that actually owns the data (unless load balancing policy that you use doesn't support it).
+For those reasons, we recommend to use one instance of `Session` per application.
```
piodul, line 39: `recommend to use` -> `recommend using`

docs/source/connecting/connecting.md
piodul: `load balancing policy` -> `the load balancing policy`

piodul: `the request` -> `requests`?

piodul (changes requested):

docs/source/connecting/connecting.md
Lorak-mmk: > `the request` -> `requests`?

Changed to `Driver will also route requests to nodes / shards that actually own the data (unless the load balancing policy that you use doesn't support it).`

Lorak-mmk: > `Driver maintains this pool`? Maybe `a pool`, or `its own pool`?

I was referring to a pool that was mentioned in a sentence before. But now that I think of it your version may be easier to understand.
Changed to `The driver maintains its own pool of connections to each node and each connection is capable of handling multiple requests in parallel.`

docs/source/connecting/connecting.md
Lorak-mmk: Dropped this section

docs/source/connecting/connecting.md
Lorak-mmk: strongly?

Lorak-mmk: I change to `strongly`, if that's also too strong I'll rewrite this sentence differently

Lorak-mmk:
v2: 
- Addressed review comments
- Changed formatting so that new section doesn't render as a block of text
- Added an adminition in order to make this section easier to notice. How it looks in Scylla docs:
image
and in mdbook:
image
This required changing preprocessor script a bit.
- Split the file into sections becaue there was too much contiguous text

Lorak-mmk:
v3: Fixed typo in preprocessor script

docs/source/connecting/connecting.md
piodul: "Strongly" sounds better to me as well. I was mostly concerned about the "short-lived" sessions given the use cases for them I mentioned in my previous comment, but with the added warning about using only one session per application I think it looks OK now.

piodul (approved):

PR #915, opened 2024-01-15
scylla: Enable `log` feature in `tracing`

Lorak-mmk:
This allows applications using `log` ecosystem to receive logs from the driver.
There shouldn't be a significant performance penalty, because `log` events are emitted only if there is not `Subscriber` active.

TODO: Look intro tracing code to see if enabling this looks costly, or setup some benchmark.

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -38,7 +38,7 @@ thiserror = "1.0"
 itertools = "0.11.0"
 bigdecimal = "0.2.0"
 num-bigint = "0.3"
-tracing = "0.1.36"
+tracing = { version = "0.1.36", features = ["log"] }
 chrono = { version = "0.4.20", default-features = false, features = ["clock"] }
 openssl = { version = "0.10.32", optional = true }
 tokio-openssl = { version = "0.6.1", optional = true }


scylla/Cargo.toml
```diff
@@ -38,7 +38,7 @@ thiserror = "1.0"
 itertools = "0.11.0"
 bigdecimal = "0.2.0"
 num-bigint = "0.3"
-tracing = "0.1.36"
+tracing = { version = "0.1.36", features = ["log"] }
```
piodul, line 41: I think it is wrong for us to unconditionally enable this feature. Our crate does not need it at all in order to function properly. The decision whether to enable the feature or not should be on the application developers (as it controls global behavior) - they can explicitly depend on `tracing` in their application and enable `log` feature.

Lorak-mmk: I'm not very familiar with how Cargo handles dependencies, resolves versions etc. I know it's possible to have 2 copies (2 versions) of the same dependency. What is user supposed to do if they want to enable `log` feature in `tracing` used by our driver and not include another `tracing` copy?
Is it possible for us to have a feature that enables `log` feature in `tracing`? I feel like this would be a bit easier to use.

Lorak-mmk:
I looked at how is this feature implemented. See the following links:
https://github.com/tokio-rs/tracing/blob/a4217633e127e9f561b4e47ec1b42aa73d475afa/tracing/src/macros.rs#L20-L54
https://github.com/tokio-rs/tracing/blob/a4217633e127e9f561b4e47ec1b42aa73d475afa/tracing/src/macros.rs#L2459-L2477
https://github.com/tokio-rs/tracing/blob/d4ca988252211b12e3beccfa6b029064ede1d6dc/tracing-core/src/dispatch.rs#L368-L376

So basically it looks like the cost is one atomic relaxed load per log call. Considering that this atomic is only written to at most once iiuc the value should be cached and there should be no contention / other problems, so overhead should be negligible.
I'll try to add a feature to our crate that enables `log` in  `tracing`. 

After reading https://docs.rs/tracing/latest/tracing/#emitting-log-records and source code I think this could be in our default features - this will allow both apps using `tracing` and `log` ecosystems to consume the driver without any problems, improving usability of this crate. For users that want to avoid this marginal cost, they can disable the feature. This is also aligned with `tracing` docs recommendation that I linked - they recommend enabling the feature for libraries that want to be usable with both `log` and `tracing`.

scylla/Cargo.toml
piodul: > I'm not very familiar with how Cargo handles dependencies, resolves versions etc. I know it's possible to have 2 copies (2 versions) of the same dependency. What is user supposed to do if they want to enable `log` feature in `tracing` used by our driver and not include another `tracing` copy?

They need to specify a version constraint such that the cargo resolver will choose the same version. I don't know all the intricacies of the algorithm either, but I'm pretty sure that if you specify:

```toml
[dependencies]
tracing = { version = "0.1", features = ["log"] }
```

then, given that we have `tracing = "0.1.36"` in our depencencies, the resolver will choose the same version and instance of the `tracing` crate. Users can verify if they succeeded by looking at `Cargo.lock`.

> Is it possible for us to have a feature that enables `log` feature in `tracing`? I feel like this would be a bit easier to use.

Yes, it is possible. However, I still think it's weird for us to add such a feature. If we were to add it, going further with that logic, all other library crates that use `tracing` internally should have such a feature as well - this does not scale. Users can easily enable the necessary feature in `tracing` for their application  with the method I outlined.

Lorak-mmk:
I'll respond in the issue, so that discussion is there and not in a PR

Lorak-mmk:
Turns out that most likely the original issue is not the fault of Scylla. I think adding the `log` feature still makes sense, just maybe not in default features. I'll update the PR shortly.

Lorak-mmk:
Turns out enabling log integration is not that easy as just enabling `log` feature on `tracing`.
I added `log` feature to Scylla, created an example copied from `logging.rs`, but replaced tracing with `env_logger`.
When running it with `RUST_LOG=trace`, I get the `info!` from the example. I also get manually added logs, both from example and from inside the crate.
But I don't get any actual driver logs. I suspect that `tracing`'s log integration doesn't play well with using `with_current_subscriber()` on futures passed to `tokio::spawn`.
So closing this for now. We should recommend users to use `tracing` for logs.

Issue #860, opened 2023-11-14
Logging in `log` is bugged, but `tracing` works


RoDmitry:
I was using `log` + `env_logger`, and logs from other crates (`sqlx` and `actix_web`) didn't print after initialization of `scylla::Session`.
Also no logs from `scylla` itself were present.

I mean if you change order of these two blocks, then you will see logs from `sqlx` (but still no logs from `scylla`):
```rust
        let session = SessionBuilder::new()
            .known_node(&env::var("SCYLLA_URI").expect("SCYLLA_URI must be present in .env"))
            .user("cassandra", "cassandra")
            .build()
            .await
            .unwrap();
```
```rust
        let pool = MySqlPoolOptions::new()
            ....
            .connect(&env::var("DATABASE_URL").expect("DATABASE_URL must be present in .env"))
            .await
            .expect("Mysql Connect error!");
```

main.rs:
```rust
#[actix_web::main]
async fn main() -> io::Result<()> {
    env::set_var("RUST_BACKTRACE", "1");
    env::set_var("RUST_LOG", "debug");
    env_logger::init();

    // and here are those blocks from above

    HttpServer::new(move || {
        App::new()
...
```
Once I replaced logging (`log` + `env_logger`) with `tracing` + `tracing-subscriber` in my app, logs are printing just fine.

Lorak-mmk:
You don't see logs from our driver because it uses `tracing` for it's logging and doesn't enable `log` feature in `tracing` dependency.
I'd encourage you  to use `tracing` ecosystem for logging - it has a crate allowing it to consume `log` events, so you'd still be compatible with libraries using `log` - see https://docs.rs/tracing/latest/tracing/#consuming-log-records

I guess we could enable `log` feature in `tracing` so that apps using `log` ecosystem can receive logs from driver, but the question is if there is any performance penalty for this. According to `tracing` docs:
```
This crate provides two feature flags, “log” and “log-always”, which will cause [spans](https://docs.rs/tracing/latest/tracing/span/index.html) and [events](https://docs.rs/tracing/latest/tracing/event/struct.Event.html) to emit log records. When the “log” feature is enabled, if no tracing Subscriber is active, invoking an event macro or creating a span with fields will emit a log record. This is intended primarily for use in libraries which wish to emit diagnostics that can be consumed by applications using tracing or log, without paying the additional overhead of emitting both forms of diagnostics when tracing is in use.
```
[...]

PR #921, opened 2024-01-26
Docs: Document `log` limitations

Lorak-mmk:
This PR describes reasons why driver doesn't work with `log` out of the box and how the user can fix this.

Fixes: #860 

## Pre-review checklist


Lorak-mmk:
@piodul I have no idea what is going on with the book job
- I added env_logger to dev dependencies so why is it complaining that it can't find it?
- What does the first error even mean, why is it trying to parse the output of this code snippet?

piodul:
> * I added env_logger to dev dependencies so why is it complaining that it can't find it?

I assume it's https://github.com/scylladb/scylla-rust-driver/pull/926

> * What does the first error even mean, why is it trying to parse the output of this code snippet?

The CI job takes the dependencies from the `examples/Cargo.toml` file, not `scylla/Cargo.toml` (`mdbook test -L target/debug/deps docs`).

docs/source/logging/logging.md
```diff
@@ -45,4 +45,71 @@ To start this example execute:
...
-The full [example](https://github.com/scylladb/scylla-rust-driver/tree/main/examples/logging.rs) is available in the `examples` folder
\ No newline at end of file
+The full [example](https://github.com/scylladb/scylla-rust-driver/tree/main/examples/logging.rs) is available in the `examples` folder
+
+## 'log' compatibility
+
+It may be surprising for some that viewing drivers logs using `log` ecosystem doesn't work out of the box despite `tracing` having a compatiblity layer
+behind `log` / `log-always` feature flags. 
+
+The problem is that this compatibility using `log` feature (which is recommended for libraries) seems to not work well with `.with_current_subscriber()` / Tokio tasks.
```
piodul, line 55: Is there an open GitHub issue about this in the `tracing` project? This sounds like a legitimate issue with the compatibility layer.

Lorak-mmk:
Still fails with the weird stdout error after rebasing on main

docs/source/logging/logging.md
Lorak-mmk: I didn't find any when I was searching some time ago, I should open one

piodul:
> Fixed, it didn't parse my stdout. It interprets code blocks without language specified as Rust and tries to run them. Obviously block with some logs didn't parse, hence the error. I'm so stupid.

Happens to the best of us.

docs/source/logging/logging.md
piodul: I'd like to know more about this before we merge this documentation page. It feels wrong to document workarounds for bugs in other projects, especially if they can just be fixed by the other project.

Maybe it is a documented limitation? If so, then perhaps we can just point to that place instead of having to maintain our own description of the workaround.

Lorak-mmk: Ok, I'll research it more, and not include this PR in 0.12

docs/source/logging/logging.md
```diff
@@ -45,4 +45,71 @@ To start this example execute:
 RUST_LOG=info cargo run
 ```
 
-The full [example](https://github.com/scylladb/scylla-rust-driver/tree/main/examples/logging.rs) is available in the `examples` folder
\ No newline at end of file
+The full [example](https://github.com/scylladb/scylla-rust-driver/tree/main/examples/logging.rs) is available in the `examples` folder
```
wprzytula, line 48: Quite a side topic, but I couldn't resist investigating it.

This usage of *folder* instead of the usual Unixy *directory* made me wonder about the very difference between these two terms, and these are some opinions:
- https://www.baeldung.com/cs/directories-vs-folders -> folder is a "leaf directory", i.e. directory without directories in it;
- https://superuser.com/questions/187900/what-is-the-difference-between-a-directory-and-a-folder -> no strong difference; *directory* more of a file system concept, *folder* more of a GUI thing.

docs/source/logging/logging.md
wprzytula: > I'd like to know more about this before we merge this documentation page.

On the other hand, in case that we don't manage to investigate this further before the next release, we should probably merge this to aid confused users, even if that's `log`'s fault that could be fixed soon.

docs/source/logging/logging.md
```diff
@@ -45,4 +45,71 @@ To start this example execute:
...
+[2024-01-26T00:51:06Z INFO  reproducer] info log
+[2024-01-26T00:51:06Z INFO  reproducer] info tracing
+[2024-01-26T00:51:06Z INFO  reproducer] spawned info log
+[2024-01-26T00:51:06Z INFO  reproducer] another info log
+```
+
+The other feature, `log-always`, works with the driver - but is not something we want to enable in this library.
+We recommend using tracing ecosystem, but if for some reason you need to stick with `log`, you can try
+enabling `log-always` feature in `tracing` by adding it to your direct dependencies (`tracing = { version = "0.1", features = [ "log-always" ] }`).
+This should enable driver log collection via `log` loggers.
```
wprzytula, line 115: Have you confirmed that this indeed works @Lorak-mmk?
If so, I suggest phrasing this as:

> This will enable ...

Let's not introduce any needless uncertainty (even though in software engineering it might come out as a suitable attitude...).

docs/source/logging/logging.md
```diff
@@ -45,4 +45,71 @@ To start this example execute:
...
+
+the output is:
+```bash
+[2024-01-26T00:51:06Z INFO  reproducer] info log
+[2024-01-26T00:51:06Z INFO  reproducer] info tracing
+[2024-01-26T00:51:06Z INFO  reproducer] spawned info log
+[2024-01-26T00:51:06Z INFO  reproducer] another info log
+```
+
+The other feature, `log-always`, works with the driver - but is not something we want to enable in this library.
```
wprzytula, line 112: A short explanation why we don't want this feature enabled might be appreciated by the audience.

examples/Cargo.toml
```diff
@@ -20,6 +20,7 @@ uuid = "1.0"
 tower = "0.4"
 stats_alloc = "0.1"
 clap = { version = "3.2.4", features = ["derive"] }
+env_logger = "0.10"
```
wprzytula, line 23: For the future clarity, let's add a comment saying that this dependency is currently used only by the Book (as one might wonder why it's put there, having checked that `cargo check --all-targets` succeeds even with that dependency removed).

Or, even better, I suggest visually extracting a section in `examples/Cargo.toml` that lists Book-only deps.

wprzytula: From a higher perspective, this possible confusion arises because the Book is actually a secondary project that is configured by the same `Cargo.toml` as the primary one (`examples` set of crates).

Lorak-mmk:
I opened an issue in tracing: https://github.com/tokio-rs/tracing/issues/2913

Issue #169, opened 2021-02-20
Use prepared statement metadata to decode results of prepared statements


piodul:
As of now (075ad419649984b7d901ae04b8bfa56783340095), when executing a statement - no matter if prepared or unprepared - we always request full result metadata to be returned in response. This metadata contains information about the type of each column returned in response, and we use it to decode rows from the response.

However, for prepared statements, this information is not necessary - the same metadata can be returned by the DB when the statement is being prepared and can be cached by the driver. When executing prepared statements, we can request the metadata to be omitted in response and use the cached metadata instead, which will save some bandwith.

There is at least one problem this approach is known to have, so we should discuss how to handle it: https://docs.datastax.com/en/developer/java-driver/3.0/manual/statements/prepared/#avoid-preparing-select-queries

cvybhu:
Requesting the full metadata on every SELECT could cause significant overhead.

Recently a user reported that on the same workload Rust driver uses 60MiB/s while the C++ driver uses only 23MiB/s. (https://scylladb-users.slack.com/archives/C01D7LCL44D/p1695956648707279)
For a SELECT with only a few int values sending the metadata could easily double the used bandwidth.
I suspect that this could be caused by us not using the `skip_metadata` flag.

PR #925, opened 2024-02-01
Use prepared statement result metadata to decode rows

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/169

## Motivation
Currently, the driver does not make a use of an optimization that CQL protocol offers during exeuction of prepared statements. The client is allowed to provide `SKIP_METADATA` flag during `Session::execute`, so the server doesn't attach a result set metadata. The driver can use the cached metadata (received in response to `Session::prepare`) to deserialize the results.

### `PreparedStatement`
As mentioned previously, the driver can cache the prepared statement's result metadata. This is why we extend the `PreparedStatementSharedData` by `result_metadata` field which holds the metadata. The metadata is obtained from the server during statement preparation (`Session::prepare`). This metadata is eventually passed to the function which deserializes the results of query execution.

### Schema changes
Before introducing the optimization, the driver was immune to the problems related to prepared statement's result metadata and schema changes.

After some schema change, Scylla and Cassandra invalidate the server-side caches and discard all of the prepared statements. When the driver executes a prepared statement after the schema change for the first time, it receives the `UNPREPARED` error frame (server does not recognize the id of the prepared statement since it was discarded). Currently, Scylla drivers handle such case by repreparation of the statement. Notice that after the schema change, the result metadata may be changed (depends on the actual change of schema). Without the optimization, Scylla always provides a valid (updated during repreparation) result metadata in the response to statement execution so we don't need to worry about it.

This raises an issue when we try to cache the metadata on the client side. It should be mutable so the driver holds a metadata that is always up to date (should be updated during reprepares). However, looking at the other drivers we can see that the client-side metadata caches are immutable. This is why we follow the same approach and not care about the schema changes.

### Config
Extended and `StatementConfig` by `skip_result_metadata` boolean flag which defaults to `false`. When set to `true`, it enables the optimization. Thanks to that, the users can enable the optimization per-statement.

## Points to discuss
- should maybe `skip_result_metadata` flag be enabled by default? Since it's an optimization, it might make sense for it to be enabled by default.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -386,7 +386,7 @@ pub struct ColumnSpec {
     pub typ: ColumnType,
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug, Clone, Default)]
 pub struct ResultMetadata {
     col_count: usize,
     pub paging_state: Option<Bytes>,
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -886,16 +886,28 @@ pub fn deser_cql_value(typ: &ColumnType, buf: &mut &[u8]) -> StdResult<CqlValue,
     })
 }
 
-fn deser_rows(buf: &mut &[u8]) -> StdResult<Rows, ParseError> {
-    let metadata = deser_result_metadata(buf)?;
+fn deser_rows(
+    buf: &mut &[u8],
+    cached_metadata: Option<&ResultMetadata>,
+) -> StdResult<Rows, ParseError> {
+    let server_metadata = deser_result_metadata(buf)?;
+
+    let metadata = match cached_metadata {
+        Some(metadata) => metadata.clone(),
+        None => {
+            // No cached_metadata provided. Server is supposed to provide the result metadata.
+            if server_metadata.col_count != server_metadata.col_specs.len() {
+                return Err(ParseError::BadIncomingData(format!(
+                    "Bad result metadata provided in the response. Expected {} column specifications, received: {}",
+                    server_metadata.col_count,
+                    server_metadata.col_specs.len()
+                )));
+            }
+            server_metadata
+        }
+    };
 
     let original_size = buf.len();
-
-    // TODO: the protocol allows an optimization (which must be explicitly requested on query by
-    // the driver) where the column metadata is not sent with the result.
-    // Implement this optimization. We'll then need to take the column types by a parameter.
-    // Beware of races; our column types may be outdated.
-    assert!(metadata.col_count == metadata.col_specs.len());
 
     let rows_count: usize = types::read_int(buf)?.try_into()?;
 
@@ -946,11 +958,14 @@ fn deser_schema_change(buf: &mut &[u8]) -> StdResult<SchemaChange, ParseError> {
     })
 }
 
-pub fn deserialize(buf: &mut &[u8]) -> StdResult<Result, ParseError> {
+pub fn deserialize(
+    buf: &mut &[u8],
+    cached_metadata: Option<&ResultMetadata>,
+) -> StdResult<Result, ParseError> {
     use self::Result::*;
     Ok(match types::read_int(buf)? {
         0x0001 => Void,
-        0x0002 => Rows(deser_rows(buf)?),
+        0x0002 => Rows(deser_rows(buf, cached_metadata)?),
         0x0003 => SetKeyspace(deser_set_keyspace(buf)?),
         0x0004 => Prepared(deser_prepared(buf)?),
         0x0005 => SchemaChange(deser_schema_change(buf)?),
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -10,7 +10,7 @@ use std::time::Duration;
 use thiserror::Error;
 use uuid::Uuid;
 
-use scylla_cql::frame::response::result::{ColumnSpec, PartitionKeyIndex};
+use scylla_cql::frame::response::result::{ColumnSpec, PartitionKeyIndex, ResultMetadata};
 
 use super::StatementConfig;
 use crate::frame::response::result::PreparedMetadata;
@@ -37,6 +37,7 @@ pub struct PreparedStatement {
 #[derive(Debug)]
 struct PreparedStatementSharedData {
     metadata: PreparedMetadata,
+    result_metadata: ResultMetadata,
     statement: String,
 }
 
@@ -59,6 +60,7 @@ impl PreparedStatement {
         id: Bytes,
         is_lwt: bool,
         metadata: PreparedMetadata,
+        result_metadata: ResultMetadata,
         statement: String,
         page_size: Option<i32>,
         config: StatementConfig,
@@ -67,6 +69,7 @@ impl PreparedStatement {
             id,
             shared: Arc::new(PreparedStatementSharedData {
                 metadata,
+                result_metadata,
                 statement,
             }),
             prepare_tracing_ids: Vec::new(),
@@ -335,6 +338,16 @@ impl PreparedStatement {
     /// Access info about partition key indexes of the bind variables of this statement
     pub fn get_variable_pk_indexes(&self) -> &[PartitionKeyIndex] {
         &self.shared.metadata.pk_indexes
+    }
+
+    /// Access metadata about the result of prepared statement returned by the database
+    pub(crate) fn get_result_metadata(&self) -> &ResultMetadata {
+        &self.shared.result_metadata
+    }
+
+    /// Access column specifications of the result set returned after the execution of this statement
+    pub fn get_result_set_col_specs(&self) -> &[ColumnSpec] {
+        &self.shared.result_metadata.col_specs
     }
 
     /// Get the name of the partitioner used for this statement.
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -10,7 +10,7 @@ use std::time::Duration;
 use thiserror::Error;
 use uuid::Uuid;
 
-use scylla_cql::frame::response::result::ColumnSpec;
+use scylla_cql::frame::response::result::{ColumnSpec, PartitionKeyIndex};
 
 use super::StatementConfig;
 use crate::frame::response::result::PreparedMetadata;
@@ -301,9 +301,19 @@ impl PreparedStatement {
         self.partitioner_name = partitioner_name;
     }
 
-    /// Access metadata about this prepared statement as returned by the database
-    pub fn get_prepared_metadata(&self) -> &PreparedMetadata {
+    /// Access metadata about the bind variables of this statement as returned by the database
+    pub(crate) fn get_prepared_metadata(&self) -> &PreparedMetadata {
         &self.shared.metadata
+    }
+
+    /// Access column specifications of the bind variables of this statement
+    pub fn get_variable_col_specs(&self) -> &[ColumnSpec] {
+        &self.shared.metadata.col_specs
+    }
+
+    /// Access info about partition key indexes of the bind variables of this statement
+    pub fn get_variable_pk_indexes(&self) -> &[PartitionKeyIndex] {
+        &self.shared.metadata.pk_indexes
     }
 
     /// Get the name of the partitioner used for this statement.
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -270,6 +270,27 @@ impl PreparedStatement {
         self.config.tracing
     }
 
+    /// Make use of cached metadata to decode results
+    /// of the statement's execution.
+    ///
+    /// If true, the driver will request the server not to
+    /// attach the result metadata in response to the statement execution.
+    ///
+    /// The driver will cache the result metadata received from the server
+    /// after statement preparation and will use it
+    /// to deserialize the results of statement execution.
+    ///
+    /// This option is false by default.
+    pub fn set_use_cached_result_metadata(&mut self, use_cached_metadata: bool) {
+        self.config.skip_result_metadata = use_cached_metadata;
+    }
+
+    /// Gets the information whether the driver uses cached metadata
+    /// to decode the results of the statement's execution.
+    pub fn get_use_cached_result_metadata(&self) -> bool {
+        self.config.skip_result_metadata
+    }
+
     /// Sets the default timestamp for this statement in microseconds.
     /// If not None, it will replace the server side assigned timestamp as default timestamp
     /// If a statement contains a `USING TIMESTAMP` clause, calling this method won't change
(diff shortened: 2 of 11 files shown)


muzarski:
I need to adjust the test case so we alter the table in some other way. The logs from CI:
```
thread 'transport::session_test::test_use_prepared_result_metadata' panicked at scylla/src/transport/session_test.rs:2017:10:
called `Result::unwrap()` on an `Err` value: DbError(Invalid, "Altering column types is no longer supported")
```

piodul:
> ## Changes

I would prefer if you summarized your changes using prose and not a bulleted list. As a reviewer, I'm interested in:

- The motivation behind the PR (A single link to an issue doesn't count),
- An overview of the solution and how it addresses the motivation,
- Decisions that you made and the rationale behind them.

Your bullet points don't give me this information, not easily at least. They concentrate too much on the details. For example, this point:

> - added `result_metadata: ArcSwap<ResultMetadata>` field to `PreparedStatementSharedData`

doesn't give any justification what it is for. After some adjustments it would serve well as a commit title (which should also be accompanied with a description unless the title itself is descriptive enough).

After reading all the bullet points I can form some idea about how the whole thing looks like, given that I'm familiar with the codebase and I created the issue that is being fixed here. However, other developers who will want to modify your code in the future might not have that luxury.

Besides helping maintainers, reviewers and other developers in the long run, I believe that preparing such a summary gives one a good opportunity to reflect on their proposed changes and help convince oneself that their changes make sense (or decide that they don't).

muzarski:
> > ## Changes
> 
> I would prefer if you summarized your changes using prose and not a bulleted list. As a reviewer, I'm interested in:
> 
>     * The motivation behind the PR (A single link to an issue doesn't count),
> 
>     * An overview of the solution and how it addresses the motivation,
> 
>     * Decisions that you made and the rationale behind them.
> 
> 
> Your bullet points don't give me this information, not easily at least. They concentrate too much on the details. For example, this point:
> 
> > * added `result_metadata: ArcSwap<ResultMetadata>` field to `PreparedStatementSharedData`
> 
> doesn't give any justification what it is for. After some adjustments it would serve well as a commit title (which should also be accompanied with a description unless the title itself is descriptive enough).
> 
> After reading all the bullet points I can form some idea about how the whole thing looks like, given that I'm familiar with the codebase and I created the issue that is being fixed here. However, other developers who will want to modify your code in the future might not have that luxury.
> 
> Besides helping maintainers, reviewers and other developers in the long run, I believe that preparing such a summary gives one a good opportunity to reflect on their proposed changes and help convince oneself that their changes make sense (or decide that they don't).

Thank you for the feedback, I'll make sure to adjust the PR description as well as commit messages that are not descriptive enough.

piodul (changes requested):
I'd like to know more about how the driver behaves when a signal to reprepare is missed (i.e. some other application reprepares and this one misses it). The scenario with `SELECT *` and adding a column is known, but I wonder what happens if we have a UDT column or a tuple and we add a field to the UDT/tuple. I think this case might actually be quite common contrary to `SELECT *` so I'm wondering whether other drivers handle it correctly and if we handle it correctly.

Could you do a test and see how our driver behaves in this scenario? I'd also like to know how other drivers handle it, for example the Java driver - it seems the most mature of the drivers I know.

muzarski:
> I'd like to know more about how the driver behaves when a signal to reprepare is missed (i.e. some other application reprepares and this one misses it). The scenario with `SELECT *` and adding a column is known, but I wonder what happens if we have a UDT column or a tuple and we add a field to the UDT/tuple. I think this case might actually be quite common contrary to `SELECT *` so I'm wondering whether other drivers handle it correctly and if we handle it correctly.
> 
> Could you do a test and see how our driver behaves in this scenario? I'd also like to know how other drivers handle it, for example the Java driver - it seems the most mature of the drivers I know.

So I've tested the UDT case you provided with both rust and java drivers. The key thing to notice is that java driver actually does NOT update the result metadata of the prepared statement during repreparations. The metadata is immutable - set only once during the first preparation of the statement. In result, the scenario you described fails even when only one client connects to the database.

The demo (on `scylla-4.x` branch):
```java
import com.datastax.oss.driver.api.core.CqlSession;
import com.datastax.oss.driver.api.core.cql.PreparedStatement;
import com.datastax.oss.driver.api.core.cql.Row;
import com.datastax.oss.driver.api.core.data.UdtValue;

public class CreateAndPopulateKeyspace {

  public static void main(String[] args) {

    CreateAndPopulateKeyspace client = new CreateAndPopulateKeyspace();

    try {
      client.connect();
      client.createSchema();
      client.prepareStatement();
      client.alterSchema();
      client.reprepareStatement();

    } catch (Exception ex) {
      ex.printStackTrace();
    } finally {
      client.close();
    }
  }

  private CqlSession session;
  private PreparedStatement preparedStatement;

  public void connect() {
    session = CqlSession.builder().build();

    System.out.printf("Connected session: %s%n", session.getName());
  }

  public void createSchema() {

    session.execute("DROP KEYSPACE IF EXISTS simplex");

    session.execute(
        "CREATE KEYSPACE IF NOT EXISTS simplex WITH replication "
            + "= {'class':'SimpleStrategy', 'replication_factor':1};");

    session.execute("CREATE TYPE simplex.my_type (a int)");

    session.execute(
        "CREATE TABLE simplex.tab (a int PRIMARY KEY, b int, c simplex.my_type)");

    session.execute("INSERT INTO simplex.tab (a, b, c) VALUES (1, 2, { a: 1 })");
  }

  public void prepareStatement() {
    preparedStatement = session.prepare("SELECT c FROM simplex.tab");
  }

  public void alterSchema() {
    session.execute("ALTER TYPE simplex.my_type ADD d blob");
  }

  public void reprepareStatement() {
    Row row = session.execute(preparedStatement.bind()).one();

    UdtValue udtValue = row.getUdtValue("c");
    System.out.println(udtValue);
  }
  
  public void close() {
    if (session != null) {
      session.close();
    }
  }
}
```

The output:
```
Using Scylla optimized driver!!!
08:36:27.544 [main] INFO  c.d.o.d.a.c.session.SessionBuilder - Using Scylla optimized driver!!!
08:36:27.770 [main] INFO  c.d.o.d.i.c.DefaultMavenCoordinates - DataStax Java driver for Apache Cassandra(R) (com.scylladb:java-driver-core) version 4.17.0.1-SNAPSHOT
08:36:27.859 [main] INFO  c.d.o.d.i.c.c.CqlPrepareAsyncProcessor - Adding handler to invalidate cached prepared statements on type changes
08:36:28.139 [s0-admin-0] INFO  c.d.o.d.internal.core.time.Clock - Using native clock for microsecond precision
Connected session: s0
08:36:36.463 [s0-io-4] WARN  c.d.o.d.i.core.cql.CqlRequestHandler - [s0|1848289347|0] Statement 0xf4bf63c95f8a8b21c40745cbc1055ff9 is not prepared on Node(endPoint=/127.0.0.1:9042, hostId=d99848a9-2754-43e9-9aa1-0d708514c8b3, hashCode=4c3d379d), repreparing
java.lang.IllegalArgumentException: Too many fields in encoded UDT value, expected 1
	at com.datastax.oss.driver.internal.core.type.codec.UdtCodec.decode(UdtCodec.java:109)
	at com.datastax.oss.driver.internal.core.type.codec.UdtCodec.decode(UdtCodec.java:32)
	at com.datastax.oss.driver.api.core.data.GettableByIndex.get(GettableByIndex.java:90)
	at com.datastax.oss.driver.api.core.data.GettableByIndex.get(GettableByIndex.java:127)
	at com.datastax.oss.driver.api.core.data.GettableByIndex.getUdtValue(GettableByIndex.java:550)
	at com.datastax.oss.driver.api.core.data.GettableByName.getUdtValue(GettableByName.java:650)
	at com.datastax.oss.driver.examples.basic.CreateAndPopulateKeyspace.reprepareStatement(CreateAndPopulateKeyspace.java:99)
	at com.datastax.oss.driver.examples.basic.CreateAndPopulateKeyspace.main(CreateAndPopulateKeyspace.java:35)

Process finished with exit code 0
```
This is the exact same reason why `SELECT *` prepared statement won't work when we add/drop the column to the table after statement preparation (even in 1-client scenario).

The same test case that uses rust driver works since we update the client-side result metadata after statement repreparation. However, this won't work in the following 2-client scenario:
- [CLIENT 1]: Prepare schema - create keyspace, udt, table (as in java example).
- [CLIENT 1]: Prepare the `SELECT c FROM simplex.tab` statement
- [CLIENT 2]: Prepare the `SELECT c FROM simplex.tab` statement
- [CLIENT 1]: Alter the user-defined type
- [CLIENT 1]: Execute the prepared statement - the repreparation takes a place, server caches the id of the prepared statement
- [CLIENT 2]: Execute the prepared statement - the server already holds an id of the prepared statement, we miss the reprepare and so we make use of the outdated metadata to deserialize the result. We end up with deserialization error (expected 1 field of UDT, received 2 fields).

Now the question is whether we want to follow the java driver's approach and make `result_metadata` immutable (then we can get rid of `ArcSwap` as well).

Unfortunately, no solution to the multi-client scenario comes to my mind. Since java driver is vulnerable to this as well, I believe there is no reasonable solution to the problem. AFAIK, the 5th version of CQL protocol offers some solution where the server informs all of its clients about the change of metadata.

piodul:
Thanks for checking.

I think it makes more sense to make the result metadata immutable. The approach with `ArcSwap` works OK only if you have only one instance of `PreparedStatement` for a given potentially-problematic query **across all running applications that share the same cluster**. I also think there might be some issues if you try to execute the same `PreparedStatement` instance and one of them gets reprepared... All in all, it only works in very specific circumstances which might be hard for the user to ensure. I think an incorrect but consistent behavior in the form of returning metadata that was returned at first prepare is better.

muzarski:
v2: rebased on top of master

muzarski:
v3 - addressed review comments:
- made `result_metadata` immutable - in result got rid of `ArcSwap`
- moved `skip_result_metadata` flag from `SessionConfig` to `StatementConfig`
- adjusted getters and setters for two types of metadata contained in `PreparedStatementSharedData`

Question related to tests:
- since the previously introduced test case tested whether the client-side metadata is updated during reprepare, we got rid of this test case as we made `result_metadata` immutable. What is the best way to test our solution? I believe we could maybe use `scylla-proxy` crate and check whether the server does not attach the metadata in the response when client sents `SKIP_METADATA` flag. Correct me if I'm wrong, as I don't have any experience with the proxy.

Note: there is still an issue with `metadata.clone()` mentioned in the review comment. I'm still investigating the issue.

muzarski:
v4: adjusted docstrings

scylla-cql/src/frame/response/result.rs
```diff
@@ -886,17 +886,29 @@ pub fn deser_cql_value(typ: &ColumnType, buf: &mut &[u8]) -> StdResult<CqlValue,
...
-fn deser_rows(buf: &mut &[u8]) -> StdResult<Rows, ParseError> {
-    let metadata = deser_result_metadata(buf)?;
+fn deser_rows(
+    buf: &mut &[u8],
+    cached_metadata: Option<&ResultMetadata>,
+) -> StdResult<Rows, ParseError> {
+    let server_metadata = deser_result_metadata(buf)?;
+
+    let metadata = match cached_metadata {
+        Some(metadata) => metadata.clone(),
```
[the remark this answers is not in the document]
piodul: I just realized that this problem will most likely go away after https://github.com/scylladb/scylla-rust-driver/issues/462 is finally implemented - `Rows` will keep data in unserialized form. Therefore, i don't think it's worth spending time on solving this until #462 is done.

piodul:
> - since the previously introduced test case tested whether the client-side metadata is updated during reprepare, we got rid of this test case as we made result_metadata immutable. What is the best way to test our solution? I believe we could maybe use scylla-proxy crate and check whether the server does not attach the metadata in the response when client sents SKIP_METADATA flag. Correct me if I'm wrong, as I don't have any experience with the proxy.

Yes, the proxy would be the right tool for this kind of test.

piodul:
And - thanks for adding the descriptions as I requested. I really appreciate that.

Don't forget to update it when you make changes to the implementation that would make it outdated (I see that it mentions `ArcSwap` right now, but AFAIK you got rid of it).

muzarski:
Rebased on top of main, so the CI passes

scylla/tests/integration/skip_metadata_optimization.rs
```diff
@@ -0,0 +1,87 @@
...
+
+            let (frame, _shard) = rx.recv().await.unwrap();
+            let mut buf = &*frame.body;
+
+            match types::read_int(&mut buf).unwrap() {
+                0x0002 => (),
+                _ => panic!("Invalid result type"),
+            }
+            let result_metadata_flags = types::read_int(&mut buf).unwrap();
+            assert!(predicate(result_metadata_flags));
```
piodul, line 68: Can we deserialize the response using `scylla-cql` instead of doing it manually here?

piodul (changes requested):

scylla/tests/integration/skip_metadata_optimization.rs
muzarski: At first, I wanted to use `scylla_cql::frame::response::result::deser_result_metadata`. The thing is that this function is private.

To use this function in the integration tests (placed in `scylla` crate), we would need to:
- make it `pub` in `scylla-cql`
- re-export it as `pub` in `scylla` - unfortunately the function needs to be visible outside of the crate to use it in the integration tests corresponding to the crate.

I've recognized this issue when working on limiting the `pub` re-exports (I'll probably post PR today). There are some utility functions such as: `scylla_cql::frame::types::read/write_int` which unfortunately have to be public, since they are used in the integration tests in `scylla` crate. The only solution to this is to add `#[doc(hidden)]` macro before the re-export, so it's not displayed in the documentation (or possibly add some mention in the docs that these are some utility functions intended for the internal usage).

I could pub re-export the `deser_result_metadata` function and use it in the test case. WDYT?

piodul: What about `Response::deserialize` (scylla-cql/src/frame/response/mod.rs)? It is already public as far as I can tell, can we use it here?

piodul: Ugh, ok. Again, as I noted in https://github.com/scylladb/scylla-rust-driver/pull/925#discussion_r1482806217 it will be possible to solve this after #462 because I'm going to change `Response::deserialize` so that it keeps the row data in serialized form.

For now we can keep it as it is, but please add a FIXME comment.

muzarski:
v5: addressed review comments

scylla/src/statement/prepared_statement.rs
```diff
@@ -270,6 +270,27 @@ impl PreparedStatement {
...
+    /// attach the result metadata in response to the statement execution.
+    ///
+    /// The driver will cache the result metadata received from the server
+    /// after statement preparation and will use it
+    /// to deserialize the results of statement execution.
+    ///
+    /// This option is false by default.
+    pub fn set_use_cached_result_metadata(&mut self, should_skip: bool) {
+        self.config.skip_result_metadata = should_skip;
+    }
```
piodul, line 289: Please change the name of the parameter.

image

The name of the parameter is shown in the documentation, and it looks confusing in relation to the method's name.

muzarski:
v6:
- resolved conflicts with main
- changed the name of the parameter of `StatementConfig::set_use_cached_result_metadata` from `should_skip` to `use_cached_metadata`

piodul (approved):

PR #926, opened 2024-02-02
Docs: Preprocessor for Python < 3.12

Lorak-mmk:
Preprocessor script was using syntax that works with Python 3.12 but not with older version.
Namely, string in format string using the same delimiter as outer string. Example: f'abc{'def'}xyz'

This commit changes delimiters of inner string to make script compatible with older versions of Python.

## Pre-review checklist

### DIFF
--- a/docs/sphinx_preprocessor.py
+++ b/docs/sphinx_preprocessor.py
@@ -48,7 +48,7 @@ def remove_sphinx_markdown(md_file_text, name):
 
 def process_section(section):
     if 'Chapter' in section:
-        print(f'Processing chapter {section['Chapter']['name']}', file=sys.stderr)
+        print(f'Processing chapter {section["Chapter"]["name"]}', file=sys.stderr)
         section['Chapter']['content'] = remove_sphinx_markdown(section['Chapter']['content'], section['Chapter']['name'])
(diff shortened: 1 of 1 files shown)


piodul:
> Preprocessor script was using syntax that works with Python 3.12 but not with older version.

Why did the CI checks pass in the PR that introduced it, then? https://github.com/scylladb/scylla-rust-driver/pull/910

As a side not, please also make sure to tick off the relevant boxes in the "Pre-review checklist". They are not optional for the maintainers.

Lorak-mmk:
> > Preprocessor script was using syntax that works with Python 3.12 but not with older version.
> 
> Why did the CI checks pass in the PR that introduced it, then? #910

Preprocessor crash is not a hard error for documentation building - it will just build using unmodified sources.

piodul:
> > > Preprocessor script was using syntax that works with Python 3.12 but not with older version.
> > 
> > 
> > Why did the CI checks pass in the PR that introduced it, then? #910
> 
> Preprocessor crash is not a hard error for documentation building - it will just build using unmodified sources.

Yeah, looks like it, and it is already reported as a bug: https://github.com/rust-lang/mdBook/issues/1872

piodul (approved):

PR #933, opened 2024-02-15
Decrease pub visibility of `scylla-cql` definitions

muzarski:
## Motivation

Some definitions in `scylla-cql` crate need to be public, so the `scylla` crate can make use of them. However, currently all of the definitions are implicitly re-exported outside and can be seen by the driver users. The idea is to explicitly re-export the types/functions that should be visible outside (and leave the rest of the public definitions for the internal usage of `scylla` crate).

## re-exported types
The main difficulty was to decide on what types should be re-exported publicly. The types that got re-exported can be seen in the documentation (`cargo doc --open`).

### Error types
I wasn't sure whether error types such as `scylla_cql::frame::frame_errors::FrameError` should be public. As of now, they are left public but this can be further discussed.

## Pre-review checklist

### DIFF
--- a/scylla/benches/benchmark.rs
+++ b/scylla/benches/benchmark.rs
@@ -1,11 +1,11 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 
 use bytes::BytesMut;
-use scylla::{
-    frame::types,
-    transport::partitioner::{calculate_token_for_partition_key, Murmur3Partitioner},
+use scylla::transport::partitioner::{calculate_token_for_partition_key, Murmur3Partitioner};
+use scylla_cql::{
+    frame::{response::result::ColumnType, types},
+    types::serialize::row::SerializedValues,
 };
-use scylla_cql::{frame::response::result::ColumnType, types::serialize::row::SerializedValues};
 
 fn types_benchmark(c: &mut Criterion) {
     let mut buf = BytesMut::with_capacity(64);
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -114,8 +114,15 @@ pub mod frame {
     }
 
     pub mod response {
+        pub use scylla_cql::frame::response::cql_to_rust;
         pub(crate) use scylla_cql::frame::response::*;
-        pub use scylla_cql::frame::response::{cql_to_rust, result};
+
+        pub mod result {
+            pub(crate) use scylla_cql::frame::response::result::*;
+            pub use scylla_cql::frame::response::result::{
+                ColumnSpec, ColumnType, CqlValue, PartitionKeyIndex, Row, TableSpec,
+            };
+        }
     }
 }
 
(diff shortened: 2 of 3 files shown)


wprzytula (approved):
Looks like a careful step in the right direction.
This PR only hides some definitions, without modyfing any import paths - which is good from the aspect of not breaking existing code.

For comparison, please see my approach to module refactor, which additionally restructures import paths in a way that makes more sense for me:
https://github.com/wprzytula/scylla-rust-driver/blob/restructure-modules/scylla/src/cql/mod.rs

scylla/benches/benchmark.rs
```diff
@@ -1,11 +1,11 @@
...
 use bytes::BytesMut;
-use scylla::{
-    frame::types,
-    transport::partitioner::{calculate_token_for_partition_key, Murmur3Partitioner},
+use scylla::transport::partitioner::{calculate_token_for_partition_key, Murmur3Partitioner};
+use scylla_cql::{
+    frame::{response::result::ColumnType, types},
+    types::serialize::row::SerializedValues,
 };
-use scylla_cql::{frame::response::result::ColumnType, types::serialize::row::SerializedValues};
```
Lorak-mmk, line 8: Why is this change necessary? Apart from formatting / grouping, I see only 1 change - importing `scylla_cql::frame::types` instead of `scylla::frame::types`. but I see in `lib.rs` that `scylla::frame::types` still exists, so why change this file at all?

Lorak-mmk (changes requested):

scylla/benches/benchmark.rs
wprzytula: @muzarski

muzarski: `scylla_cql::frame::types` has a lot of utility functions such as `read/write_int`. These should not be public. OTOH, it contains the definitions of `Consistency` and `SerialConsistency` - these obviously should be public. This is the change we introduce to `scylla/src/lib.rs`.

Why do we need the change in `scylla/benches/benchmark.rs`? The answer is that we cannot use private definitions of `scylla` crate there. Benchmarks use some of the utility functions that I just mentioned above (which were just made private). This is why we change `scylla::frame::types` to `scylla_cql::frame::types` (in this crate, these utility functions are public).

Lorak-mmk (approved):

muzarski:
v2: resolved conflicts with main

scylla/src/lib.rs
```diff
@@ -102,7 +102,30 @@ pub mod macros;
...
+    pub mod types {
+        pub use scylla_cql::frame::types::{Consistency, SerialConsistency};
+    }
+
+    pub mod response {
+        pub use scylla_cql::frame::response::cql_to_rust;
+        pub(crate) use scylla_cql::frame::response::*;
+
+        pub mod result {
+            pub(crate) use scylla_cql::frame::response::result::*;
```
wprzytula, line 121: I'm not sure if I'm happy with the glob imports here. Being `pub(crate)` ones, they are relatively benign though.

wprzytula (approved):

PR #934, opened 2024-02-15
Remove `strum` and `strum_macros` dependencies

muzarski:

## Motivation
`strum` and `strum_macros` are unstable crates and their usages need to be removed from the public API to stabilize it.

These crates are responsible for generating `FromStr` implementations for the types.

## Changes
Replaced `strum` usages for `NativeType` and `ColumnKind` types with our own implementations of `FromStr`.

## Pre-review checklist


piodul:
The semver checks job silently failed here for the reason described here: https://github.com/scylladb/scylla-rust-driver/issues/936. It looks like `/dev/tty` very recently stopped being available in our job runner's environment.

PR #935, opened 2024-02-15
Remove `histogram::Histogram` from public API

muzarski:

## Motivation
`histogram` is an unstable crate and the types introduced by this crate need to be removed from our public API. The only type being used is `histogram::Histogram`.

It's exposed publicly via `MetricsError::Poison`. However, the `MetricsError::Poison` value is never constructed since we always unwrap the `PoisonError` when trying to acquire the lock.

## Changes
Removed `MetricsError::Poison` and refactored `MetricsError` to be a struct holding an error cause (`&'static str` returned from the `histogram::Histogram` API).

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/metrics.rs
+++ b/scylla/src/transport/metrics.rs
@@ -1,30 +1,23 @@
 use histogram::Histogram;
 use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::{Arc, Mutex, MutexGuard, PoisonError};
+use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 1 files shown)


piodul (approved):

piodul:
Curious, this was quite an obvious breaking change (removal of an enum variant in a public struct) but the semver checks job didn't report anything.

piodul:
Something went wrong when running the check:

image

Not sure what went wrong, but the CI job with the semver check should fail and not pass. I'll create an issue about that.

In any case, this is not a blocker for merge.

PR #939, opened 2024-02-22
Rename `chrono`, `time` and `secret` features

muzarski:

## Motivation

To follow the same naming conventions as for other features, we want to add a crate version numbers to the remaining dependencies hidden behind crate features: `chrono`, `time` and `secrecy`.

### Dependecies versions
I bumped `secrecy` crate version from `0.7.0` to `0.8.0` which is the most recent version.

### Feature names
Renamed following features (and dependencies):
- feature `chrono` => feature `chrono-04`, dependency `chrono` => dependency `chrono_04` (scylla-cql)
- feature `time` => feature `time-03`, dependency `time` => dependency `time_03` (scylla-cql)
- feature `secret` => feature `secrecy-08`, dependency `secrecy` => dependency `secrecy_08` 

## Pre-review checklist

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -13,7 +13,7 @@ rustyline-derive = "0.6"
 scylla = { path = "../scylla", features = [
     "ssl",
     "cloud",
-    "chrono",
+    "chrono-04",
     "time-03",
     "num-bigint-03",
     "num-bigint-04",
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -14,7 +14,7 @@ scylla = { path = "../scylla", features = [
     "ssl",
     "cloud",
     "chrono",
-    "time",
+    "time-03",
     "num-bigint-03",
     "num-bigint-04",
     "bigdecimal-04",
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -15,7 +15,7 @@ byteorder = "1.3.4"
 bytes = "1.0.1"
 # FIXME: Remove <1.38 once https://github.com/tokio-rs/tokio/issues/6610 is fixed
 tokio = { version = ">=1.34, <1.38", features = ["io-util", "time"] }
-secrecy = { version = "0.7.0", optional = true }
+secrecy = { version = "0.8", optional = true }
 snap = "1.0"
 uuid = "1.0"
 thiserror = "1.0"
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -22,7 +22,7 @@ thiserror = "1.0"
 num-bigint-03 = { package = "num-bigint", version = "0.3", optional = true }
 num-bigint-04 = { package = "num-bigint", version = "0.4", optional = true }
 bigdecimal-04 = { package = "bigdecimal", version = "0.4", optional = true }
-chrono = { version = "0.4.32", default-features = false, optional = true }
+chrono-04 = { package = "chrono", version = "0.4.32", default-features = false, optional = true }
 lz4_flex = { version = "0.11.1" }
 async-trait = "0.1.57"
 serde = { version = "1.0", features = ["derive"], optional = true }
@@ -42,12 +42,12 @@ harness = false
 [features]
 secrecy-08 = ["dep:secrecy-08"]
 time-03 = ["dep:time-03"]
-chrono = ["dep:chrono"]
+chrono-04 = ["dep:chrono-04"]
 num-bigint-03 = ["dep:num-bigint-03"]
 num-bigint-04 = ["dep:num-bigint-04"]
 bigdecimal-04 = ["dep:bigdecimal-04"]
 full-serialization = [
-    "chrono",
+    "chrono-04",
     "time-03",
     "secrecy-08",
     "num-bigint-03",
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -15,7 +15,7 @@ byteorder = "1.3.4"
 bytes = "1.0.1"
 # FIXME: Remove <1.38 once https://github.com/tokio-rs/tokio/issues/6610 is fixed
 tokio = { version = ">=1.34, <1.38", features = ["io-util", "time"] }
-secrecy = { version = "0.8", optional = true }
+secrecy-08 = { package = "secrecy", version = "0.8", optional = true }
 snap = "1.0"
 uuid = "1.0"
 thiserror = "1.0"
@@ -40,7 +40,7 @@ name = "benchmark"
 harness = false
 
 [features]
-secret = ["secrecy"]
+secrecy-08 = ["dep:secrecy-08"]
 time = ["dep:time"]
 chrono = ["dep:chrono"]
 num-bigint-03 = ["dep:num-bigint-03"]
@@ -49,7 +49,7 @@ bigdecimal-04 = ["dep:bigdecimal-04"]
 full-serialization = [
     "chrono",
     "time",
-    "secret",
+    "secrecy-08",
     "num-bigint-03",
     "num-bigint-04",
     "bigdecimal-04",
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -26,13 +26,13 @@ chrono = { version = "0.4.32", default-features = false, optional = true }
 lz4_flex = { version = "0.11.1" }
 async-trait = "0.1.57"
 serde = { version = "1.0", features = ["derive"], optional = true }
-time = { version = "0.3", optional = true }
+time-03 = { package = "time", version = "0.3", optional = true }
 
 [dev-dependencies]
 assert_matches = "1.5.0"
 criterion = "0.4"        # Note: v0.5 needs at least rust 1.70.0
 # Use large-dates feature to test potential edge cases
-time = { version = "0.3.21", features = ["large-dates"] }
+time-03 = { package = "time", version = "0.3.21", features = ["large-dates"] }
 uuid = { version = "1.0", features = ["v4"] }
 
 [[bench]]
@@ -41,14 +41,14 @@ harness = false
 
 [features]
 secrecy-08 = ["dep:secrecy-08"]
-time = ["dep:time"]
+time-03 = ["dep:time-03"]
 chrono = ["dep:chrono"]
 num-bigint-03 = ["dep:num-bigint-03"]
 num-bigint-04 = ["dep:num-bigint-04"]
 bigdecimal-04 = ["dep:bigdecimal-04"]
 full-serialization = [
     "chrono",
-    "time",
+    "time-03",
     "secrecy-08",
     "num-bigint-03",
     "num-bigint-04",
--- a/scylla-cql/src/frame/response/cql_to_rust.rs
+++ b/scylla-cql/src/frame/response/cql_to_rust.rs
@@ -8,9 +8,6 @@ use std::net::IpAddr;
 use thiserror::Error;
 use uuid::Uuid;
 
-#[cfg(feature = "chrono")]
-use chrono::{DateTime, NaiveDate, NaiveTime, Utc};
-
 #[derive(Error, Debug, Clone, PartialEq, Eq)]
 pub enum FromRowError {
     #[error("{err} in the column with index {column}")]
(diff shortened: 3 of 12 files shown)


scylla/Cargo.toml
```diff
@@ -39,7 +39,7 @@ rand = "0.8.3"
 thiserror = "1.0"
 itertools = "0.11.0"
 tracing = "0.1.36"
-chrono = { version = "0.4.20", default-features = false, features = ["clock"] }
+chrono = { version = "0.4", default-features = false, features = ["clock"] }
```
Lorak-mmk, line 42: Did you check if the driver builds with `0.4.0`?
I see that it was introduced in this commit: https://github.com/scylladb/scylla-rust-driver/commit/10debf60725e8531b1e99904a013da9837570765
And that in `0.4.0` there are no `MIN / MAX` items in `NaiveData` which are used by us here: https://github.com/scylladb/scylla-rust-driver/blob/87908f3c73cdece258d28b5b2191a843e7da3b98/scylla/src/transport/cql_types_test.rs#L301
See docs for chrono: https://docs.rs/chrono/0.4.0/chrono/naive/struct.NaiveDate.html

So before dropping requirements `chrono` / `time` or other requirements version, please check if everything still compiles and passes with such versions. You can do that e.g. by using exact match in Cargo.toml, e.g. `chrono = { version = "=0.4.0", default-features = false, features = ["clock"] }`

scylla-cql/Cargo.toml
```diff
@@ -14,33 +14,33 @@ scylla-macros = { version = "0.4.0", path = "../scylla-macros" }
...
-secrecy = { version = "0.7.0", optional = true }
+secrecy-08 = { package = "secrecy", version = "0.8", optional = true }
 snap = "1.0"
 uuid = "1.0"
 thiserror = "1.0"
 num-bigint-03 = { package = "num-bigint", version = "0.3", optional = true }
 num-bigint-04 = { package = "num-bigint", version = "0.4", optional = true }
 bigdecimal-04 = { package = "bigdecimal", version = "0.4", optional = true }
-chrono = { version = "0.4.27", default-features = false, optional = true }
+chrono-04 = { package = "chrono", version = "0.4", default-features = false, optional = true }
```
Lorak-mmk, line 24: It's the same thing here, please check that scylla-cql really doesn't require any version higher that `0.4.0`

scylla-cql/Cargo.toml
```diff
@@ -14,33 +14,33 @@ scylla-macros = { version = "0.4.0", path = "../scylla-macros" }
...
 async-trait = "0.1.57"
 serde = { version = "1.0", features = ["derive"], optional = true }
-time = { version = "0.3", optional = true }
+time-03 = { package = "time", version = "0.3", optional = true }
 
 [dev-dependencies]
 criterion = "0.4" # Note: v0.5 needs at least rust 1.70.0
 # Use large-dates feature to test potential edge cases
-time = { version = "0.3.21", features = ["large-dates"] }
+time-03 = { package = "time", version = "0.3", features = ["large-dates"] }
```
Lorak-mmk, line 33: And here again

Lorak-mmk (changes requested):

scylla-cql/Cargo.toml
muzarski: Same for this dependency, the `time::PrimitiveDateTime::MAX/MIN` are missing in version `0.3.0`. I'll revert the changes dropping the minimum req version of `time` crate as well.

muzarski:
v2: reverted the changes that dropped the minimum req version of `time` and `chrono` dependencies

scylla-cql/Cargo.toml
```diff
@@ -14,33 +14,33 @@ scylla-macros = { version = "0.4.0", path = "../scylla-macros" }
 byteorder = "1.3.4"
 bytes = "1.0.1"
 tokio = { version = "1.12", features = ["io-util", "time"] }
-secrecy = { version = "0.7.0", optional = true }
+secrecy-08 = { package = "secrecy", version = "0.8", optional = true }
```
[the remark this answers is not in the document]
wprzytula: Having read about [Cargo dependency resolver rules](https://doc.rust-lang.org/cargo/reference/resolver.html) once again, I think there's no difference between specifying `0.8` and `0.8.0`.

examples/Cargo.toml
```diff
@@ -10,7 +10,7 @@ futures = "0.3.6"
 openssl = "0.10.32"
 rustyline = "9"
 rustyline-derive = "0.6"
-scylla = {path = "../scylla", features = ["ssl", "cloud", "chrono", "time", "num-bigint-03", "num-bigint-04", "bigdecimal-04"]}
+scylla = {path = "../scylla", features = ["ssl", "cloud", "chrono-04", "time-03", "num-bigint-03", "num-bigint-04", "bigdecimal-04"]}
```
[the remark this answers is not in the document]
muzarski: It's not like both of these features cannot be used at the same time. The user might (for some weird reason) want to support (de)serialization for both versions of num_bigint::BigInt.

There is other problem, though - `BigInt` is not used in the examples at all... This means that we do not make use of any of these features anyway. I'm not sure why I added these features to examples. Probably for some reason related to vscode rust-analyzer. As I remember, the vscode' analyzer compiler the driver code with the features used in the examples/ by default. However, this can be configured in the workspace settings.

I think we could remove unused features from examples, but is it really necessary?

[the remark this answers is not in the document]
wprzytula: If we really want to have all features enabled in examples, I'd suggest using "full_serialization" feature instead of listing them all manually.

Lorak-mmk (changes requested):
I'd really like to see support for 2 major versions of each lib (as is done with num-bigint) - that way we would know that we are actually able to support more than one

scylla-cql/src/frame/response/result.rs
```diff
@@ -161,8 +161,8 @@ impl CqlValue {
         self.as_cql_date().and_then(|date| date.try_into().ok())
     }
 
-    #[cfg(feature = "time")]
-    pub fn as_date(&self) -> Option<time::Date> {
+    #[cfg(feature = "time-03")]
+    pub fn as_date(&self) -> Option<time_03::Date> {
         self.as_cql_date().and_then(|date| date.try_into().ok())
     }
```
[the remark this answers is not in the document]
muzarski: Well, it looks like it should be removed then. Otherwise, we would need to introduce some weird version-specific names for the corresponding methods. WDYT?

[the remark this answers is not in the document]
wprzytula: @Lorak-mmk It seems your answer is awaited.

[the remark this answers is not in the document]
Lorak-mmk: We obviously can't just remove the method - how would we handle dates then? I'm not sure what the best solution is here. Current one is not feasible - why put those crates behing per-version features when we can't add a new one? Any ideas?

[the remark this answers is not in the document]
Lorak-mmk: I don't immediately see any better solution than per-version methods.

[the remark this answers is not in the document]
muzarski: > We obviously can't just remove the method - how would we handle dates then?

Notice that there are no corresponding methods for `num_bigint::BigInt`. These methods are not really necessary - they are just handy utility functions. Removing them would require the users to explicitly convert `CqlValue` to `value::CqlDate` (using `as_cql_date` method), which then, could be converted to specific `time::Date` version (using the `TryInto` implementation).

Knowing that we could possibly remove the function, do you still think that introducing per-version methods is a better solution?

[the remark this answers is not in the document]
Lorak-mmk: In that case yes, I think we should remove the method.

muzarski:
> I'd really like to see support for 2 major versions of each lib (as is done with num-bigint) - that way we would know that we are actually able to support more than one

With @Lorak-mmk, we had a discussion offline and decided not to introduce support for older versions of `chrono` and `time` crates. The commits that introduce support for `time 0.3` and `chrono 0.4` are basically a showcase that the support for future versions of these crates can be easily introduced. These commits should be dropped before merge.

Lorak-mmk (approved):

wprzytula:
@muzarski Please remove mock commits, then I'll review the final version.

muzarski:
v3:
- adjusted job names in CI
- dropped mock commits

scylla-cql/src/frame/response/cql_to_rust.rs
```diff
@@ -8,11 +8,8 @@ use std::net::IpAddr;
 use thiserror::Error;
 use uuid::Uuid;
 
-#[cfg(feature = "chrono")]
-use chrono::{DateTime, NaiveDate, NaiveTime, Utc};
-
-#[cfg(feature = "secret")]
-use secrecy::{Secret, Zeroize};
+#[cfg(feature = "secrecy-08")]
+use secrecy_08::{Secret, Zeroize};
```
wprzytula, line 13: Why remove `chrono` `use`s but preserve `secrecy` uses at the same time?

scylla-cql/src/frame/response/cql_to_rust.rs
```diff
@@ -568,43 +565,49 @@ mod tests {
...
+            Err(FromCqlValError::BadVal),
+            time_03::Date::from_cql(max_date)
+        );
 
         // Different CQL type. Since value can't be casted, it should return `BadCqlType` error
         let bad_type = CqlValue::Double(0.5);
         assert_eq!(
             Err(FromCqlValError::BadCqlType),
-            time::Date::from_cql(bad_type)
+            time_03::Date::from_cql(bad_type)
```
wprzytula, line 610: Not related to this PR, but please fix it as a bonus:
`casted` -> `cast`

scylla-cql/src/frame/response/cql_to_rust.rs
```diff
@@ -669,48 +672,51 @@ mod tests {
...
-            time::Time::from_cql(bad_time2)
+            time_03::Time::from_cql(bad_time2)
         );
 
         // Different CQL type. Since value can't be casted, it should return `BadCqlType` error
         let bad_type = CqlValue::Double(0.5);
         assert_eq!(
             Err(FromCqlValError::BadCqlType),
-            time::Time::from_cql(bad_type)
+            time_03::Time::from_cql(bad_type)
```
wprzytula, line 719: ditto: `casted` -> `cast`

scylla-cql/src/frame/response/cql_to_rust.rs
```diff
@@ -744,55 +750,55 @@ mod tests {
...
-            time::OffsetDateTime::from_cql(max_timestamp)
+            time_03::OffsetDateTime::from_cql(max_timestamp)
         );
 
         // Different CQL type. Since value can't be casted, it should return `BadCqlType` error
         let bad_type = CqlValue::Double(0.5);
         assert_eq!(
             Err(FromCqlValError::BadCqlType),
-            time::OffsetDateTime::from_cql(bad_type)
+            time_03::OffsetDateTime::from_cql(bad_type)
```
wprzytula, line 801: ditto: `casted` -> `cast`

scylla-cql/src/frame/response/result.rs
```diff
@@ -1428,21 +1431,21 @@ mod tests {
...
+        use time_03::Time;
 
         // 0 when converted to NaiveTime is 0:0:0.0
         let midnight = Time::from_hms_nano(0, 0, 0, 0).unwrap();
         let time =
             super::deser_cql_value(&ColumnType::Time, &mut (0i64).to_be_bytes().as_ref()).unwrap();
 
         dbg!(&time);
-        assert_eq!(time.as_time(), Some(midnight));
+        assert_eq!(time.as_time_03(), Some(midnight));
```
wprzytula, line 1448: This `dbg!` shouldn't be here. Let's remove it.

scylla-cql/src/types/serialize/value.rs
```diff
@@ -9,11 +9,8 @@ use std::sync::Arc;
 use thiserror::Error;
 use uuid::Uuid;
 
-#[cfg(feature = "chrono")]
-use chrono::{DateTime, NaiveDate, NaiveTime, Utc};
-
-#[cfg(feature = "secret")]
-use secrecy::{ExposeSecret, Secret, Zeroize};
+#[cfg(feature = "secrecy-08")]
+use secrecy_08::{ExposeSecret, Secret, Zeroize};
```
wprzytula, line 14: ditto: why is `chrono` no longer `use`d, but `secrecy` still is?

wprzytula:
Is the new deserialization code that uses `time`, `chrono` and `secrecy` deps adjusted as well?

muzarski:
> Is the new deserialization code that uses `time`, `chrono` and `secrecy` deps adjusted as well?

Good point.

muzarski:
v3.1:
- addressed minor issues from review comments
- adjusted deserialization module to make use of proper feature flags

wprzytula (approved):

muzarski:
@Lorak-mmk @wprzytula It has 2 approves. Can we finally merge this?

Issue #936, opened 2024-02-16
CI: semver checks job breaks on `tee /dev/tty`, and it shouldn't silently pass


piodul:
Happened in: https://github.com/scylladb/scylla-rust-driver/actions/runs/7918304923/job/21616399359?pr=935
Full logs from the job: [logs_12589.zip](https://github.com/scylladb/scylla-rust-driver/files/14308264/logs_12589.zip)

![image](https://github.com/scylladb/scylla-rust-driver/assets/4950936/91b84cab-4699-4394-b5c5-ce7565d9f889)

There are two issues here:

- `/dev/tty` seems to be unavailable in the environment used to run the job, causing the command to fail,
- In case of such errors, the job should not succeed but rather fail.

cc: @Lorak-mmk

PR #942, opened 2024-02-28
Fix semver checks

Lorak-mmk:
Previously semver checks CI used `/dev/tty` which seems unavailable in Github Actions.
Changed it to `/proc/self/fd/2` which I expect to be available pretty much anywhere.
Changed this CI step so that the job fails on errors. It wouldn't prevent #936 , because this command is part of pipeline, so it won't be affected by `set -e`, but I think it's a good change anyway.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/936

## Pre-review checklist

### DIFF
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -66,7 +66,7 @@ jobs:
         set +e
         echo "output<<SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
         # Weird sed strip ANSI colors from output
-        make semver-rev rev="$PR_BASE" |& tee /dev/tty | sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g" >> $GITHUB_OUTPUT
+        make semver-rev rev="$PR_BASE" |& tee /proc/self/fd/2 | sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g" >> $GITHUB_OUTPUT
         exitcode=${PIPESTATUS[0]}
         echo "SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
 
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -72,8 +72,6 @@ jobs:
 
         echo "Semver checks exitcode: " $exitcode
         echo "exitcode=$exitcode" >> $GITHUB_OUTPUT
-        exit "$exitcode"
-      continue-on-error: true
 
   semver-pull-request-label:
     runs-on: ubuntu-latest
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -63,11 +63,16 @@ jobs:
     - name: Verify the API compatibilty with PR base
       id: semver-pr-check
       run: |
-        set +e
+        set -e # So that failed commands exit the script
+        set -o pipefail # So that if a command in a pipe fails, the whole command fails
+
         echo "output<<SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
+        SEMVER_REV_OUTPUT=$(make semver-rev rev="$PR_BASE" 2>&1) && true # "&& true" preserves exit code but cancels effects of set -e
+        exitcode=$?
+
         # Weird sed strip ANSI colors from output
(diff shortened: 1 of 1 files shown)


Lorak-mmk:
Tested that it works now in https://github.com/scylladb/scylla-rust-driver/pull/943

piodul:
> It wouldn't prevent #936 , because this command is part of pipeline, so it won't be affected by `set -e`, but I think it's a good change anyway.

So, what kind of new errors will we catch now? Why can't we catch errors such as those which manifested in #936 ?

piodul:
> I'm not sure how to best prevent such errors. I guess we could manually check output of steps of the pipeline using `PIPESTATUS` array and manually error out if second or third step failed. One disadvantage is that it would be more bash code that could fail. Would you like me to do this?

I educated myself on bash error handling a little bit and I think we both make it work and make it readable. I applied what I learnt to your code and I think it should look like this after modifications:

```bash
set -e # So that failed commands exit the script
set -o pipefail # So that if a command in a pipe fails, the whole command fails

echo "output<<SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
SEMVER_REV_OUTPUT="$(semver-rev rev="$PR_BASE" 2>&1)" && true # "&& true" preserves exit code but cancels effects of set -e
exitcode=$?

# Weird sed strip ANSI colors from output
# If any of the commands below fail, `set -e` and `set -o pipefail` should exit the script
echo "${SEMVER_REV_OUTPUT}" | tee /proc/self/fd/2 | sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g" >> $GITHUB_OUTPUT
echo "Semver checks exitcode: " $exitcode
echo "SEMVER_STDOUT_EOF" >> $GITHUB_OUTPUT
echo "exitcode=$exitcode" >> $GITHUB_OUTPUT
```

Note that I didn't test this code, this is just to demonstrate the ideas.

wprzytula:
> ```shell
> SEMVER_REV_OUTPUT="$(semver-rev rev="$PR_BASE" 2>&1)" && true # "&& true" preserves exit code but cancels effects of set -e
> ```

Are you sure you meant `&& true`, not `|| true` ? IIUC, what we want to achieve here is that even if `semver-rev` command fails, `set -e` won't trigger an exit at this point.

Lorak-mmk:
> > ```shell
> > SEMVER_REV_OUTPUT="$(semver-rev rev="$PR_BASE" 2>&1)" && true # "&& true" preserves exit code but cancels effects of set -e
> > ```
> 
> Are you sure you meant `&& true`, not `|| true` ? IIUC, what we want to achieve here is that even if `semver-rev` command fails, `set -e` won't trigger an exit at this point.

I tried it and `&&` seems correct. There is a slight error in @piodul's code - `"` shouldn't be there, so the line should look like this:
```
SEMVER_REV_OUTPUT=$(semver-rev rev="$PR_BASE" 2>&1) && true # "&& true" preserves exit code but cancels effects of set -e
```

If we used `||`, it would not preserve exit code.

```
karolbaryla@localhost:~$ set -e
karolbaryla@localhost:~$ set_status() { return "$1"; }
karolbaryla@localhost:~$ $(set_status 2) && true
karolbaryla@localhost:~$ echo $?
2
```

```
karolbaryla@localhost:~$ set -e
karolbaryla@localhost:~$ set_status() { return "$1"; }
karolbaryla@localhost:~$ $(set_status 2) || true
karolbaryla@localhost:~$ echo $?
0
```

piodul:
> > ```shell
> > SEMVER_REV_OUTPUT="$(semver-rev rev="$PR_BASE" 2>&1)" && true # "&& true" preserves exit code but cancels effects of set -e
> > ```
> 
> Are you sure you meant `&& true`, not `|| true` ? IIUC, what we want to achieve here is that even if `semver-rev` command fails, `set -e` won't trigger an exit at this point.

Yes, I meant `&& true`. https://stackoverflow.com/questions/2870992/automatic-exit-from-bash-shell-script-on-error/2871034#2871034

Lorak-mmk:
Updated the PR according to @piodul 's suggestion

wprzytula (approved):

piodul (approved):

piodul:
@Lorak-mmk although this is already merged, please update the PR description to reflect reality (the "It wouldn't prevent" part looks outdated)

PR #944, opened 2024-03-02
Shard selecting load balancing

Lorak-mmk:
This is a more polished version of https://github.com/scylladb/scylla-rust-driver/pull/791
The changes from #791 are:
- Totally reworked commit structure, so that each commit actually compiles and passes all tests and checks. Doing this took quite a lot of time. Original commit structure was... a bit chaotic.
- Fixed all the warnings
- Removed some functions that were no longer used.
- Updated a comment that became incorrect

Posting the original description, slightly modified, as it is still relevant:

# Motivation
Most of our drivers, being inherited from Cassandra, load balance only over nodes, not specific shards. Multiple ideas have arised that could benefit from having a shard-selecting load balancing. Among them:
- **shard-aware batching** (#738);
- **tablets support**:
with tablets enabled (ATM experimental in ScyllaDB), target shard is not derived from token (computed from partition key), but rather read from `system.tablets`. Therefore, a load balancer should be able to decide a target shard on its own, by abstracting over either token ring or tablets being used for cluster topology.
- **overloaded shard optimisation**:
some tests have shown that sometimes, when a shard is particularly overloaded, it may be beneficial (performance-wise) to send the request to the proper node, **but** a wrong shard. That shard would then do part of the work that the overloaded shard would else have to do itself.

# Design
- LB policy now is to return a `(NodeRef, Shard)` pair, enabling finer-grained control over targeted shards.
- regarding tablets support: `ReplicaLocator` is the place where the abstraction over either token ring or tablets is to be implemented. Ideally, the LB policy does not have to be aware of the actual mechanism (token ring or tablets) being used for a particular query.

# What's done
- internal and public load-balancing-related interfaces are changed from `NodeRef` to `(NodeRef, Shard)` pair,
- shard selection logic is removed from `NodeConnectionPool`; a method is added there that returns a connection to a specific shard,
- `Session`'s logic propagates the load balancing policy's target shard down to the connection pool,
- a stub implementation of shard selection is added to `ReplicaLocator`. At the moment, it simply computes the shard based on the token, the same way as it was done in the connection pool layer before.

## Pre-review checklist

### DIFF
--- a/docs/source/load-balancing/load-balancing.md
+++ b/docs/source/load-balancing/load-balancing.md
@@ -2,8 +2,8 @@
 
 ## Introduction
 
-The driver uses a load balancing policy to determine which node(s) to contact
-when executing a query. Load balancing policies implement the
+The driver uses a load balancing policy to determine which node(s) and shard(s)
+to contact when executing a query. Load balancing policies implement the
 `LoadBalancingPolicy` trait, which contains methods to generate a load
 balancing plan based on the query information and the state of the cluster.
 
@@ -12,12 +12,14 @@ being opened. For a node connection blacklist configuration refer to
 `scylla::transport::host_filter::HostFilter`, which can be set session-wide
 using `SessionBuilder::host_filter` method.
 
+In this chapter, "target" will refer to a pair `<node, optional shard>`.
+
 ## Plan
 
 When a query is prepared to be sent to the database, the load balancing policy
-constructs a load balancing plan. This plan is essentially a list of nodes to
+constructs a load balancing plan. This plan is essentially a list of targets to
 which the driver will try to send the query. The first elements of the plan are
-the nodes which are the best to contact (e.g. they might be replicas for the
+the targets which are the best to contact (e.g. they might be replicas for the
 requested data or have the best latency).
 
 ## Policy
@@ -84,17 +86,16 @@ first element of the load balancing plan is needed, so it's usually unnecessary
 to compute entire load balancing plan. To optimize this common case, the
 `LoadBalancingPolicy` trait provides two methods: `pick` and `fallback`.
 
-`pick` returns the first node to contact for a given query, which is usually
-the best based on a particular load balancing policy. If `pick` returns `None`,
-then `fallback` will not be called.
+`pick` returns the first target to contact for a given query, which is usually
+the best based on a particular load balancing policy.
 
-`fallback`, returns an iterator that provides the rest of the nodes in the load
-balancing plan. `fallback` is called only when using the initial picked node
-fails (or when executing speculatively).
+`fallback`, returns an iterator that provides the rest of the targets in the
+load balancing plan. `fallback` is called when using the initial picked
+target fails (or when executing speculatively) or when `pick` returned `None`.
 
-It's possible for the `fallback` method to include the same node that was
+It's possible for the `fallback` method to include the same target that was
 returned by the `pick` method. In such cases, the query execution layer filters
-out the picked node from the iterator returned by `fallback`.
+out the picked target from the iterator returned by `fallback`.
 
 ### `on_query_success` and `on_query_failure`:
 
--- a/scylla/src/transport/connection_pool.rs
+++ b/scylla/src/transport/connection_pool.rs
@@ -1,7 +1,7 @@
 #[cfg(feature = "cloud")]
 use crate::cloud::set_ssl_config_for_scylla_cloud_host;
 
-use crate::routing::{Shard, ShardCount, Sharder, Token};
+use crate::routing::{Shard, ShardCount, Sharder};
 use crate::transport::errors::QueryError;
 use crate::transport::{
     connection,
@@ -28,7 +28,7 @@ use std::time::Duration;
 
 use tokio::sync::{broadcast, mpsc, Notify};
 use tracing::instrument::WithSubscriber;
-use tracing::{debug, trace, warn};
+use tracing::{debug, error, trace, warn};
 
 /// The target size of a per-node connection pool.
 #[derive(Debug, Clone, Copy)]
@@ -235,22 +235,25 @@ impl NodeConnectionPool {
         .unwrap_or(None)
     }
 
-    pub(crate) fn connection_for_token(&self, token: Token) -> Result<Arc<Connection>, QueryError> {
-        trace!(token = token.value, "Selecting connection for token");
+    pub(crate) fn connection_for_shard(&self, shard: Shard) -> Result<Arc<Connection>, QueryError> {
+        trace!(shard = shard, "Selecting connection for shard");
         self.with_connections(|pool_conns| match pool_conns {
             PoolConnections::NotSharded(conns) => {
                 Self::choose_random_connection_from_slice(conns).unwrap()
             }
             PoolConnections::Sharded {
-                sharder,
                 connections,
+                sharder
             } => {
-                let shard: u16 = sharder
-                    .shard_of(token)
+                let shard = shard
                     .try_into()
-                    .expect("Shard number doesn't fit in u16");
-                trace!(shard = shard, "Selecting connection for token");
-                Self::connection_for_shard(shard, sharder.nr_shards, connections.as_slice())
+                    // It's safer to use 0 rather that panic here, as shards are returned by `LoadBalancingPolicy`
+                    // now, which can be implemented by a user in an arbitrary way.
+                    .unwrap_or_else(|_| {
+                        error!("The provided shard number: {} does not fit u16! Using 0 as the shard number. Check your LoadBalancingPolicy implementation.", shard);
+                        0
+                    });
+                Self::connection_for_shard_helper(shard, sharder.nr_shards, connections.as_slice())
             }
         })
     }
@@ -266,13 +269,13 @@ impl NodeConnectionPool {
                 connections,
             } => {
                 let shard: u16 = rand::thread_rng().gen_range(0..sharder.nr_shards.get());
-                Self::connection_for_shard(shard, sharder.nr_shards, connections.as_slice())
+                Self::connection_for_shard_helper(shard, sharder.nr_shards, connections.as_slice())
             }
         })
     }
 
     // Tries to get a connection to given shard, if it's broken returns any working connection
-    fn connection_for_shard(
+    fn connection_for_shard_helper(
         shard: u16,
         nr_shards: ShardCount,
         shard_conns: &[Vec<Arc<Connection>>],
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -3,7 +3,7 @@ pub use self::latency_awareness::LatencyAwarenessBuilder;
 
 use super::{FallbackPlan, LoadBalancingPolicy, NodeRef, RoutingInfo};
 use crate::{
-    routing::Token,
+    routing::{Shard, Token},
     transport::{cluster::ClusterData, locator::ReplicaSet, node::Node, topology::Strategy},
 };
 use itertools::{Either, Itertools};
@@ -70,13 +70,14 @@ enum StatementType {
 /// It can be configured to be datacenter-aware and token-aware.
 /// Datacenter failover for queries with non local consistency mode is also supported.
 /// Latency awareness is available, althrough not recommended.
+#[allow(clippy::type_complexity)]
 pub struct DefaultPolicy {
     preferences: NodeLocationPreference,
     is_token_aware: bool,
     permit_dc_failover: bool,
-    pick_predicate: Box<dyn Fn(&NodeRef) -> bool + Send + Sync>,
+    pick_predicate: Box<dyn Fn(&(NodeRef<'_>, Shard)) -> bool + Send + Sync>,
     latency_awareness: Option<LatencyAwareness>,
-    fixed_shuffle_seed: Option<u64>,
+    fixed_seed: Option<u64>,
 }
 
 impl fmt::Debug for DefaultPolicy {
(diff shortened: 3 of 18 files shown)


scylla/src/transport/load_balancing/default.rs
```diff
@@ -297,29 +302,37 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
...
                     NodeLocationCriteria::DatacenterAndRack(dc, rack),
                 );
-                Either::Left(Self::round_robin_nodes(local_nodes, rack_predicate))
+                Either::Left(self.round_robin_nodes_with_shards(local_nodes, rack_predicate))
             } else {
-                Either::Right(std::iter::empty::<NodeRef<'a>>())
+                Either::Right(std::iter::empty::<(NodeRef<'a>, Shard)>())
             };
-        let robined_local_nodes = Self::round_robin_nodes(local_nodes, Self::is_alive);
+        let robined_local_nodes = self.round_robin_nodes_with_shards(local_nodes, Self::is_alive);
```
wprzytula, line 309: Nit: this variable should be properly named `robinned`, or else it would be pronounced as *combined*.

scylla/src/transport/load_balancing/default.rs
```diff
@@ -297,29 +302,37 @@ or refrain from preferring datacenters (which may ban all other datacenters, if
...
 
         let all_nodes = cluster.replica_locator().unique_nodes_in_global_ring();
 
         // If a datacenter failover is possible, loosen restriction about locality.
         let maybe_remote_nodes = if self.is_datacenter_failover_possible(&routing_info) {
-            let robined_all_nodes = Self::round_robin_nodes(all_nodes, Self::is_alive);
+            let robined_all_nodes = self.round_robin_nodes_with_shards(all_nodes, Self::is_alive);
```
wprzytula, line 315: ditto

scylla/src/transport/load_balancing/default.rs
```diff
@@ -70,13 +70,14 @@ enum StatementType {
...
+#[allow(clippy::type_complexity)]
 pub struct DefaultPolicy {
     preferences: NodeLocationPreference,
     is_token_aware: bool,
     permit_dc_failover: bool,
-    pick_predicate: Box<dyn Fn(&NodeRef) -> bool + Send + Sync>,
+    pick_predicate: Box<dyn Fn(&(NodeRef<'_>, Shard)) -> bool + Send + Sync>,
     latency_awareness: Option<LatencyAwareness>,
-    fixed_shuffle_seed: Option<u64>,
+    fixed_seed: Option<u64>,
```
wprzytula, line 80: Justification for this name change: the seed is now not only used for **shuffling** nodes, but also for sampling a random shard in case we don't know the proper one.

scylla/tests/integration/execution_profiles.rs
```diff
@@ -46,9 +47,13 @@ impl<const NODE: u8> BoundToPredefinedNodePolicy<NODE> {
...
 impl<const NODE: u8> LoadBalancingPolicy for BoundToPredefinedNodePolicy<NODE> {
-    fn pick<'a>(&'a self, _info: &'a RoutingInfo, cluster: &'a ClusterData) -> Option<NodeRef<'a>> {
+    fn pick<'a>(
+        &'a self,
+        _info: &'a RoutingInfo,
+        cluster: &'a ClusterData,
+    ) -> Option<(NodeRef<'a>, Shard)> {
         self.report_node(Report::LoadBalancing);
-        cluster.get_nodes_info().iter().next()
+        cluster.get_nodes_info().iter().next().map(|node| (node, 0)) // FIXME: hardcoded shard 0
```
wprzytula, line 56: Let's not leave this `TODO` here, but rather discuss it straight and remove.
In the particular case of this test, hardcoded shard 0 seems to work good enough. Another option is to sample a random shard with a fixed seed, but I'm not convinced it's beneficial for the test in any way.

scylla/src/transport/connection_pool.rs
```diff
@@ -235,22 +235,25 @@ impl NodeConnectionPool {
...
+                let shard = shard
                     .try_into()
-                    .expect("Shard number doesn't fit in u16");
-                trace!(shard = shard, "Selecting connection for token");
-                Self::connection_for_shard(shard, sharder.nr_shards, connections.as_slice())
+                    // It's safer to use 0 rather that panic here, as shards are returned by `LoadBalancingPolicy`
+                    // now, which can be implemented by a user in an arbitrary way.
+                    .unwrap_or_else(|_| {
+                        error!("The provided shard number: {} does not fit u16! Using 0 as the shard number.", shard);
+                        0
```
wprzytula, line 254: The error message could also say: "Check your `LoadBalancingPolicy` implementation.", to hint users about the suspected culprit.

scylla/src/transport/session.rs
```diff
@@ -1507,28 +1492,23 @@ impl Session {
     }
 
     // This method allows to easily run a query using load balancing, retry policy etc.
-    // Requires some information about the query and two closures
-    // First closure is used to choose a connection
-    // - query will use node.random_connection()
-    // - execute will use node.connection_for_token()
-    // The second closure is used to do the query itself on a connection
+    // Requires some information about the query and a closure.
+    // The closure is used to do the query itself on a connection
```
wprzytula, line 1496: Nit: add a colon at the end.

scylla/tests/integration/execution_profiles.rs
Lorak-mmk: This is just a test, so I think using shard 0 is ok. I'll remove the comment.

wprzytula (approved):

wprzytula:
> `cargo semver-checks` found no API-breaking changes in this PR! 🎉🥳

Well, that's unfortunately not true, but it's [a known limitation of `semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks/issues/637#issuecomment-1896054494)...

The key takeaway is to not trust `semver-checks` fully, as there can be quite a lot of false negatives.

piodul:
> > `cargo semver-checks` found no API-breaking changes in this PR! 🎉🥳
> 
> Well, that's unfortunately not true, but it's [a known limitation of `semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks/issues/637#issuecomment-1896054494)...
> 
> The key takeaway is to not trust `semver-checks` fully, as there can be quite a lot of false negatives.

The message indeed sounds too confident, perhaps we should add some caution about the possible false negatives.

docs/source/load-balancing/load-balancing.md
```diff
@@ -15,11 +15,13 @@ using `SessionBuilder::host_filter` method.
 ## Plan
 
 When a query is prepared to be sent to the database, the load balancing policy
-constructs a load balancing plan. This plan is essentially a list of nodes to
+constructs a load balancing plan. This plan is essentially a list of pairs of node and optional shard to
```
wprzytula, line 18: Nit:
1) please keep maximum line width consistent
2) I suggest: *a list of pairs <node, optional shard>*

docs/source/load-balancing/load-balancing.md
```diff
@@ -15,11 +15,13 @@ using `SessionBuilder::host_filter` method.
...
 which the driver will try to send the query. The first elements of the plan are
-the nodes which are the best to contact (e.g. they might be replicas for the
+the nodes+shards which are the best to contact (e.g. they might be replicas for the
 requested data or have the best latency).
 
+In this chapter, "element" will refer to a pair of node and optional shard.
```
wprzytula, line 23: *pair <node, optional shard>*

wprzytula: Maybe replace too generic *element* with *target* or *candidate* ?

docs/source/load-balancing/load-balancing.md
```diff
@@ -2,7 +2,7 @@
```
wprzytula, line 2: Please update the `LoadBalancingPolicy` trait's docstring as well.

docs/source/load-balancing/load-balancing.md
```diff
@@ -84,15 +86,15 @@ first element of the load balancing plan is needed, so it's usually unnecessary
 to compute entire load balancing plan. To optimize this common case, the
 `LoadBalancingPolicy` trait provides two methods: `pick` and `fallback`.
 
-`pick` returns the first node to contact for a given query, which is usually
+`pick` returns the first element to contact for a given query, which is usually
 the best based on a particular load balancing policy. If `pick` returns `None`,
 then `fallback` will not be called.
```
wprzytula, line 91: This is no longer true, see 88f46058ca9db98df0e54809127a925da926ec5d. If you can, please fix my oversight. The correct behaviour of `pick/fallback` is described in the `LoadBalancingPolicy`'s docstring.

wprzytula (changes requested):

docs/source/load-balancing/load-balancing.md
Lorak-mmk: Target sounds very good, thanks

docs/source/load-balancing/load-balancing.md
Lorak-mmk: Done

docs/source/load-balancing/load-balancing.md
Lorak-mmk: Done

docs/source/load-balancing/load-balancing.md
Lorak-mmk: Done

scylla/src/transport/load_balancing/mod.rs
```diff
@@ -41,16 +41,17 @@ pub struct RoutingInfo<'a> {
...
 ///
 /// When a query is prepared to be sent to ScyllaDB/Cassandra, a `LoadBalancingPolicy`
-/// implementation constructs a load balancing plan. That plan is a list of nodes to which
-/// the driver will try to send the query. The first elements of the plan are the nodes which are
+/// implementation constructs a load balancing plan. That plan is a list of
+/// targets (target is a node + an optional shard) to which
+/// the driver will try to send the query. The first elements of the plan are the targets which are
 /// the best to contact (e.g. they might have the lowest latency).
 ///
 /// Most queries are send on the first try, so the query execution layer rarely needs to know more
```
wprzytula, line 52: /send/sent (let's fix this)

scylla/src/transport/load_balancing/mod.rs
```diff
@@ -41,16 +41,17 @@ pub struct RoutingInfo<'a> {
...
-/// than one node from plan. To better optimize that case, `LoadBalancingPolicy` has two methods:
-/// `pick` and `fallback`. `pick` returns a first node to contact for a given query, `fallback`
+/// than one target from plan. To better optimize that case, `LoadBalancingPolicy` has two methods:
+/// `pick` and `fallback`. `pick` returns a first target to contact for a given query, `fallback`
```
wprzytula, line 54: a first -> the first

docs/source/load-balancing/load-balancing.md
```diff
@@ -84,15 +86,14 @@ first element of the load balancing plan is needed, so it's usually unnecessary
...
-then `fallback` will not be called.
+`pick` returns the first target to contact for a given query, which is usually
+the best based on a particular load balancing policy.
 
-`fallback`, returns an iterator that provides the rest of the nodes in the load
-balancing plan. `fallback` is called only when using the initial picked node
-fails (or when executing speculatively).
+`fallback`, returns an iterator that provides the rest of the targets in the
+load balancing plan. `fallback` is called when using the initial picked
+element fails (or when executing speculatively) or when `pick` returned `None`.
```
wprzytula, line 94: element -> target

docs/source/load-balancing/load-balancing.md
```diff
@@ -84,15 +86,14 @@ first element of the load balancing plan is needed, so it's usually unnecessary
...
 
-It's possible for the `fallback` method to include the same node that was
+It's possible for the `fallback` method to include the same target that was
 returned by the `pick` method. In such cases, the query execution layer filters
 out the picked node from the iterator returned by `fallback`.
```
wprzytula, line 98: filters out the picked node or the picked target? Please check in `plan.rs` implementation.

scylla/src/transport/load_balancing/mod.rs
```diff
@@ -41,16 +41,17 @@ pub struct RoutingInfo<'a> {
...
 
-/// Policy that decides which nodes to contact for each query.
+/// Policy that decides which nodes and shards to contact for each query.
 ///
 /// When a query is prepared to be sent to ScyllaDB/Cassandra, a `LoadBalancingPolicy`
-/// implementation constructs a load balancing plan. That plan is a list of nodes to which
-/// the driver will try to send the query. The first elements of the plan are the nodes which are
+/// implementation constructs a load balancing plan. That plan is a list of
+/// targets (target is a node + an optional shard) to which
+/// the driver will try to send the query. The first elements of the plan are the targets which are
```
wprzytula, line 49: nit: line width

wprzytula (approved):
Mostly LGTM, some nits only.

docs/source/load-balancing/load-balancing.md
Lorak-mmk: It filters out targets. I'll change the doc

scylla/src/transport/load_balancing/mod.rs
Lorak-mmk: Not sure what you mean. This line has 100 characters, it doesn't seem excessive. There are also multiple other lines that exceed 80, so clearly it's not the limit we are using.

wprzytula: OK, nvm then.

wprzytula (approved):

PR #956, opened 2024-03-13
Decouple table and column specs

wprzytula:
## Motivation
Every column in a `Response::Result` comes from the same table (and it follows that from the same keyspace as well). It's thus redundant to store a copy of `TableSpec` (table and keyspace names) in each `ColumnSpec` (name and type of a column), which was done before.

## What's done
This PR moves `TableSpec` out of `ColumnSpec` and only allocates `TableSpec` once per each query response, effectively saving `2 * (C-1)` string allocations, where `C` denotes the number of columns returned in the response.
`TableSpec` is now stored in `ResultMetadata` and `PreparedMetadata.`

As table spec is no longer available in column specs, a public field in `QueryResult` is added for users to still
be able to retrieve this information from `QueryResult`. Keep in mind that this is a temporary measure, because `QueryResult` in the current form will be deprecated soon as part of the upcoming deserialization refactor (#462).

## Notes to reviewers
Please pay special attention to how user's experience changes after this API change. Don't they lose access to some information?

## Pre-review checklist

- ~~<[ ] I added appropriate `Fixes:` annotations to PR description.~~


scylla-cql/src/frame/response/result.rs
```diff
@@ -441,13 +442,19 @@ pub enum Result {
...
+    buf: &mut &[u8],
+    known_spec: Option<TableSpec>,
+) -> StdResult<TableSpec, ParseError> {
+    let ks_name = types::read_string(buf)?;
+    let table_name = types::read_string(buf)?;
+
+    Ok(known_spec.unwrap_or_else(|| TableSpec {
+        ks_name: ks_name.to_owned(),
+        table_name: table_name.to_owned(),
+    }))
```
Lorak-mmk, line 457: I'd like those equality checks to be a part of this PR so that we have more chances of catching any issues caused by this change.

As you know CQL protocol allows table spec to be either global or per column.
Do you know why that is? I assume there is some reason for the server to not always send the global one, and right now this PR assumes there is no reason and we can just use table spec of last column everywhere.

If there really is no reason for those per-column specs to exist, then that should be explained in the comment, and there should still be panicking checks in case this assumption turns out to be false.

Lorak-mmk (changes requested):

scylla-cql/src/frame/response/result.rs
wprzytula: Right, let's mind the Chesterton’s Fence: _don’t ever take down a fence until you know why it was put up._ 
I'll investigate those per-column table specs.

wprzytula: [Python driver makes the same assumption](https://github.com/scylladb/python-driver/blob/f3567166a2c151a1d8a44e1281b17bd612b2032d/cassandra/query.py#L557-L560), i.e. uses the table spec of the first column for all columns.

Lorak-mmk: This may be a good question for #engineering channel on Slack. I'd like to know when will Scylla / Cassandra send global spec and when per-column spec, and if it's posibble for columns specs to differ. If it's not possible, then learning a bit of history would be beneficial here imo: was it possible in the past? What was the case for it?

wprzytula: @nyh This is what I've talked to you about.

wprzytula:
v2:
- the assumption of the same table specs for every code is checked and  else `ParseError` is returned.

scylla-cql/src/frame/response/result.rs
Lorak-mmk: One more thing to consider: how likely is it that in some future version Cassandra / Scylla adds some form of joins / multi table queries?
AFAIK Cassandra already added ACID transactions (using their ACCORD algorithm), it doesn't seem so improbable for them to add something that queries more than 1 table in the future.
As those structs you modify are public, supporting this would require a breaking change. 

Do you think we could use Cow / Arc or something else to make this future-proof? That way we could have global table spec in `ResultMetadata`, but also have per-column spec if necessary.

wprzytula: Adding joins to CQL would be such a giant change in ScyllaDB - wide-column DBs aren't made for joins - that I strongly doubt it will ever happen.
`Arc`s and `Cow`s have considerable overhead that I don't deem worth incurring here just due to an extremely unprobable scenario.

wprzytula:
As realised in #1134, `TableSpec`s may differ between columns in the case of `PreparedMetadata`. Thus, we must **not** hold a single table spec per all columns, so this PR is obsolete. Closing.

Lorak-mmk:
It is obsolete for PreparedMetadata (because bind markers may be for different tables), but what about ResultMetadata?

PR #958, opened 2024-03-14
Remove unreachable pubs

Lorak-mmk:
Up until now some structs / fields etc were marked as pub despite not being a part of public API.
This is confusing when reading code: in order to know if something is truly public you need to check all the modules up to lib.rs to verify that each appears in some `pub use`.

This PR removes all such occurrences - now something is public if and only if it is marked as pub.
It also enables rustc lint `unreachable-pub` to prevent such occurences in the future.
## Pre-review checklist

### DIFF
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -44,3 +44,6 @@ num-bigint-03 = ["dep:num-bigint-03"]
 num-bigint-04 = ["dep:num-bigint-04"]
 bigdecimal-04 = ["dep:bigdecimal-04"]
 full-serialization = ["chrono", "time", "secret", "num-bigint-03", "num-bigint-04", "bigdecimal-04"]
+
+[lints.rust]
+unreachable_pub = "warn"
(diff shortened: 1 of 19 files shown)


scylla-proxy/src/actions.rs
```diff
@@ -409,67 +409,67 @@ impl RequestReaction {
 
 struct ExampleDbErrors;
 impl ExampleDbErrors {
-    pub fn syntax_error() -> DbError {
+    fn syntax_error() -> DbError {
         DbError::SyntaxError
     }
```
wprzytula, line 414: This `struct` should be `pub` instead; it's intended as a public API even though not yet used by any external code.

wprzytula (changes requested):

scylla-proxy/src/actions.rs
Lorak-mmk: Done

scylla-cql/Cargo.toml
```diff
@@ -44,3 +44,6 @@ num-bigint-03 = ["dep:num-bigint-03"]
 num-bigint-04 = ["dep:num-bigint-04"]
 bigdecimal-04 = ["dep:bigdecimal-04"]
 full-serialization = ["chrono", "time", "secret", "num-bigint-03", "num-bigint-04", "bigdecimal-04"]
+
+[lints.rust]
+unreachable_pub = "deny"
```
wprzytula, line 49: Let's `warn` instead of `deny`ing, just as other warnings work. They I converted to errors only in CI, which facilitates development.

wprzytula (approved):

PR #959, opened 2024-03-14
Show tracing output only for failed tests

wprzytula:
## Motivation

Flaky tests keep disrupting our CI. Those test failures are often difficult to reproduce locally. A way is needed to debug tests that failed in CI.
On the other hand, error traces bloat outputs.

## What's done
- by using `tracing_subscriber`'s `TestWriter,` tracing's prints are now captured and only shown if a test fails;
- for all tests to easily employ the captured tracing, a function `setup_tracing()` is introduced and added to most tests in `scylla` and `scylla-proxy` crates.

## Pre-review checklist


scylla/tests/integration/lwt_optimisation.rs
```diff
@@ -110,7 +115,7 @@ async fn if_lwt_optimisation_mark_offered_then_negotiatied_and_lwt_routed_optima
             println!("NOBODY was queried!");
         }
```
Lorak-mmk, line 116: changes in this file (apart from adding `setup_tracing()`) should be in a separate commit.

CONTRIBUTING.md
```diff
@@ -49,6 +49,19 @@ The above commands will leave a running ScyllaDB cluster in the background.
...
+This is a bit problematic for us, because driver uses tracing for logs, and without
+setting up a subscriber traces are not printed. That's why we have a helper function
+for tests: `setup_tracing`. It sets up tracing subscriber with env filter (so
+you can filter traces using env variables) and with writer that is compatible
+with test framework's output capture.
+
+Most of the tests already call this function, and any new tests should too.
+If you want to see tracing output from failing test and it doesn't call this function,
+please add the call at the beginning of the test.
+
```
Lorak-mmk, line 64: You could add me as a co-author of this commit.

Lorak-mmk (changes requested):
I see you use `tracing_subscriber::EnvFilter::from_default_env()`, but we don't set `RUST_LOG` var in CI. That means only `ERROR` traces will be printed. Please add `RUST_LOG` to all CI `cargo test` calls so that all traces are printed (as we discussed).

Lorak-mmk (approved):

Issue #957, opened 2024-03-14
`test_tracing` is probably flaky with Cassandra


wprzytula:
It failed twice in an unrelated pull request (https://github.com/scylladb/scylla-rust-driver/pull/955).

Logs:
https://github.com/scylladb/scylla-rust-driver/actions/runs/8267645061
https://github.com/scylladb/scylla-rust-driver/actions/runs/8269633516

Example:
```
---- transport::session_test::test_tracing stdout ----
Unique name: test_rust_1710345586_63
thread 'transport::session_test::test_tracing' panicked at scylla/src/transport/session_test.rs:1062:5:
No rows for tracing with this session id!
stack backtrace:
// initial frames omitted
  19:     0x563ae10baa96 - scylla::transport::session_test::assert_in_tracing_table::{{closure}}::hd4586b50cedb3ae9
                               at /home/runner/work/scylla-rust-driver/scylla-rust-driver/scylla/src/transport/session_test.rs:1062:5
  20:     0x563ae10b6de4 - scylla::transport::session_test::test_tracing_prepare::{{closure}}::hf62a9c69a433c71e
                               at /home/runner/work/scylla-rust-driver/scylla-rust-driver/scylla/src/transport/session_test.rs:924:54
  21:     0x563ae10d3dc2 - scylla::transport::session_test::test_tracing::{{closure}}::h1f54523201107fee
                               at /home/runner/work/scylla-rust-driver/scylla-rust-driver/scylla/src/transport/session_test.rs:854:48
  22:     0x563ae0e4c8f8 - <core::pin::Pin<P> as core::future::future::Future>::poll::hd7cfb3aeff23cadd
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/future/future.rs:124:9
  23:     0x563ae0e4cd8b - <core::pin::Pin<P> as core::future::future::Future>::poll::hf117c9ba3dcd69b6
                               at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/future/future.rs:124:9
  24:     0x563ae076f834 - tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}::{{closure}}::h0b3a7e8eafd583ce
                               at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.36.0/src/runtime/scheduler/current_thread/mod.rs:659:57
[...]

Issue #709, opened 2023-04-21
Default retry delay for getting tracing info may be too low


havaker:
https://github.com/scylladb/scylla-rust-driver/blob/0c13134225a0a6da0fc44908423ac4dc82965971/scylla/src/tracing.rs#L67

Currently, the default retry delay for retrieving tracking information is set to 3ms. However, issue #708 has shown that in certain cases, this delay is inadequate.

We should research and determine a new default retry delay that is more reasonable.

PR #966, opened 2024-03-19
session_test: Increase tracing info timeout

Lorak-mmk:
Default values for fetching tracing info are quite strict in the driver: it will make 5 attempts to fetch the data, waiting 3ms between attempts.

For Scylla this is enough, but for Cassandra it is not. For this reason our tests sometimes fail.

This was partially (for one test) fixed previously: https://github.com/scylladb/scylla-rust-driver/commit/61ad8781f1a01c9ec5ff087296229c2245c429b3
This PR fixes other test case that might be affected.
Previous fix used 50 attempts with 200ms wait, I changed it to 200 attempts with 50ms wait in order to not slow down Scylla tests unnecessarily.

Increased default retry attempts in Session from 5 to 10 to make tracing fetching more reliable for Scylla users without increasing latency.
I didn't touch default sleep between attempts - for Scylla new values should be ok, and Cassandra users can change them in config. I added a warning in docs so that they are aware of the problem.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/957
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/709

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -329,7 +329,7 @@ impl SessionConfig {
             #[cfg(feature = "cloud")]
             cloud_config: None,
             enable_write_coalescing: true,
-            tracing_info_fetch_attempts: NonZeroU32::new(5).unwrap(),
+            tracing_info_fetch_attempts: NonZeroU32::new(10).unwrap(),
             tracing_info_fetch_interval: Duration::from_millis(3),
             tracing_info_fetch_consistency: Consistency::One,
             cluster_metadata_refresh_interval: Duration::from_secs(60),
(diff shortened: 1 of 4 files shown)


Lorak-mmk:
While at it, we should decide what to do about https://github.com/scylladb/scylla-rust-driver/issues/709
I'd be in favor of closing it, because this default seems good for Scylla (I don't think tracing tests ever failed for Scylla in CI) so I don't want to increase latency for Scylla users because of Cassandra.

Lorak-mmk:
> While at it, we should decide what to do about #709 I'd be in favor of closing it, because this default seems good for Scylla (I don't think tracing tests ever failed for Scylla in CI) so I don't want to increase latency for Scylla users because of Cassandra.

Added comments about Cassandra in session builder, and increased retries amount from 5 to 10, I think we can close https://github.com/scylladb/scylla-rust-driver/issues/709 with this.

wprzytula (approved):

wprzytula:
> Added comments about Cassandra in session builder, and increased retries amount from 5 to 10, I think we can close #709 with this.

I agree, let's close #709.

wprzytula (approved):

Issue #974, opened 2024-04-03
Enhancement Proposal: Simplifying Token Calculation for High-Frequency Append-Log Style Operations


lvboudre:
Enhancement Proposal: Simplifying Token Calculation for High-Frequency Append-Log Style Operations

## Overview:

In our current project utilizing ScyllaDB, we're implementing a high-frequency, append-log style architecture to handle concurrent append requests. To optimize performance and minimize network traffic, we're batching these requests similar to how Kafka API operates, sending batches to ScyllaDB every 10 milliseconds.

To ensure efficient batching and minimize network overhead, it's crucial to group insert requests that will ultimately end up on the same node within ScyllaDB. This necessitates the computation of tokens for each insert statement, enabling us to determine their placement within the token ring.

## Current Challenge:

Presently, the existing API poses challenges in efficiently computing the token of a `PreparedStatement` without incurring significant performance overhead. The process involves invoking `Session::calculate_token`, which necessitates serializing a row (resulting in memory allocation), extracting the partition key, and then computing the token. Subsequently, when batching these statements using `Session::batch`, each row undergoes serialization again, effectively doubling memory allocation and serialization overhead.

## Immediate Solution

To streamline this process and enhance performance, we propose making Session::calculate_token_untyped public instead of keeping it pub(crate). By exposing this method publicly, we can pre-serialize every row, thereby reusing the serialization results to compute tokens and seamlessly integrate them into our batching process.

## Additionnal Note

In addition to the proposed enhancement of making `Session::calculate_token_untyped` public, we suggest making the `PartitionHasher` publicly accessible as well. This would empower users to compute results in advance without having to go through the serialization process of `SerializeRow` and `PreparedStatement`.

Considering that many ScyllaDB use cases involve key-value stores where the partition key is often known early on, exposing PartitionHasher would facilitate more efficient pre-computation of tokens, enhancing overall performance and developer experience.

Lorak-mmk:
I'm reluctant to expose such functions publicly - I'd rather think about the solution that avoids the allocation while preserving type safety.
BoundStatement could possibly be such a solution ( https://github.com/scylladb/scylla-rust-driver/issues/941 ) -  it would allow the values to be serialized only once and you could compute token of such statement without serializing again.

In order to decrease the impact for now you could use `ClusterData::compute_token` - that way you'll only need to allocate and serialize partition key twice, not all values for the query.

[...]

Issue #941, opened 2024-02-27
Introduce `BoundStatement`


Lorak-mmk:
After releasing serialization refactor several users reported use cases that are not well-supported by our new APIs.
The problems come from the need to have specific (non-generic) owned type for query values.
Previously `SerializedValues` was such a type. You could use it to serialize values ahead of time and then pass it around as you wish.
After refactor you would need to use types like `Vec<Box<dyn SerializeCql>>` or `Box<dyn SerializeRow>`, depending on your use case. There are some disadvantages
- Additional allocations.
- Need the values type to be owned - which is a problem if they come from e.g. zero-copy deserialization because you need to clone them.
- If you need specific values to also be dynamic or add them one-by-one, then you need to use something like `Vec<Box<dyn SerializeCql>>` and put that in a box - that's a lot more allocations and indirections.

So in short: migration to new API is possible in every case (I think), but sometimes the performance cost is high.

I propose to introduce new query type (like `Query` and `PreparedStatement`): `BoundStatement`. This is a prepared statement that also stores serialized values.
This would allow users to serialize parameters ahead of time while remaining type-safe and avoiding unnecessary allocations.

My first idea for an API is a a `bind(&self, values: impl SerializeRow) -> BoundStatement` method on `PreparedStatement`.
We can also think about allowing to bind values one by one. This would complicate the API and implementation and introduce some runtime errors (e.g. if user first binds value by order and then by name), but would be more flexible.

Main problem I see is how to adapt our Session API to this - creating another set of methods, like for `Query` and `PreparedStatement` seems bloaty.
There are more reasons to work on our Session API (potentially removing `Query` with values, bad naming etc), so I think we should discuss Session API changes in https://github.com/scylladb/scylla-rust-driver/issues/713
Right now probably the most reasonable solution that would allow us to support `BoundStatement` is `Executable` trait suggested in this issue.

Relevant issues / discussions:
https://github.com/scylladb/scylla-rust-driver/issues/890
https://github.com/scylladb/scylla-rust-driver/issues/713
https://github.com/scylladb/scylla-rust-driver/issues/108
https://scylladb-users.slack.com/archives/C01D7LCL44D/p1708948081839569

wprzytula:
I'm convinced that with the new strongly typed serialization framework we really need `BoundStatement` for both compatibility and performance purposes.
I believe we should work on it together with exploring possible implementation of `Executable` trait (#713), so that both those new things are compatible.

[...]

PR #975, opened 2024-04-03
Make `calculate_token_untyped` public

lvboudre:
Fixes: #974

## Pre-review checklist


wprzytula:
> I'm reluctant to expose such functions publicly - I'd rather think about the solution that avoids the allocation while preserving type safety.
BoundStatement could possibly be such a solution ( https://github.com/scylladb/scylla-rust-driver/issues/941 ) - it would allow the values to be serialized only once and you could compute token of such statement without serializing again.

As said here, we are not going to expose unsafe (in the sense of non-type-safety) API.
Instead, we're going to introduce `BoundStatement`, which will pre-serialize values and thus know its token before execution.

PR #980, opened 2024-04-16
docs: adding context to timeuuid using v1 feature

DanielHe4rt:
## Motivation

I decided to give more context for newcomers that want to try Timeuuid in the Rust Driver. Since I didn't had that much experience in the beginning and struggled for a while until understand how to use it.

Also I still didn't understand when and how I should generate the the node_id on `uuid::new_v1([])` parameters and IMHO it would be a great addition to the docs. 

### Fixes
- [x] Add context to `uuid::v1` usage
- [x] (Optional) Add context to `new_v1()` node_id numbers

## Pre-review checklist

### DIFF
--- a/docs/source/data-types/timeuuid.md
+++ b/docs/source/data-types/timeuuid.md
@@ -1,8 +1,8 @@
 # Timeuuid
 
-`Timeuuid` is represented as `value::CqlTimeuuid`.
-`value::CqlTimeuuid` is a wrapper for `uuid::Uuid` (using the `v1` feature) with custom ordering logic
-which follows Scylla/Cassandra semantics.
+The `Timeuuid` type is represented as `value::CqlTimeuuid`.
+
+Also, `value::CqlTimeuuid` is a wrapper for `uuid::Uuid`  with custom ordering logic which follows Scylla/Cassandra semantics.
 
 ```rust
 # extern crate scylla;
@@ -29,7 +29,9 @@ if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.ro
 # }
 ```
 
-To use the Timeuuid on `uuid` crate, enable the feature `v1` in your crate using:
+## Creating your own Timeuuid
+
+To create your own `Timeuuid` objects from timestamp-based `uuid` v1, you need to enable the feature `v1` of `uuid` crate using:
 
 ```shell
 cargo add uuid -F v1
@@ -48,9 +50,13 @@ use scylla::frame::value::CqlTimeuuid;
 use uuid::Uuid;
 # async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
 
-// Insert some timeuuid into the table
-let to_insert = CqlTimeuuid::from(Uuid::now_v1(&[1, 2, 3, 4, 5, 6]));
-session  
+// Tip: you can use random stable numbers or your MAC Address
+let node_id = [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC];
+
+// Build your Timeuuid with the current timestamp
+let to_insert = CqlTimeuuid::from(Uuid::now_v1(&node_id));
+
+session
     .query("INSERT INTO keyspace.table (a) VALUES(?)", (to_insert,))  
     .await?;
 
@@ -63,3 +69,5 @@ if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.ro
 # Ok(())
 # }
 ```
+
+Learn more about UUID::v1 [here](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_(date-time_and_MAC_address)).
--- a/docs/source/data-types/timeuuid.md
+++ b/docs/source/data-types/timeuuid.md
@@ -2,7 +2,7 @@
 
 The `Timeuuid` type is represented as `value::CqlTimeuuid`.
 
-Also, `value::CqlTimeuuid` is a wrapper for `uuid::Uuid`  with custom ordering logic which follows Scylla/Cassandra semantics.
+Also, `value::CqlTimeuuid` is a wrapper for `uuid::Uuid` with custom ordering logic which follows Scylla/Cassandra semantics.
 
 ```rust
 # extern crate scylla;
@@ -15,15 +15,18 @@ use scylla::frame::value::CqlTimeuuid;
 
 // Insert some timeuuid into the table
 let to_insert: CqlTimeuuid = CqlTimeuuid::from_str("8e14e760-7fa8-11eb-bc66-000000000001")?;
+
 session
     .query("INSERT INTO keyspace.table (a) VALUES(?)", (to_insert,))
     .await?;
 
-// Read timeuuid from the table
-if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.rows {
-    for row in rows.into_typed::<(CqlTimeuuid,)>() {
-        let (timeuuid_value,): (CqlTimeuuid,) = row?;
-    }
+// Read Timeuuid from the table
+let result = session.query("SELECT a FROM keyspace.table", &[]).await?;
+
+let mut iter = result.rows_typed::<(CqlTimeuuid, )>()?;
+
+while let Some((timeuuid,)) = iter.next().transpose()? {
+    println!("Read a value from row: {}", timeuuid);
 }
 # Ok(())
 # }
@@ -60,11 +63,13 @@ session
     .query("INSERT INTO keyspace.table (a) VALUES(?)", (to_insert,))  
     .await?;
 
-// Read timeuuid from the table
-if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.rows {
-    for row in rows.into_typed::<(CqlTimeuuid,)>() {
-        let (timeuuid_value,): (CqlTimeuuid,) = row?;
-    }
+// Read Timeuuid from the table
+let result = session.query("SELECT a FROM keyspace.table", &[]).await?;
+
+let mut iter = result.rows_typed::<(CqlTimeuuid, )>()?;
+
+while let Some((timeuuid,)) = iter.next().transpose()? {
+    println!("Read a value from row: {}", timeuuid);
 }
 # Ok(())
 # }
--- a/docs/source/data-types/timeuuid.md
+++ b/docs/source/data-types/timeuuid.md
@@ -1,7 +1,7 @@
 # Timeuuid
 
 `Timeuuid` is represented as `value::CqlTimeuuid`.
-`value::CqlTimeuuid` is a wrapper  for `uuid::Uuid` with custom ordering logic
+`value::CqlTimeuuid` is a wrapper for `uuid::Uuid` (using the `v1` feature) with custom ordering logic
 which follows Scylla/Cassandra semantics.
 
 ```rust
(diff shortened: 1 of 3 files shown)


wprzytula:
Unfortunately, the error message from CI tells me nothing. As it's Sphinx task-related, it's @dgarcia360 who should have a clue about that.

wprzytula:
> Also I still didn't understand when and how I should generate the the node_id on `uuid::new_v1([])` parameters and IMHO it would be a great addition to the docs.

My impression (I'm not an expert in CQL, though) is that it is only ScyllaDB itself that should generate Timeuuid values, and the driver side should have no interest in creating their own ones; the driver should probably only use Timeuuid as opaque identifiers.
@nyh Am I right?

docs/source/data-types/timeuuid.md
```diff
@@ -19,6 +19,41 @@ session
...
+To use the Timeuuid on `uuid` crate, enable the feature `v1` in your crate using:
+
+```shell
+cargo add uuid -F v1
+```
+
+and now you're gonna be able to use the `uuid::v1` features: 
+
+```rust
+# extern crate uuid;
```
DanielHe4rt, line 41: I just pushed the example without the ignore. The problem is that Timeuuid uses the `v1` feature inside `uuid` package. But to create a example using that is not possible since the "extern crate" flag doesn't allow to add a feature, making the example fails. 

My first try was using the `ignore` flag to not pass in this validation, but didn't worked. 

The goal is to have something like this:
```suggestion
# extern crate uuid::v1; // Just a example
```

wprzytula: > examples/Cargo.toml is the one used to configure deps for the docs. You can call cargo add uuid --dev -F v1 and the feature will be accessible from the docs.

> Ah, right. Then adding v1 to uuid's feature list in examples/Cargo.toml and removing this ignore should fix the problem.

docs/source/data-types/timeuuid.md
```diff
@@ -19,6 +19,41 @@ session
...
+# use scylla::Session;
+# use std::error::Error;
+# use std::str::FromStr;
+use scylla::IntoTypedRows;
+use scylla::frame::value::CqlTimeuuid;
+use uuid::Uuid;
+# async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
+
+// Insert some timeuuid into the table
+let to_insert = CqlTimeuuid::from(Uuid::now_v1(&[1, 2, 3, 4, 5, 6]));
```
wprzytula, line 52: This can't be left here as-is; `node_id` argument for `now_v1` must be some reasonable number. We need to understand what is commonly chosen to served as `node_id` in existing applications.

wprzytula (changes requested):

docs/source/data-types/timeuuid.md
```diff
@@ -19,6 +19,41 @@ session
...
+if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.rows {
+    for row in rows.into_typed::<(CqlTimeuuid,)>() {
+        let (timeuuid_value,): (CqlTimeuuid,) = row?;
+    }
+}
+# Ok(())
+# }
+```
+
+To use the Timeuuid on `uuid` crate, enable the feature `v1` in your crate using:
```
Lorak-mmk, line 32: You don't need to to this unless you want to create your own uuid v1 objects based on timestamp. Fetching objects from DB works perfectly fine without it.
I think this should be more like "To create your own `Timeuuid` objects from timstamp-base `uuid` v1, you need to enable the feature `v1` of `uuid` crate using:"

Lorak-mmk (changes requested):

docs/source/data-types/timeuuid.md
DanielHe4rt:
```rust
// Tip: you can use random stable numbers or your MAC Address
let node_id = [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC];
```

Just added more context to this line just in case. Maybe we can add a new method that generates the Uuid v1 with the MAC Address or some other stable number around Hardware ID.

Maybe I'm a lazy developer, however I would love to only call a `uuid::now_v1()` without any parameter needed.

wprzytula: > Maybe I'm a lazy developer, however I would love to only call a uuid::now_v1() without any parameter needed.

It's unfortunately hard to have strong guarantees without significant effort from the developer side...

wprzytula: > ```rust
> // Tip: you can use random stable numbers or your MAC Address
> let node_id = [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC];
> ```
> 
> Just added more context to this line just in case. Maybe we can add a new method that generates the Uuid v1 with the MAC Address or some other stable number around Hardware ID.

@Lorak-mmk WDYT?

docs/source/data-types/timeuuid.md
```diff
@@ -1,7 +1,7 @@
 # Timeuuid
 
 `Timeuuid` is represented as `value::CqlTimeuuid`.
-`value::CqlTimeuuid` is a wrapper  for `uuid::Uuid` with custom ordering logic
+`value::CqlTimeuuid` is a wrapper for `uuid::Uuid` (using the `v1` feature) with custom ordering logic
```
wprzytula, line 4: The _using the `v1` feature_ part should only be said be needed when _constructing_ `CqlTimeuuid`; using them does not require the feature. Please rewrite this accordingly.

docs/source/data-types/timeuuid.md
```diff
@@ -19,6 +19,45 @@ session
...
+# use scylla::Session;
+# use std::error::Error;
+# use std::str::FromStr;
+use scylla::IntoTypedRows;
+use scylla::frame::value::CqlTimeuuid;
+use uuid::Uuid;
+# async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
+
+// Tip: you can use random stable numbers or your MAC Address
+let node_id = [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC];
```
wprzytula, line 52: Perhaps leaving here a link to Wikipedia on UUID v1 is a good idea:
https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_(date-time_and_MAC_address)

wprzytula (changes requested):

docs/source/data-types/timeuuid.md
DanielHe4rt: After some chatting with @codelieutenant he gave the idea to also point as an environment variable. Since it's a thing that it's not supposed to change maybe it's suppose to be a good fit.

What do you think about this approach?

wprzytula: Seems legit, though again - I have no idea what people normally do to acquire the node id in their applications.

DanielHe4rt:
@annastuchlik Can you drop a review on it? Just in case I missed something :p

docs/source/data-types/timeuuid.md
Lorak-mmk: I 

> > ```rust
> > // Tip: you can use random stable numbers or your MAC Address
> > let node_id = [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC];
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Just added more context to this line just in case. Maybe we can add a new method that generates the Uuid v1 with the MAC Address or some other stable number around Hardware ID.
> 
> @Lorak-mmk WDYT?

I think this is an example that presents how to create one's own CqlTimeuuid value from uuid, not uuid library best practices. Randomly generated node id is fine, hardcoded (like in the current version) is ok too.

docs/source/data-types/timeuuid.md
```diff
@@ -27,4 +27,47 @@ if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.ro
 }
 # Ok(())
 # }
-```
\ No newline at end of file
+```
+
+## Creating your own Timeuuid
+
+To create your own `Timeuuid` objects from timstamp-base `uuid` v1, you need to enable the feature `v1` of `uuid` crate using:
```
Lorak-mmk, line 34: `base` -> `based`

Lorak-mmk (changes requested):
One minor typo, apart from this looks good to me

docs/source/data-types/timeuuid.md
DanielHe4rt: there was also the "timstamp" lol

fixed.

DanielHe4rt:
Ready to merge!

docs/source/data-types/timeuuid.md
```diff
@@ -1,8 +1,8 @@
 # Timeuuid
 
-`Timeuuid` is represented as `value::CqlTimeuuid`.
-`value::CqlTimeuuid` is a wrapper  for `uuid::Uuid` with custom ordering logic
-which follows Scylla/Cassandra semantics.
+The `Timeuuid` type is represented as `value::CqlTimeuuid`.
+
+Also, `value::CqlTimeuuid` is a wrapper for `uuid::Uuid`  with custom ordering logic which follows Scylla/Cassandra semantics.
```
wprzytula, line 5: Redundant double space

docs/source/data-types/timeuuid.md
```diff
@@ -27,4 +27,47 @@ if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.ro
...
+let node_id = [0x12, 0x34, 0x56, 0x78, 0x9A, 0xBC];
+
+// Build your Timeuuid with the current timestamp
+let to_insert = CqlTimeuuid::from(Uuid::now_v1(&node_id));
+
+session
+    .query("INSERT INTO keyspace.table (a) VALUES(?)", (to_insert,))  
+    .await?;
+
+// Read timeuuid from the table
```
wprzytula, line 63: `Timeuuid` is once capitalised, another time not. Please choose one convention.

docs/source/data-types/timeuuid.md
```diff
@@ -27,4 +27,47 @@ if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.ro
...
+if let Some(rows) = session.query("SELECT a FROM keyspace.table", &[]).await?.rows {
+    for row in rows.into_typed::<(CqlTimeuuid,)>() {
+        let (timeuuid_value,): (CqlTimeuuid,) = row?;
+    }
+}
```
wprzytula, line 68: Don't use raw `rows` field on `QueryResult`; prefer `rows_typed::<(CqlTimeuuid,)>()` method:

```suggestion
let mut iter = session.query("SELECT a FROM keyspace.table", &[]).await?.rows_typed::<(CqlTimeuuid,)>();
while let Some((timeuuid_value,)) = iter.next().transpose()? {
  // do something with timeuuid_value
}
```

wprzytula (changes requested):

docs/source/data-types/timeuuid.md
DanielHe4rt: I'll update, but can you explain in details why I shouldn't use it?

DanielHe4rt: Fixed! Ready to merge.

wprzytula: `rows` make you interact with an untyped, materialised vec of columns. At the present state of the driver, the main drawback is reduced type safety; as you call `into_typed` anyway, this negative aspect is mitigated.
Another reason for avoiding using `rows` ATM, which is relevant here, is the ongoing deserialization refactor. We ditch the materialized vec of `CqlValue` at all to reduce unnecessary allocations. The new deserialization framework is lazy, which means that the values are going to be deserialized on-the-fly straight to the end type, based on the type provided in the iterator (`rows_typed::<T>`).
With the new framework having landed in the driver, `rows` will be removed at all (although their capabilities can be emulated for the users that *really* need them).

See https://github.com/scylladb/scylla-rust-driver/pull/955

Issue #852, opened 2023-10-20
FromRow is not working for type wrapper 


DylanVerstraete:
I have this custom type wrapper for uuid:

```
#[derive(Debug, Clone, FromRow)]
pub struct FileId(Uuid);
```
And I am getting error: `derive(FromRow) works only for structs with named fields`

Is there a reason that this does not work or is there a workaround?

piodul:
> Is there a reason that this does not work or is there a workaround?

I'm not entirely sure why we chose to support named structs instead of tuple structs for FromRow, especially that this trait doesn't do anything with the Rust struct field names. I suppose it's a good idea to add support for it to the driver.

As a workaround, you can implement the `FromRow` trait yourself - if your tuple struct is short enough (up to 16 fields) then it's just the matter of delegating the implementation to a regular tuple, for example:

```rust
use scylla::cql_to_rust::FromRowError;
use scylla::FromRow;
use scylla_cql::frame::response::result::Row;

struct MyTupleStruct(i32, String, i64);
impl FromRow for MyTupleStruct {
    fn from_row(row: Row) -> Result<Self, FromRowError> {
        let (a, b, c) = <_ as FromRow>::from_row(row)?;
        Ok(Self(a, b, c))
    }
}
```

DylanVerstraete:
Awesome! Thanks

PR #985, opened 2024-04-19
derive `FromRow` for structs with unnamed fields

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/852

## Motivation

For some reason, users were not able to derive `FromRow` trait for structs with unnamed fields.
This PR fixes that.

## Changes
Adjusted `from_row_derive` function from `scylla-macros` so that it supports structs with unnamed fields as well.

## Pre-review checklist

### DIFF
--- a/scylla-macros/src/from_row.rs
+++ b/scylla-macros/src/from_row.rs
@@ -6,33 +6,43 @@ use syn::{spanned::Spanned, DeriveInput};
 pub(crate) fn from_row_derive(tokens_input: TokenStream) -> Result<TokenStream, syn::Error> {
     let item = syn::parse::<DeriveInput>(tokens_input)?;
     let path = crate::parser::get_path(&item)?;
-    let struct_fields = crate::parser::parse_named_fields(&item, "FromRow")?;
+    let struct_fields = crate::parser::parse_struct_fields(&item, "FromRow")?;
 
     let struct_name = &item.ident;
     let (impl_generics, ty_generics, where_clause) = item.generics.split_for_impl();
 
-    // Generates tokens for field_name: field_type::from_cql(vals_iter.next().ok_or(...)?), ...
-    let set_fields_code = struct_fields.named.iter().map(|field| {
-        let field_name = &field.ident;
-        let field_type = &field.ty;
+    // Generates a token that sets the values of struct fields: field_type::from_cql(...)
+    let (fill_struct_code, fields_count) = match struct_fields {
+        crate::parser::StructFields::Named(fields) => {
+            let set_fields_code = fields.named.iter().map(|field| {
+                let field_name = &field.ident;
+                let field_type = &field.ty;
 
-        quote_spanned! {field.span() =>
-            #field_name: {
-                let (col_ix, col_value) = vals_iter
-                    .next()
-                    .unwrap(); // vals_iter size is checked before this code is reached, so
-                               // it is safe to unwrap
+                quote_spanned! {field.span() =>
+                    #field_name: {
+                        let (col_ix, col_value) = vals_iter
+                            .next()
+                            .unwrap(); // vals_iter size is checked before this code is reached, so
+                                    // it is safe to unwrap
 
-                <#field_type as FromCqlVal<::std::option::Option<CqlValue>>>::from_cql(col_value)
-                    .map_err(|e| FromRowError::BadCqlVal {
-                        err: e,
-                        column: col_ix,
-                    })?
-            },
+                        <#field_type as FromCqlVal<::std::option::Option<CqlValue>>>::from_cql(col_value)
+                            .map_err(|e| FromRowError::BadCqlVal {
+                                err: e,
+                                column: col_ix,
+                            })?
+                    },
+                }
+            });
+
+            // This generates: { field1: {field1_code}, field2: {field2_code}, ...}
+            let fill_struct_code = quote! {
+                {#(#set_fields_code)*}
+            };
+            (fill_struct_code, fields.named.len())
         }
-    });
+        crate::parser::StructFields::Unnamed(_fields) => todo!(),
+    };
 
-    let fields_count = struct_fields.named.len();
     let generated = quote! {
         impl #impl_generics #path::FromRow for #struct_name #ty_generics #where_clause {
             fn from_row(row: #path::Row)
@@ -49,9 +59,7 @@ pub(crate) fn from_row_derive(tokens_input: TokenStream) -> Result<TokenStream, 
                 }
                 let mut vals_iter = row.columns.into_iter().enumerate();
 
-                Ok(#struct_name {
-                    #(#set_fields_code)*
-                })
+                Ok(#struct_name #fill_struct_code)
             }
         }
     };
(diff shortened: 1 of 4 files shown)


wprzytula:
> I thought that deserialization refactor will make this macro obsolete and later (before 1.0) we're going to remove it. What is the point of spending time on improvements to this macro? @wprzytula

That's a valid point. Once we have this PR ready, though, it's probably worthy to include it in the next release. 1.0 isn't going to be released very soon, so let's give users this enhancement as early as we can.

scylla-macros/src/parser.rs
```diff
@@ -35,6 +35,46 @@ pub(crate) fn parse_named_fields<'a>(
...
+    Named(&'a FieldsNamed),
+    Unnamed(&'a FieldsUnnamed),
+}
+
+/// Parses the tokens_input to a DeriveInput and returns the struct name from which it derives and
+/// the fields
+pub(crate) fn parse_struct_fields<'a>(
+    input: &'a DeriveInput,
+    current_derive: &str,
+) -> Result<StructFields<'a>, syn::Error> {
```
wprzytula, line 48: Shouldn't it be:
_Parses the tokens_input **from** a DeriveInput and **~~returns~~** the struct name from which it derives and **returns** the fields_
?

scylla-macros/src/parser.rs
```diff
@@ -35,6 +35,46 @@ pub(crate) fn parse_named_fields<'a>(
...
+pub(crate) enum StructFields<'a> {
+    Named(&'a FieldsNamed),
+    Unnamed(&'a FieldsUnnamed),
+}
+
+/// Parses the tokens_input to a DeriveInput and returns the struct name from which it derives and
+/// the fields
+pub(crate) fn parse_struct_fields<'a>(
+    input: &'a DeriveInput,
+    current_derive: &str,
```
wprzytula, line 47: This is going to contain the trait name, not the struct name? Then I don't understand the docstring about the struct name.

scylla-macros/src/parser.rs
```diff
@@ -35,6 +35,46 @@ pub(crate) fn parse_named_fields<'a>(
...
+    match &input.data {
+        Data::Struct(data) => match &data.fields {
+            Fields::Named(named_fields) => Ok(StructFields::Named(named_fields)),
+            Fields::Unnamed(unnamed_fields) => Ok(StructFields::Unnamed(unnamed_fields)),
+            Fields::Unit => Err(syn::Error::new_spanned(
+                data.struct_token,
+                format!(
+                    "derive({}) works only for structs with fields",
+                    current_derive
+                ),
```
wprzytula, line 58: Maybe extract this to a function?

scylla-macros/src/from_row.rs
```diff
@@ -6,33 +6,68 @@ use syn::{spanned::Spanned, DeriveInput};
...
+                    },
+                }
+            });
+
+            // This generates: { field1: {field1_code}, field2: {field2_code}, ...}
+            let fill_struct_code = quote! {
+                {#(#set_fields_code)*}
+            };
+            (fill_struct_code, fields.named.len())
+        }
```
wprzytula, line 42: One suggestion for splitting into commits: when changing some single case into an enum that covers multiple cases, it is a good practise to first introduce the original case as a single variant of the enum, proving that nothing changes in the original case, and add more cases (variants of the enum) in the next commits.

scylla/src/macros.rs
```diff
@@ -1,6 +1,7 @@
-/// #[derive(FromRow)] derives FromRow for struct
+/// Derive macro for the [`FromRow`](crate::frame::response::cql_to_rust::FromRow) trait
+/// which deserializes a row to given Rust structure.
 ///
-/// Works only on simple structs without generics etc
+/// It is supported for structs with either named or unnamed fields.
```
wprzytula, line 4: Please keep mention of the limitation to structs without generics.

scylla-macros/src/parser.rs
muzarski: What exactly do you suggest to extract?

scylla-macros/src/parser.rs
muzarski: Yeah, now that I read the comment it does not make much sense. I copied it from the very similar function which extracts only NAMED fields:
```
/// Parses the tokens_input to a DeriveInput and returns the struct name from which it derives and
/// the named fields
```

scylla-macros/src/parser.rs
wprzytula: The whole `format!` expression, parametrised by `current_derive`.

scylla/src/macros.rs
Lorak-mmk: > I removed this mention for a reason. IIRC, the following compiles:
> 
> ```rust
> #[derive(FromRow)]
> struct GenericRow<T: FromCqlVal<CqlValue>> {
>     col: T,
> }
> ```
> 
> Do you see any other use cases for generics in structs that would derive `FromRow`?

What about if you remove `FromCqlVal<CqlValue>` bound?
Typical derives (`Clone` etc) work in this case:
```rust
#[derive(Clone)]
struct Generic<T> {
   val: T,
}
```

Then `Clone` will be implemented only for `T`'s that implement `Clone`. Does `FromRow` work the same way? If not, then it doesn't really support generics.

muzarski: In such case, we obviously get the error that `T` does not implement `FromCqlVal<CqlValue>` since it's required to deserialize the value.

> Typical derives (Clone etc) work in this case

Now that I think about it, you are right. I'll keep the mention then.

muzarski: One way we could solve it is to generate the bounds for generics ourselves, but it's not related to this PR + I don't think it's so important to support generics anyway.

scylla-macros/src/from_row.rs
muzarski: I've split it into two commits

scylla-macros/src/parser.rs
muzarski: Extracted it to the closure. Did the same for `parse_named_fields` utility function.

scylla-macros/src/parser.rs
muzarski: I adjusted the docstring comment.

scylla-macros/src/from_row.rs
```diff
@@ -6,33 +6,43 @@ use syn::{spanned::Spanned, DeriveInput};
...
+            });
+
+            // This generates: { field1: {field1_code}, field2: {field2_code}, ...}
+            let fill_struct_code = quote! {
+                {#(#set_fields_code)*}
+            };
+            (fill_struct_code, fields.named.len())
         }
-    });
+        crate::parser::StructFields::Unnamed(_fields) => todo!(),
```
wprzytula: The approach here is better for readability that the previous one, but what I had in mind involved adding the `Unnamed` variant to the enum only in the next commit, together with the handling logic, so that `todo!()` and nonworking-yet-compiled code can be avoided at all.

Lorak-mmk (approved):

wprzytula (approved):

Issue #713, opened 2023-04-25
Replace `Query` with a more fitting name


piodul:
Confusingly, the driver calls unprepared statement a `Query` (contrary to `PreparedStatement`). It's not a correct term and can lead to confusion. "Queries" are statements that _ask_ for data and receive some data in return. "Statement" is a broader term and encompasses items that don't return data but rather modify data or schema, e.g. `INSERT` or `CREATE`. The `Query` object has no problems with statements that are not queries.

I suggest to rename `Query` to something that fits its purpose better, e.g. `UnpreparedStatement` or `SimpleStatement` - need to decide upon something, it's probably the best to look at the naming in other drivers. To ease transition to the new name, we will re-introduce `Query` as a type alias, then deprecate and eventually remove at some point.

The documentation will have to be adjusted as well as it seems to use "query" and "statement" interchangeably, which is wrong due to the reasons described above but was probably caused by confusion resulting from the existing `Query` name.

nyh:
By the way, the same logic also applies to the names of the `Session` methods - `query()` and `execute()`. When I started using the Rust driver (yesterday ;-)) it wasn't immediately obvious why the unprepared version of `execute()` is called "query" and not `execute_unprepared`, or something like it. It's not a huge problem (I just read the whole `Session` documentation and found everything) but it was strange.

Sadly, Rust doesn't seem to have function overloading, so it's not possible to allow the same name `execute()` for both prepared and unprepared versions, as done in Python for example, but at least the names could be closer, and perhaps also physically closer in the sorted documentation.

piodul:
> By the way, the same logic also applies to the names of the `Session` methods - `query()` and `execute()`. When I started using the Rust driver (yesterday ;-)) it wasn't immediately obvious why the unprepared version of `execute()` is called "query" and not `execute_unprepared`, or something like it. It's not a huge problem (I just read the whole `Session` documentation and found everything) but it was strange.
>
> Sadly, Rust doesn't seem to have function overloading, so it's not possible to allow the same name `execute()` for both prepared and unprepared versions, as done in Python for example, but at least the names could be closer, and perhaps also physically closer in the sorted documentation.

Function overloading can be simulated to some extent via traits. For example, `PreparedStatement` and `UnpreparedStatement` could implement a trait (let's say it's called `Executable`), and then you could have:

```rust
pub async fn execute(&self, stmt: &impl Executable) {
    // ...
}
```

Maybe this could be extended to batches as well.

[...]

PR #986, opened 2024-04-19
prepared: docs for `PreparedStatement`

muzarski:
In response to https://github.com/scylladb/scylla-rust-driver/pull/970#discussion_r1570902278.

Added docstring for `PreparedStatement`, which mentions:
- how to prepare a statement
- benefits of using prepared statements
- statement repreparation that is handled by the driver
- prepared statement cloning
- correlation between prepared statements and schema altering 

## Pre-review checklist

### DIFF
--- a/docs/source/queries/prepared.md
+++ b/docs/source/queries/prepared.md
@@ -3,6 +3,12 @@
 Prepared queries provide much better performance than simple queries,
 but they need to be prepared before use.
 
+Benefits that prepared statements have to offer:
+- Type safety - thanks to metadata provided by the server, the driver can verify bound values' types before serialization. This way, we can be always sure that the Rust type provided by the user is compatible (and if not, the error is returned) with the destined native type. The same applies for deserialization.
+- Performance - when executing a simple query with non-empty values list, the driver
+prepares the statement before execution. The reason for this is to provide type safety for simple queries. However, this implies 2 round trips per simple query execution. On the other hand, the cost of prepared statement's execution is only 1 round trip.
+- Improved load-balancing - using the statement metadata, the driver can compute a set of destined replicas for current statement execution. These replicas will be preferred when choosing the node (and shard) to send the request to. For more insight on this, see [performance section](#performance).
+
 ```rust
 # extern crate scylla;
 # use scylla::Session;
--- a/scylla/src/statement/prepared_statement.rs
+++ b/scylla/src/statement/prepared_statement.rs
@@ -22,6 +22,68 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
 use crate::transport::partitioner::{Partitioner, PartitionerHasher, PartitionerName};
 
 /// Represents a statement prepared on the server.
+///
+/// To prepare a statement, simply execute [`Session::prepare`](crate::transport::session::Session::prepare).
+///
+/// If you plan on reusing the statement, or bounding some values to it during execution, always
+/// prefer using prepared statements over [`Session::query`](crate::transport::session::Session::query).
+///
+/// Benefits that prepared statements have to offer:
+/// * Performance - a prepared statement holds information about metadata
+/// that allows to carry out a statement execution in a type safe manner.
+/// When [`Session::query`](crate::transport::session::Session::query) is called with
+/// non-empty bound values, the driver has to prepare the statement before execution (to provide type safety).
+/// This implies 2 round trips per [`Session::query`](crate::transport::session::Session::query).
+/// On the other hand, the cost of [`Session::execute`](crate::transport::session::Session::execute) is only 1 round trip.
+/// * Increased type-safety - bound values' types are validated with
+/// the [`PreparedMetadata`] received from the server during the serialization.
+/// * Improved load balancing - thanks to statement metadata, the driver is able
+/// to compute a set of destined replicas for the statement execution. These replicas
+/// will be preferred when choosing the node (and shard) to send the request to.
+/// * Result deserialization optimization - see [`PreparedStatement::set_use_cached_result_metadata`].
+///
+/// # Clone implementation
+/// Cloning a prepared statement is a cheap operation. It only
+/// requires copying a couple of small fields and some [Arc] pointers.
+/// Always prefer cloning over executing [`Session::prepare`](crate::transport::session::Session::prepare)
+/// multiple times to save some roundtrips.
+///
+/// # Statement repreparation
+/// When schema is updated, the server is supposed to invalidate its
+/// prepared statement caches. Then, if client tries to execute a given statement,
+/// the server will respond with an error. Users should not worry about it, since
+/// the driver handles it properly and tries to reprepare the statement.
+/// However, there are some cases when client-side prepared statement should be dropped
+/// and prepared once again via [`Session::prepare`](crate::transport::session::Session::prepare) -
+/// see the mention about altering schema below.
+///
+/// # Altering schema
+/// If for some reason you decided to alter the part of schema that corresponds to given prepared
+/// statement, then the corresponding statement (and its copies obtained via [`PreparedStatement::clone`]) should
+/// be dropped. The statement should be prepared again.
+///
+/// There are two reasons for this:
+///
+/// ### CQL v4 protocol limitations
+/// The driver only supports CQL version 4.
+///
+/// In multi-client scenario, only the first client which reprepares the statement
+/// will receive the updated metadata from the server.
+/// The rest of the clients will still hold on the outdated metadata.
+/// In version 4 of CQL protocol there is currently no way for the server to notify other
+/// clients about prepared statement's metadata update.
+///
+/// ### Client-side metadata immutability
+/// The decision was made to keep client-side metadata immutable.
+/// Mainly because of the CQLv4 limitations mentioned above. This means
+/// that metadata is not updated during statement repreparation.
+/// This raises two issues:
+/// * bound values serialization errors - since [`PreparedMetadata`] is not updated
+/// * result deserialization errors - when [`PreparedStatement::set_use_cached_result_metadata`] is enabled,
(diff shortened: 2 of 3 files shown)


scylla/src/statement/prepared_statement.rs
```diff
@@ -283,6 +283,17 @@ impl PreparedStatement {
     /// after statement preparation and will use it
     /// to deserialize the results of statement execution.
     ///
+    /// NOTICE: Users should NOT alter the schema after enablind
```
wprzytula, line 286: enabling

wprzytula (changes requested):

scylla/src/statement/prepared_statement.rs
```diff
@@ -283,6 +283,17 @@ impl PreparedStatement {
...
+    /// this optimisation on the statement.
+    /// To be more precise:
+    /// they should not alter the tables and UDTs that are used in the given statement.
+    /// The reason for this is that the decision was made to keep client-side metadata immutable,
+    /// and so, it is not updated during statement repreparation.
+    /// Altering the schema, and then, executing the statement with optimisation enabled may lead
+    /// to runtime errors. In the worst case, it could lead to
```
wprzytula, line 293: and then ~~,~~ executing (remove the comma)

scylla/src/statement/prepared_statement.rs
```diff
@@ -283,6 +283,17 @@ impl PreparedStatement {
...
+    /// some data being deserialized to wrong Rust type contained
+    /// in outdated metadata.
```
wprzytula, line 295: contained -> specified

wprzytula (changes requested):

docs/source/queries/unprepared.md
```diff
@@ -92,15 +92,15 @@ while let Some(read_row) = iter.next().transpose()? {
 # Ok(())
 # }
 ```
-> In cases where page size is set, simple query returns only a single page of results.\
+> In cases where page size is set, unpreapred query returns only a single page of results.\
```
wprzytula, line 95: unprepared

docs/source/index.md
```diff
@@ -15,7 +15,7 @@ Although optimized for Scylla, the driver is also compatible with [Apache Cassan
 * [Quick start](quickstart/quickstart.md) - Setting up a Rust project using `scylla-rust-driver` and running a few queries
 * [Migration guides](migration-guides/migration-guides.md) - How to update the code that used an older version of this driver
 * [Connecting to the cluster](connecting/connecting.md) - Configuring a connection to scylla cluster
-* [Making queries](queries/queries.md) - Making different types of queries (simple, prepared, batch, paged)
+* [Making queries](queries/queries.md) - Making different types of queries (unprepared, prepared, batch, paged)
```
wprzytula, line 18: Let's ditch the notion of _query_ from the driver and use _statement_ exclusively. Here, I would replace _making queries_ with _creating and executing statements_.

scylla/src/statement/prepared_statement.rs
```diff
@@ -22,6 +22,60 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
...
+/// ### CQLv4 protocol limitations
+/// In multi-client scenario, only the first client which reprepares the statement
+/// will receive the updated metadata from the server.
+/// The rest of the clients will still hold on the outdated metadata.
+/// In version 4 of CQL protocol there is currently no way for the server to notify other
+/// clients about prepared statement's metadata update.
+///
+/// ### Client-side metadata immutability
+/// The decision was made, to keep client-side metadata immutable.
+/// Mainly because of the CQLv4 limitations mentioned above. This means,
```
wprzytula, line 73: was made to keep

scylla/src/statement/prepared_statement.rs
```diff
@@ -22,6 +22,60 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
...
+/// that metadata is not updated during statement repreparation.
+/// This raises two issues:
+/// * bound values serialization errors - since [`PreparedMetadata`] is not updated
+/// * result deserialization errors - when [`PreparedStatement::set_use_cached_result_metadata`] is enabled,
+/// since [`ResultMetadata`] is not updated
```
wprzytula, line 78: I suggest summarising this by reiterating the conclusion:
_So, to mitigate those issues, drop the outdated `PreparedStatement` manually and prepare it again against the new schema._

wprzytula (changes requested):

docs/source/index.md
Lorak-mmk: That would require a change in naming (replacing `Query` with `UnpreparedStatement`), otherwise you would have a `Query` type but not of the docs would mention queries. What do yuo think about splitting in in 2 PRs - in this one let's improve prepared statement docs, and in the next one let's replace Query with UnpreparedStatement (and fix docs), which would fix https://github.com/scylladb/scylla-rust-driver/issues/713

docs/source/queries/prepared.md
```diff
@@ -1,8 +1,12 @@
 # Prepared query
 
-Prepared queries provide much better performance than simple queries,
+Prepared queries provide much better performance than unprepared queries,
 but they need to be prepared before use.
 
+There are other benefits that prepared queries have to offer:
+- Increased type-safety - thanks to metadata provided by the server, the driver can verify bound values' types before serialization. This way, we can be always sure that the Rust type provided by the user is compatible (and if not, the error is returned) with the destined native type. The same applies for deserialization.
```
Lorak-mmk, line 7: Unprepared statements are type safe too because they are prepared before execution. They return the same errors.
I'd argue they are safer because they don't suffer from the problem with schema changes - ser / deser will always use current metadata (unless schema is changed between preparation and execution)

scylla/src/statement/prepared_statement.rs
```diff
@@ -22,6 +22,60 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
...
 /// Represents a statement prepared on the server.
+///
+/// To prepare a statement, simply execute [`Session::prepare`](crate::transport::session::Session::prepare).
+///
+/// If you plan on reusing the statement, or bounding some values to it during execution, always
+/// prefer using prepared statements over [`Session::query`](crate::transport::session::Session::query).
+///
+/// Benefits that prepared statements have to offer:
+/// * Increased type-safety - bound values' types are validated with
+/// the [`PreparedMetadata`] received from the server during the serialization.
```
Lorak-mmk, line 33: ditto - this statement should be formulated differently.
Also the imo biggest advantage is not mentioned - only 1 round trip per execution instead of 2, for statements with values

scylla/src/statement/prepared_statement.rs
```diff
@@ -22,6 +22,60 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
...
+/// * Result deserialization optimization - see [`PreparedStatement::set_use_cached_result_metadata`].
+///
+/// # Clone implementation
+/// Cloning a prepared statement is a cheap operation. It only
+/// requires copying a couple of small fields and some [Arc] pointers.
+/// Always prefer cloning over executing [`Session::prepare`](crate::transport::session::Session::prepare)
+/// multiple times to save some roundtrips.
+///
+/// # Statement repreparation
+/// When schema is somehow updated, the server is supposed to invalidate its
```
Lorak-mmk, line 46: Why "somehow"? It sounds like schema change is something unusual / undesirable

docs/source/index.md
muzarski: Sounds good to me, I can take care of that.

scylla/src/statement/prepared_statement.rs
```diff
@@ -22,6 +22,60 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
...
+/// prepared statement caches. Then, if client tries to execute a given statement,
+/// the server will respond with an error. Users should not worry about it, since
+/// the driver handles it properly and tries to reprepare the statement.
+/// However, there are some cases when client-side prepared statement should be dropped
+/// and prepared once again via [`Session::prepare`](crate::transport::session::Session::prepare) -
+/// see the mention about altering schema below.
+///
+/// # Altering schema
+/// Do NOT alter the schema after preparing a statement.
+/// To be more precise:
```
Lorak-mmk, line 56: First statement is not imprecise, it's incorrect.
We should not discourage people from altering schema - it is a normal operation.
We should explain how to do this safely and what can go wrong.

Lorak-mmk (changes requested):

scylla/src/statement/prepared_statement.rs
```diff
@@ -22,6 +22,66 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
...
+/// be dropped. The statement should be prepared again.
+///
+/// There are two reasons for this:
+///
+/// ### CQLv4 protocol limitations
+/// In multi-client scenario, only the first client which reprepares the statement
+/// will receive the updated metadata from the server.
+/// The rest of the clients will still hold on the outdated metadata.
+/// In version 4 of CQL protocol there is currently no way for the server to notify other
+/// clients about prepared statement's metadata update.
```
wprzytula, line 72: Please state explicitly that we only support CQL v4.

scylla/src/statement/prepared_statement.rs
```diff
@@ -22,6 +22,66 @@ use crate::transport::execution_profile::ExecutionProfileHandle;
...
+///
+/// ### Client-side metadata immutability
+/// The decision was made to keep client-side metadata immutable.
+/// Mainly because of the CQLv4 limitations mentioned above. This means,
+/// that metadata is not updated during statement repreparation.
```
wprzytula, line 77: This means that (no comma should be here)

wprzytula (approved):

Lorak-mmk (approved):

PR #992, opened 2024-05-02
Remove `.with_current_subscriber()` calls

Lorak-mmk:
Those calls were introduced in e5570956758503a8f0a7ed0ad1c649909f370a73,
with message:
```
Now, when the driver spawns a task to run a new future on it, that
future will use the same subscriber as the code that spawned the task in
the first place.
```

There is unfortunately no explanation about when it is necessary.
I don't see any problems after removing those - logs are still collected correctly using a tracing subscriber.
Those calls however have a harmful side-effect: they prevent usage of `log` loggers to listen to driver logs using `log` feature in `tracing` crate, which was reported in https://github.com/scylladb/scylla-rust-driver/issues/860 .
After reporting the problem to `tracing` crate (https://github.com/tokio-rs/tracing/issues/2913) one of maintainers said that using `.with_current_subscriber()` in a library is not necessary in general.

As I don't see any issue caused by removing these calls, but their existence cause an issue, they are removed in this PR.
The PR also introduces an example showing usage of the driver with `env_logger` and adjusts documentation about logging.

Fixes: #860 

@piodul if you remember why those calls were added please explain, because I have no idea in which scenario they could be necessary.

## Pre-review checklist

### DIFF
--- a/docs/source/logging/logging.md
+++ b/docs/source/logging/logging.md
@@ -1,9 +1,15 @@
 # Logging
 
 The driver uses the [tracing](https://github.com/tokio-rs/tracing) crate for all logs.\
-To view the logs you have to create a `tracing` subscriber to which all logs will be written.
+There are two ways to view the logs:
+- Create a `tracing` subscriber to which all logs will be written (recommended).
+- Enable `log` feature on `tracing` crate and use some logger from `log` ecosystem. \
+Only do this if you can't use `tracing` subscriber for some reason.
 
-To just print the logs you can use the default subscriber:
+## Using tracing subscriber
+
+To print the logs you can use the default subscriber:
+
 ```rust
 # extern crate scylla;
 # extern crate tokio;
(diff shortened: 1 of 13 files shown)


piodul:
> @piodul if you remember why those calls were added please explain, because I have no idea in which scenario they could be necessary.

I'm not entirely sure. I might have discussed the original problem long ago with Gor and/or @wprzytula. I think it might have been necessary for the cpp driver.

One explanation that comes to my mind is:
1. You spawn a task with a subscriber which is different than the global one,
2. You create a session in that task,
3. For me, it sounds reasonable that the tasks spawned internally by the session should inherit the tracing subscriber of the task that the session was created in. Without `.with_current_subscriber()`, that wouldn't be the case.

Lorak-mmk:
If I understand correctly this is not in issue when global default subscriber is used, because it is a default for all threads. This is what cpp-rust-driver currently does.
To make sure, I've partially updated cpp-rust-driver to use current rust driver and run `examples/basic` (with added `cass_log_set_level(CASS_LOG_TRACE)` at the beggining of main).
Logs are properly printed both with main and this PR.

examples/logging_log.rs
```diff
@@ -0,0 +1,27 @@
...
+use scylla::SessionBuilder;
+use std::env;
+use tracing::info;
+
+// To run this example, and view logged messages, RUST_LOG env var needs to be set
+// This can be done using shell command presented below
+// RUST_LOG=info cargo run --example logging_log
+#[tokio::main]
+async fn main() -> Result<()> {
+    // Driver uses tracing for logging purposes, but it's possible to use `log`
```
wprzytula, line 12: nit: add backticks to `tracing`, same as you did with `log`.

examples/logging_log.rs
```diff
@@ -0,0 +1,27 @@
...
+    // ecosystem to view the messages. This requires adding `tracing` crate to
+    // dependencies and enabling it's "log" feature. Then you will be able to use
```
wprzytula, line 14: its

wprzytula (commented):
`Cargo.lock.msrv` needs to be updated. Otherwise, looks acceptable to me, although I'm still reluctant to believe that cpp-rust-driver's logging really works without those calls; they really had some purpose and I feel that there is some trap that we forget about.

Lorak-mmk:
> `Cargo.lock.msrv` needs to be updated. Otherwise, looks acceptable to me, although I'm still reluctant to believe that cpp-rust-driver's logging really works without those calls; they really had some purpose and I feel that there is some trap that we forget about.

If there is something that I didn't notice I'd prefer to work around it in cpp-rust-driver (even if it's inconvenient or even requires a patched version of the driver) rather than break the experience for Rust Driver users.

docs/source/logging/logging.md
```diff
@@ -45,4 +51,44 @@ To start this example execute:
...
+You can run it from main folder of driver repository using `RUST_LOG=trace SCYLLA_URI=<scylla_ip>:9042 cargo run --example logging`.
+
+## Using log
+
+To collect tracing events using log collector you first need to enable `log` feature on `tracing` crate.
+You can use `cargo add tracing -F log` or edit `Cargo.toml`:
+```toml
+tracing = { version = "0.1.40" , features = ["log"] }
+```
+then you can setup `env_logger` os some other logger and it will output logs from the driver:
```
wprzytula, line 64: Will enabling `log` feature in the end crate's Cargo.toml suffice, or is it required to have it enabled for scylla crate as well?
Another question. What if the end user crate uses `tracing` crate in version that is incompatible with the version that `scylla` crate uses?

Lorak-mmk: 1. Yes, it is enough
2. Not sure what you are asking. Are you asking if adding hypotetical tracing 0.2 to end user Cargo.toml and enabling log feature there will enable log feature in our tracing 0.1? No, it won't

Lorak-mmk: Quoting comment from the tracing issue I opened (https://github.com/tokio-rs/tracing/issues/2913#issuecomment-2007815042 ):
> A library should not generally enable the log or log-always feature flags, and generally should never be responsible for setting the default subscriber. Instead, user code that depends on the library should enable tracing's log crate support if it wants to, and set the default subscriber if it wishes to use tracing. I don't think the calls to with_current_subscriber are necessary in a library in general.

wprzytula: 2. I'm asking about whether mismatched versions of `tracing` in `scylla` crate and the end crate can cause incompatility issues, resulting in misbehaving tracing.

Lorak-mmk: No idea, probably. It's a question to tracing crate maintainers, how are they going to handle new major version to avoid ecosystem split.

wprzytula (approved):

PR #996, opened 2024-05-10
treewide: silence mock structs unused warning

wprzytula:
With a new version of Rust toolchain (stable 1.78.0), our never constructed structs used for testing macros began to trigger rustc warnings. Now, they are silenced explicitly using `#[allow(unused)]`.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types/serialize/row.rs
+++ b/scylla-cql/src/types/serialize/row.rs
@@ -1141,6 +1141,7 @@ mod tests {
     // Do not remove. It's not used in tests but we keep it here to check that
     // we properly ignore warnings about unused variables, unnecessary `mut`s
     // etc. that usually pop up when generating code for empty structs.
+    #[allow(unused)]
     #[derive(SerializeRow)]
     #[scylla(crate = crate)]
     struct TestRowWithNoColumns {}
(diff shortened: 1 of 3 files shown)


wprzytula:
If considerably many users are already using the new toolchain, we may want to consider releasing a patch version to silence those warnings.

Lorak-mmk:
> If considerably many users are already using the new toolchain, we may want to consider releasing a patch version to silence those warnings.

I see only changes in tests, it should not affect users of the driver so I don't see the point at all.


When are the warnings fixed by this PR triggered? I didn't see any failed CI runs because of this. 
Also, instead of saying `new version of Rust toolchain` please specify the version - someone reading this in the future won't know what version you had in mind without checking release dates.

wprzytula:
> I see only changes in tests, it should not affect users of the driver so I don't see the point at all.

Agreed.

> When are the warnings fixed by this PR triggered? I didn't see any failed CI runs because of this.

They pop up when I run `cargo check --all-targets`.

> Also, instead of saying `new version of Rust toolchain` please specify the version - someone reading this in the future won't know what version you had in mind without checking release dates.

Done.

Lorak-mmk:
Ah, it's probably not failing because we use Rust provided on CI runner images instead of setting up our own and I guess they were not updated yet.
We should probably change this at some point to use latest stable always.

Lorak-mmk (approved):

Lorak-mmk:
Eh, I forgot to check the commit message and you use `new version of Rust toolchain` there too.
Well, it's too late now and it's not that important

PR #1004, opened 2024-05-21
Pre-deserialization-refactor preparatory changes

wprzytula:
This is a bunch of preparatory changes to come before the proper deserialization refactor.
To avoid the PR with the first deserialization refactor portion being so big, I extracted them here to be merged sooner.

### What's done:
- `types::write_bytes_opt()` becomes more generic;
- types: return `std::io::Error` instead of `ParseError` where possible - don't unnecessarily broaden the codomain of those functions (not to have to unnecessarily pattern match on their returned error type);
- scylla-cql: add `assert_matches` as dev-dep. It will be helpful in testing deserialization framework, and will also be used in serialization instead of `assert!(matches!(...))`;
- codewide: ditch assert!(matches!(...)) - as assert_matches!(...) provide superior experience (providing helpful debugging messages with how the mismatched type looked like), all `assert!(matches!(...))` occurences across the code were changed to `assert_matches!(...)`;
- serialize: deduplicate `do_serialize[_err]`;
- serialize errors:
  - names and error messages are changed to better reflect reality;
  - a bunch of typos are fixed, and some names are changed to be more explicit;
  - where possible, `write!(f, ...)` is replaced with `f.write_str(...)` for performance. 

## Pre-review checklist

### DIFF
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -1524,6 +1524,7 @@ dependencies = [
 name = "scylla-cql"
 version = "0.2.0"
 dependencies = [
+ "assert_matches",
  "async-trait",
  "bigdecimal",
  "byteorder",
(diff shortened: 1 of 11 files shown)


wprzytula:
Rebased on main to fix clippy complaints.

muzarski (approved):

Issue #1001, opened 2024-05-15
Use empty `CollectionType<T>` instead of `Option<CollectionType<T>>`


Zk2u:
While [ScyllaDB doesn’t distinguish between null and and empty set](https://opensource.docs.scylladb.com/stable/cql/types.html#sets), it’s not very ergonomic to represent these collections as `Option<CollectionType<T>>`. For example, an empty set in Scylla should be deserialised into an empty `HashSet<T>`, not a `None::<CollectionType<T>>`. 

This would be a breaking API change.

wprzytula:
I'll include this functionality in the ongoing deserialization refactor, planned for release 0.14.

wprzytula:
@617a7aa Please take a look at #1007. Please confirm that the behaviour showcased by the added test is in line with your expectations.

617a7aa:
ACK, looks good to me.

PR #1007, opened 2024-05-21
WIP: session_test: regression test for empty collections deserialization

wprzytula:
ScyllaDB does not distinguish empty collections from nulls. That is, INSERTing an empty collection is equivalent to nullifying the corresponding column.
As pointed out in #1001, it's a nice QoL feature to be able to deserialize empty CQL collections to empty Rust collections instead of `None::<RustCollection>`. A test is added that checks that.

## NOTICE
The test fails with the current deserialization framework implementation. It should work, however, once the new deserialization framework is introduced and the test is migrated to use it.
The goal of this PR at this point is to showcase what is the desired behaviour and remind that it has not yet been reached.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -28,11 +28,12 @@ use assert_matches::assert_matches;
 use bytes::Bytes;
 use futures::{FutureExt, StreamExt, TryStreamExt};
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


617a7aa (approved):
ACK, this behaviour matches the issue's requirements.

wprzytula:
Rebased on main.

wprzytula:
Done in #1057.

PR #1013, opened 2024-06-18
CI: fix clippy issues

muzarski:
There were two clippy checks that were failing:
- not using `unwrap_or_default` in the code generated by `darling` crate. See the corresponding issue: https://github.com/TedDriggs/darling/issues/293
- usage of legacy numeric constant - `std::i16::MAX`, instead of `i16::MAX`

This PR fixes the issues and makes CI pass.

## Pre-review checklist

### DIFF
--- a/scylla-macros/src/serialize/row.rs
+++ b/scylla-macros/src/serialize/row.rs
@@ -1,3 +1,6 @@
+// See: https://github.com/TedDriggs/darling/issues/293
+#![allow(clippy::manual_unwrap_or_default)]
+
 use std::collections::HashMap;
 
 use darling::FromAttributes;
--- a/.github/workflows/authenticate_test.yml
+++ b/.github/workflows/authenticate_test.yml
@@ -27,5 +27,7 @@ jobs:
         options: --health-cmd "cqlsh --username cassandra --password cassandra --debug" --health-interval 5s --health-retries 30
     steps:
     - uses: actions/checkout@v3
+    - name: Update rust toolchain
+      run: rustup update
     - name: Run tests
       run: RUST_LOG=trace cargo test --verbose authenticate_superuser -- custom_authentication --ignored
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -27,6 +27,8 @@ jobs:
         options: --health-cmd "cqlsh --debug scylladb" --health-interval 5s --health-retries 10
     steps:
     - uses: actions/checkout@v3
+    - name: Update rust toolchain
+      run: rustup update
     - name: Install mdbook
       run: cargo install mdbook --no-default-features
     - name: Build the project
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -25,6 +25,8 @@ jobs:
       run: |
         docker compose -f test/cluster/cassandra/docker-compose.yml up -d --wait
       # A separate step for building to separate measuring time of compilation and testing
+    - name: Update rust toolchain
+      run: rustup update
     - name: Build the project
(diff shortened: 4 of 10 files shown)


scylla-macros/src/serialize/row.rs
```diff
@@ -1,3 +1,6 @@
+// See: https://github.com/TedDriggs/darling/issues/293
+#![allow(clippy::manual_unwrap_or_default)]
```
Lorak-mmk, line 2: We need to remove those after the issue is fixed in darling. Please open an issue about it so we don't forget

Lorak-mmk: Also do we need it on the whole file? Isn't it possible to only have it in problematic places?

Lorak-mmk (changes requested):

muzarski:
It looks like GH actions used the old version of clippy: `clippy 0.1.78 (9b00956 2024-04-29)`.
My version locally is: `clippy 0.1.79 (129f3b9 2024-06-10)`.

That's weird, because the runners used for CI in `cql-stress` already use newest version of clippy.

scylla-macros/src/serialize/row.rs
muzarski: > Also do we need it on the whole file? Isn't it possible to only have it in problematic places?

I tried putting it in the following places:
- next to the fields marked with `#[darling(default)]` attribute (tried both after and before darling attribute)
- next to the struct with marked fields

None of these worked. Do you have any more ideas?

piodul: You can also try putting the `allow` before or after the `#[derive(FromAttributes)]` part.

muzarski: > You can also try putting the `allow` before or after the `#[derive(FromAttributes)]` part.

Unfortunately, this does not work

Lorak-mmk:
Afaik rustup is present in runner images. Maybe we should explicitly install latest stable using rustup and set it as default in CI instead of relying on what is the current default on the runners.

wprzytula (approved):

muzarski:
> Afaik rustup is present in runner images. Maybe we should explicitly install latest stable using rustup and set it as default in CI instead of relying on what is the current default on the runners.

v2: Added a step which updates rust toolchain in CI.

muzarski:
> > Afaik rustup is present in runner images. Maybe we should explicitly install latest stable using rustup and set it as default in CI instead of relying on what is the current default on the runners.
> 
> v2: Added a step which updates rust toolchain in CI.

@wprzytula @Lorak-mmk Do you think that we should do the same for other workflows that use rust toolchain? As of now, we only update it for `Rust / build` workflow, which does all of the clippy and format checks. However, there are other workflows that use rust compiler, e.g. `Cassandra tests / build` or `Book / build` (they do not use `clippy/rustfmt`, tho).

wprzytula:
> Do you think that we should do the same for other workflows that use rust toolchain?

I can see two approaches. 
1. Unified CI, if we always assume that we only check on two Rust versions: latest stable and MSRV. This will hopefully result in less CI failures due to such problems as toolchain version difference locally vs in CI.
2. More than one recent stable toolchain version is used. E.g. latest and the last-before-the-latest. I think that it's a shot in a foot, because even more inconvenience will follow from such dualism.

Lorak-mmk:
I agree, let's use latest stable everywhere, and msrv in msrv pipeline

muzarski:
v2: using latest stable version of rust toolchain in CI (min_rust is exception - using MSRV)

wprzytula (approved):

wprzytula:
@Lorak-mmk Please re-review and merge.

Issue #519, opened 2022-08-18
Better names and higher granularity for errors


wprzytula:
### Problem statement
There are quite a few places in driver's code where the errors' names and/or messages are inappropriate. Unjustified `ProtocolError`'s usage is prevalent. We lack an error for faults triggered by user's logic, such as unwise custom `RetryPolicy` being used.

### Proposed solution
Add more error variants (mainly for `QueryError`), fix the usaged to make names and messages appropriate. Stop abusing `ProtocolError` when no guilt is on received messages.

roydahan:
What other drivers do for it and the question is if we need to conform into a practice that all drivers agreed upon?

wprzytula:
> What other drivers do for it and the question is if we need to conform into a practice that all drivers agreed upon?

With drivers, error handling depends strongly on the language used.
- Java driver uses exceptions (I suppose Python too);
- gocql uses `err` type together with various textual error messages;
- Cpp driver returns error code without any additional context;
- Rust driver has a hierarchy of `enum`s that builds up a tree of error types; this could be described in some document (as I asked @muzarski for) and discussed about. In Rust driver, we want strongly typed errors to be able to pattern match on them.

What this issue is about **is not** making error handling in Rust driver "conformant" to other drivers (because there's no such convention whatsoever), yet it is about changes to the error types hierarchy to make it more reasonable.

roydahan:
@muzarski / @wprzytula what are the minimal set of errors we would like to handle for this for the 0.14.0 milestone or in general as the "API breaking" item.
I doubt we need to handle all cases right now.

In any case, if easy let's divide them to issues under an umbrella issue and target them to milestones.

wprzytula:
In the Rust world, all alterations to current errors must be, unfortunately, considered API-breaking. Only after the introduced changes will the error framework be well-suited for changes in a non-API-breaking way. That means that we must do that before 1.0.
But we can, of course, divide the changes into smaller parts and attach to multiple milestones.

roydahan:
@muzarski what's left in order to close this one?

muzarski:
> [@muzarski](https://github.com/muzarski) what's left in order to close this one?

[...]

PR #1017, opened 2024-06-20
errors: Low level (de)ser errors

muzarski:

# Motivation

This commit is a small step forward to resolving the issues related to errors in rust driver. The main goal of this PR is to get rid of `ParseError` from `frame::types` module.

As a bonus, the newly introduced deserialization module is adjusted as well. Thanks to that, we remove the cycle in error types (`DeserializationError <-> ParseError`).

# Changes
- introduced new enum error type returned from types module: `LowLevelDeserializationError`
- changed the return error type of `types::read_*` functions to `LowLevelDeserializationError`
-  replaced `value::BuiltinDeserializationErrorKind::GenericParseError` with new variants
- changed the return error type of `deser_cql_value` to DeserializationError
- replaced `TabletsParsingError::Parse` with `TabletsParsingError::DeserializationError` 

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -35,6 +35,8 @@ pub enum FrameError {
 
 #[derive(Error, Debug)]
 pub enum ParseError {
+    #[error("Low-level deserialization failed: {0}")]
+    LowLevelDeserializationError(#[from] LowLevelDeserializationError),
     #[error("Could not serialize frame: {0}")]
     BadDataToSerialize(String),
     #[error("Could not deserialize frame: {0}")]
@@ -51,4 +53,35 @@ pub enum ParseError {
     SerializationError(#[from] SerializationError),
     #[error(transparent)]
     CqlTypeError(#[from] CqlTypeError),
+}
+
+/// A low level deserialization error.
+///
+/// This type of error is returned when deserialization
+/// of some primitive value fails.
+///
+/// Possible error kinds:
+/// - generic io error - reading from buffer failed
+/// - out of range integer conversion
+/// - conversion errors - e.g. slice-to-array or primitive-to-enum
+/// - not enough bytes in the buffer to deserialize a value
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum LowLevelDeserializationError {
+    #[error(transparent)]
+    IoError(#[from] std::io::Error),
+    #[error(transparent)]
+    TryFromIntError(#[from] std::num::TryFromIntError),
+    #[error("Failed to convert slice into array: {0}")]
+    TryFromSliceError(#[from] std::array::TryFromSliceError),
+    #[error("Not enough bytes! expected: {expected}, received: {received}")]
+    TooFewBytesReceived { expected: usize, received: usize },
+    #[error("Invalid value length: {0}")]
+    InvalidValueLength(i32),
+    #[error("Unknown consistency: {0}")]
+    UnknownConsistency(#[from] TryFromPrimitiveError<u16>),
+    #[error("Invalid inet bytes length: {0}")]
+    InvalidInetLength(u8),
+    #[error("UTF8 deserialization failed: {0}")]
+    UTF8DeserializationError(#[from] std::str::Utf8Error),
 }
--- a/scylla-cql/src/frame/types.rs
+++ b/scylla-cql/src/frame/types.rs
@@ -1,5 +1,6 @@
 //! CQL binary protocol in-wire types.
 
+use super::frame_errors::LowLevelDeserializationError;
 use super::frame_errors::ParseError;
 use super::TryFromPrimitiveError;
 use byteorder::{BigEndian, ReadBytesExt};
@@ -160,13 +161,15 @@ impl<'a> RawValue<'a> {
     }
 }
 
-fn read_raw_bytes<'a>(count: usize, buf: &mut &'a [u8]) -> Result<&'a [u8], ParseError> {
+fn read_raw_bytes<'a>(
+    count: usize,
+    buf: &mut &'a [u8],
+) -> Result<&'a [u8], LowLevelDeserializationError> {
     if buf.len() < count {
-        return Err(ParseError::BadIncomingData(format!(
-            "Not enough bytes! expected: {} received: {}",
-            count,
-            buf.len(),
-        )));
+        return Err(LowLevelDeserializationError::TooFewBytesReceived {
+            expected: count,
+            received: buf.len(),
+        });
     }
     let (ret, rest) = buf.split_at(count);
     *buf = rest;
@@ -182,7 +185,7 @@ pub fn write_int(v: i32, buf: &mut impl BufMut) {
     buf.put_i32(v);
 }
 
-pub fn read_int_length(buf: &mut &[u8]) -> Result<usize, ParseError> {
+pub fn read_int_length(buf: &mut &[u8]) -> Result<usize, LowLevelDeserializationError> {
     let v = read_int(buf)?;
     let v: usize = v.try_into()?;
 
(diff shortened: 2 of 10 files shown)


muzarski:
v1.1: rebased on main

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -52,3 +54,11 @@ pub enum ParseError {
     #[error(transparent)]
     CqlTypeError(#[from] CqlTypeError),
 }
+
+/// A low level serialization error.
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum LowLevelSerializationError {
+    #[error("Integer conversion out of range: {0}")]
+    TryFromIntError(#[from] std::num::TryFromIntError),
```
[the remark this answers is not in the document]
muzarski: The error message is: "out of range integral type conversion attempted". I'll leave it as `error(transparent)` then, since we cant provide more context in this case.

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -60,9 +60,7 @@ impl<'frame> DeserializeValue<'frame> for CqlValue {
         v: Option<FrameSlice<'frame>>,
     ) -> Result<Self, DeserializationError> {
         let mut val = ensure_not_null_slice::<Self>(typ, v)?;
-        let cql = deser_cql_value(typ, &mut val).map_err(|err| {
-            mk_deser_err::<Self>(typ, BuiltinDeserializationErrorKind::GenericParseError(err))
-        })?;
+        let cql = deser_cql_value(typ, &mut val).map_err(deser_error_replace_rust_name::<Self>)?;
         Ok(cql)
```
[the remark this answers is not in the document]
muzarski: If I remove this, the following test fails:
```failures:

---- types::deserialize::row::tests::test_row_errors stdout ----
thread 'types::deserialize::row::tests::test_row_errors' panicked at scylla-cql/src/types/deserialize/row.rs:651:13:
assertion `left == right` failed
  left: "i64"
 right: "scylla_cql::frame::response::result::CqlValue"
```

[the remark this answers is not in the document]
muzarski: I set the `CqlValue` as `rust_type` in `deser_cql_value` in case of nested types (e.g. `CqlValue::Tuple`).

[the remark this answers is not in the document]
wprzytula: OK, got it.

muzarski:
v2:
- addressed review comments
- converted error type for all `types::[read/write]_` to `LL(De)SerializationError`. Previously, I forgot to unify the return error types for all of the utility functions. This is helpful, so the higher level errors won't need to handle `std::io::Error/std::num::TryFromIntError` separately.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -52,3 +56,48 @@ pub enum ParseError {
...
+/// of some primitive value fails.
+///
+/// Possible error kinds:
+/// - generic io error - reading from buffer failed
+/// - out of range integer conversion
+/// - conversion errors - e.g. slice-to-array or primitive-to-enum
+/// - not enough bytes in the buffer to deserialize a value
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum LowLevelDeserializationError {
```
wprzytula, line 86: Now I'm very happy with the docstrings :)

wprzytula (approved):

wprzytula:
What about adding context to LLDeserError in new deserialization framework?

scylla-cql/src/frame/types.rs
```diff
@@ -161,20 +162,22 @@ impl<'a> RawValue<'a> {
...
+            received: buf.len(),
+        });
     }
     let (ret, rest) = buf.split_at(count);
     *buf = rest;
     Ok(ret)
 }
 
-pub fn read_int(buf: &mut &[u8]) -> Result<i32, std::io::Error> {
+pub fn read_int(buf: &mut &[u8]) -> Result<i32, LowLevelDeserializationError> {
```
Lorak-mmk, line 180: This makes returned type less specific. If this function can only return io error then why force the caller to handle other errors?

scylla-cql/src/frame/types.rs
```diff
@@ -207,7 +210,7 @@ fn type_int() {
     }
 }
 
-pub fn read_long(buf: &mut &[u8]) -> Result<i64, std::io::Error> {
+pub fn read_long(buf: &mut &[u8]) -> Result<i64, LowLevelDeserializationError> {
```
Lorak-mmk, line 213: This applies also here and other places where `std::io::Error` was replaced

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -1656,10 +1651,18 @@ impl Display for BuiltinDeserializationErrorKind {
...
     }
 }
 
+impl From<LowLevelDeserializationError> for BuiltinDeserializationErrorKind {
+    #[inline]
+    fn from(err: LowLevelDeserializationError) -> Self {
+        Self::LowLevelDeserializationError(err)
+    }
+}
+
```
Lorak-mmk, line 1665: In other places we are using `thiserror` crate to generate the conversions. Why not do the same here?

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -37,6 +37,8 @@ pub enum FrameError {
 pub enum ParseError {
```
Lorak-mmk, line 37: Please remind me, is the current plan to get rid of ParseError completely?
If not, we need to remember to add non_exhaustive to it, and other errors (like FrameError above).

Lorak-mmk (changes requested):

Lorak-mmk (changes requested):
I love it when github decides to post my review but without the main comment I wrote :) Posting again...

I think those are good changes in the right direction in general.

Please explain, because I don't remember from our discussion and I can't understand from commits: why do we have a distinction between deserialization error and low level deserialization error? The errors are a public API, why should a user of a library care which internal layer of the driver returned the error and not just what the error is?


Note: If we don't have something like this, whole error refactor should be finished with tests that will present and assert the error messages in various conditions. That way we can see them without writing one-off code to trigger those errors.

scylla-cql/src/frame/frame_errors.rs
muzarski: I ultimately would like to get rid of `ParseError`. It will be decoupled into multiple error types, which will be more specific and what's the most important - they will be typed. As of now, most of the `ParseError`s are stringified (e.g. converted to `QueryError::InvalidMessage`).

There is a similar issue with a `FrameError`, however I'm not planning to remove it. I'd like to decouple it and introduce a distinction between `FrameSerializationError` (for client frames) and `FrameDeserializationError` (for server frames). These two could be then represented as variants of `{Query/NewSession}Error` (and not stringified to `QueryError::InvalidMessage`, which is something we currently do).

wprzytula:
> Please explain, because I don't remember from our discussion and I can't understand from commits: why do we have a distinction between deserialization error and low level deserialization error? The errors are a public API, why should a user of a library care which internal layer of the driver returned the error and not just what the error is?

LowLevelDeserializationError is strictly connected to the CQL protocol types deserialization layer. On top of it, there is DeserializationError, which carries the meaning of a logical error during deserialization.
For example, the underlying CQL protocol type structure can be perfectly valid (so no LowLevelDeserializationError involved), but invalid for the expected logical structure of, say, a CQL Set. (e.g., an `int` present when `bytes` or `string` was expected). While the example is not perfect, I hope you get the difference.

muzarski:
v3:
- rebased on main
- removed `LowLevelSerializationError` at all. The reason is that `types::write_*` functions all return `std::num::TryFromIntError`, so currently there is no need to wrap it with an additional error type.
- narrowed the return type of `types::read_*` functions to `std::io::Error` (if possible)
- added more context to deserialization framework. Introduced three new variants which wrap `LowLevelDeserializationError`:
  - BadDate - deserialization of some date's field failed (months, days, or nanoseconds)
  - BadDecimalScale - deserialization of decimal's scale failed
  - RawCqlBytesReadError - failed to read raw cql value bytes from a frame slice

Lorak-mmk:
>     * removed `LowLevelSerializationError` at all. The reason is that `types::write_*` functions all return `std::num::TryFromIntError`, so currently there is no need to wrap it with an additional error type.
> 

I didn't read the changes yet, so my question may be inaccurate.
Isn't it possible for this situation to change in the future?
If it is, then `LowLevelSerializationError` is better because we can add variants to it.
If it can't then nvm

Lorak-mmk (approved):

wprzytula (approved):

PR #1019, opened 2024-06-24
Remove tokio version bounds

rukai:
This is fine now since https://github.com/tokio-rs/tokio/issues/6610 is resolved via https://github.com/becheran/ntest/pull/28 which has now been released in ntest 0.9.3
This PR also bumps the itertools + base64 versions, since they are trivial upgrades

The latest git main branch of scylla-rust-driver fails to build in my project due to the version restriction, these kinds of version restrictions often cause problems like this, so it would be bad for such a version restriction to end up in a release.

## Pre-review checklist

### DIFF
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -961,9 +961,9 @@ dependencies = [
 
 [[package]]
 name = "ntest"
-version = "0.9.0"
+version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "da8ec6d2b73d45307e926f5af46809768581044384637af6b3f3fe7c3c88f512"
+checksum = "fb183f0a1da7a937f672e5ee7b7edb727bf52b8a52d531374ba8ebb9345c0330"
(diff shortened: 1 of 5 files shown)


Lorak-mmk:
Thanks for the contribution!
If the issue was resolved via a new version of ntest, then
1. There is no need to bump minimum version of Tokio and we could still use `version = "1.34"`, right?
2. ntest should be bumped to `0.9.3` to avoid problems for people who have lockfiles locally.

(nit) could you split itrertools + base64 update to a separate commit?

rukai:
The feedback should be addressed now

Lorak-mmk:
What I meant was to also update ntest in Cargo.toml files - we need to require new version because old version doesn't work with new Tokio.
After you bump those versions I think it's good to go

rukai:
I've now updated the PR to also update ntest in the Cargo.toml not just in the Cargo.lock.msrv

wprzytula (approved):

Lorak-mmk (approved):

Issue #1014, opened 2024-06-19
CQL Vector type


pkolaczk:
https://cassandra.apache.org/doc/latest/cassandra/reference/vector-data-type.html

I noticed that any attempt to perform a query on a table that contains a vector column fails, because the schema is rejected. 

Any hints where I should modify the code to add support for this type?

mykaul:
ScyllaDB doesn't yet support Vector types.

wprzytula:
Reopening as support for CQL Vector is part of Cassandra compatibility, which we strive towards. Contributions are welcome that would add support for it.

wprzytula:
> Any hints where I should modify the code to add support for this type?

For adding `CqlVector` type, see [frame/value.rs](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/frame/value.rs) (e.g. similarly to `CqlTime`).
Also, add a variant to [`CqlValue`](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/frame/response/result.rs#L75) and implement [`FromCqlVal`](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/frame/response/cql_to_rust.rs#L141).

Then, the ser/de capabilities should be added by implementing `{De,S}erializeValue` in  [serialize/value.rs](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/types/serialize/value.rs) and [deserialize/value.rs](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/types/deserialize/value.rs).

mykaul:
> > Any hints where I should modify the code to add support for this type?
> 
> For adding `CqlVector` type, see [frame/value.rs](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/frame/value.rs) (e.g. similarly to `CqlTime`). Also, add a variant to [`CqlValue`](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/frame/response/result.rs#L75) and implement [`FromCqlVal`](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/frame/response/cql_to_rust.rs#L141).
> 
[...]

PR #1020, opened 2024-06-24
cassandra 5.0 vector type CREATE/INSERT support

rukai:
makes progress towards: https://github.com/scylladb/scylla-rust-driver/issues/1014

The vector type is introduced by the currently in beta cassandra 5.
See: https://cassandra.apache.org/doc/latest/cassandra/reference/vector-data-type.html
Scylla does not support vector types and so the tests are setup to only compile/run with a new cassandra_tests config.

This commit does not add support for retrieving the data via a SELECT.
That was omitted to reduce scope and will be implemented in follow up work.

I have split off support for CREATE/INSERT since I value the early feedback of small PRs.
I didnt realize that @pkolaczk had shown interest in implementing this until after I'd went to submit my PR.
Hopefully I've not stepped on any toes.

The cassandra docker-compose.yaml is changed to 5.0 beta cassandra, if this is undesirable I can introduce a second docker-compose.yaml specifically for 5.0

image
Confirmed that the test is running in CI

## Pre-review checklist

### DIFF
--- a/.github/workflows/cassandra.yml
+++ b/.github/workflows/cassandra.yml
@@ -31,7 +31,7 @@ jobs:
       run: cargo build --verbose --tests --features "full-serialization"
     - name: Run tests on cassandra
       run: |
-        CDC='disabled' RUST_LOG=trace SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose --features "full-serialization" -- --skip test_views_in_schema_info --skip test_large_batch_statements
+        CDC='disabled' RUSTFLAGS="--cfg cassandra_tests" RUST_LOG=trace SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose --features "full-serialization" -- --skip test_views_in_schema_info --skip test_large_batch_statements
     - name: Stop the cluster
       if: ${{ always() }}
       run: docker compose -f test/cluster/cassandra/docker-compose.yml stop
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -3165,3 +3165,108 @@ async fn test_api_migration_session_sharing() {
         assert!(matched);
     }
 }
+
+#[cfg(cassandra_tests)]
+#[tokio::test]
+async fn test_vector_type_metadata() {
+    setup_tracing();
+    let session = create_new_session_builder().build().await.unwrap();
+    let ks = unique_keyspace_name();
+
+    session.query_unpaged(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
+    session
+        .query_unpaged(
+            format!(
+                "CREATE TABLE IF NOT EXISTS {}.t (a int PRIMARY KEY, b vector<int, 4>, c vector<text, 2>)",
+                ks
+            ),
+            &[],
+        )
+        .await
+        .unwrap();
+
+    session.refresh_metadata().await.unwrap();
+    let metadata = session.get_cluster_data();
+    let columns = &metadata.keyspaces[&ks].tables["t"].columns;
+    assert_eq!(
+        columns["b"].type_,
+        CqlType::Vector {
+            type_: Box::new(CqlType::Native(NativeType::Int)),
+            dimensions: 4,
+        },
+    );
+    assert_eq!(
+        columns["c"].type_,
+        CqlType::Vector {
+            type_: Box::new(CqlType::Native(NativeType::Text)),
+            dimensions: 2,
+        },
+    );
+}
+
+#[cfg(cassandra_tests)]
+#[tokio::test]
+async fn test_vector_type_unprepared() {
+    setup_tracing();
+    let session = create_new_session_builder().build().await.unwrap();
+    let ks = unique_keyspace_name();
+
+    session.query_unpaged(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
+    session
+        .query_unpaged(
+            format!(
+                "CREATE TABLE IF NOT EXISTS {}.t (a int PRIMARY KEY, b vector<int, 4>, c vector<text, 2>)",
+                ks
+            ),
+            &[],
+        )
+        .await
+        .unwrap();
+
+    session
+        .query_unpaged(
+            format!(
+                "INSERT INTO {}.t (a, b, c) VALUES (1, [1, 2, 3, 4], ['foo', 'bar'])",
+                ks
+            ),
+            &[],
+        )
+        .await
+        .unwrap();
+
+    // TODO: Implement and test SELECT statements and bind values (`?`)
+}
+
+#[cfg(cassandra_tests)]
+#[tokio::test]
+async fn test_vector_type_prepared() {
+    setup_tracing();
+    let session = create_new_session_builder().build().await.unwrap();
+    let ks = unique_keyspace_name();
+
+    session.query_unpaged(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
+    session
+        .query_unpaged(
+            format!(
+                "CREATE TABLE IF NOT EXISTS {}.t (a int PRIMARY KEY, b vector<int, 4>, c vector<text, 2>)",
+                ks
+            ),
+            &[],
+        )
+        .await
+        .unwrap();
+
+    let prepared_statement = session
+        .prepare(format!(
+            "INSERT INTO {}.t (a, b, c) VALUES (?, [11, 12, 13, 14], ['afoo', 'abar'])",
+            ks
+        ))
+        .await
+        .unwrap();
+    session
+        .execute_unpaged(&prepared_statement, &(2,))
+        .await
+        .unwrap();
+
+    // TODO: Implement and test SELECT statements and bind values (`?`)
+}
--- a/scylla/src/transport/topology.rs
+++ b/scylla/src/transport/topology.rs
@@ -188,6 +188,12 @@ enum PreCqlType {
         type_: PreCollectionType,
     },
     Tuple(Vec<PreCqlType>),
+    Vector {
+        type_: Box<PreCqlType>,
+        /// matches the datatype used by the java driver:
+        /// <https://github.com/apache/cassandra-java-driver/blob/85bb4065098b887d2dda26eb14423ce4fc687045/core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java#L77>
+        dimensions: i32,
+    },
     UserDefinedType {
         frozen: bool,
         name: String,
@@ -211,6 +217,10 @@ impl PreCqlType {
                     .map(|t| t.into_cql_type(keyspace_name, udts))
                     .collect(),
             ),
+            PreCqlType::Vector { type_, dimensions } => CqlType::Vector {
+                type_: Box::new(type_.into_cql_type(keyspace_name, udts)),
+                dimensions,
+            },
             PreCqlType::UserDefinedType { frozen, name } => {
                 let definition = match udts
                     .get(keyspace_name)
@@ -236,6 +246,12 @@ pub enum CqlType {
         type_: CollectionType,
     },
     Tuple(Vec<CqlType>),
+    Vector {
+        type_: Box<CqlType>,
+        /// matches the datatype used by the java driver:
+        /// <https://github.com/apache/cassandra-java-driver/blob/85bb4065098b887d2dda26eb14423ce4fc687045/core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java#L77>
+        dimensions: i32,
+    },
     UserDefinedType {
         frozen: bool,
         // Using Arc here in order not to have many copies of the same definition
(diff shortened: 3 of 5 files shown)


scylla/src/transport/topology.rs
```diff
@@ -1602,6 +1615,28 @@ fn parse_cql_type(p: ParserState<'_>) -> ParseResult<(PreCqlType, ParserState<'_
...
+        let p = p.skip_white();
+        let p = p.accept(",")?;
+        let p = p.skip_white();
+
+        let (size, p) = p.take_while(|c| c.is_numeric());
+        let size = size.parse().map_err(|_| {
+            p.error(ParseErrorCause::Other(
+                "Expected integer but found non-numeric character",
+            ))
+        })?;
```
wprzytula, line 1630: The error message is slightly misleading. After all, `4.0` is a valid number, so `.` is a numeric character, but it's never a part of an integer.

scylla/src/transport/topology.rs
```diff
@@ -184,6 +184,10 @@ enum PreCqlType {
         type_: PreCollectionType,
     },
     Tuple(Vec<PreCqlType>),
+    Vector {
+        type_: Box<PreCqlType>,
+        size: u16,
+    },
```
wprzytula, line 190: Why did you choose specifically `u16` for size? Can you share your reference documentation?

test/cluster/cassandra/docker-compose.yml
```diff
@@ -10,12 +10,12 @@ networks:
         - subnet: 172.42.0.0/16
 services:
   cassandra1:
-    image: cassandra
+    image: cassandra:5.0-beta1
     healthcheck:
```
wprzytula, line 14: I'm unsure if we are OK with testing only with newest beta cassandra instead of stable. WDYT @Lorak-mmk?

scylla/src/transport/session_test.rs
```diff
@@ -2885,3 +2885,36 @@ async fn test_manual_primary_key_computation() {
...
+    session
+        .query(
+            format!(
+                "INSERT INTO {}.t (a, b, c) VALUES (1, [1, 2, 3, 4], ['foo', 'bar'])",
+                ks
+            ),
+            &[],
+        )
+        .await
+        .unwrap();
```
wprzytula, line 2917: Important: missing test for a prepared statement combined with Vector.

wprzytula: Question: is Vector insertable as bind value?
I.e., is the following query valid?
`INSERT INTO {}.t (a, b, c) VALUES (1, ?, ?)`
If so, `ColumnType::Vector` variant must be added and ser/de for it must be supported.

wprzytula (commented):
What I see, looks good.

However, what I see in this PR is support for parsing schema containing Vector type. What is completely missing is serialization and deserialization support (the latter, you said, you leave for a follow-up). See https://github.com/scylladb/scylla-rust-driver/issues/1014#issuecomment-2178186151 for hints.

scylla/src/transport/topology.rs
rukai: Good point, Ive now added the reference as a comment above that line.

scylla/src/transport/session_test.rs
rukai: I added a test using prepared statements.

I've added bind values to the TODO at the bottom of the test. Lets address them in the follow up work, since when we add `ColumnType::Vector` we will need to add support for all the other vector functionality at the same time.

scylla/src/transport/topology.rs
```diff
@@ -184,6 +184,12 @@ enum PreCqlType {
         type_: PreCollectionType,
     },
     Tuple(Vec<PreCqlType>),
+    Vector {
+        type_: Box<PreCqlType>,
+        /// as per <https://cassandra.apache.org/doc/latest/cassandra/reference/vector-data-type.html>
+        /// vectors are limited to a size of 8192
+        size: u16,
```
wprzytula, line 191: Now I can see that this is the size limit of a Vector, but what interests me is how Cassandra represents Vector in the system tables. It surely does not use unsigned int, because AFAIK unsigned numbers are not supported by neither ScyllaDB nor Cassandra.
The idea of `PreCqlType` in `topology.rs` is to represent the raw type definition read from system tables. Only later can we convert it to more convenient representation, involving unsigned ints.

wprzytula: I believe it makes more sense to store size in the same type as Cassandra does.

rukai: The java driver uses an int, so I'll do the same: https://github.com/apache/cassandra-java-driver/blob/85bb4065098b887d2dda26eb14423ce4fc687045/core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java#L77

rukai: I adjusted the code to use an i32 to match the java driver.
I also renamed the `size` to `dimensions` as that is a more accurate name.

scylla/src/transport/session_test.rs
```diff
@@ -2885,3 +2885,45 @@ async fn test_manual_primary_key_computation() {
...
+
+    let prepared_statement = session
+        .prepare(format!(
+            "INSERT INTO {}.t (a, b, c) VALUES (2, [11, 12, 13, 14], ['afoo', 'abar'])",
+            ks
+        ))
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
rukai: I have changed query string to bind the partition key.

I was already executing the statement, but I've adjusted the execute to provide the binded value.

scylla/src/utils/parse.rs
```diff
@@ -87,6 +87,21 @@ impl<'s> ParserState<'s> {
...
     }
 
+    /// Parses a sequence of digits and '-' as an integer.
+    /// Consumes characters until it finds a character that is not a digit or '-'.
+    ///
+    /// An error is returned if:
+    /// * The first character is not a digit or '-'
+    /// * The the integer is larger than i32
+    pub(crate) fn parse_i32(self) -> ParseResult<(i32, Self)> {
+        let (digits, p) = self.take_while(|c| c.is_ascii_digit() || c == '-');
```
wprzytula, line 97: This has a minor flaw of accepting strings such as `00-21-37`, but such strings will be rejected below anyway, so no problem here.

scylla/src/utils/parse.rs
```diff
@@ -87,6 +87,21 @@ impl<'s> ParserState<'s> {
         me
     }
 
+    /// Parses a sequence of digits and '-' as an integer.
+    /// Consumes characters until it finds a character that is not a digit or '-'.
+    ///
+    /// An error is returned if:
+    /// * The first character is not a digit or '-'
+    /// * The the integer is larger than i32
```
wprzytula, line 95: The the -> The

scylla/src/transport/session_test.rs
```diff
@@ -2885,3 +2885,45 @@ async fn test_manual_primary_key_computation() {
...
+        .prepare(format!(
+            "INSERT INTO {}.t (a, b, c) VALUES (?, [11, 12, 13, 14], ['afoo', 'abar'])",
+            ks
+        ))
+        .await
+        .unwrap();
+    session.execute(&prepared_statement, &(2,)).await.unwrap();
+
+    // TODO: Implement and test SELECT statements and bind values (`?`)
+}
```
wprzytula, line 2929: Could you add a test for schema metadata? Something like this (probably a bit more checks here):
```rust
    let metadata = session.get_cluster_data();
    let columns = &metadata.keyspaces[&ks].tables["t"].columns;
    assert_eq!(
        columns["b"].type_,
        CqlType::Vector(...);
    );
```

rukai: good idea! done.

scylla/src/utils/parse.rs
rukai: done

wprzytula (dismissed):

test/cluster/cassandra/docker-compose.yml
Lorak-mmk: We can use this in addition to stable, but not instead of stable

rukai: I've moved cassandra-5.0-beta1 into a separate compose file.

scylla/src/transport/topology.rs
Lorak-mmk: Unresolving this comment because we talked about that with @wprzytula 
I'm not aware of any case were size of vector is stored in some table (as a cql type). IIUC we always parse type names from strings - so the argument about system tables doesn't apply.
Is it possible to create vector with negative length? I'd be very surprised if it was possible. 
If it is not we can and should use unsigned integer for the size (and simplify the parsing function thanks to that).

Should it be u16? I'm not sure. If documentation says that size is limited to 8192 then yes, but I can't find this information is the link in the comment. Could you link the source?

.github/workflows/cassandra.yml
```diff
@@ -40,3 +40,19 @@ jobs:
...
+      run: |
+        CDC='disabled' RUSTFLAGS="--cfg cassandra_tests" RUST_LOG=trace SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --verbose --features "full-serialization" -- --skip test_views_in_schema_info --skip test_large_batch_statements
+    - name: Stop the cluster
+      if: ${{ always() }}
+      run: docker compose -f test/cluster/cassandra5/docker-compose.yml stop
+    - name: Print the cluster logs
+      if: ${{ always() }}
+      run: docker compose -f test/cluster/cassandra5/docker-compose.yml logs
+    - name: Remove cluster
+      run: docker compose -f test/cluster/cassandra5/docker-compose.yml down
```
Lorak-mmk, line 58: This will be done sequentially - first we perform tests on normal cassandra, then on beta.
Cassandra tests are already longest part of our CI. Maybe it would be better to create a separate workflow for this?

Lorak-mmk (changes requested):
Thanks for the contribution! It looks good, I left a few minor comments (and responded to a past discussion about a type of vector size).

I believe that the PR title and description are a bit imprecise.

The PR does not add CREATE or INSERT support for Vector type. CREATE is and always was possible, so was INSERT with hardcoded values.

The PR enables the driver to fetch and represent metadata containing Vector type, which is required to fully connect to cluster (because if the driver can't fetch metadata then it also won't fetch peer list).

scylla/src/transport/topology.rs
rukai: The limit of 2^13 = 8192 is taken from https://cassandra.apache.org/doc/latest/cassandra/reference/vector-data-type.html

However it can be observed that the documentation is incorrect.

From functionally testing, the highest value accepted when creating a table is `2**31 -1 = 2147483647`.
The reason would be that its the highest value representable by a signed 32 bit integer
From functional testing it can also be seen that negative numbers and `0` are rejected.

Do you want me to swap the value from an i32 to a u32?

scylla/src/transport/session_test.rs
```diff
@@ -2885,3 +2885,61 @@ async fn test_manual_primary_key_computation() {
...
+    assert_eq!(
+        columns["c"].type_,
+        CqlType::Vector {
+            type_: Box::new(CqlType::Native(NativeType::Text)),
+            dimensions: 2,
+        },
+    );
+
+    // TODO: Implement and test SELECT statements and bind values (`?`)
+}
```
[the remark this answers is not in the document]
rukai: I have split up the tests as requested.

.github/workflows/cassandra.yml
rukai: I have moved it into a separate workflow as requested.

rukai:
> I believe that the PR title and description are a bit imprecise.
>
> The PR does not add CREATE or INSERT support for Vector type. CREATE is and always was possible, so was INSERT with hardcoded values.
>
> The PR enables the driver to fetch and represent metadata containing Vector type, which is required to fully connect to cluster (because if the driver can't fetch metadata then it also won't fetch peer list).

I'm not quite following what you want.
To save both of us the trouble, can you please provide the exact commit message that you want.

rukai:
Hi, any news on this one?

wprzytula:
> Hi, any news on this one?

Sorry for the delay, both @Lorak-mmk and I have been on vacation. Now let @Lorak-mmk take a fresh look.

Lorak-mmk (commented):
The PR looks much better now. I am however reluctant to merge it before Cassandra version with Vector support is officially released. Mostly because:
- implementation could change before release
- cql specification seems incomplete w.r.t vectors. They are mentioned in section 5.2 of protocol v5 docs, so we know their serialized form. I don't however see anything about them in section 4.2.5.2 which describes prepared metadata.

Regarding the above: I totally forgot that prepared metadata is a place where we parse the type not from string but from binary form, sorry about that :(
This is also a reason why I favor a bit larger PRs - this would come up when trying to add support for selecting vectors using a prepared statement.

@wprzytula because spec is incomplete, and type used for size depends on it, I think we should wait until C* releases the feature. wdyt?

rukai:
cassandra 5 is released, any interest in landing this?

rukai:
oh right, theres some TODO's in here for when cassandra 5 releases. I'll address those

rukai:
The TODO's have been addressed: the cassandra 5 specific workflow and docker-compose.yaml have been removed, instead we just use the regular cassandra workflow (which is automatically version 5 now since it does not specify a version)

rukai:
oh right, a recent rust release has disallowed that kind of feature check, I'll look into alternatives

rukai:
I applied the same fix as https://github.com/scylladb/scylla-rust-driver/pull/1049 and CI is green now

wprzytula (dismissed):

scylla/src/utils/parse.rs
```diff
@@ -87,6 +87,21 @@ impl<'s> ParserState<'s> {
...
+    /// An error is returned if:
+    /// * The first character is not a digit or '-'
+    /// * The integer is larger than i32
+    pub(crate) fn parse_i32(self) -> ParseResult<(i32, Self)> {
+        let (digits, p) = self.take_while(|c| c.is_ascii_digit() || c == '-');
+        if let Ok(value) = digits.parse() {
+            Ok((value, p))
+        } else {
+            Err(p.error(ParseErrorCause::Expected("integer of max length 2**32")))
+        }
```
Lorak-mmk, line 102: Two minor issues:
- I don't think we expect integer of LENGTH 2**32 ;) The message should be something like "Expected 32-bit signed integer".
- Looking at other usages and Display definition, I think this should use `ParseErrorCause::Other`.

Lorak-mmk (changes requested):
Left one minor comment  - after you fix it we can merge I think, and fix potential issue when introducing ser/deser

scylla/src/utils/parse.rs
rukai: Thankyou for the review, I have addressed your feedback.

Lorak-mmk (approved):

PR #1026, opened 2024-07-01
errors: Replace `ParseError` with `CqlResponseParseError` in `frame::response` module

muzarski:

Depends on: https://github.com/scylladb/scylla-rust-driver/pull/1017

# Motivation
We ultimately want to get rid of the `ParseError`. This error type is overloaded with a lot of variants which do not provide much context to the users. In addition, a lot of errors are stringified.

This PR gets rid of `ParseError` from the `frame::response` module which is responsible for CQL response deserialization. This is the place where the usage of `ParseError` is abused the most.

# Changes
## Types of response
There are multiple types of responses. We introduce a new error type for each of them, e.g. `CqlResultParseError` or `CqlSupportedParseError`. I follow a bottom-to-top approach, starting from implementing an error types for CQL column type deserialization.

Each commit introduces a new error type, which is firstly encapsulated by `ParseError` (so the higher level functions returning `ParseError` can apply ? operator). Once we exhaust all of the error variants that can be returned by some function, we introduce a new error type which encapsulates error types introduced in previous commits. These error types are then moved out of the `ParseError` (if possible, sometimes the functions are not exclusive to `frame::response` module which is the only module addressed in this PR).

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -39,6 +39,8 @@ pub enum ParseError {
     SetKeyspaceParseError(#[from] SetKeyspaceParseError),
     #[error("Schema change event deserialization failed: {0}")]
     SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
+    #[error("Table spec deserialization failed: {0}")]
+    TableSpecParseError(#[from] TableSpecParseError),
     #[error("Low-level deserialization failed: {0}")]
     LowLevelDeserializationError(#[from] LowLevelDeserializationError),
     #[error("Could not serialize frame: {0}")]
@@ -89,6 +91,17 @@ pub enum SchemaChangeEventParseError {
     FunctionArgumentParseError(LowLevelDeserializationError),
     #[error("Unknown target of schema change: {0}")]
     UnknownTargetOfSchemaChange(String),
+}
+
+/// An error type returned when deserialization
+/// of table specification fails.
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum TableSpecParseError {
+    #[error("Malformed keyspace name: {0}")]
+    MalformedKeyspaceName(LowLevelDeserializationError),
+    #[error("Malformed table name: {0}")]
+    MalformedTableName(LowLevelDeserializationError),
 }
 
 /// A low level deserialization error.
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -41,6 +41,8 @@ pub enum ParseError {
     SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
     #[error("Table spec deserialization failed: {0}")]
     TableSpecParseError(#[from] TableSpecParseError),
+    #[error(transparent)]
+    TypeParseError(#[from] CqlTypeParseError),
     #[error("Low-level deserialization failed: {0}")]
     LowLevelDeserializationError(#[from] LowLevelDeserializationError),
     #[error("Could not serialize frame: {0}")]
@@ -51,8 +53,6 @@ pub enum ParseError {
     DeserializationError(#[from] DeserializationError),
     #[error(transparent)]
     IoError(#[from] std::io::Error),
-    #[error("type not yet implemented, id: {0}")]
-    TypeNotImplemented(u16),
     #[error(transparent)]
     SerializeValuesError(#[from] SerializeValuesError),
     #[error(transparent)]
@@ -102,6 +102,28 @@ pub enum TableSpecParseError {
     MalformedKeyspaceName(LowLevelDeserializationError),
     #[error("Malformed table name: {0}")]
     MalformedTableName(LowLevelDeserializationError),
+}
+
+/// An error type returned when deserialization of CQL type name fails.
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum CqlTypeParseError {
+    #[error("Malformed type id: {0}")]
+    TypeIdParseError(LowLevelDeserializationError),
+    #[error("Malformed custom type name: {0}")]
+    CustomTypeNameParseError(LowLevelDeserializationError),
+    #[error("Malformed name of UDT keyspace: {0}")]
+    UdtKeyspaceNameParseError(LowLevelDeserializationError),
+    #[error("Malformed UDT name: {0}")]
+    UdtNameParseError(LowLevelDeserializationError),
+    #[error("Malformed UDT fields count: {0}")]
+    UdtFieldsCountParseError(LowLevelDeserializationError),
+    #[error("Malformed UDT's field name: {0}")]
+    UdtFieldNameParseError(LowLevelDeserializationError),
+    #[error("Malformed tuple length: {0}")]
+    TupleLengthParseError(LowLevelDeserializationError),
+    #[error("CQL Type not yet implemented, id: {0}")]
+    TypeNotImplemented(u16),
 }
 
 /// A low level deserialization error.
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -80,7 +80,7 @@ pub enum LowLevelDeserializationError {
     InvalidValueLength(i32),
     #[error("Unknown consistency: {0}")]
     UnknownConsistency(#[from] TryFromPrimitiveError<u16>),
-    #[error("Invalid inet bytes length: {0}")]
+    #[error("Invalid inet bytes length: {0}. Accepted lengths are 4 and 16 bytes.")]
     InvalidInetLength(u8),
     #[error("UTF8 deserialization failed: {0}")]
     UTF8DeserializationError(#[from] std::str::Utf8Error),
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -36,6 +36,8 @@ pub enum FrameError {
 #[derive(Error, Debug)]
 pub enum ParseError {
     #[error(transparent)]
+    CqlErrorParseError(#[from] CqlErrorParseError),
+    #[error(transparent)]
     CqlAuthChallengeParseError(#[from] CqlAuthChallengeParseError),
     #[error(transparent)]
     CqlAuthSuccessParseError(#[from] CqlAuthSuccessParseError),
@@ -63,6 +65,22 @@ pub enum ParseError {
     SerializationError(#[from] SerializationError),
     #[error(transparent)]
     CqlTypeError(#[from] CqlTypeError),
+}
+
+/// An error type returned when deserialization of ERROR response fails.
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum CqlErrorParseError {
+    #[error("Malformed error code: {0}")]
+    ErrorCodeParseError(LowLevelDeserializationError),
+    #[error("Malformed error reason: {0}")]
+    ReasonParseError(LowLevelDeserializationError),
+    #[error("Malformed error field {field} of DB error {db_error}: {err}")]
+    MalformedErrorField {
+        db_error: &'static str,
+        field: &'static str,
+        err: LowLevelDeserializationError,
+    },
 }
 
 /// An error type returned when deserialization of AUTH_CHALLENGE response fails.
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -37,6 +37,8 @@ pub enum FrameError {
 pub enum ParseError {
     #[error("Set keyspace response deserialization failed: {0}")]
     SetKeyspaceParseError(#[from] SetKeyspaceParseError),
+    #[error("Schema change event deserialization failed: {0}")]
+    SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
     #[error("Low-level deserialization failed: {0}")]
     LowLevelDeserializationError(#[from] LowLevelDeserializationError),
     #[error("Could not serialize frame: {0}")]
@@ -62,6 +64,31 @@ pub enum ParseError {
 pub enum SetKeyspaceParseError {
     #[error("Malformed keyspace name: {0}")]
     MalformedKeyspaceName(#[from] LowLevelDeserializationError),
+}
+
+/// An error type returned when deserialization of
+/// `RESULT::Schema_change` response fails.
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum SchemaChangeEventParseError {
+    #[error("Malformed schema change type string: {0}")]
+    TypeOfChangeParseError(LowLevelDeserializationError),
+    #[error("Malformed schema change target string:: {0}")]
+    TargetTypeParseError(LowLevelDeserializationError),
+    #[error("Malformed name of keyspace affected by schema change: {0}")]
+    AffectedKeyspaceParseError(LowLevelDeserializationError),
+    #[error("Malformed name of the table affected by schema change: {0}")]
+    AffectedTableNameParseError(LowLevelDeserializationError),
+    #[error("Malformed name of the target affected by schema change: {0}")]
+    AffectedTargetNameParseError(LowLevelDeserializationError),
+    #[error(
+        "Malformed number of arguments of the function/aggregate affected by schema change: {0}"
+    )]
+    ArgumentCountParseError(LowLevelDeserializationError),
+    #[error("Malformed argument of the function/aggregate affected by schema change: {0}")]
+    FunctionArgumentParseError(LowLevelDeserializationError),
+    #[error("Unknown target of schema change: {0}")]
+    UnknownTargetOfSchemaChange(String),
 }
 
 /// A low level deserialization error.
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use super::TryFromPrimitiveError;
 use crate::cql_to_rust::CqlTypeError;
 use crate::frame::value::SerializeValuesError;
@@ -69,7 +71,7 @@ pub enum ParseError {
 
 /// An error type returned when deserialization of ERROR response fails.
 #[non_exhaustive]
-#[derive(Error, Debug)]
+#[derive(Error, Debug, Clone)]
 pub enum CqlErrorParseError {
     #[error("Malformed error code: {0}")]
     ErrorCodeParseError(LowLevelDeserializationError),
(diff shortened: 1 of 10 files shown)


wprzytula:
You haven't checked nor crossed out the _I added relevant tests for new features and bug fixes_ option. Do you believe any tests are relevant here?

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -108,6 +108,24 @@ pub enum PreparedParseError {
...
+    ResultMetadataParseError(#[from] ResultMetadataParseError),
+    #[error("Invalid result metadata, expected {expected_col_count} col specs, received {received_col_count}.")]
+    ColumnCountMismatch {
+        expected_col_count: usize,
+        received_col_count: usize,
+    },
+    #[error("Malformed rows count: {0}")]
+    RowsCountParseError(LowLevelDeserializationError),
+    #[error("Data deserialization failed: {0}")]
+    DataDeserializationError(#[from] DeserializationError),
```
wprzytula, line 232: Is it specific enough? Does the upper layer add information that it's about data received in RESULT::ROWS response?

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -63,6 +59,24 @@ pub enum ParseError {
...
+    #[error("Unknown RESULT response id: {0}")]
+    UnknownResultId(i32),
+    #[error("'Set_keyspace' response deserialization failed: {0}")]
+    SetKeyspaceParseError(#[from] SetKeyspaceParseError),
+    #[error("'Schema_change' response deserialization failed: {0}")]
+    SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
+    #[error("'Prepared' response deserialization failed: {0}")]
+    PreparedParseError(#[from] PreparedParseError),
+    #[error("'Rows' response deserialization failed: {0}")]
+    RowsParseError(#[from] RowsParseError),
```
wprzytula, line 140: OK, at this point I can see enough context added (this is a follow-up to my previous comment)

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -109,6 +111,18 @@ pub enum SchemaChangeEventParseError {
...
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum ClusterChangeEventParseError {
+    #[error("Malformed type of status change: {0}")]
+    TypeOfChangeParseError(LowLevelDeserializationError),
+    #[error("Malformed node address: {0}")]
+    NodeAddressParseError(LowLevelDeserializationError),
+    #[error("Unknown type of status change: {0}")]
+    UnknownTypeOfChange(String),
+}
```
wprzytula: You mention both `Topology` and `Status` in the docstring but only mention `status` in error messages. Why so?

wprzytula: At least `TypeOfChangeParseError` should have `cluster change` instead of `status change`.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -87,7 +85,24 @@ pub enum SetKeyspaceParseError {
...
+    #[error("Malformed event type string: {0}")]
+    EventTypeParseError(LowLevelDeserializationError),
+    #[error("Unknown event type: {0}")]
+    UnknownEventType(String),
+    #[error("Failed to deserialize schema change event: {0}")]
+    SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
+    #[error("Failed to deserialize topology change event: {0}")]
+    TopologyChangeEventParseError(ClusterChangeEventParseError),
+    #[error("Failed to deserialize status change event: {0}")]
+    StatusChangeEventParseError(ClusterChangeEventParseError),
```
wprzytula, line 164: Why did you decide to unify Topology and Status errors into common Cluster error? Could you summarise the reasons?

wprzytula: I think not all variants of ClusterChangeEventParseError apply to both Topology and Status.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -67,6 +69,18 @@ pub enum ParseError {
...
+/// An error type returned when deserialization of ERROR response fails.
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum CqlErrorParseError {
+    #[error("Malformed error code: {0}")]
+    ErrorCodeParseError(LowLevelDeserializationError),
+    #[error("Malformed error reason: {0}")]
+    ReasonParseError(LowLevelDeserializationError),
+    #[error(transparent)]
+    GenericDeserializationError(LowLevelDeserializationError),
```
wprzytula, line 90: Maybe more context?

wprzytula:
```suggestion
    #[error("Malformed error field {field} of error with code {code}: {err})]
    MalformedErrorField {
        code: i32,
        field: &'static str,
        err: LowLevelDeserializationError,
    },
```

wprzytula: As some errors are very rare, having plenty of information to identify the exact error and its problematic field can be invaluable in debugging.

wprzytula (changes requested):
Awesome changes!

Mainly some nits.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -66,6 +68,31 @@ pub enum SetKeyspaceParseError {
     StringDeserializationError(#[from] LowLevelDeserializationError),
 }
 
+/// An error type returned when deserialization of
+/// `RESULT::Schema_change` response fails.
```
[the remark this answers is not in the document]
muzarski: I don't know. I just based it on a `Table of Contents` of CQL protocol (https://cassandra.apache.org/_/native_protocol.html):
```
      4.2.5. RESULT
        4.2.5.1. Void
        4.2.5.2. Rows
        4.2.5.3. Set_keyspace
        4.2.5.4. Prepared
        4.2.5.5. Schema_change
```

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -41,8 +41,8 @@ pub enum ParseError {
     SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
     #[error("Failed to deserialize a table spec: {0}")]
     TableSpecParseError(#[from] TableSpecParseError),
-    #[error(transparent)]
-    TypeParseError(#[from] TypeParseError),
```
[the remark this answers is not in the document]
muzarski: Why not? The `CqlTypeParseError` is only used during column specs deserialization. No need to hold it in `ParseError` at this point. It is moved to `ColumnSpecsParseError`.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -106,6 +106,31 @@ pub enum TableSpecParseError {
...
+/// of table column specifications fails.
+#[non_exhaustive]
+#[derive(Error, Debug)]
+pub enum ColumnSpecsParseError {
+    #[error("Column spec deserialization failed, column index: {column_index}, error: {kind}")]
+    ColumnSpecParseError {
+        column_index: usize,
+        kind: ColumnSpecParseErrorKind,
+    },
+}
```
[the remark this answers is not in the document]
muzarski: 1. I introduced it as an enum to have it marked as `non_exhaustive` - so yes, for api extendability. WDYT @Lorak-mmk .

2. I'll come up with some better naming.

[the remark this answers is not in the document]
muzarski: Acutally, I don't think it needs to be an enum. `ColumnSpecParseErrorKind` is already an enum.

scylla-cql/src/frame/frame_errors.rs
muzarski: All of the variants are shared between status and topology changes. I decided to unify them to reduce the copy-paste.

scylla-cql/src/frame/frame_errors.rs
muzarski: Ok, I agree. I'll add some context.

scylla-cql/src/frame/response/result.rs
```diff
@@ -865,7 +865,7 @@ fn deser_prepared(buf: &mut &[u8]) -> StdResult<Prepared, ParseError> {
 }
 
 #[allow(clippy::unnecessary_wraps)]
-fn deser_schema_change(buf: &mut &[u8]) -> StdResult<SchemaChange, ParseError> {
+fn deser_schema_change(buf: &mut &[u8]) -> StdResult<SchemaChange, SchemaChangeEventParseError> {
     Ok(SchemaChange {
         event: SchemaChangeEvent::deserialize(buf)?,
     })
```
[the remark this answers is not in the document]
muzarski: I removed it in a separate commit.

muzarski:
v2:
- rebased on current version of `low_lever_deser_errors`
- addressed review comments

wprzytula (commented):
Awesome work! This is a huge step in direction of clean and informative error handling.

scylla-cql/src/frame/request/auth_response.rs
```diff
@@ -12,6 +12,6 @@ impl SerializableRequest for AuthResponse {
     const OPCODE: RequestOpcode = RequestOpcode::AuthResponse;
 
     fn serialize(&self, buf: &mut Vec<u8>) -> Result<(), ParseError> {
-        write_bytes_opt(self.response.as_ref(), buf)
+        Ok(write_bytes_opt(self.response.as_ref(), buf)?)
     }
```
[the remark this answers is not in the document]
Lorak-mmk: How is the former more readable?!!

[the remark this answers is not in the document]
wprzytula: Well, it reads from left to right cleanly, whereas expression of `Ok(expr)` gives the impression that the result is always `Ok()`, but then you read the contents to find a question mark inside that performs an early return with `Err()`. I strongly prefer the monadic operations: you have a Result, then you map its error by wrapping it in an inferred supertype, and return that.

muzarski:
Rebased on main

scylla-cql/src/frame/request/auth_response.rs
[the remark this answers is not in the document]
muzarski: Yeah, I'll leave it as is. I believe we already made use of `Ok(expr?)` many times throughout the code.

muzarski:
v2.1: addressed nit review comments

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -61,6 +57,24 @@ pub enum ParseError {
...
+#[derive(Error, Debug)]
+pub enum CqlResultParseError {
+    #[error("Malformed RESULT response id: {0}")]
+    ResultIdParseError(LowLevelDeserializationError),
+    #[error("Unknown RESULT response id: {0}")]
+    UnknownResultId(i32),
+    #[error("'Set_keyspace' response deserialization failed: {0}")]
+    SetKeyspaceParseError(#[from] SetKeyspaceParseError),
+    #[error("'Schema_change' response deserialization failed: {0}")]
+    SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
```
Lorak-mmk, line 138: @wprzytula @muzarski Do you think that this variant should be present here?
CQL events are always sent on a separate channel (-1) which handles only that - so there should be no way for a user to get this error

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -39,6 +39,8 @@ pub enum ParseError {
     CqlResultParseError(#[from] CqlResultParseError),
     #[error("Schema change event deserialization failed: {0}")]
     SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
+    #[error("Failed to deserialize cluster change event: {0}")]
+    ClusterChangeEventParseError(#[from] ClusterChangeEventParseError),
```
Lorak-mmk: And here the new event parse error is only added to ParseError, not `CqlResultParseError` - I think it should be this way for all errors.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -85,7 +83,24 @@ pub enum SetKeyspaceParseError {
...
+    #[error("Malformed event type string: {0}")]
+    EventTypeParseError(LowLevelDeserializationError),
+    #[error("Unknown event type: {0}")]
+    UnknownEventType(String),
+    #[error("Failed to deserialize schema change event: {0}")]
+    SchemaChangeEventParseError(#[from] SchemaChangeEventParseError),
+    #[error("Failed to deserialize topology change event: {0}")]
+    TopologyChangeEventParseError(ClusterChangeEventParseError),
+    #[error("Failed to deserialize status change event: {0}")]
+    StatusChangeEventParseError(ClusterChangeEventParseError),
```
Lorak-mmk: Ok, I see you already added this type. Then why is there still `SchemaChangeEventParseError` variant in `CqlResultParseError`?

scylla-cql/src/frame/frame_errors.rs
Lorak-mmk, line 1: Regarding commit `f_errors: derive Clone `: can you explain why we need `Clone` on our error types?

muzarski: It's because `QueryError` and `SessionError` are `Clone`'able. I'll add this to commit message.

Lorak-mmk: I'm asking why QueryError and SessionError need to be cloneable

scylla-cql/src/frame/frame_errors.rs
Lorak-mmk: And yes, it may be that schema change event has the same binary form as schema change response - but they are different things anyway and we should not mix them up.

scylla-cql/src/frame/frame_errors.rs
muzarski: I believe the main reason is this piece of code from `Connection::router`:
```rust
        for (_, handler) in response_handlers {
            // Ignore sending error, request was dropped
            let _ = handler.response_sender.send(Err(error.clone()));
        }

        // If someone is listening for connection errors notify them
        let _ = error_sender.send(error);
```

muzarski: There are also multiple places in `history` module where we make use of it.

Lorak-mmk: The later part (`let _ = error_sender.send(error);`) seems to be a mechanism used only 1 test  (at least I can't find any other usages): `connection_is_closed_on_no_response_to_keepalives`.
I don't like that we push this channels trough multiple layers of the driver only to be used by 1 test.

And I don't understand what is the point of using it in this test:
```
        for _ in 0..3 {
            tokio::time::sleep(Duration::from_millis(500)).await;
            conn.query_single_page("SELECT host_id FROM system.local")
                .await
                .unwrap();
        }
        // As everything is normal, no error should have been reported.
        assert_matches!(
            error_receiver.try_recv(),
            Err(tokio::sync::oneshot::error::TryRecvError::Empty)
        );
```

If there were errors then the queries would fail, so unwrap would panic. This assert seems to check nothing.

cc: @wprzytula

Lorak-mmk: > There are also multiple places in `history` module where we make use of it.

That makes way more sense and I think is good enough reason for the Clone. Could you explain this in the commit message?

scylla-cql/src/frame/frame_errors.rs
muzarski: I think that you meant `CqlResponseParseError::CqlEventParseError` variant. I think that you are right. If so, then I'll remove this variant from public API.

scylla-cql/src/frame/frame_errors.rs
muzarski: I changed the variant name to `SchemaChangeParseError`

scylla-cql/src/frame/frame_errors.rs
wprzytula: > And I don't understand what is the point of using it in this test:
> 
> ```rust
>         for _ in 0..3 {
>             tokio::time::sleep(Duration::from_millis(500)).await;
>             conn.query_single_page("SELECT host_id FROM system.local")
>                 .await
>                 .unwrap();
>         }
>         // As everything is normal, no error should have been reported.
>         assert_matches!(
>             error_receiver.try_recv(),
>             Err(tokio::sync::oneshot::error::TryRecvError::Empty)
>         );
> ```

Your reasoning is that you want to get rid of the channel. However, assuming it's needed for `PoolRefiller` anyway, then it makes sense to assert that it indeed returns no errors at any point of a valid query.

Lorak-mmk: > > The later part (`let _ = error_sender.send(error);`) seems to be a mechanism used only 1 test (at least I can't find any other usages): `connection_is_closed_on_no_response_to_keepalives`.
> > I don't like that we push this channels trough multiple layers of the driver only to be used by 1 test.
> 
> `PoolRefiller` seems to be using this channel at least in `wait_for_error` function.

You're right, I missed it, sorry.

scylla-cql/src/frame/frame_errors.rs
wprzytula: I see no easy way to decouple event handling from ordinary response handling. Thus, we need a common error type for both.

muzarski: It's true. Now the question is whether we want to introduce new public error type excluding this specific variant? I'm not sure if there is much sense to do so.

muzarski: I'd leave it as is for now.

wprzytula: I see your point (why would handling events have its own error variants in QueryError), but the size of such refactor makes it out of this PR's scope IMO.

wprzytula:
ref: #1056

scylla-cql/src/frame/frame_errors.rs
Lorak-mmk: It's not only about Events. `CqlResponseParseError` has the following variants:
```
    CqlErrorParseError
    CqlAuthChallengeParseError
    CqlAuthSuccessParseError
    CqlAuthenticateParseError
    CqlSupportedParseError
    CqlEventParseError
    CqlResultParseError
```
If I'm not mistaken only first and last are possible for the user to encounter in QueryError (only possible in NewSessionError). I understand that it may increase the amount of changes, but I think we should avoid creating such types anyway. It makes it much harder for users who want to handle errors in a different way than just printing them.

wprzytula: OK, this is quite convincing. I suggest that @muzarski analyse all possible error routes in the driver and try to redesign the error hierarchy so that user-facing error types don't contain variants that the user is never supposed to see. If needed, we can introduce two layers of error handling:
- a lower-lever one, resembling one introduced in this PR,
- a higher-level one, restricting variants to those that a user really might see. It should be organised with the user's perspective in mind rather than reflect the driver's inner architecture.

Lorak-mmk (approved):

Issue #1035, opened 2024-07-09
batch: counter batches are not working properly


muzarski:
Here is the test that reproduces the issue:
```rust
#[tokio::test]
async fn test_counter_batch() {
    setup_tracing();
    let session = Arc::new(create_new_session_builder().build().await.unwrap());
    let ks = unique_keyspace_name();

    session.query(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks), &[]).await.unwrap();
    session
        .query(
            format!(
                "CREATE TABLE IF NOT EXISTS {}.t_batch (key int PRIMARY KEY, value counter)",
                ks
            ),
            &[],
        )
        .await
        .unwrap();

    let q = format!("UPDATE {}.t_batch SET value = value + ? WHERE key = ?", ks);

    let mut my_batch = Batch::new(BatchType::Counter);
    assert_eq!(BatchType::Counter, my_batch.get_type());
    my_batch.append_statement(&q[..]);
    my_batch.append_statement(&q[..]);
    my_batch.append_statement(&q[..]);

    session
        .batch(
            &my_batch,
            ((Counter(1), 2), (Counter(1), 2), (Counter(1), 2)),
        )
        .await
        .unwrap();
}
```

This test fails with a server error:
```
---- transport::session_test::test_my_batch stdout ----
Unique name: test_rust_1720524333_0
thread 'transport::session_test::test_counter_batch' panicked at scylla/src/transport/session_test.rs:404:10:
called `Result::unwrap()` on an `Err` value: DbError(Invalid, "Cannot include a counter statement in a logged batch")
```

The driver sends a LOGGED batch, even though the user specified the batch type to be a COUNTER.

The issue was discovered by a faling integration test in cpp-rust-driver's CI after updating rust-driver's version to 0.13: https://github.com/scylladb/cpp-rust-driver/pull/138.

muzarski:
I found the culprit: https://github.com/scylladb/scylla-rust-driver/commit/98b75421fe8ff1cab735297ad6ed3c5bbd177367. I'll investigate it further and open a PR fixing the issue.

PR #1038, opened 2024-07-10
conn: copy a batch type in prepare_batch

muzarski:
Fix: https://github.com/scylladb/scylla-rust-driver/issues/1035

Previously, there was a bug in prepare_batch function. We wouldn't inherit a BatchType from the `initial_batch` - in result, we would always send LOGGED batches (since LOGGED is a default batch type).

This would result in server errors when including a counter statements in a LOGGED batch. I added a test case which checks that it works correctly now, and the driver sends a COUNTER batch when user asks to.

## Pre-review checklist


scylla/src/transport/connection.rs
```diff
@@ -914,7 +914,7 @@ impl Connection {
             prepared_queries.insert(query, prepared);
         }
 
-        let mut batch: Cow<Batch> = Cow::Owned(Default::default());
+        let mut batch: Cow<Batch> = Cow::Owned(Batch::new(init_batch.get_type()));
```
Lorak-mmk, line 917: Are we sure that there are no more config values that need to be copied?
Maybe a bit cleaner solution would to to introduce `new_from` method on Batch?

muzarski:
v2: Added a `Batch::new_from` constructor which inherits a type and config values from existing batch.

wprzytula (approved):

Lorak-mmk (approved):

Issue #1028, opened 2024-07-09
Expose setters for STARTUP options


wprzytula:
1. cpp-rust-driver needs to set DRIVER_NAME and DRIVER_VERSION according to its versioning instead of Rust driver's values, so that we can tell apart the bare Rust driver and cpp-rust-driver running on top of it.
2. cpp-driver has API to configure APPLICATION_NAME, APPLICATION_VERSION, CLIENT_ID, so they should be supported as well.

### Suggestion
Let's group all the configurable STARTUP options in one struct, for better structure.

Migrated to Jira: [DRIVER-236](https://scylladb.atlassian.net/browse/DRIVER-236)

[DRIVER-236]: https://scylladb.atlassian.net/browse/DRIVER-236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1039, opened 2024-07-10
connection: expose setters for STARTUP options

wprzytula:
Resolves: #1028

## Motivation
See #1028.

# What's done

### Reduce allocations by using `Cow<str>` for SUPPORTED/STARTUP options
- `types::write_string_map` was genericised to accept `HashMap`s with keys and values being any form of strings, including `Cow<str>`.
- as majority of keys and significant number of values of STARTUP options are known at compile time and thus statically allocated, `Startup` request now holds `HashMap<Cow<str>, Cow<str>`  instead of `HashMap<String, String>` to reduce needless allocations.

### refactors for readability
- `setup_tcp_keepalive()` was extracted from `Connection::new()` to reduce clutter.
- keys of SUPPORTED/STARTUP options were extracted as constants to `options` module to have them in one place and avoid using plain string literals in code.
- resolved a dated TODO about whether logic in `open_named_connection` should be part of `Connection::new` - answered the question, justified the answer and added docstrings to convey that answer.
- added many comments to `open_named_connection`, as the function is big and crucial for driver operation, so future readers will benefit.

### Compression-related small enhancements
- `Compression` name is now no longer allocated, as its name is a `&'static str` anyway.
- warning is now emitted if requested Compression is not supported; before, it would silently fall back to no compression.

## `SelfIdentity` struct and API
- grouped self-identity-related STARTUP options in `SelfIdentity` struct. It includes:
  - `DRIVER_NAME`
  - `DRIVER_VERSION`
  - `APPLICATION_NAME`
  - `APPLICATION_VERSION`
  - `CLIENT_ID`
  
  `DRIVER_NAME` and `DRIVER_VERSION` are set by default to _"ScyllaDB Rust Driver"_ (previously _"scylla-rust-driver"_, changed to be consistent with _"ScyllaDB Python Driver"_) and Rust driver's package version as declared by Cargo. The remaining ones are unset by default.

- exposed `SelfIdentity`'s API in `SessionBuilder`, and added a docstring with an example usage.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/request/options.rs
+++ b/scylla-cql/src/frame/request/options.rs
@@ -20,6 +20,9 @@ pub const COMPRESSION: &str = "COMPRESSION";
 pub const CQL_VERSION: &str = "CQL_VERSION";
 pub const DRIVER_NAME: &str = "DRIVER_NAME";
 pub const DRIVER_VERSION: &str = "DRIVER_VERSION";
+pub const APPLICATION_NAME: &str = "APPLICATION_NAME";
+pub const APPLICATION_VERSION: &str = "APPLICATION_VERSION";
+pub const CLIENT_ID: &str = "CLIENT_ID";
 
 /* Value names for options in SUPPORTED/STARTUP */
 pub const DEFAULT_CQL_PROTOCOL_VERSION: &str = "4.0.0";
--- a/scylla-cql/src/frame/request/options.rs
+++ b/scylla-cql/src/frame/request/options.rs
@@ -20,3 +20,8 @@ pub const COMPRESSION: &str = "COMPRESSION";
 pub const CQL_VERSION: &str = "CQL_VERSION";
 pub const DRIVER_NAME: &str = "DRIVER_NAME";
 pub const DRIVER_VERSION: &str = "DRIVER_VERSION";
+
+/* Value names for options in SUPPORTED/STARTUP */
+pub const DEFAULT_CQL_PROTOCOL_VERSION: &str = "4.0.0";
+pub const DEFAULT_DRIVER_NAME: &str = "scylla-rust-driver";
+pub const DEFAULT_DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
--- a/scylla-cql/src/frame/request/options.rs
+++ b/scylla-cql/src/frame/request/options.rs
@@ -11,3 +11,12 @@ impl SerializableRequest for Options {
         Ok(())
     }
 }
+
+/* Key names for options in SUPPORTED/STARTUP */
+pub const SCYLLA_SHARD_AWARE_PORT: &str = "SCYLLA_SHARD_AWARE_PORT";
+pub const SCYLLA_SHARD_AWARE_PORT_SSL: &str = "SCYLLA_SHARD_AWARE_PORT_SSL";
+
+pub const COMPRESSION: &str = "COMPRESSION";
+pub const CQL_VERSION: &str = "CQL_VERSION";
+pub const DRIVER_NAME: &str = "DRIVER_NAME";
+pub const DRIVER_VERSION: &str = "DRIVER_VERSION";
--- a/scylla-cql/src/frame/request/options.rs
+++ b/scylla-cql/src/frame/request/options.rs
@@ -23,5 +23,5 @@ pub const DRIVER_VERSION: &str = "DRIVER_VERSION";
 
 /* Value names for options in SUPPORTED/STARTUP */
 pub const DEFAULT_CQL_PROTOCOL_VERSION: &str = "4.0.0";
-pub const DEFAULT_DRIVER_NAME: &str = "scylla-rust-driver";
+pub const DEFAULT_DRIVER_NAME: &str = "ScyllaDB Rust Driver";
 pub const DEFAULT_DRIVER_VERSION: &str = env!("CARGO_PKG_VERSION");
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -524,7 +524,7 @@ impl Connection {
 
     pub(crate) async fn startup(
         &self,
-        options: HashMap<String, String>,
+        options: HashMap<Cow<'_, str>, Cow<'_, str>>,
     ) -> Result<Response, QueryError> {
         Ok(self
             .send_request(&request::Startup { options }, false, false, None)
@@ -1517,8 +1517,8 @@ pub(crate) async fn open_connection(
         addr,
         source_port,
         config,
-        Some("scylla-rust-driver".to_string()),
-        option_env!("CARGO_PKG_VERSION").map(|v| v.to_string()),
+        Some("scylla-rust-driver"),
+        option_env!("CARGO_PKG_VERSION"),
     )
     .await
 }
@@ -1527,8 +1527,8 @@ pub(crate) async fn open_named_connection(
     addr: SocketAddr,
     source_port: Option<u16>,
     config: &ConnectionConfig,
-    driver_name: Option<String>,
-    driver_version: Option<String>,
+    driver_name: Option<&str>,
+    driver_version: Option<&str>,
 ) -> Result<(Connection, ErrorReceiver), QueryError> {
     // TODO: shouldn't all this logic be in Connection::new?
     let (mut connection, error_receiver) =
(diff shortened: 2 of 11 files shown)


muzarski (changes requested):
Good job on adding comments to `connection` module. I always feel like I'm lost trying to understand the logic that resides there.

The commit message mentions:
```
Settable identity include:
- DRIVER_NAME
- DRIVER_VERSION
- APPLICATION_NAME
- APPLICATION_VERSION
- CLIENT_ID

All five are visible in Cassandra's `system_views.clients` in
`client_options` column.

DRIVER_NAME & DRIVER_VERSION are visible in ScyllaDB's `system.clients`.
```

Maybe we should add some test cases that verify it? Especially for the newly introduced STARTUP options.

Lorak-mmk:
Where does `APPLICATION_NAME` and `APPLICATION_VERSION` come from?
PR description reads like those attributes are as supported and known as `DRIVER_NAME` / `DRIVER_VERSION`, but:
- there are no columns in system.clients dedicated to those attributes
-  `APPLICATION_NAME` / `APPLICATION_VERSION` are not present in Scylla's source code at all
- only 2 results in Cassandra source is some example in docs: https://github.com/apache/cassandra/blob/d3cbf9c1f72057d2a5da9df8ed567d20cd272931/doc/modules/cassandra/pages/managing/operating/virtualtables.adoc?plain=1#L218 . `APPLICATION_NAME` and `APPLICATION_VERSION` appears in `client_options` which IIUC is an arbitrary dict where client can send any keys.
- driver variables are mentioned in protocol v5 (https://github.com/apache/cassandra/blob/d3cbf9c1f72057d2a5da9df8ed567d20cd272931/doc/native_protocol_v5.spec#L480 ), application variables are not.

Please explain in the PR description (and maybe commit message too):
- Why support those specific keys instead of arbitrary keys
- The difference in support between driver variables and application variables
- Where those application variables come from.

scylla-cql/src/frame/request/options.rs
```diff
@@ -11,3 +11,19 @@ impl SerializableRequest for Options {
...
+pub const COMPRESSION: &str = "COMPRESSION";
+pub const CQL_VERSION: &str = "CQL_VERSION";
+pub const DRIVER_NAME: &str = "DRIVER_NAME";
+pub const DRIVER_VERSION: &str = "DRIVER_VERSION";
+
+/* Value names for options in SUPPORTED/STARTUP */
+pub const DEFAULT_CQL_PROTOCOL_VERSION: &str = "4.0.0";
+pub const DEFAULT_DRIVER_NAME: &str = "scylla-rust-driver";
+// TODO: consider using `env!`, which raises compile-time error if constant is not defined.
+// I think the constant should be always defined, no need to allow for it to be undefined.
```
Lorak-mmk, line 31: Docs agree with you: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates
If the version is not in the manifest then the variable will be set to an empty string. I think you can use `env!`

scylla/src/transport/connection.rs
```diff
@@ -1617,6 +1617,10 @@ pub(crate) async fn open_named_connection(
             );
         } else {
             // Fall back to no compression
+            tracing::warn!(
+                "Requested compression <{}> is not supported by the cluster. Falling back to no compression",
+                compression_str
+            );
```
Lorak-mmk, line 1744: when can this happen?

scylla/src/transport/session_builder.rs
```diff
@@ -942,6 +943,37 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
         self.config.cluster_metadata_refresh_interval = interval;
         self
     }
+
+    /// Set the custom identity of the driver/application/instance,
+    /// to be sent as options in STARTUP message.
+    ///
+    /// By default driver name and version are sent;
+    /// application name and version and client id are not sent.
+    /// # Example
```
Lorak-mmk, line 952: nit: maybe add a newline before `# Example`?

scylla/src/transport/connection.rs
```diff
@@ -1515,6 +1517,11 @@ async fn maybe_translated_addr(
...
 }
 
+/// Opens a connection and performs its setup on CQL level:
+/// - performs OPTIONS/STARTUP handshake (chooses desired connections options);
+/// - registers for all event types using REGISTER request (if this is control connection).
+///
+/// At the beginning, translates node's address, if it is subject to address translation.
 pub(crate) async fn open_connection(
     endpoint: UntranslatedEndpoint,
     source_port: Option<u16>,
```
Lorak-mmk, line 1661: I know that `open_named_connection` is later removed and the name is taken from identity, so this is not important, but: open_connection always sets name to `scylla-rust-driver` and it is the only caller of `open_named_connection`, so driver name will always be `scylla-rust-driver` - however previous commit claimed to change default name. How is it changed if it is never used? Idk, maybe I'm reading the code incorrectly.

scylla/src/transport/connection.rs
```diff
@@ -1578,22 +1583,29 @@ pub(crate) async fn open_named_connection(
...
         shard_aware_port,
         protocol_features,
     };
     connection.set_features(features);
 
+    /* Prepare options that the driver opts-in in STARTUP frame. */
+    let mut options = HashMap::new();
+    protocol_features.add_startup_options(&mut options);
+
+    // The only CQL protocol version supported by the driver.
```
Lorak-mmk: This change is not only adding comments. But this reordering seems harmelss, so let's ignore it.

Lorak-mmk (changes requested):

wprzytula:
Rebased on main.

scylla/src/transport/connection.rs
wprzytula: That's a correct observation. I'll adjust the code correspondingly.

scylla/src/transport/connection.rs
wprzytula: This may happen if SUPPORTED message sent by a node does not contain the particular compression string in the value for the `"COMPRESSION"` key.

wprzytula:
> Where does `APPLICATION_NAME` and `APPLICATION_VERSION` come from? PR description reads like those attributes are as supported and known as `DRIVER_NAME` / `DRIVER_VERSION`, but:
> 
>     * there are no columns in system.clients dedicated to those attributes
> 
>     * `APPLICATION_NAME` / `APPLICATION_VERSION` are not present in Scylla's source code at all
> 
>     * only 2 results in Cassandra source is some example in docs: https://github.com/apache/cassandra/blob/d3cbf9c1f72057d2a5da9df8ed567d20cd272931/doc/modules/cassandra/pages/managing/operating/virtualtables.adoc?plain=1#L218 . `APPLICATION_NAME` and `APPLICATION_VERSION` appears in `client_options` which IIUC is an arbitrary dict where client can send any keys.
> 
>     * driver variables are mentioned in protocol v5 (https://github.com/apache/cassandra/blob/d3cbf9c1f72057d2a5da9df8ed567d20cd272931/doc/native_protocol_v5.spec#L480 ), application variables are not.
> 
> 
> Please explain in the PR description (and maybe commit message too):
> 
>     * Why support those specific keys instead of arbitrary keys
> 
>     * The difference in support between driver variables and application variables
> 
>     * Where those application variables come from.

They come directly from CPP driver. See https://github.com/scylladb/cpp-driver/blob/fa0f27069a625057984d1fa58f434ea99b86c83f/include/cassandra.h#L2916.
As we want to support as big subset of its API as possible in cpp-rust-driver, I decided to expose API for setting those particular key-value pairs, similarly to what cpp-driver does, and not an API to set arbitrary key-value pairs.

Allowing users to set arbitrary options could break the driver by overwriting options that bear special meaning, e.g. the shard-aware port. Therefore, I'm against such liberal API. OTOH, we need to expose APPLICATION_NAME, APPLICATION_VERSION and CLIENT_ID for cpp-rust-driver.

Does this convince you @Lorak-mmk?

Lorak-mmk:
> > Where does `APPLICATION_NAME` and `APPLICATION_VERSION` come from? PR description reads like those attributes are as supported and known as `DRIVER_NAME` / `DRIVER_VERSION`, but:
> > ```
> > * there are no columns in system.clients dedicated to those attributes
> > 
> > * `APPLICATION_NAME` / `APPLICATION_VERSION` are not present in Scylla's source code at all
> > 
> > * only 2 results in Cassandra source is some example in docs: https://github.com/apache/cassandra/blob/d3cbf9c1f72057d2a5da9df8ed567d20cd272931/doc/modules/cassandra/pages/managing/operating/virtualtables.adoc?plain=1#L218 . `APPLICATION_NAME` and `APPLICATION_VERSION` appears in `client_options` which IIUC is an arbitrary dict where client can send any keys.
> > 
> > * driver variables are mentioned in protocol v5 (https://github.com/apache/cassandra/blob/d3cbf9c1f72057d2a5da9df8ed567d20cd272931/doc/native_protocol_v5.spec#L480 ), application variables are not.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please explain in the PR description (and maybe commit message too):
> > ```
> > * Why support those specific keys instead of arbitrary keys
> > 
> > * The difference in support between driver variables and application variables
> > 
> > * Where those application variables come from.
> > ```
> 
> They come directly from CPP driver. See https://github.com/scylladb/cpp-driver/blob/fa0f27069a625057984d1fa58f434ea99b86c83f/include/cassandra.h#L2916. As we want to support as big subset of its API as possible in cpp-rust-driver, I decided to expose API for setting those particular key-value pairs, similarly to what cpp-driver does, and not an API to set arbitrary key-value pairs.
> 
> Allowing users to set arbitrary options could break the driver by overwriting options that bear special meaning, e.g. the shard-aware port. Therefore, I'm against such liberal API. OTOH, we need to expose APPLICATION_NAME, APPLICATION_VERSION and CLIENT_ID for cpp-rust-driver.
> 
> Does this convince you @Lorak-mmk?

It does. Please include this explanation in commit message and in a comment.

wprzytula:
v2:
- addressed review suggestions;
- made `SelfIdentity` more convenient by accepting `impl Into<Cow<...>>` instead of only `Cow<...>`;
- wrote a test using proxy that asserts proper self identity being set in STARTUP frame.

muzarski (approved):

Lorak-mmk (approved):

Issue #1023, opened 2024-06-26
Don't fail metadata fetch when encountering invalid peer


Lorak-mmk:
When the row in system.peers is invalid (e.g. has rpc_address == null) the driver will encounter an error when trying to convert the row into Rust struct and this error will cause whole metadata fetch to fail.
Such invalid rows may happen in Scylla (and judging by the code in other drivers - in Cassandra too) sometimes, but should be a transient condition.
The correct behavior would be to log info about invalid row and skip it. It seems that all other driver (at least the ones I checked - Python, Java 3/4, Go) behave in this way.
As long as Scylla send topology change event after a corresponding node is fully available (so its row is correct in system.peers) it should not prevent us from implementing https://github.com/scylladb/scylla-rust-driver/issues/1008

vponomaryov:
Our latte stress command failed to start in multi-dc setup facing this problem with provided `preferred region` parameter:

```
[2m2024-07-28T09:14:31.177595Z[0m [33m WARN[0m [2mscylla::transport::topology[0m[2m:[0m \
  Initial metadata read failed, proceeding with metadata consisting only of the initial peer list and dummy tokens. \
    This might result in suboptimal performance and schema information not being available. [3merror[0m[2m=[0m\
      ProtocolError("system.peers or system.local has invalid column type")
info: Loading workload script /scylla-qa-internal/custom_d1/workload2/latte/custom_d1_workload2.rn...
[2m2024-07-28T09:14:31.180326Z[0m [33m WARN[0m [2mscylla::transport::load_balancing::default[0m[2m:[0m \
  Datacenter specified as the preferred one (us-east1-s1) does not exist!
info: Connecting to ["10.142.0.59", "10.142.0.121", "10.142.0.125", "10.142.0.131", "10.142.0.142", "10.142.0.154", "10.142.0.159", "10.142.0.160", "10.142.0.5"]... 
[2m2024-07-28T09:14:31.180348Z[0m [33m WARN[0m [2mscylla::transport::load_balancing::default[0m[2m:[0m \
  Datacenter specified as the preferred one (us-east1-s1) does not exist!
error: Cassandra error: Failed to execute query "SELECT cluster_name, release_version FROM system.local" with params []: \
  Protocol Error: Empty query plan - driver bug!
[2m2024-07-28T09:14:31.180354Z[0m [31mERROR[0m [2mscylla::transport::load_balancing::plan[0m[2m:[0m \
[...]

Issue #1021, opened 2024-06-24
Be more verbose when failing to fetch metadata 


fee-mendes:
As seen in https://github.com/scylladb/scylladb/issues/19107, sometimes the driver may fail to fetch metadata from its peers and print messages in the logs such as:

```rust
2024-06-24T19:05:16.729306Z  WARN scylla::transport::topology: Failed to establish control connection and fetch metadata on all known peers. Falling back to initial contact points.
2024-06-24T19:05:16.729347Z  WARN scylla::transport::topology: Failed to fetch metadata using current control connection control_connection_address="172.31.29.124:9042" error=Protocol Error: system.peers or system.local has invalid column type
2024-06-24T19:05:16.733118Z ERROR scylla::transport::topology: Could not fetch metadata error=Protocol Error: system.peers or system.local has invalid column type
```

It is wise to know exactly which invalid data was found, such as from which `host_id`, which column was found to be invalid and which value was returned. 

Similar work has been done for the Python driver in https://github.com/scylladb/python-driver/issues/303

mykaul:
The root cause might be due to https://github.com/scylladb/scylla-rust-driver/issues/1023 btw.

dkropachev:
@fee-mendes, original problem is solved in https://github.com/scylladb/scylla-rust-driver/issues/1023, and verbose logging is going to be implemented at https://github.com/scylladb/scylla-rust-driver/issues/462

wprzytula:
This has been included in release 0.13.2.

PR #1045, opened 2024-07-29
topology: skip invalid peers instead of failing

wprzytula:
## Motivation

It appears that in some clusters there sometimes emerges a transient state that there are some incomplete/invalid peer entries in system tables. Under such circumstances, the driver would throw an error and fail the metadata refresh altogether, eventually leading to critical conditions such as no hosts to query.

## Solution

To fix this in line with how other drivers approach this, a warning is emitted upon invalid peer entry read, and the entry is skipped with no error returned.

Fixes: #1023
Fixes: #235

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/topology.rs
+++ b/scylla/src/transport/topology.rs
@@ -819,10 +819,16 @@ async fn query_peers(conn: &Arc<Connection>, connect_port: u16) -> Result<Vec<Pe
 
     let translated_peers_futures = untranslated_rows.map(|row_result| async {
         let (source, raw_row) = row_result?;
-        let row = raw_row.into_typed().map_err(|_| {
-            QueryError::ProtocolError("system.peers or system.local has invalid column type")
-        })?;
-        create_peer_from_row(source, row, local_address).await
+        match raw_row.into_typed() {
+            Ok(row) => create_peer_from_row(source, row, local_address).await,
+            Err(err) => {
+                warn!(
+                    "system.peers or system.local has an invalid row, skipping it: {}",
+                    err
+                );
+                Ok(None)
+            }
+        }
     });
 
     let peers = translated_peers_futures


scylla/src/transport/topology.rs
```diff
@@ -819,10 +819,16 @@ async fn query_peers(conn: &Arc<Connection>, connect_port: u16) -> Result<Vec<Pe
...
-        let row = raw_row.into_typed().map_err(|_| {
-            QueryError::ProtocolError("system.peers or system.local has invalid column type")
-        })?;
-        create_peer_from_row(source, row, local_address).await
+        match raw_row.into_typed() {
+            Ok(row) => create_peer_from_row(source, row, local_address).await,
+            Err(err) => {
+                warn!(
+                    "system.peers or system.local has an invalid row, skipping it: {}",
+                    err
```
fee-mendes, line 827: I really would like to see https://github.com/scylladb/scylla-rust-driver/issues/1021 here. IIUC this would most always return `system.peers or system.local has invalid column type`, no? 

Either way, this LGTM as it addresses the skip part.

fee-mendes (approved):

scylla/src/transport/topology.rs
wprzytula: The `warn!` prints a pretty message contained by `err`, so we should get the deepest possible explanation of what went wrong.

vponomaryov (approved):
LGTM

Lorak-mmk:
You can mark https://github.com/scylladb/scylla-rust-driver/issues/235 as fixed by this.

wprzytula:
Rebased on main.

vponomaryov:
@wprzytula , @Lorak-mmk 

Do you plan to make a new tag any time soon after merging this bugfix?

wprzytula:
> @wprzytula , @Lorak-mmk
> 
> Do you plan to make a new tag any time soon after merging this bugfix?

We are thinking of introducing some paging API changes soon and then releasing.

PR #1046, opened 2024-07-29
codewide: appease clippy again

wprzytula:
This time clippy got angry about bad indentation in docstrings' lists.

## Pre-review checklist

### DIFF
--- a/scylla/src/macros.rs
+++ b/scylla/src/macros.rs
@@ -39,6 +39,7 @@ pub use scylla_cql::macros::IntoUserType;
 /// - serialization will succed if suffix of UDT fields is missing. If there are missing fields in the
 ///   middle it will fail. Note that if "skip_name_checks" is enabled, and the types happen to match,
 ///   it is possible for serialization to succeed with unexpected result.
+///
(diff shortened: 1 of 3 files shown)


Lorak-mmk (approved):

Lorak-mmk:
I approved the current changes, but they seem to not be enough to make clippy happy.

wprzytula:
> I approved the current changes, but they seem to not be enough to make clippy happy.

The remaining errors come not from clippy, but from rustc itself. A new version of rustc denied support for arbitrary cfg parameters. Those must be declared, as the error message says.

I would merge this as-is, solving the `cfg` issue in another PR, because it has nothing to do with clippy.

wprzytula:
> Is there a reason for doing this in another PR?

Just naming, and keeping PRs doing one thing, not multiple.

> I'm reluctant to merge without passing CI.

I see no point in being that religious about passing CI. If the only changes the PR introduces are related to indentation in docstrings, nothing can get worse because of the PR.

> Can't this PR fix both issues?

Well, it can. Any PR can fix any number of issues.

PR #1047, opened 2024-07-31
types: make vint_[en/de]code functions pub

muzarski:
This is just for the sake of cpp-rust-driver
which implements custom serialization.

## Pre-review checklist

- ~[] I added relevant tests for new features and bug fixes.~

### DIFF
--- a/scylla-cql/src/frame/types.rs
+++ b/scylla-cql/src/frame/types.rs
@@ -678,11 +678,11 @@ fn unsigned_vint_decode(buf: &mut &[u8]) -> Result<u64, std::io::Error> {
     Ok(v)
 }
 
-pub(crate) fn vint_encode(v: i64, buf: &mut Vec<u8>) {
+pub fn vint_encode(v: i64, buf: &mut Vec<u8>) {
     unsigned_vint_encode(zig_zag_encode(v), buf)
 }
 
-pub(crate) fn vint_decode(buf: &mut &[u8]) -> Result<i64, std::io::Error> {
+pub fn vint_decode(buf: &mut &[u8]) -> Result<i64, std::io::Error> {
     unsigned_vint_decode(buf).map(zig_zag_decode)
 }
 


wprzytula:
@muzarski Please investigate the failing CI, especially cassandra tests.

Lorak-mmk:
Is this still needed after yesterdays changes to serialization in cpp-rust-driver?

wprzytula:
> Is this still needed after yesterdays changes to serialization in cpp-rust-driver?

As I see it, one of `scylla_cql`'s roles is to be a low-level API for interacting with the DB, also to be used by interested parties independently from `scylla` crate. For that to be possible, all its functionality should be public.

Lorak-mmk:
> > Is this still needed after yesterdays changes to serialization in cpp-rust-driver?
> 
> As I see it, one of `scylla_cql`'s roles is to be a low-level API for interacting with the DB, also to be used by interested parties independently from `scylla` crate. For that to be possible, all its functionality should be public.

We can export a function after 1.0 if needed, but we can't un-export it, and exporting everything makes it much harder to make any changes.
If those functions are not needed by either `scylla` or `cpp-rust-driver` then I see no reason to export them until there is a use case for that.

wprzytula:
> > > Is this still needed after yesterdays changes to serialization in cpp-rust-driver?
> > 
> > 
> > As I see it, one of `scylla_cql`'s roles is to be a low-level API for interacting with the DB, also to be used by interested parties independently from `scylla` crate. For that to be possible, all its functionality should be public.
> 
> We can export a function after 1.0 if needed, but we can't un-export it, and exporting everything makes it much harder to make any changes. If those functions are not needed by either `scylla` or `cpp-rust-driver` then I see no reason to export them until there is a use case for that.

`scylla_cql`'s public API is not public API of `scylla`, so I don't see any hazard in exporting everything from `scylla_cql`. It's on `scylla` crate to control re-exports from `scylla_cql` properly.

Lorak-mmk:
> > > > Is this still needed after yesterdays changes to serialization in cpp-rust-driver?
> > > 
> > > 
> > > As I see it, one of `scylla_cql`'s roles is to be a low-level API for interacting with the DB, also to be used by interested parties independently from `scylla` crate. For that to be possible, all its functionality should be public.
> > 
> > 
> > We can export a function after 1.0 if needed, but we can't un-export it, and exporting everything makes it much harder to make any changes. If those functions are not needed by either `scylla` or `cpp-rust-driver` then I see no reason to export them until there is a use case for that.
> 
> `scylla_cql`'s public API is not public API of `scylla`, so I don't see any hazard in exporting everything from `scylla_cql`. It's on `scylla` crate to control re-exports from `scylla_cql` properly.

https://rust-lang.github.io/api-guidelines/necessities.html#public-dependencies-of-a-stable-crate-are-stable-c-stable

When we release `scylla 1.0` `scylla-cql` will need to already be `1.0` and so have its API stable. Good luck making any changes after that if you export everything.

`scylla-cql`, like any other library in existance, has its public API. In case of `scylla-cql` public API regarding vints is `SerializeValue` implementation and implementations for deserialization. 
`vint_[en/de]code` are internal implementation detail and unless there is a need for them to be public they should not be.

muzarski:
I'll close it. I agree that for now, there is no need to make `vint_` util functions public.

PR #1054, opened 2024-08-07
errors: remove ParseError

muzarski:

This PR solves multiple issues:
## No context for request serialization
I introduced a `CqlRequestSerializationError`, with the corresponding variants for each request type. Previously, the request serialization module would return `ParseError`, which did not provide much context.

## Distinguishing between frame ser and deser errors
Decoupled `FrameError` (which originally was overloaded with variants corresponding to serialization and deserialization) into:
- `CqlRequestSerializationError` - in fact, I only moved the variant regarding frame body compression from `FrameError` to `CqlRequestSerializationError`
- `FrameHeaderParseError` - it corresponds to the errors that appear when `Connection::reader()` async task receives a frame, and tries to deserialize its header. It happens before the actual response handler is chosen (based on the stream id).
- `FrameBodyExtensionsParseError` - as the name suggests, it is returned when the driver fails to deserialize body extensions (e.g. frame warnings, or custom payload).

## Test utilities making use of ParseError
There were a couple of test utilities used by `scylla-proxy` that would depend on ParseError. In some cases, if needed, I introduced new error types (e.g. `RequestDeserializationError`). I provided a docstring, noting that this is intended for driver's internal use and testing (it has to be `pub`, because of `scylla-proxy` crate).

## Removing ParseError
With this PR, we could finally remove `ParseError` altogether! I believe that there are still some error-related issues to be addressed, but this is a huge step forward, as this was the most problematic error type.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 pub use super::request::{
     execute::ExecuteSerializationError,
+    prepare::PrepareSerializationError,
     query::{QueryParametersSerializationError, QuerySerializationError},
 };
 
@@ -43,6 +44,8 @@ pub enum FrameError {
 
 #[derive(Error, Debug)]
 pub enum ParseError {
+    #[error("Failed to serialize PREPARE request: {0}")]
+    PrepareSerialization(#[from] PrepareSerializationError),
     #[error("Failed to serialize EXECUTE request: {0}")]
     ExecuteSerialization(#[from] ExecuteSerializationError),
     #[error("Failed to serialize QUERY request: {0}")]
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -13,7 +13,6 @@ pub use super::request::{
 
 use super::response::CqlResponseKind;
 use super::TryFromPrimitiveError;
-use crate::cql_to_rust::CqlTypeError;
 use crate::frame::value::SerializeValuesError;
 use crate::types::deserialize::{DeserializationError, TypeCheckError};
 use crate::types::serialize::SerializationError;
@@ -110,8 +109,6 @@ pub enum ParseError {
     SerializeValuesError(#[from] SerializeValuesError),
     #[error(transparent)]
     SerializationError(#[from] SerializationError),
-    #[error(transparent)]
-    CqlTypeError(#[from] CqlTypeError),
 }
 
 /// An error that occurred during CQL request serialization.
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -22,8 +22,6 @@ use thiserror::Error;
 #[derive(Error, Debug)]
 pub enum FrameError {
     #[error(transparent)]
-    CqlRequestSerialization(#[from] CqlRequestSerializationError),
-    #[error(transparent)]
     Parse(#[from] ParseError),
     #[error("Frame is compressed, but no compression negotiated for connection.")]
     NoCompressionNegotiated,
@@ -39,8 +37,6 @@ pub enum FrameError {
     StdIoError(#[from] std::io::Error),
     #[error("Unrecognized opcode{0}")]
     TryFromPrimitiveError(#[from] TryFromPrimitiveError<u8>),
-    #[error("Snap compression error: {0}")]
-    SnapCompressError(Arc<dyn Error + Sync + Send>),
     #[error("Snap decompression error: {0}")]
     SnapDecompressError(Arc<dyn Error + Sync + Send>),
     #[error("Error decompressing lz4 data {0}")]
@@ -100,6 +96,10 @@ pub enum CqlRequestSerializationError {
     /// Failed to serialize QUERY request.
     #[error("Failed to serialize QUERY request: {0}")]
     QuerySerialization(#[from] QuerySerializationError),
+
+    /// Request body compression failed.
+    #[error("Snap compression error: {0}")]
+    SnapCompressError(Arc<dyn Error + Sync + Send>),
 }
 
 /// An error type returned when deserialization of CQL
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 pub use super::request::{
+    auth_response::AuthResponseSerializationError,
     batch::{BatchSerializationError, BatchStatementSerializationError},
     execute::ExecuteSerializationError,
     prepare::PrepareSerializationError,
@@ -45,6 +46,8 @@ pub enum FrameError {
 
 #[derive(Error, Debug)]
 pub enum ParseError {
+    #[error("Failed to serialize AUTH_RESPONSE request: {0}")]
+    AuthResponseSerialization(#[from] AuthResponseSerializationError),
     #[error("Failed to serialize BATCH request: {0}")]
     BatchSerialization(#[from] BatchSerializationError),
     #[error("Failed to serialize PREPARE request: {0}")]
--- a/scylla-cql/src/frame/frame_errors.rs
+++ b/scylla-cql/src/frame/frame_errors.rs
@@ -29,24 +29,44 @@ pub enum FrameError {
     WarningsListParse(LowLevelDeserializationError),
     #[error("Malformed custom payload map: {0}")]
     CustomPayloadMapParse(LowLevelDeserializationError),
-    #[error("Received frame marked as coming from a client")]
-    FrameFromClient,
-    #[error("Received frame marked as coming from the server")]
-    FrameFromServer,
-    #[error("Received a frame from version {0}, but only 4 is supported")]
-    VersionNotSupported(u8),
-    #[error("Connection was closed before body was read: missing {0} out of {1}")]
-    ConnectionClosed(usize, usize),
-    #[error("Failed to read the frame header: {0}")]
-    HeaderIoError(std::io::Error),
-    #[error("Failed to read a chunk of response body. Expected {0} more bytes, error: {1}")]
-    BodyChunkIoError(usize, std::io::Error),
-    #[error("Unrecognized opcode{0}")]
-    TryFromPrimitiveError(#[from] TryFromPrimitiveError<u8>),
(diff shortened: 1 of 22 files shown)


wprzytula:
ref: #1056

muzarski:
Rebased on https://github.com/scylladb/scylla-rust-driver/pull/1067

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -32,8 +32,10 @@ pub enum FrameDeserializationError {
     VersionNotSupported(u8),
     #[error("Connection was closed before body was read: missing {0} out of {1}")]
     ConnectionClosed(usize, usize),
-    #[error(transparent)]
-    StdIoError(#[from] std::io::Error),
+    #[error("Failed to read a frame header: {0}")]
+    HeaderIoError(Arc<std::io::Error>),
```
wprzytula, line 33: nit: `a frame header` -> `the frame header` (because there's exactly one header per frame)

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -42,6 +44,16 @@ pub enum FrameDeserializationError {
...
 
+impl FrameDeserializationError {
+    pub fn new_header_io_error(err: std::io::Error) -> Self {
+        Self::HeaderIoError(Arc::new(err))
+    }
+
+    pub fn new_body_chunk_io_error(expected_bytes: usize, err: std::io::Error) -> Self {
+        Self::BodyChunkIoError(expected_bytes, Arc::new(err))
+    }
+}
```
wprzytula, line 52: Do these have to be `pub`?

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -66,20 +64,12 @@ pub enum FrameSerializationError {
...
-    #[error(transparent)]
-    DeserializationTypeCheckError(#[from] TypeCheckError),
-    #[error(transparent)]
     IoError(#[from] std::io::Error),
     #[error(transparent)]
-    SerializeValuesError(#[from] SerializeValuesError),
-    #[error(transparent)]
-    SerializationError(#[from] SerializationError),
+    TryFromIntError(#[from] std::num::TryFromIntError),
 }
```
wprzytula: You add `TryFromIntError` probably because some removed variant contained it. Doesn't this remove the context that the variant provided, leaving us with plain `TryFromIntError` without context?

scylla-cql/src/frame/request/mod.rs
```diff
@@ -74,7 +77,28 @@ pub trait SerializableRequest {
...
+    UnknownFlags { flags: u8 },
+    #[error("Named values in frame are currently unsupported")]
+    NamedValuesUnsupported,
+    #[error("Expected SerialConsistency, got regular Consistency: {0}")]
+    ExpectedSerialConsistency(Consistency),
+    #[error(transparent)]
+    BatchTypeParse(#[from] BatchTypeParseError),
+    #[error("Unexpected batch statement kind: {0}")]
+    UnexpectedBatchStatementKind(u8),
 }
```
wprzytula, line 102: If it's not intended to be exposed to users, maybe let's make it `doc(hidden)` ?

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -60,9 +60,6 @@ pub enum FrameSerializationError {
     SnapCompressError(snap::Error),
 }
 
-#[derive(Error, Debug)]
-pub enum ParseError {}
-
 #[non_exhaustive]
```
wprzytula: Dieeeee, it's high time.

scylla-cql/src/types/deserialize/mod.rs
```diff
@@ -46,30 +46,31 @@
 //!
 //! ```rust
 //! # use scylla_cql::frame::response::result::ColumnType;
-//! # use scylla_cql::frame::frame_errors::ParseError;
 //! # use scylla_cql::types::deserialize::{DeserializationError, FrameSlice, TypeCheckError};
 //! # use scylla_cql::types::deserialize::value::DeserializeValue;
+//! # use thiserror::Error;
```
wprzytula, line 51: Let's not hide this (don't use `#`), because otherwise readers can be confused about the `Error` trait - where it comes from.

scylla-cql/src/types/deserialize/mod.rs
```diff
@@ -85,11 +86,18 @@
 //! For example:
 //!
 //! ```rust
-//! # use scylla_cql::frame::frame_errors::ParseError;
 //! # use scylla_cql::frame::response::result::ColumnType;
 //! # use scylla_cql::types::deserialize::{DeserializationError, FrameSlice, TypeCheckError};
 //! # use scylla_cql::types::deserialize::value::DeserializeValue;
+//! # use thiserror::Error;
```
wprzytula, line 92: ditto

scylla-cql/src/types/deserialize/mod.rs
```diff
@@ -135,32 +137,36 @@
 //! Example:
 //!
 //! ```rust
-//! # use scylla_cql::frame::frame_errors::ParseError;
 //! # use scylla_cql::frame::response::result::ColumnType;
 //! # use scylla_cql::types::deserialize::{DeserializationError, FrameSlice, TypeCheckError};
 //! # use scylla_cql::types::deserialize::value::DeserializeValue;
 //! # use bytes::Bytes;
+//! # use thiserror::Error;
```
wprzytula, line 144: ditto

wprzytula (changes requested):
Mostly good direction, mainly nits.

scylla-cql/src/frame/frame_errors.rs
muzarski: These are used by `scylla-proxy`, but we could make an exception there and just simply construct corresponding variants directly. I'll do it, since I agree that it does not make much sense for these util functions to be public.

wprzytula (approved):

muzarski:
v1.1: fixed a typo in the `RequestDeserializationError` docstring

Lorak-mmk (changes requested):

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -40,10 +43,10 @@ pub enum FrameError {
...
 pub enum ParseError {
+    #[error("Failed to serialize EXECUTE request: {0}")]
+    ExecuteSerialization(#[from] ExecuteSerializationError),
     #[error("Failed to serialize QUERY request: {0}")]
     QuerySerialization(#[from] QuerySerializationError),
-    #[error("Failed to serialize query parameters: {0}")]
-    QueryParametersSerialization(#[from] QueryParametersSerializationError),
     #[error("Low-level deserialization failed: {0}")]
     LowLevelDeserializationError(#[from] LowLevelDeserializationError),
     #[error("Could not serialize frame: {0}")]
```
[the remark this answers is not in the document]
muzarski: > I assume it was introduced in first commit just to make commits atomic?

Exactly. I firstly introduced `QueryParametersSerializationError` - since I changed the return type of `QueryParameters::serialize`, I needed to add it to `ParseError` (returned by `SerializableRequest::serialize`). Then, I introduced two error types that depend on `QPSerializationError`, and thanks to that I could finally remove it from `ParseError`.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -43,7 +43,13 @@ pub enum FrameBodyExtensionsParseError {
...
     #[error("Error decompressing lz4 data {0}")]
-    Lz4DecompressError(#[from] lz4_flex::block::DecompressError),
+    Lz4DecompressError(#[from] Arc<lz4_flex::block::DecompressError>),
+}
+
+impl From<lz4_flex::block::DecompressError> for FrameBodyExtensionsParseError {
+    fn from(value: lz4_flex::block::DecompressError) -> Self {
+        FrameBodyExtensionsParseError::Lz4DecompressError(Arc::new(value))
+    }
 }
```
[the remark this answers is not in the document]
muzarski: Good catch

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -34,16 +34,16 @@ pub enum FrameError {
...
     #[error("Unrecognized opcode{0}")]
     TryFromPrimitiveError(#[from] TryFromPrimitiveError<u8>),
+    #[error("Snap compression error: {0}")]
+    SnapCompressError(snap::Error),
     #[error("Error compressing lz4 data {0}")]
     Lz4CompressError(#[from] lz4_flex::block::CompressError),
+    #[error("Snap decompression error: {0}")]
+    SnapDecompressError(snap::Error),
     #[error("Error decompressing lz4 data {0}")]
     Lz4DecompressError(#[from] lz4_flex::block::DecompressError),
```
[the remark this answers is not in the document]
muzarski: What would you suggest then? We could have a common `(De)CompressError` variant for both snap and lz4 - it would not provide any additional context, or maybe it could include a stringified error returned from the corresponding library (and maybe some enum of (de)compression kind - either lz4 or snap).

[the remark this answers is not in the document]
Lorak-mmk: Why not Arc<dyn Error>?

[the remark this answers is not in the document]
Lorak-mmk: My phone skipped a part of message - I meant `Arc<dyn Error>`.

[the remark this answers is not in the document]
muzarski: Yes, I got that. I somehow forgot about this possibility and was trying to come up with some other solution

muzarski:
v1.2: addressed @Lorak-mmk 's comments:
- removed `snap::Error` and `lz4_flex::block::DecompressError` types from public API. Replaced them with dynamic `Arc<dyn Error>` in public enum variants.
- added docstring for `FrameBodyExtensionsParseError`

muzarski:
@Lorak-mmk I also checked if we use any of these libraries' types in other parts of public API, but didn't find anything. You can double check that, if possible.

Lorak-mmk (approved):

muzarski:
@Lorak-mmk Rerun the CI, please. A known flaky test `lwt_optimisation::if_lwt_optimisation_mark_offered_then_negotiatied_and_lwt_routed_optimally` failed:
```
`Result::unwrap()` on an `Err` value: DbError(ServerError, "Failed to apply group 0 change due to concurrent modification")
```

muzarski:
Can we merge?

PR #1055, opened 2024-08-12
Remove Problematic Unwraps In Macros

Kailokk:
Using unwraps in macros means the code is compiled as part of the end users build.
This therefore means any lints a user uses will check the macro code as well.
These unwraps therefore the [clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/unwrap_used)  `unwrap_used` will throw a warning/error if the user has enabled it.

I wanted to enable this in a project that uses scylla, so i've created this merge request to try and fix this in the least invasive way possible.

I have returned error states where the unwraps were used, but if this feels more in the realm of a `panic` state, then it can simply be swapped for `panic`

## Pre-review checklist

### DIFF
--- a/scylla-macros/src/from_row.rs
+++ b/scylla-macros/src/from_row.rs
@@ -22,9 +22,7 @@ pub(crate) fn from_row_derive(tokens_input: TokenStream) -> Result<TokenStream, 
                     #field_name: {
                         let (col_ix, col_value) = vals_iter
                             .next()
-                            .unwrap(); // vals_iter size is checked before this code is reached, so
-                                    // it is safe to unwrap
-
+                            .expect("BUG: Size validated iterator did not contain the expected number of values"); 
                         <#field_type as FromCqlVal<::std::option::Option<CqlValue>>>::from_cql(col_value)
                             .map_err(|e| FromRowError::BadCqlVal {
                                 err: e,
@@ -48,8 +46,7 @@ pub(crate) fn from_row_derive(tokens_input: TokenStream) -> Result<TokenStream, 
                     {
                         let (col_ix, col_value) = vals_iter
                             .next()
-                            .unwrap(); // vals_iter size is checked before this code is reached, so
-                                    // it is safe to unwrap
+                            .expect("BUG: Size validated iterator did not contain the expected number of values"); 
 
                         <#field_type as FromCqlVal<::std::option::Option<CqlValue>>>::from_cql(col_value)
                             .map_err(|e| FromRowError::BadCqlVal {
--- a/scylla-macros/src/from_user_type.rs
+++ b/scylla-macros/src/from_user_type.rs
@@ -30,7 +30,7 @@ pub(crate) fn from_user_type_derive(tokens_input: TokenStream) -> Result<TokenSt
                     // field)
                     if let Some(received_field_name) = received_field_name {
                         if received_field_name == stringify!(#field_name) {
-                            let (_, value) = fields_iter.next().unwrap();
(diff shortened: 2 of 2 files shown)


Lorak-mmk (changes requested):
As far as I can tell this lint only complains about unwraps, and `.exepct()` is fine.
Imo if we are sure that .unwraps() can't fail, then they should remain panicking - because triggering this panic means a programming error on our part which should be reported and fixed. Unless @wprzytula disagrees, could you change those unwraps to `.expect`'s with proper error messages?

Kailokk:
> As far as I can tell this lint only complains about unwraps, and `.exepct()` is fine.
> 
> Imo if we are sure that .unwraps() can't fail, then they should remain panicking - because triggering this panic means a programming error on our part which should be reported and fixed. Unless @wprzytula disagrees, could you change those unwraps to `.expect`'s with proper error messages?

This is true, but it bubbles the problem up to the `expect_used` lint instead. https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used

This is definitely preferable, but personally i would like to see it removed, just given that this code is expanded into users of the drivers code.

Happy to change it if not panicking feels strange to you, i can absolutely understand the reason you have given

Lorak-mmk:
I think it would be better to change it to expects, but let's wait for @wprzytula 's opinion before doing it.

wprzytula:
> Imo if we are sure that .unwraps() can't fail, then they should remain panicking - because triggering this panic means a programming error on our part which should be reported and fixed.

100%, we use `unwrap` where a runtime occurence of an error would signify that our assumptions were wrong, and so we need to fix it.

> Unless @wprzytula disagrees, could you change those unwraps to .expect's with proper error messages?

I fully agree. `expect()`s are widely used in library code, too - see [tokio's usages of `expect()`](https://github.com/search?q=repo%3Atokio-rs%2Ftokio%20expect(&type=code). Also, this is the approach I took in new deserialization macros.

Kailokk:
@wprzytula 

I think that's reasonable. I have one more alternative to suggest:

If we use `.unwrap_or_else(|| panic!(...));` in place of expect.
Normally this is equivalent code, but in this case, using `.unwrap_or_else` will prevent lint warnings in end users code. This is only a problem in my eyes because these are macros.

If you don't find this acceptable, i will implement the `.expect` calls as discussed.

Lorak-mmk:
> @wprzytula
> 
> I think that's reasonable. I have one more alternative to suggest:
> 
> If we use `.unwrap_or_else(|| panic!(...));` in place of expect. Normally this is equivalent code, but in this case, using `.unwrap_or_else` will prevent lint warnings in end users code. This is only a problem in my eyes because these are macros.

If someone enables this lint that means they don't want any `expect` calls in their code - and I suspect they have reasons for that other than the name of this method. Tricking the lint does not seem helpful in this case - it obfuscates the code and prevents the user of the lint from achieving the goal of panic prevention.

> 
> If you don't find this acceptable, i will implement the `.expect` calls as discussed.

I think that would be the best solution.

Kailokk:
> I think that would be the best solution.

Then so it shall be :)

Kailokk:
@Lorak-mmk should be okay now.

wprzytula (approved):

Lorak-mmk:
CI fail will be fixed by https://github.com/scylladb/scylla-rust-driver/pull/1060 so we can ignore the fail here.
I'll squash-merge because of unclean commit history.

Kailokk:
@Lorak-mmk thank you for your time

PR #1060, opened 2024-08-19
tests: disable tablets in CQL Counter's tests

wprzytula:
As tablets don't support counters yet, they need to be disabled in Counter type tests.
See https://github.com/scylladb/scylladb/commit/c70f321c6f581357afdf3fd8b4fe8e5c5bb9736e.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/cql_types_test.rs
+++ b/scylla/src/transport/cql_types_test.rs
@@ -3,7 +3,7 @@ use crate::cql_to_rust::FromCqlVal;
 use crate::frame::response::result::CqlValue;
 use crate::frame::value::{Counter, CqlDate, CqlTime, CqlTimestamp};
 use crate::macros::FromUserType;
-use crate::test_utils::{create_new_session_builder, setup_tracing};
+use crate::test_utils::{create_new_session_builder, scylla_supports_tablets, setup_tracing};
 use crate::transport::session::Session;
 use crate::utils::test_utils::unique_keyspace_name;
 use itertools::Itertools;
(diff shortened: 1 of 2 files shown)


scylla/src/transport/cql_types_test.rs
```diff
@@ -16,23 +16,27 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
...
 // Used to prepare a table for test
-// Creates a new keyspace
+// Creates a new keyspace, without tablets if requested and the ScyllaDB instance supports them.
 // Drops and creates table {table_name} (id int PRIMARY KEY, val {type_name})
-async fn init_test(table_name: &str, type_name: &str) -> Session {
+async fn init_test_maybe_without_tablets(
+    table_name: &str,
+    type_name: &str,
+    should_disable_tablets: bool,
+) -> Session {
```
Lorak-mmk, line 25: I don't think the name change is needed, it doesn't help with anything.

It would be better to change parameter to `enable_tablets` to avoid unnecessary negation - it is easier for a human to reason about.

Lorak-mmk (changes requested):

scylla/src/transport/cql_types_test.rs
wprzytula: What name change do you have in mind? Even though GH shows this as a name change, what I actually did here is introducing a new function. The old function's logic is delegated here. Thanks to that, we don't have to do a huge lot of name changes.

Lorak-mmk: Right, I didn't notice that `init_test` still exists.

The part about negated semantics still stands. It's actually the same point you made in review of my PR that introduced tablets: https://github.com/scylladb/scylla-rust-driver/pull/937#discussion_r1549144940

Lorak-mmk (approved):

Issue #987, opened 2024-04-24
Make paging state strongly typed


wprzytula:
Currently, paging state is represented both in the driver's internals and in the public API as `Bytes` blob. As it's no use allowing the users to construct paging state on their own, I suggest creating an opaque newtype for it, with a private constructor.

Additionally, the core benefit that `Bytes` bring - being able to subslice a bigger allocation - does not apply in the case of paging state. I believe that paging state could be better represented as `Arc<[u8]>`.

Lorak-mmk:
> As it's no use allowing the users to construct paging state on their own, I suggest creating an opaque newtype for it, with a private constructor.

They may want to serialize paging state and deserialize it later

wprzytula:
> > As it's no use allowing the users to construct paging state on their own, I suggest creating an opaque newtype for it, with a private constructor.
> 
> They may want to serialize paging state and deserialize it later

Can't this be said about any struct that we intentionally restrict access to?

Lorak-mmk:
No, this is actually existing use case, that's how you can implement paging e.g. in your web service.
Even Java Driver docs mention this use case: https://java-driver.docs.scylladb.com/scylla-3.11.2.x/manual/paging/index.html
It's hardly comparable to serializing any other random struct.

wprzytula:
Then it would make sense to expose constructor (`from_raw`) and accessor (`into_raw`/`into_inner`).

wprzytula:
> What is the problem with current version (just using Bytes)?

The answer is:
> Additionally, the core benefit that Bytes bring - being able to subslice a bigger allocation - does not apply in the case of paging state. I believe that paging state could be better represented as Arc<[u8]>.

When I see `Bytes`, I expect some shared ownership of the whole frame. This is misleading, because paging state's `Bytes` are currently created from an exclusively owned buffer that is distinct from the frame `Bytes`.

wprzytula:
> Custom struct will force users to create their own newtype wrapper to be able to serialize it

Not if we provide the two functions I mentioned:

> Then it would make sense to expose constructor (from_raw) and accessor (into_raw/into_inner).

Lorak-mmk:
> > Custom struct will force users to create their own newtype wrapper to be able to serialize it
> 
> Not if we provide the two functions I mentioned:

Yes, even then, because people use frameworks for serialization (serde, rkyv etc)
> 
> > Then it would make sense to expose constructor (from_raw) and accessor (into_raw/into_inner).

[...]

Issue #940, opened 2024-02-23
query_paged is a misleading name


Lorak-mmk:
`query_paged` suggests that the executed query will be paged, while this is not the case.
Whether the query is paged or not is controlled by querie's `page_size` attribute.
The only difference between query and query_paged is that the former accepts custom paging_state, so something like `query_with_paging_state` would probably be a better name.

roydahan:
@wprzytula I see this one is in "High Priority" but it doesn't have a milestone.
Is it planned to pre-1.0 milestones?

Issue #978, opened 2024-04-11
Request execution API changes - umbrella issue


wprzytula:
This is an umbrella issue to group all changes regarding the request execution API, as there are a number of them planned before the API gets stable.

```[tasklist]
### Tasks
- [ ] https://github.com/scylladb/scylla-rust-driver/issues/713 - we're currently planning to implement the `Executable` trait
- [ ] https://github.com/scylladb/scylla-rust-driver/issues/262 - necessarily consider this when implementing `Executable`
- [ ] https://github.com/scylladb/scylla-rust-driver/issues/417 - this needs to be resolved if we decide to forbid executing unprepared requests with values (which we want to, in order to make requests type-safe)
- [ ] #940 - covered by `Executable`
- [ ] #941 - solves multiple issues, e.g. #974
- [ ] #108 - possibly solved by #941
```

PR #1061, opened 2024-08-20
Query paging API: hardening, robustness, explicitness

wprzytula:
# Motivation

Query paging API has shown to be unclear for users:
- `{query,execute}_paged` had misleading name (#940),
- `paging_state` was a magic `Bytes` instance (#987),
- `{query,execute}` would silently return only the first page if page size was specified on a statement,
-  users would naturally resort to `{query,execute}` as the methods with the simplest names, even though they should be used with SELECTs with caution:
  - with no page size set, an unpaged SELECT query would have high latency and memory footprint, and could event time out,
  - with page size set, only first page would be returned (see the previous big point).
- it was easy to forget about iterating over pages based on `paging_state`, as it was just a field on `QueryResult`,
- there was lack of examples of a loop together with `{query,execute}_paged`.

This PR seeks to address those problems before we attack the **Giant Request Execution Refactor™ (GRER™)** (#978).

# What's done

### The core idea incorporated:

**Shift responsibility about whether to page or not from statements** (by setting `page_size` to `Some` or `None`) **to `Session's` methods**.

## Detailed changes
1. **Strongly typed `PageSize` is introduced**, which restricts the inner `i32`'s range to positive integers. This is important, because passing nonpositive integer as a page size silently makes the query unpaged. All APIs now accept `PageSize`, and it's user's responsibility to create a `PageSize`, handling possible errors thrown upon its construction.
1. **Paging state is made strongly typed.** Two new types are introduced: `PagingState` and `PagingStateResponse`. `PagingState` is, underneath, just an `Option<Arc>` over the raw bytes from the DB. The idea is that `PagingState` is passed as an argument to request execution API to ask for resuming the query (_continuing it_) from the point specified by given `PagingState`. `PagingStateResponse` is returned always by the DB from paged requests - with information about either more or no more pages being available. `PagingStateResponse` is easily convertible into `PagingState`, so that user can use it in in a loop.
1. **`PageSize` is made mandatory on statements.** Henceforth, paged API will always use it and unpaged API will always ignore it.
1. **`{query,execute}` methods have `_unpaged` prefix appended** and they always perform an unpaged query (ignoring the `PageSize` specified on a statement).
1. **`{query,execute}_paged` are renamed to `{query,execute}_single_page`**. The name `%_paged` has long confused users. The new name indicates explicitly that only a single page is fetched with a single call to those methods.
1. **`{query,execute}_single_page` pass page state explicitly.** As an attempt to make paged queries' API more explicit, robust and self-explanatory, `PagingStateResponse` is decoupled from `QueryResult`. Instead, it's returned from `{query,execute}_single_page` as a second field of a pair, helping users not to forget about it.
  Moreover, internal and external APIs which expect that the query is not paged now warn if a "more pages" page state is returned from a query.
1. `select-paging` example is modified to showcase the suggested way to use `{query,execute}_single_page`.
1. Documentation is adjusted to the introduced changes.

Fixes: #940
Fixes: #987

## Pre-review checklist


scylla-cql/src/frame/request/query.rs
```diff
@@ -204,3 +210,58 @@ impl<'q> QueryParameters<'q> {
...
+    #[inline]
+    pub(crate) fn new_from_raw_bytes(raw_bytes: &[u8]) -> Self {
+        Self(Arc::from(raw_bytes))
+    }
+
+    #[inline]
+    pub fn as_byte_slice(&self) -> &[u8] {
+        self.0.deref()
+    }
+}
```
Lorak-mmk, line 268: I have to admit that I'm not exactly the fan of this API:
- The naming isn't great: `PagingContinuation` / `into_paging_continuation_opt` are imo quite unfriendly names with not a very obvious meaning.
- `into_paging_continuation_opt` returns a value that also represents a beggining of paging query. Consider the example from session_test:
```rust
        let mut paging_continuation: Option<PagingContinuation> = None;
        let mut watchdog = 0;
        loop {
            let rs_manual = session
                .execute_paged(&prepared_paged, &[], paging_continuation)
                .await
                .unwrap();
            results_from_manual_paging.append(&mut rs_manual.rows.unwrap());
            if watchdog > 30 || rs_manual.paging_state.finished() {
                break;
            }
            watchdog += 1;
            paging_continuation = rs_manual.paging_state.into_paging_continuation_opt();
        }
```

if user forgot about checking `rs_manual.paging_state.finished() {` they would go into infinite loop.

Ideas:
- Rename `PagingState` to `PagingResponse` or `PagingStateResponse` - it better conveys the semantics that this is something received from server.
- Then rename `PagingContinuation` to `PagingState` - this is the type that user will be using most (possibly storing it somewhere etc), so it's good to have a name that will be familiar from Scylla docs / other drivers.
- In next points I'll refer to those structures by the names I proposed above.
- `PagingState` could internally be an enum, or an Option, then our methods would accept `PagingState` instead of `Option<PagingState>`.
- Get rid of `into_paging_continuation_opt`
- Nit: `new_from_raw_bytes` could accept an `Arc`. If user already has an Arc then it avoids allocation. If they don't then creating it is not a big overhead.
- Rename paging_state field on QueryResult to paging_response (doesn't matter, will be removed in future commits anyway, right?)

To sum up (I didn't compile this, it's just a rough idea):
```rust
pub enum PagingStateReponse {
    NoMorePages,
    HasMorePages(PagingState)
}

pub struct PagingState {
    Start,
    Continue(Arc<[u8]>)
}

impl PagingState {
    pub fn new() -> PagingState {// or start() 
        PagingState::Start
    }

    #[inline]
    pub(crate) fn new_from_raw_bytes(raw_bytes: &[u8]) -> Self {
        if raw_bytes.len() == 0 {
             PagingState::Start
        } else {
            PagingState::Continue(Arc::from(raw_bytes))
    }

    #[inline]
    pub fn as_byte_slice(&self) -> &[u8] {
        match self {
            PagingState::Start => &[],
            PagingState::Continue(b) => b.deref()
        }
    }
}

pub async fn query_paged(
        &self,
        query: impl Into<Query>,
        values: impl SerializeRow,
        paging_continuation: PagingContinuation,
```


Then the example before would look like this (or similar):
```rust
        let mut paging_state: PagingState = PagingState::start();
        let mut watchdog = 0;
        loop {
            let rs_manual = session
                .execute_paged(&prepared_paged, &[], paging_state)
                .await
                .unwrap();
            results_from_manual_paging.append(&mut rs_manual.rows.unwrap());
            if watchdog > 30 {
                break;
            }
            watchdog += 1;
            match  rs_manual.paging_response { // or into_paging_control_flow, not important here
                PagingResponse::NoMorePages => break,
                PagingResponse::HasMorePages(new_state) => paging_state = new_state,
            }
        }
```

While it may be a line or to longer, I think this version is cleaner and more obvious to readers.

scylla/src/transport/connection.rs
```diff
@@ -274,14 +283,26 @@ impl NonErrorQueryResponse {
...
-        })
+        ))
+    }
+
+    pub(crate) fn into_query_result(self) -> Result<QueryResult, QueryError> {
+        let (result, paging_state) = self.into_query_result_and_paging_state()?;
+
+        if !paging_state.finished() {
+            warn!("Internal driver API misuse: nonfinished paging state discarded by `NonErrorQueryResponse::into_query_result`");
+        }
```
Lorak-mmk, line 303: As far as I can tell this is used only for batches (and in one test, but it doesn't matter here), and batches don't have paging. In that case, if we encounter this warning this mean that either
- There is some serious bug in the driver
- There is bug in Scylla

both those cases warrant a panic imo.

scylla/src/transport/session.rs
```diff
@@ -625,7 +625,12 @@ impl Session {
         values: impl SerializeRow,
     ) -> Result<QueryResult, QueryError> {
         let query = query.into();
-        self.query_inner(&query, values, None, None).await
+        let (result, paging_state) = self.query_inner(&query, values, None, None).await?;
+        if !paging_state.finished() {
+            warn!("Unpaged query's response hints that there are more pages! Consider using Session API for paging queries,
+                   such as Session::query_iter or Session::query_single_page in a loop.");
+        }
```
Lorak-mmk, line 641: Here it's the same - this warning means that either driver or Scylla are broken, right? We should imo panic in such situation.

scylla/src/transport/session.rs
```diff
@@ -1007,16 +1012,22 @@ impl Session {
...
         let serialized_values = prepared.serialize_values(&values)?;
-        self.execute_inner(prepared, &serialized_values, None, None)
-            .await
+        let (result, paging_state) = self
+            .execute_inner(prepared, &serialized_values, None, None)
+            .await?;
+        if !paging_state.finished() {
+            warn!("Unpaged query's response hints that there are more pages! Consider using Session API for paging queries,
+                   such as Session::execute_iter or Session::execute_single_page in a loop.");
+        }
```
Lorak-mmk, line 1072: ditto

Lorak-mmk (changes requested):

docs/source/queries/result.md
```diff
@@ -1,6 +1,7 @@
 # Query result
```
muzarski, line 1: I think the changes in this file (and other `.md` files) should belong to the next commit.

muzarski (approved):
Some nits, otherwise LGTM

scylla-cql/src/frame/request/query.rs
```diff
@@ -204,3 +206,76 @@ impl<'q> QueryParameters<'q> {
         })
     }
 }
+
+#[derive(Debug, Clone)]
+pub enum PagingStateResponse {
+    HasMorePages { next_page_handle: PagingState },
+    NoMorePages,
+}
```
Lorak-mmk, line 215: Why 'handle'?

scylla/src/transport/connection.rs
```diff
@@ -801,6 +801,7 @@ impl Connection {
     pub(crate) async fn query_single_page(
         &self,
         query: impl Into<Query>,
+        page_size: PageSize,
     ) -> Result<QueryResult, QueryError> {
```
Lorak-mmk: Commit: `connection: pass page size through request API `
Could you explain here (and preferably in commit message too) why is there a need to have page_size argument for functions in Connection?
It was my understanding that page_size is attached to a statement (prepared or not) - in case of a function that I'm commenting on it would be in the `query: impl Into<Query>,`.

scylla/src/transport/connection.rs
```diff
@@ -273,14 +282,29 @@ impl NonErrorQueryResponse {
...
+
+    pub(crate) fn into_query_result(self) -> Result<QueryResult, QueryError> {
+        let (result, paging_state) = self.into_query_result_and_paging_state()?;
+
+        if !paging_state.finished() {
+            let error_msg = "Internal driver API misuse or a server bug: nonfinished paging state\
+                             discarded by `NonErrorQueryResponse::into_query_result`";
+            error!(error_msg);
+            return Err(QueryError::ProtocolError(error_msg));
+        }
```
Lorak-mmk, line 305: `discarded by` doesn't sound right now that the whole query is errored out.
`passed to`?

scylla/src/transport/connection.rs
```diff
@@ -1199,21 +1223,24 @@ impl Connection {
...
-                    QueryError::ProtocolError("system.local query returned a wrong number of rows")
-                }
-                SingleRowTypedError::FromRowError(_) => {
-                    QueryError::ProtocolError("Row is not uuid type as it should be")
-                }
-            })?;
+            .await?;
+        if !paging_state.finished() {
+            warn!("Fetching local schema version returned nonfinished paging state");
+        }
```
Lorak-mmk: Not strictly related to this PR: I'm wondering if it's ever possible for that to happen. Maybe unpaged / iter should be used here for correctness?

Lorak-mmk (changes requested):
I don't like the first commit. Using invalid values for page size was never shown to be a problem / footgun for our users, so advantages from strong typing are quite limited. Otoh it introduces unnecessary complications in each call to set_page_size.
Imo in case of such obscure error, which also implies a bug in user code, a panic wuld be better - we still catch the error, but don't introduce complications for users.
So let's drop PageSize type and perform panicking check in set_page_size methods.

PR description is out of date - it mentions `PagingContinuation`.

In checklist you marked that `All commits compile, pass static checks and pass test.` - is that really the case? I would expect commit `session: make {query,execute} methods unpaged ` to break something if it really prevents usage of unpaged queries. If it doesn't it sounds like our tests are lacking.

scylla/src/transport/connection.rs
wprzytula: `would be discarded` -> this is the idea. Does it sound good enough?

scylla/src/transport/connection.rs
wprzytula: I agree, I see no reason why we should use `query_single_page` here.

scylla-cql/src/frame/request/query.rs
wprzytula: Because it can be used to obtain the next page, this way serving as its handle.

wprzytula:
> So let's drop PageSize type and perform panicking check in set_page_size methods.

I suggest a slightly different thing: let's make `PageSize` not occur in the public API, but still be passed around in the internal codebase. This way we make sure that we won't forget about a panic anywhere - because we must handle an error on `try_into()` - I suggest `unwrap`ping it.

scylla/src/transport/connection.rs
Lorak-mmk: ok

wprzytula:
> In checklist you marked that All commits compile, pass static checks and pass test. - is that really the case? I would expect commit session: make {query,execute} methods unpaged to break something if it really prevents usage of unpaged queries. If it doesn't break any test then it sounds like our tests are lacking.

Good catch. Let's see if any test breaks.

Lorak-mmk:
> > So let's drop PageSize type and perform panicking check in set_page_size methods.
> 
> I suggest a slightly different thing: let's make `PageSize` not occur in the public API, but still be passed around in the internal codebase. This way we make sure that we won't forget about a panic anywhere - because we must handle an error on `try_into()` - I suggest `unwrap`ping it.

Makes sense

wprzytula:
> > In checklist you marked that All commits compile, pass static checks and pass test. - is that really the case? I would expect commit session: make {query,execute} methods unpaged to break something if it really prevents usage of unpaged queries. If it doesn't break any test then it sounds like our tests are lacking.
> 
> Good catch. Let's see if any test breaks.

Sad news... everything has passed.

examples/select-paging.rs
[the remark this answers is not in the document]
wprzytula: I cannot find it neither in the mentioned commit nor in the last commit.

scylla/src/transport/connection.rs
wprzytula: It seems that I forgot to make `Connection`'s inner API follow the idea introduced for `Session`. Namely, `unpaged` method not taking `PageSize`, `single_page` taking `PageSize`, and both those methods based on top of a common method taking `Option<PageSize>`.

wprzytula: I'll adjust it, thanks for pointing this out.

examples/select-paging.rs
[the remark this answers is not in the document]
Lorak-mmk: image

scylla-cql/src/frame/request/query.rs
Lorak-mmk: For me a handle is some (usually small / simple) object that allows access to some (bigger / allocated, not sure what is the right word here) resource - like a file handle, or a PHANDLE in win32 APIs.
Why not `HasMorePages { state: PagingState },`?

examples/select-paging.rs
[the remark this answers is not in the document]
wprzytula: Ah, it's in commit message. Makes sense.

wprzytula:
v1.1: addressed all Karol's comments apart from PageSize panics, which is TODO.

wprzytula:
v1.1.1: Clippy fix

README.md
```diff
@@ -19,7 +19,7 @@ let uri = "127.0.0.1:9042";
 
 let session: Session = SessionBuilder::new().known_node(uri).build().await?;
 
-let result = session.query("SELECT a, b, c FROM ks.t", &[]).await?;
+let result = session.query_unpaged("SELECT a, b, c FROM ks.t", &[]).await?;
```
fee-mendes, line 22: Well, this is precisely the kind of query and place you want people NOT to use `query_unpaged`. :-)

docs/source/data-types/blob.md
```diff
@@ -13,11 +13,11 @@ use scylla::IntoTypedRows;
 // We can insert it by reference to not move the whole blob
 let to_insert: Vec<u8> = vec![1, 2, 3, 4, 5];
 session
-    .query("INSERT INTO keyspace.table (a) VALUES(?)", (&to_insert,))
+    .query_unpaged("INSERT INTO keyspace.table (a) VALUES(?)", (&to_insert,))
     .await?;
 
 // Read blobs from the table
-let result = session.query("SELECT a FROM keyspace.table", &[]).await?;
+let result = session.query_unpaged("SELECT a FROM keyspace.table", &[]).await?;
```
fee-mendes, line 20: Also a full scan, just page.

docs/source/data-types/collections.md
```diff
@@ -63,11 +63,11 @@ use std::collections::HashSet;
 // Insert a set of ints into the table
 let my_set: HashSet<i32> = vec![1, 2, 3, 4, 5].into_iter().collect();
 session
-    .query("INSERT INTO keyspace.table (a) VALUES(?)", (&my_set,))
+    .query_unpaged("INSERT INTO keyspace.table (a) VALUES(?)", (&my_set,))
     .await?;
 
 // Read a set of ints from the table
-let result = session.query("SELECT a FROM keyspace.table", &[]).await?;
+let result = session.query_unpaged("SELECT a FROM keyspace.table", &[]).await?;
```
fee-mendes, line 70: Paged.

README.md
wprzytula: ...which explains why we need this refactor that much :P

docs/source/data-types/collections.md
```diff
@@ -116,11 +116,11 @@ let mut my_map: HashMap<String, i32> = HashMap::new();
 my_map.insert("abcd".to_string(), 16);
 
 session
-    .query("INSERT INTO keyspace.table (a) VALUES(?)", (&my_map,))
+    .query_unpaged("INSERT INTO keyspace.table (a) VALUES(?)", (&my_map,))
     .await?;
 
 // Read a map from the table
-let result = session.query("SELECT a FROM keyspace.table", &[]).await?;
+let result = session.query_unpaged("SELECT a FROM keyspace.table", &[]).await?;
```
fee-mendes, line 123: Paged.

docs/source/data-types/collections.md
```diff
@@ -142,11 +142,11 @@ let mut my_map: BTreeMap<String, i32> = BTreeMap::new();
...
```
fee-mendes, line 149: Paged.

docs/source/data-types/counter.md
```diff
@@ -11,7 +11,7 @@ use scylla::IntoTypedRows;
 use scylla::frame::value::Counter;
 
 // Read counter from the table
-let result = session.query("SELECT c FROM keyspace.table", &[]).await?;
+let result = session.query_unpaged("SELECT c FROM keyspace.table", &[]).await?;
```
fee-mendes, line 14: Paged.

docs/source/data-types/blob.md
wprzytula: There are so many unpaged queries present in our docs... I suggest postponing rewrite of them to a separate follow-up PR.

fee-mendes:
Yeah, I guess it is fine to postpone all docs expressions, but ideally we want people to use paging and live on the safe side unless they know what they are doing.

Quickly walked through most of the actual changes and they look fine to me. But let's not forget to update docs and user-facing info so actually prevent them from ending up in the same state as they were prior to this series. 

Sorry for the spam :)

wprzytula:
> Yeah, I guess it is fine to postpone all docs expressions, but ideally we want people to use paging and live on the safe side unless they know what they are doing.
> 
> Quickly walked through most of the actual changes and they look fine to me. But let's not forget to update docs and user-facing info so actually prevent them from ending up in the same state as they were prior to this series.
> 
> Sorry for the spam :)

You made great job pointing out those bad practices in docs, thanks!

wprzytula:
v1.2: removed `PageSize` from public API, panicking on non-positive integer passed.

wprzytula:
v 1.2.1: mentioned that setters panic with non-positive numbers given.

wprzytula:
v1.2.2: **fully** removed `PageSize` from public API - I forgot about getters - and made it `pub(crate)`.

Lorak-mmk (approved):

muzarski (approved):

PR #1062, opened 2024-08-21
default_lb: Enrich comments & docstrings

wprzytula:
The DefaultPolicy is a complex machinery. In order to grasp it fully, extensive comments are essential.
Some internal functions' names were changed, to make them more better reflect the functions' semantics.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/load_balancing/default.rs
+++ b/scylla/src/transport/load_balancing/default.rs
@@ -58,15 +58,28 @@ impl NodeLocationPreference {
     }
 }
 
+/// An ordering requirement for replicas.
 #[derive(Clone, Copy)]
 enum ReplicaOrder {
+    /// No requirement. Replicas can be returned in arbitrary order.
     Arbitrary,
-    RingOrder,
+
+    /// A requirement for the order to be deterministic, not only across statement executions
+    /// but also across drivers. This is used for LWT optimisation, to avoid Paxos conflicts.
+    Deterministic,
 }
 
+/// Statement kind, used to enable specific load balancing patterns for certain cases.
+///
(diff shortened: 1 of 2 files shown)


Lorak-mmk (changes requested):
Great PR!
I'd add a description for `DefaultPolicyTargetComparator` because it is definitely not obvious what that is or why it is present.
I propose the following, lmk what you think:
```
Plan should return unique elements. It is not however obvious what it means because some nodes in the plan may have shards and some may not.
This helper structure defines equality of plan elements.
How the comparison works:
- If at least one of elements is shard-less then compare just nodes.
- If both elements have shards, then compare both nodes and shards.
Why is it implemented this way?
Driver should not attempt to send the same request to the same destination twice.
If plan element doesn't have shard specified, then a random shard will be chosen by the driver.
If the plan also contains the same node but with the shard present, and we randomly chose the same shard for the shard-less element, then we have duplication.
Example: plan is `[(Node1, Some(1)), (Node1, None)]` - if the driver uses the second element and randomly chooses shard 1, then we have a duplication.

On the other hand if the plan has a duplicate node, but with different shards, then we want to use both elements - so we can't just make the list unique by node, and so this struct was created.
```

wprzytula:
v1.1: justified why latency awareness is not recommended.

wprzytula:
> I propose the following, lmk what you think:

Wonderful, I'll include it in the PR.

wprzytula:
v1.2: added  `DefaultPolicyTargetComparator` docstring.

Lorak-mmk (approved):

wprzytula:
v1.2.1: clippy fixes.

PR #1065, opened 2024-08-22
New deserialization API - preparations

wprzytula:
This PR consists the first bunch of minor preparatory steps for the main leap of the deserialization refactor, extracted from #1057 as @Lorak-mmk requested.

For explanation and broader context see #1057.
Ref: #462

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/iterator.rs
+++ b/scylla/src/transport/iterator.rs
@@ -415,7 +415,7 @@ mod checked_channel_sender {
             response::result::{ResultMetadata, Rows},
         },
     };
-    use std::marker::PhantomData;
+    use std::{marker::PhantomData, sync::Arc};
     use tokio::sync::mpsc;
     use uuid::Uuid;
 
@@ -456,7 +456,7 @@ mod checked_channel_sender {
         ) {
             let empty_page = ReceivedPage {
                 rows: Rows {
-                    metadata: ResultMetadata::mock_empty(),
+                    metadata: Arc::new(ResultMetadata::mock_empty()),
                     paging_state_response: PagingStateResponse::NoMorePages,
                     rows_count: 0,
                     rows: Vec::new(),
--- a/scylla/src/transport/iterator.rs
+++ b/scylla/src/transport/iterator.rs
@@ -410,7 +410,10 @@ impl RowIterator {
 mod checked_channel_sender {
     use scylla_cql::{
         errors::QueryError,
-        frame::response::result::{ResultMetadata, Rows},
+        frame::{
+            request::query::PagingStateResponse,
+            response::result::{ResultMetadata, Rows},
+        },
     };
     use std::marker::PhantomData;
     use tokio::sync::mpsc;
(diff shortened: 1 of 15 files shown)


scylla/src/transport/query_result.rs
```diff
@@ -311,20 +314,12 @@ mod tests {
...
         QueryResult {
             rows: None,
             warnings: vec![],
             tracing_id: None,
             paging_state: None,
-            col_specs: vec![column_spec],
+            metadata: None,
             serialized_size: 0,
         }
     }
```
Lorak-mmk, line 327: This changes the semantics of this function. Doesn't it break tests?

scylla/src/transport/iterator.rs
```diff
@@ -853,8 +853,8 @@ where
             let result = (self.fetcher)(paging_state).await?;
             let response = result.into_non_error_query_response()?;
             match response.response {
-                NonErrorResponse::Result(result::Result::Rows(mut rows)) => {
-                    paging_state = rows.paging_state.take();
+                NonErrorResponse::Result(result::Result::Rows(rows)) => {
+                    paging_state = rows.paging_state.clone();
                     let (proof, send_result) = self
```
Lorak-mmk, line 858: Is there a reason for this change? Why we can't `.take()` anymore? I fail to see how does it relate to this commit (`codewide: share ResultMetadata by an Arc `)

scylla/src/transport/iterator.rs
```diff
@@ -689,7 +689,7 @@ where
                     .load_balancing_policy
                     .on_query_success(&self.statement_info, elapsed, node);
 
-                self.paging_state = rows.paging_state.take();
```
muzarski: Here as well (regarding @Lorak-mmk 's comment). I think this should belong to other commit. I am not even sure why we were modifying received `rows` (by moving the paging state out of them), even though we send them via some channel later.

muzarski (changes requested):

scylla/src/transport/iterator.rs
wprzytula: This is a leftover from before I extracted paging state from ResultMetadata. I'll fix this.

scylla/src/transport/iterator.rs
wprzytula: We can do that because the iterator manages paging state itself, and it does not expose any getters of it.

scylla/src/transport/query_result.rs
wprzytula: The test case was invalid. `rows` being `None` imply empty `col_specs`. Fortunately, no tests depended on them being nonempty.

wprzytula:
v1.1: reverted `paging_state.clone()` to `paging_state.take()` in iterator.

scylla/src/transport/query_result.rs
Lorak-mmk: From what I can see it is used in methods that make the rows non empty: `make_rows_query_result`, `make_string_rows_query_result`. Now those methods produce QueryResult with rows, but with empty col_specs. Isn't that invalid?

wprzytula: It indeed is invalid. I'll fix it in make_rows_query_result, make_string_rows_query_result.

wprzytula:
v1.2: rebased on main.

wprzytula:
v1.2.1: fixed compile error on MSRV due to: `error[E0277]: the trait bound `[u8]: Default` is not satisfied`

Lorak-mmk (approved):

muzarski (approved):

PR #1068, opened 2024-08-27
docs: Eradicate unpaged SELECTs and warn about their drawbacks

wprzytula:
Follow-up of #1061.

# Motivation

Unpaged SELECTs should be discouraged due to the following drawbacks:
> However, `query_unpaged` will return all results in one, possibly giant, piece
> (unless a timeout occurs due to high load incurred by the cluster).
> This:
> - increases latency,
> - has large memory footprint,
> - puts high load on the cluster,
> - is more likely to time out (because big work takes more time than little work,
>   and returning one large piece of data is more work than returning one chunk of data).

# What's done

- the above caveat is added in the documentation book,
- existing examples are modified to use `query_iter` instead,
- giant bonus: a broad overview over CQL statement kinds is added, together with best practices wrt preparing, batching and paging.

## Note to reviewers

Please make sure (by reading both crate docs and the docs book) that in all places it is clear enough that SELECTs by default should be done using `query_iter` rather than `query_unpaged`, and why so.

## Pre-review checklist

### DIFF
--- a/docs/source/queries/paged.md
+++ b/docs/source/queries/paged.md
@@ -80,7 +80,7 @@ On a `Query`:
 use scylla::query::Query;
 
 let mut query: Query = Query::new("SELECT a, b FROM ks.t");
-query.set_page_size(16.try_into().unwrap());
+query.set_page_size(16);
 
 let _ = session.query_iter(query, &[]).await?; // ...
 # Ok(())
@@ -99,7 +99,7 @@ let mut prepared: PreparedStatement = session
     .prepare("SELECT a, b FROM ks.t")
     .await?;
 
-prepared.set_page_size(16.try_into().unwrap());
+prepared.set_page_size(16);
 
 let _ = session.execute_iter(prepared, &[]).await?; // ...
 # Ok(())
@@ -121,7 +121,7 @@ use scylla::query::Query;
 use scylla::statement::{PagingState, PagingStateResponse};
 use std::ops::ControlFlow;
 
-let paged_query = Query::new("SELECT a, b, c FROM ks.t").with_page_size(6.try_into().unwrap());
+let paged_query = Query::new("SELECT a, b, c FROM ks.t").with_page_size(6);
 
 let mut paging_state = PagingState::start();
 loop {
@@ -165,7 +165,7 @@ use scylla::statement::{PagingState, PagingStateResponse};
 use std::ops::ControlFlow;
 
 let paged_prepared = session
-    .prepare(Query::new("SELECT a, b, c FROM ks.t").with_page_size(7.try_into().unwrap()))
+    .prepare(Query::new("SELECT a, b, c FROM ks.t").with_page_size(7))
     .await?;
 
 let mut paging_state = PagingState::start();
--- a/docs/source/queries/paged.md
+++ b/docs/source/queries/paged.md
@@ -2,9 +2,31 @@
 Sometimes query results might be so big that one prefers not to fetch them all at once,
 e.g. to reduce latency and/or memory footprint.
 Paged queries allow to receive the whole result page by page, with a configurable page size.
+In fact, most SELECTs queries should be done with paging, to avoid big load on cluster and large memory footprint.
 
-`Session::query_iter` and `Session::execute_iter` take a [simple query](simple.md)
-or a [prepared query](prepared.md) and return an `async` iterator over result `Rows`.
+> ***Warning***\
+> Issuing unpaged SELECTs (`Session::query_unpaged` or `Session::execute_unpaged`)
+> may have dramatic performance consequences! **BEWARE!**\
+> If the result set is big (or, e.g., there are a lot of tombstones), those atrocities can happen:
+> - cluster may experience high load,
+> - queries may time out,
+> - the driver may devour a lot of RAM,
+> - latency will likely spike.
+>
+> Stay safe. Page your SELECTs.
+
+## `RowIterator`
+
+The automated way to achieve that is `RowIterator`. It always fetches and enables access to one page,
+while prefetching the next one. This limits latency and is a convenient abstraction.
+
+> ***Note***\
+> `RowIterator` is quite heavy machinery, introducing considerable overhead. Therefore,
+> don't use it for statements that do not benefit from paging. In particular, avoid using it
+> for non-SELECTs.
+
+On API level, `Session::query_iter` and `Session::execute_iter` take a [simple query](simple.md)
+or a [prepared query](prepared.md), respectively, and return an `async` iterator over result `Rows`.
 
 > ***Warning***\
 > In case of unprepared variant (`Session::query_iter`) if the values are not empty
@@ -22,7 +44,6 @@ Use `query_iter` to perform a [simple query](simple.md) with paging:
 # use scylla::Session;
 # use std::error::Error;
 # async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
-use scylla::IntoTypedRows;
 use futures::stream::StreamExt;
 
 let mut rows_stream = session
@@ -45,7 +66,6 @@ Use `execute_iter` to perform a [prepared query](prepared.md) with paging:
 # use scylla::Session;
 # use std::error::Error;
 # async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
-use scylla::IntoTypedRows;
 use scylla::prepared_statement::PreparedStatement;
 use futures::stream::StreamExt;
 
@@ -106,10 +126,10 @@ let _ = session.execute_iter(prepared, &[]).await?; // ...
 # }
 ```
 
-### Passing the paging state manually
-It's possible to fetch a single page from the table, extract the paging state
-from the result and manually pass it to the next query. That way, the next
-query will start fetching the results from where the previous one left off.
+## Manual paging
+It's possible to fetch a single page from the table, and manually pass paging state
+to the next query. That way, the next query will start fetching the results
+from where the previous one left off.
 
 On a `Query`:
 ```rust
@@ -197,5 +217,18 @@ loop {
 ```
 
 ### Performance
-Performance is the same as in non-paged variants.\
 For the best performance use [prepared queries](prepared.md).
+See [query types overview](queries.md).
+
+## Best practices
+
+| Query result fetching   | Unpaged                                                                                                                 | Paged manually                                                                                       | Paged automatically                                                                               |
+|-------------------------|-------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------|
+| Exposed Session API     | `{query,execute}_unpaged`                                                                                               | `{query,execute}_single_page`                                                                        | `{query,execute}_iter`                                                                            |
+| Working                 | get all results in a single CQL frame, into a single Rust struct                                                        | get one page of results in a single CQL frame, into a single Rust struct                             | upon high-level iteration, fetch consecutive CQL frames and transparently iterate over their rows |
+| Cluster load            | potentially **HIGH** for large results, beware!                                                                         | normal                                                                                               | normal                                                                                            |
+| Driver overhead         | low - simple frame fetch                                                                                                | low - simple frame fetch                                                                             | considerable - `RowIteratorWorker` is a separate tokio task                                       |
+| Feature limitations     | none                                                                                                                    | none                                                                                                 | speculative execution not supported                                                               |
+| Driver memory footprint | potentially **BIG** - all results have to be stored at once!                                                            | small - only one page stored at a time                                                               | small - at most constant number of pages stored at a time                                         |
+| Latency                 | potentially **BIG** - all results have to be generated at once!                                                         | considerable on page boundary - new page needs to be fetched                                         | small - next page is always pre-fetched in background                                             |
+| Suitable operations     | - in general: operations with empty result set (non-SELECTs)</br> - as possible optimisation: SELECTs with LIMIT clause | - for advanced users who prefer more control over paging, with less overhead of `RowIteratorWorker`  | - in general: all SELECTs                                                                         |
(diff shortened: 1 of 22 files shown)


wprzytula:
NOTE: examples in `examples/*.rs` are still not adjusted this way. I suggest a separate PR for them.

fee-mendes (approved):
In general it LGTM. 

One thing I just realized is that `paged.md` still contains the old performance disclaimer. I thought `query_iter` vs the new `execute_unpaged` had their own caveats. 

Either way, I guess it is fine to follow-up somewhere else and it shouldn't block the existing work.

muzarski:
In `docs/execution-profiles/priority.md`:
```rust
let mut query = Query::from("SELECT * FROM ks.table");

// Query is not assigned any specific profile, so session's profile is applied.
// Therefore, the query will be executed with Consistency::One.
session.query_unpaged(query.clone(), ()).await?;

query.set_execution_profile_handle(Some(query_profile.into_handle()));
// Query's profile is applied.
// Therefore, the query will be executed with Consistency::Two.
session.query_unpaged(query.clone(), ()).await?;

query.set_consistency(Consistency::Three);
// An option is set directly on the query.
// Therefore, the query will be executed with Consistency::Three.
session.query_unpaged(query, ()).await?;
```

I'd use `Session::query_single_page` here.

muzarski (approved):
But otherwise, LGTM

wprzytula:
> In `docs/execution-profiles/priority.md`:
> 
> ```rust
> let mut query = Query::from("SELECT * FROM ks.table");
> 
> // Query is not assigned any specific profile, so session's profile is applied.
> // Therefore, the query will be executed with Consistency::One.
> session.query_unpaged(query.clone(), ()).await?;
> 
> query.set_execution_profile_handle(Some(query_profile.into_handle()));
> // Query's profile is applied.
> // Therefore, the query will be executed with Consistency::Two.
> session.query_unpaged(query.clone(), ()).await?;
> 
> query.set_consistency(Consistency::Three);
> // An option is set directly on the query.
> // Therefore, the query will be executed with Consistency::Three.
> session.query_unpaged(query, ()).await?;
> ```
> 
> I'd use `Session::query_single_page` here.

Wouldn't `query_single_page` (comparing to `query_iter`) introduce more noise, as it involves manual passing paging state?

fee-mendes:
I think it is fine. As long as we clarify the Performance section in paged.md as I pointed out earlier and discuss these points in depth there.

Probably that's how you want to introduce the pages queries doc

wprzytula:
> One thing I just realized is that `paged.md` still contains the old performance disclaimer. I thought `query_iter` vs the new `execute_unpaged` had their own caveats.

What disclaimer do you have in mind? All that I can see seem to be up-to-date.

muzarski:
> > In `docs/execution-profiles/priority.md`:
> > ```rust
> > let mut query = Query::from("SELECT * FROM ks.table");
> > 
> > // Query is not assigned any specific profile, so session's profile is applied.
> > // Therefore, the query will be executed with Consistency::One.
> > session.query_unpaged(query.clone(), ()).await?;
> > 
> > query.set_execution_profile_handle(Some(query_profile.into_handle()));
> > // Query's profile is applied.
> > // Therefore, the query will be executed with Consistency::Two.
> > session.query_unpaged(query.clone(), ()).await?;
> > 
> > query.set_consistency(Consistency::Three);
> > // An option is set directly on the query.
> > // Therefore, the query will be executed with Consistency::Three.
> > session.query_unpaged(query, ()).await?;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'd use `Session::query_single_page` here.
> 
> Wouldn't `query_single_page` (comparing to `query_iter`) introduce more noise, as it involves manual passing paging state?

That's true. My only concern was that we use `query_unpaged` for a `SELECT *` statement. But this doc wasn't actually created to showcase the `query_unpaged`/`query_iter` differences (the result is not even used here), so I think it's fine to leave it as `query_unpaged`.

muzarski:
> A general thought:
> 
> Now all SELECTs are recommended to be done using `{query,execute}_iter`. This means that all SELECTs would introduce considerable overhead of using the `RowIteratorWorker`, even if the SELECT only returns a single row. This isn't optimal.
> 
> However, let's stick to this approach, because it guarantees correctness wrt paging. In GRER (Giant Request Execution Refactor) we'll think about redesigning iteration over pages so that it's not that expensive, so by 0.17 we should have this solved. WDYT @fee-mendes @Lorak-mmk @muzarski ?

In examples refactor (that I'm currently working on), I decided to make use of `query_single_page` when we are sure that result will consist of a single row.

wprzytula:
> > A general thought:
> > Now all SELECTs are recommended to be done using `{query,execute}_iter`. This means that all SELECTs would introduce considerable overhead of using the `RowIteratorWorker`, even if the SELECT only returns a single row. This isn't optimal.
> > However, let's stick to this approach, because it guarantees correctness wrt paging. In GRER (Giant Request Execution Refactor) we'll think about redesigning iteration over pages so that it's not that expensive, so by 0.17 we should have this solved. WDYT @fee-mendes @Lorak-mmk @muzarski ?
> 
> In examples refactor (that I'm currently working on), I decided to make use of `query_single_page` when we are sure that result will consist of a single row (or when we want to showcase the `QueryResult` API).

Remember that even if the result is going to consist of a single row, in some situations (e.g. when there are a lot of tombstones - @fee-mendes please confirm) there can be a number of empty pages received before we finally get one with our expected row.

fee-mendes:
> Remember that even if the result is going to consist of a single row, in some situations (e.g. when there are a lot of tombstones - @fee-mendes please confirm) there can be a number of empty pages received before we finally get one with our expected row.

Correct. If the query is paged, empty replica pages will kick in. If it is unpaged, then you retrieve the full result set.

In general, it is hard to guarantee a SELECT will return only a single row, or a single page. `LIMIT 1`, aggregations (`COUNT()`), and key-value tables are probably the only exceptions.

wprzytula:
> I am reading [paged.md](https://github.com/scylladb/scylla-rust-driver/blob/main/docs/source/queries/paged.md) which says:
> 
> ```
> ### Performance
> Performance is the same as in non-paged variants.\
> For the best performance use [prepared queries](prepared.md).
> ```

Right. This particular statement was never true wrt `{query,execute}_iter`, which are always at least a bit less performant than `{query,execute}_{single_page,unpaged}`. I'll correct those recommendations as you suggest.

fee-mendes:
> Right. This particular statement was never true wrt `{query,execute}_iter`, which are always at least a bit less performant than `{query,execute}_{single_page,unpaged}`. I'll correct those recommendations as you suggest.

Well, be mindful though - that performance is subjective even in the `{query,execute}_{single_page,unpaged}` case. I'd argue that `single_page` is always more performant than `unpaged`, as it requires less memory footprint given that `unpaged` queries may return a large `QueryResult` when scanning large keys, not to mention the pressure on the server-side. 

Hence why I think the focus should shift from "Performance" to our recommended "Best Practices", with both client/server in mind.

docs/source/queries/result.md
```diff
@@ -3,6 +3,21 @@
 `Session::query_unpaged`, `Session::query_single_page`, `Session::execute_unpaged` and `Session::execute_single_page`
 return a `QueryResult` with rows represented as `Option<Vec<Row>>`.
 
+> ***Note***\
+> Using unpaged queries for SELECTs is discouraged in general.
+> Query results may be so big that it is not preferable to fetch them all at once.
+> However, `query_unpaged` will return all results in one, possibly giant, piece
+> (unless a timeout occurs due to high load incurred by the cluster).
```
Lorak-mmk, line 10: It would be good to add a note that even if results are small, there may be a lot of tombstones that will also increase processing time on DB side.

docs/source/queries/result.md
```diff
@@ -3,6 +3,21 @@
...
+> This:
+> - increases latency,
+> - has large memory footprint,
+> - puts high load on the cluster,
+> - is more likely to time out (because big work takes more time than little work,
+>   and returning one large piece of data is more work than returning one chunk of data).
+
+> To sum up, for SELECTs that may return return a lot of data prefer paged queries,
+> e.g. with `Session::query_iter()` (see [Paged queries](paged.md)).
+
```
Lorak-mmk, line 20: Maybe remove a remark about `a lot of data` because of the tombstone problem?

docs/source/queries/paged.md
```diff
@@ -80,7 +80,7 @@ On a `Query`:
 use scylla::query::Query;
 
 let mut query: Query = Query::new("SELECT a, b FROM ks.t");
-query.set_page_size(16.try_into().unwrap());
+query.set_page_size(16);
 
 let _ = session.query_iter(query, &[]).await?; // ...
 # Ok(())
```
Lorak-mmk, line 86: I'm wondering why this even passed CI. Shouldn't those examples fail to compile before this change?

docs/source/queries/result.md
wprzytula: Or maybe rephase it as `(especially those that may return a lot of data)`?

docs/source/queries/paged.md
wprzytula: There are blanket impls ;)
`impl TryFrom<T> for T { ... }`

wprzytula:
New commit: giant overview of statements and methods of executing them, together with best practices.

wprzytula:
Addresses Karol's comments.

docs/source/queries/simple.md
```diff
@@ -84,6 +84,19 @@ Each row can be parsed as a tuple of rust types using `rows_typed`:
...
+// However, `query_unpaged` will return all results in one, possibly giant, piece
+// (unless a timeout occurs due to high load incurred by the cluster).
+// This:
+// - increases latency,
+// - has large memory footprint,
+// - puts high load on the cluster,
+// - is more likely to time out (because big work takes more time than little work,
+//   and returning one large piece of data is more work than returning one chunk of data).
+// To sum up, for SELECTs that may return return a lot of data prefer paged queries,
+// e.g. with `Session::query_iter()`.
```
Lorak-mmk, line 98: This comment should be changed as was the one in `docs/source/queries/result.md`

wprzytula: You mean adding note about tombstones?

docs/source/queries/paged.md
```diff
@@ -197,5 +217,18 @@ loop {
...
+| Cluster load            | potentially **HIGH** for large results, beware!                                                                         | normal                                                                                               | normal                                                                                            |
+| Driver overhead         | low - simple frame fetch                                                                                                | low - simple frame fetch                                                                             | considerable - `RowIteratorWorker` is a separate tokio task                                       |
+| Feature limitations     | none                                                                                                                    | none                                                                                                 | speculative execution not supported                                                               |
```
Lorak-mmk, line 231: I'm wondering if we should print warning / erorr query if speculative execution is set on execution profile used in {query ,execute}_iter.

docs/source/queries/paged.md
```diff
@@ -197,5 +217,18 @@ loop {
...
+| Driver memory footprint | potentially **BIG** - all results have to be stored at once!                                                            | small - only one page stored at a time                                                               | small - at most constant number of pages stored at a time                                         |
+| Latency                 | potentially **BIG** - all results have to be generated at once!                                                         | considerable on page boundary - new page needs to be fetched                                         | small - next page is always pre-fetched in background                                             |
+| Suitable operations     | - in general: operations with empty result set (non-SELECTs)</br> - as possible optimisation: SELECTs with LIMIT clause | - for advanced users who prefer more control over paging, with less overhead of `RowIteratorWorker`  | - in general: all SELECTs                                                                         |
```
Lorak-mmk, line 234: Isn't SELECT with limit still susceptible to tombstones problem?

docs/source/queries/paged.md
wprzytula: Hmm, that might be a good idea, but OTOH forcing using different profiles for iter and non-iter execution and thus being inconvenient (for speculative execution users).

docs/source/queries/paged.md
wprzytula: @fee-mendes ?

docs/source/queries/queries.md
```diff
@@ -1,26 +1,80 @@
...
+| Memory footprint      | potentially big - all results have to be stored at once                                                                 | small - at most constant number of pages are stored by the driver at the same time                                                                                   |
+| Latency               | potentially big - all results have to be generated at once                                                              | small - at most one chunk of data must be generated at once, so latency of each chunk is small                                                                       |
+| Suitable operations   | - in general: operations with empty result set (non-SELECTs)</br> - as possible optimisation: SELECTs with LIMIT clause | - in general: all SELECTs                                                                                                                                            |
+
+For more detailed comparison and more best practices, see [doc page about paging](paged.md).
```
Lorak-mmk, line 72: ditto about LIMIT

docs/source/queries/simple.md
Lorak-mmk: Yes

docs/source/queries/paged.md
Lorak-mmk: I worry about situation where user wants to use speculative execution,  but silently doesn't because of `_iter` methods. You raised good point, I'm not sure how to best handle this possible issue.

docs/source/queries/queries.md
```diff
@@ -1,26 +1,80 @@
...
+
+Driver supports all kinds of statements supported by ScyllaDB. The following tables aim to bridge between DB concepts and driver's API.
+They include recommendations on which API to use in what cases.
+
+## Kinds of CQL statements (from the CQL protocol point of view):
+
+| Kind of CQL statement | Single              | Batch                                    |
+|-----------------------|---------------------|------------------------------------------|
+| Prepared              | `PreparedStatement` | `Batch` filled with `PreparedStatement`s |
+| Unprepared            | `Query`             | `Batch` filled with `Query`s             |
```
muzarski, line 11: Hmm, maybe it's worth mentioning that batch unprepared statements are being prepared by the driver if their bound value list is not empty. Actually, the same applies to single unprepared queries.

muzarski (approved):

fee-mendes (approved):

PR #1072, opened 2024-09-16
clippy: map -> inspect when original item is returned

muzarski:
Fixing a clippy lints regarding a usage of `map()` instead of `inspect()`, when original item (closure argument) is returned from the closure.

## Pre-review checklist

- ~[ ] All commits compile, pass static checks and pass test.~


muzarski:
`min_rust` is not passing unfortunately. Should we simply ignore these clippy lints (`#[allow(clippy::manual_inspect)]`) or bump MSRV?

Lorak-mmk:
The error that I see in CI is that `inspect` is unstable on MSRV version. In that case you can stop using it: `let socket = socket.connect(real_addr).await; /* do something with socket */; socket`.

muzarski:
> The error that I see in CI is that `inspect` is unstable on MSRV version. In that case you can stop using it: `let socket = socket.connect(real_addr).await; /* do something with socket */; socket`.

Done

wprzytula (approved):

Lorak-mmk (approved):

PR #1074, opened 2024-09-18
errors: move non-frame error types to scylla crate and adjust their visibility

muzarski:
Depends on https://github.com/scylladb/scylla-rust-driver/pull/1067

## Motivation
Most of the error types defined in `scylla-cql::errors` module should not belong there. Only 3 types are directly used in `scylla-cql` crate (`DbError`, `OperationType` and `WriteType`). The rest can be safely moved to `scylla` crate. It would allow us to modify the visibility of error types much easier. In scylla-cql crate, all error types need to be public - we would need to adjust their pub-visibility via re-exporting in scylla crate which is a tedious work. In addition, it does not make much sense for e.g. `QueryError` to reside in `scylla-cql`.

## Changes
- moved most of the error types from `scylla_cql::errors` to `scylla::errors` module
- adjusted pub-visibility of some error types. Especially, made transition error types, such as `UserRequestError`, `pub(crate)`
- moved `DbError`, `OperationType` and `WriteType` types to `frame::response::error` module. There is no need for additional `scylla_cql::errors` module just for them
- removed `scylla_cql::errors` module altogether
- bonus: adjusted/added some docstrings for public error types
- bonus: added more context to `TranslationError`

## Questions
Q: In result, the users can now find non-frame error types under `scylla::errors` instead of `scylla::transport::errors`. I'm not sure whether it's a good change or not. Should I move the `errors` module to `scylla::transport` module, so we don't introduce too many API breaking changes for the users? 

A: We decided to retain the original path. I'm moving the `errors` module to `scylla::transport` in last commit.

## Pre-review checklist


wprzytula:
In result, the users can now find non-frame error types under scylla::errors instead of scylla::transport::errors. I'm not sure whether it's a good change or not. Should I move the errors module to scylla::transport module, so we don't introduce too many API breaking changes for the users?

Let's retain module path compatibility for now (i.e., please stick to `crate::transport::errors`). Once I revive _the modules refactor_, we'll clean up all module paths.

scylla/src/errors.rs
```diff
@@ -291,12 +296,15 @@ impl ConnectionError {
 }
 
 /// Error caused by failed address translation done before establishing connection
-#[derive(Debug, Copy, Clone, Error)]
+#[derive(Debug, Clone, Error)]
 pub enum TranslationError {
```
wprzytula: I think it may be beneficial in the future to have this marked as `non_exhaustive`. WDYT?

muzarski:
> > In result, the users can now find non-frame error types under scylla::errors instead of scylla::transport::errors. I'm not sure whether it's a good change or not. Should I move the errors module to scylla::transport module, so we don't introduce too many API breaking changes for the users?
> 
> Let's retain module path compatibility for now (i.e., please stick to `crate::transport::errors`). Once I revive _the modules refactor_, we'll clean up all module paths.

Done. I did it in a separate commit at the end. I hope it's acceptable, since rebasing and moving contents of each commit to other file would be a tedious work...

scylla/src/errors.rs
```diff
@@ -265,6 +265,22 @@ pub enum BadQuery {
     Other(String),
 }
 
+/// Invalid keyspace name given to `Session::use_keyspace()`
+#[derive(Debug, Error, Clone)]
+pub enum BadKeyspaceName {
```
wprzytula: Let's consider adding `non_exhaustive` here. I don't immediately see any possible extension with another variant here, though.

scylla-cql/src/errors.rs
```diff
@@ -277,37 +277,6 @@ pub enum WriteType {
     Other(String),
 }
```
wprzytula: Typo in commit message: `request: move CqlRequestKind enum to request module `
`belond` -> `belong`

scylla-cql/src/frame/response/mod.rs
```diff
@@ -10,13 +10,43 @@ use std::sync::Arc;
...
 use crate::frame::protocol_features::ProtocolFeatures;
 use crate::frame::response::result::ResultMetadata;
 use crate::frame::TryFromPrimitiveError;
 
 use super::frame_errors::CqlResponseParseError;
 
+/// Possible CQL responses received from the server
+#[derive(Debug, Copy, Clone)]
+#[non_exhaustive]
+pub enum CqlResponseKind {
```
wprzytula, line 22: Is this (and `CqlRequestKind`) going to remain `pub`? Does it have to?

scylla-cql/src/frame/response/error.rs
```diff
@@ -1,9 +1,10 @@
-use crate::errors::{DbError, OperationType, WriteType};
```
wprzytula: Typo in `errors: remove scylla-cql::errors module` commit msg:
`coresponds` -> `corresponds`

scylla/src/errors.rs
```diff
@@ -362,32 +385,60 @@ pub struct ConnectionSetupRequestError {
...
+    /// An error returned by [`AuthenticatorSession::evaluate_challenge`](crate::authentication::AuthenticatorSession::evaluate_challenge).
     #[error("Failed to evaluate auth challenge on client side: {0}")]
     AuthChallengeEvaluationError(AuthError),
+
+    /// An error returned by [`AuthenticatorSession::success`](crate::authentication::AuthenticatorSession::success).
     #[error("Failed to finish auth challenge on client side: {0}")]
     AuthFinishError(AuthError),
+
+    /// User did not provide authentication. See [`SessionBuilder::user`](crate::transport::session_builder::SessionBuilder::user)
+    /// and/or [`SessionBuilder::authenticator_provider`](crate::transport::session_builder::SessionBuilder::authenticator_provider).
```
wprzytula: Suggested clarification:
`User did not provide authentication.` -> `User did not provide authentication while the cluster requires it.`

wprzytula (changes requested):

scylla/src/errors.rs
muzarski: Sure. What about `[Query/NewSession]Error`s?

scylla-cql/src/frame/response/mod.rs
muzarski: Yes, `CqlRequestKind` needs to be pub because of the `ConnectionSetupRequestError::request_kind` field. Same for `CqlResponseKind` and e.g. `ConnectionSetupRequestErrorKind::UnexpectedResponse` variant.

muzarski: Nvm, actually `ConnectionSetupRequestError::request_kind` is private. Should we make fields of this error type public?

wprzytula: I think that Display is enough public API for users in this case.

scylla/src/errors.rs
wprzytula: Them both definitely.

scylla-cql/src/frame/response/mod.rs
muzarski: `CqlRequestKind` and `CqlResponseKind` are public in scylla-cql. They are re-exported as `pub(crate)` in scylla, though.

muzarski:
Rebased on main.

scylla/src/errors.rs
```diff
@@ -354,8 +354,7 @@ impl ConnectionSetupRequestError {
 }
 
 #[derive(Error, Debug)]
-#[non_exhaustive]
-pub enum BrokenConnectionErrorKind {
+pub(crate) enum BrokenConnectionErrorKind {
     #[error("Timed out while waiting for response to keepalive request on connection to node {0}")]
     KeepaliveTimeout(IpAddr),
     #[error("Failed to execute keepalive query: {0}")]
```
Lorak-mmk: Commit: `errors: make BrokenConnectionErrorKind pub(crate) `

I don't agree with the reasoning in the commit message. I agree that using `Arc<dyn Error>` means that usually user should not care what the specific type is. Note the word "usually". Using `Arc<dyn Error>` is enough of a deterrent from using specific type - precisely because it requires downcasting and not just matching.
If however a user wants to use the specific type and match the reason they should be able to. And for that to be possible the error needs to be public (and non_exhaustive).

Lorak-mmk (changes requested):

muzarski:
v1.2: addressed @Lorak-mmk comments
- made `BrokenConnectionErrorKind`, `CqlEventHandlingError` and `RequestError` public
- provided docstrings for the types mentioned above
- adjusted error reason getters for `BrokenConnectionError`

Lorak-mmk:
I see you introduced a new method:
```
    /// Retrieve a typed error reason.
    pub fn get_reason_typed(&self) -> &BrokenConnectionErrorKind {
        self.0.downcast_ref::<BrokenConnectionErrorKind>().unwrap()
    }
```

I believe it should not be present.
1. It lifts the burden of downcasting from user, which is not the goal.
2. Prevents us from changing the type later. If we just return `Arc<dyn Error>` we can change the underlying error type later without breaking semver. With this method present that is no longer possible.

muzarski:
> I see you introduced a new method:
> 
> ```
>     /// Retrieve a typed error reason.
>     pub fn get_reason_typed(&self) -> &BrokenConnectionErrorKind {
>         self.0.downcast_ref::<BrokenConnectionErrorKind>().unwrap()
>     }
> ```
> 
> I believe it should not be present.
> 
>     1. It lifts the burden of downcasting from user, which is not the goal.
> 
>     2. Prevents us from changing the type later. If we just return `Arc<dyn Error>` we can change the underlying error type later without breaking semver. With this method present that is no longer possible.

Makes sense, I'll remove it.

muzarski:
> > I see you introduced a new method:
> > ```
> >     /// Retrieve a typed error reason.
> >     pub fn get_reason_typed(&self) -> &BrokenConnectionErrorKind {
> >         self.0.downcast_ref::<BrokenConnectionErrorKind>().unwrap()
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I believe it should not be present.
> > ```
> > 1. It lifts the burden of downcasting from user, which is not the goal.
> > 
> > 2. Prevents us from changing the type later. If we just return `Arc<dyn Error>` we can change the underlying error type later without breaking semver. With this method present that is no longer possible.
> > ```
> 
> Makes sense, I'll remove it.

Done

muzarski:
v1.3: fixed a docstring comment of `RequestError::UnableToAllocStreamId`.

Lorak-mmk (approved):

wprzytula (approved):

PR #1075, opened 2024-09-19
errors: fix driver's logic that bases on error variants returned from query execution

muzarski:

## Motivation
The error refactor that I'm currently working on, surprisingly changed driver's logic in some cases where the decision is made based on the error returned from query execution. There are three places where driver's logic was silently altered:
- retry_policy -> driver should retry execution in case of IoError
- use_keyspace -> if query execution returned IoErrors for some (not all) nodes/connections, these errors should be ignored
- speculative_exeuction -> driver should ignore IoErrors in this context

During error refactor I changed all places where `QueryError::IoError` was constructed. Now all of these IoErrors are represented as `BrokenConnectionError` and `ConnectionPoolError` types (and their variants). These types represent an error which implies that a corresponding connection/node is broken/unreachable.

## Changes
- Adjusted driver's logic in places mentioned above.
- Removed `[Query/NewSession]Error::IoError` variants altogether. They are no longer constructed anywhere - they were replaced by `ConnectionPoolError` and `BrokenConnection` variants during previous refactors.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -785,9 +785,9 @@ pub(crate) fn use_keyspace_result(
         match result {
             Ok(()) => was_ok = true,
             Err(err) => match err {
-                QueryError::IoError(_)
-                | QueryError::BrokenConnection(_)
-                | QueryError::ConnectionPoolError(_) => broken_conn_error = Some(err),
+                QueryError::BrokenConnection(_) | QueryError::ConnectionPoolError(_) => {
+                    broken_conn_error = Some(err)
+                }
                 _ => return Err(err),
             },
         }
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -772,20 +772,22 @@ impl ClusterWorker {
 pub(crate) fn use_keyspace_result(
     use_keyspace_results: impl Iterator<Item = Result<(), QueryError>>,
 ) -> Result<(), QueryError> {
-    // If there was at least one Ok and the rest were IoErrors we can return Ok
+    // If there was at least one Ok and the rest were broken connection errors we can return Ok
     // keyspace name is correct and will be used on broken connection on the next reconnect
 
-    // If there were only IoErrors then return IoError
-    // If there was an error different than IoError return this error - something is wrong
+    // If there were only broken connection errors then return broken connection error.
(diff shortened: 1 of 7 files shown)


scylla/src/transport/connection_pool.rs
```diff
@@ -1108,20 +1108,22 @@ impl PoolRefiller {
...
                 match result {
                     Ok(()) => was_ok = true,
                     Err(err) => match err {
-                        QueryError::IoError(io_err) => io_error = Some(io_err),
+                        QueryError::IoError(_)
+                        | QueryError::BrokenConnection(_)
+                        | QueryError::ConnectionPoolError(_) => broken_conn_error = Some(err),
                         _ => return Err(err),
                     },
                 }
```
wprzytula: This logic looks duplicated - see another place modified in this commit, in `cluster.rs`. Can we somehow deduplicate it?

Lorak-mmk:
The problems you mentioned sound like serious bugs. Are they present in 0.14 or only on main?

wprzytula:
> The problems you mentioned sound like serious bugs. Are they present in 0.14 or only on main?

From what I see, they are directly caused by PRs merged in recent days.

scylla/src/transport/connection_pool.rs
muzarski: I think some iterator API could be used here to make it less verbose.

muzarski: Or maybe not. I'll just extract it to function

muzarski:
v1.1: Deduplicated the logic shared between `ClusterWorker` and `PoolRefiller`

scylla/src/transport/cluster.rs
```diff
@@ -788,3 +764,35 @@ impl ClusterWorker {
...
+            },
+        }
+    }
+
+    if was_ok {
+        return Ok(());
+    }
+
+    // We can unwrap io_error because use_keyspace_results must be nonempty
+    Err(QueryError::IoError(io_error.unwrap()))
```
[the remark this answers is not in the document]
wprzytula: BTW, it's a shame that we don't yet have NonEmptyIterator or NonEmptyVec types in stdlib, especially that we already have NonZero unsigned integers.

wprzytula (approved):

Lorak-mmk (approved):

PR #1077, opened 2024-09-21
Fix clippy `expect_used` complains for `derive(FromRow)` macro

Mr-Leshiy:
# Description

Disables `clippy::expect_used` clippy rule for the `derive(FromRow)` macro.
As right now inside `derive(FromRow)` definition, as a part of the produced by this macro code `expect()` is used,
so it also violates and affects the clients code.
For that cases when the consumer of this library enabled `clippy::expect_used` for their project,
it requires to somehow fix it and disable it for the `derive(FromRow)` produced code.

## Dependent issue
https://github.com/input-output-hk/catalyst-voices/issues/828

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/cql_to_rust.rs
+++ b/scylla-cql/src/frame/response/cql_to_rust.rs
@@ -941,6 +941,11 @@ mod tests {
         );
     }
 
+    // Enabling `expect_used` clippy lint,
+    // validates that `derive(FromRow)` macro definition does do not violates such rule under the hood.
+    // Could be removed after such rule will be applied for the whole crate.
+    // <https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used>
+    #[deny(clippy::expect_used)]
     #[test]
     fn struct_from_row() {
         #[derive(FromRow)]
--- a/scylla-cql/src/frame/response/cql_to_rust.rs
+++ b/scylla-cql/src/frame/response/cql_to_rust.rs
@@ -941,6 +941,11 @@ mod tests {
         );
     }
 
+    // Enabling `expect_used` clippy lint,
+    // validates that `derive(FromRow)` macro definition does do not violates such rule under the hood.
+    // Could be removed after such rule will be applied for the whole crate.
+    // <https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used>
+    #[deny(clippy::expect_used)]
     #[test]
     fn struct_from_row() {
         #[derive(FromRow)]
@@ -965,6 +970,11 @@ mod tests {
         assert_eq!(my_row.c, Some(vec![1, 2]));
     }
 
+    // Enabling `expect_used` clippy lint,
+    // validates that `derive(FromRow)` macro definition does do not violates such rule under the hood.
+    // Could be removed after such rule will be applied for the whole crate.
+    // <https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used>
+    #[deny(clippy::expect_used)]
     #[test]
     fn struct_from_row_wrong_size() {
         #[derive(FromRow, PartialEq, Eq, Debug)]
@@ -1004,6 +1014,11 @@ mod tests {
         );
     }
 
+    // Enabling `expect_used` clippy lint,
+    // validates that `derive(FromRow)` macro definition does do not violates such rule under the hood.
+    // Could be removed after such rule will be applied for the whole crate.
+    // <https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used>
+    #[deny(clippy::expect_used)]
     #[test]
     fn unnamed_struct_from_row() {
         #[derive(FromRow)]
--- a/scylla-cql/src/frame/response/cql_to_rust.rs
+++ b/scylla-cql/src/frame/response/cql_to_rust.rs
@@ -941,11 +941,6 @@ mod tests {
         );
     }
 
-    // Enabling `expect_used` clippy lint,
-    // validates that `derive(FromRow)` macro definition does do not violates such rule under the hood.
-    // Could be removed after such rule will be applied for the whole crate.
-    // <https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used>
-    #[deny(clippy::expect_used)]
     #[test]
     fn struct_from_row() {
         #[derive(FromRow)]
--- a/scylla-macros/src/from_row.rs
+++ b/scylla-macros/src/from_row.rs
@@ -20,7 +20,9 @@ pub(crate) fn from_row_derive(tokens_input: TokenStream) -> Result<TokenStream, 
 
                 quote_spanned! {field.span() =>
                     #field_name: {
-                        // to avoid unnecessary copy `std::mem::take` is used
+                        // To avoid unnecessary copy `std::mem::take` is used.
+                        // Using explicit indexing operation is safe because `row_columns` is an array and `col_ix` is a litteral.
+                        // <https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/builtin/static.UNCONDITIONAL_PANIC.html>
                         let col_value = ::std::mem::take(&mut row_columns[#col_ix]);
                         <#field_type as FromCqlVal<::std::option::Option<CqlValue>>>::from_cql(col_value)
                             .map_err(|e| FromRowError::BadCqlVal {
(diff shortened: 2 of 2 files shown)


Lorak-mmk:
There was a recent PR in this area: https://github.com/scylladb/scylla-rust-driver/pull/1055

`expect_used` is allowed bu default - so for it to fail a user of the library has to enable it explicitly.
Enabling such lint means that a user does not want `expect` calls in the code - or at least wants to be warned about them and review them.
If we add such allows in produced code then users will not be warned about those calls, which defeats the purpose of this lint.

Mr-Leshiy:
I can't agree with you. Have a opposite opinion on this :slightly_smiling_face:.
Yes, the macro itself technically expanded and the macro code inserted into the place where this macros is used.
And as such macro defined by the library (as any other primitives like structs, function) from the client code it does not make any sense to know how it works and leaking potential linting problems, because from the client side it is just impossible to fix them.
As a result from the clients code to fulfil the linter as it was enabled (like in our case), it leads to the awful code like this:
```Rust
#[allow(clippy::expect_used)]
mod result {
    use scylla::FromRow;
    #[derive(FromRow)]
    pub(crate) struct YourStruct {
    }
}
use result:: YourStruct;
```

Generally for me seems, that the best solution would be eliminate the usage of the `.expect()` at all, but it will leads to the major refactor of this macro implementation. 
If you dont mind, I can try to modify this derive implementation.

But in general I would somehow fix this part, to reduce a headache on the clients side, because right its annoying, I can tell from our team experience :slightly_smiling_face: .

Lorak-mmk:
> @Lorak-mmk I can't agree with you. Have a opposite opinion on this 🙂. Yes, the macro itself technically expanded and the macro code inserted into the place where this macros is used. And as such macro defined by the library (as any other primitives like structs, function) from the client code it does not make any sense to know how it works and leaking potential linting problems, because from the client side it is just impossible to fix them.

If your goal in denying use of `expect` is to avoid panics, then library code is just as important as your own code (it can panic just as well) so it should be reviewed just as carefully.
It is not impossible to deal with badly behaved library:
- You can contribute the fix
- You can use your own fork
- You can decide to use different library
- You can rewrite part of the library yourself.

> As a result from the clients code to fulfil the linter as it was enabled (like in our case), it leads to the awful code like this:
> 
> ```rust
> #[allow(clippy::expect_used)]
> mod result {
>     use scylla::FromRow;
>     #[derive(FromRow)]
>     pub(crate) struct YourStruct {
>     }
> }
> use result:: YourStruct;
> ```

Yeah that's ugly :/ It would be much better if this worked:
```rust
    #[derive(Debug, FromRow)]
    #[allow(clippy::expect_used)]
    struct RowData {
    }
```

It would be good to allow this (unless it turns out to be impossible for some reason). After it is done it would be my recommended fix here instead of putting this `allow` in macro-generated code.
Why? Someone who is OK with `expect` in library code can add this directive - the cost is low, just 1 additional line of code.
Someone who is careful about `expect` in their code and library code (and would like to carefully review each such call) will not be warned about the presence of it at all if macro generated those `allow` directives - and there is no workaround for that.

> 
> In general, for me it seems that the best solution would be eliminate the usage of the `.expect()` at all, but it will leads to the major refactor of this macro implementation. If you dont mind, I can try to modify this derive implementation.

I agree that removing `exepect` / `unwrap` etc to completely avoid the possibility of panic is probably the best option. Is it feasible? I don't know, I don't remember this code that well. I can see that new macros (`DeserializeRow` / `DeserializeValue`) also use some `unwrap`'s and `expect`'s. @wprzytula do you think it's possible to avoid them completely?

> 
> But in general I would somehow fix this part, to reduce a headache on the clients side, because right its annoying, I can tell from our team experience 🙂 .

wprzytula:
> Yeah that's ugly :/ It would be much better if this worked:
> 
> ```rust
>     #[derive(Debug, FromRow)]
>     #[allow(clippy::expect_used)]
>     struct RowData {
>     }
> ```
> 
> It would be good to allow this (unless it turns out to be impossible for some reason). After it is done it would be my recommended fix here instead of putting this `allow` in macro-generated code.

Agreed. I support this, especially that it would allow solving other similar problems arising from macro expansion, too, by adding corresponding `#[allow]` attributes.
So what could be done (preferably by external contributors, as the maintainers are busy with higher-priority items now) is modifying `FromRow`, FromCqlVal`, `DeserializeRow`, `DeserializeValue` so that those macros put decorated struct definitions into a new module and decorate the module with #[allow] attributes specified after `[derive]`.

scylla-macros/src/from_row.rs
```diff
@@ -14,19 +14,18 @@ pub(crate) fn from_row_derive(tokens_input: TokenStream) -> Result<TokenStream,
...
                 let field_name = &field.ident;
                 let field_type = &field.ty;
 
                 quote_spanned! {field.span() =>
                     #field_name: {
-                        let (col_ix, col_value) = vals_iter
-                            .next()
-                            .expect("BUG: Size validated iterator did not contain the expected number of values"); 
+                        // to avoid unnecessary copy `std::mem::take` is used
+                        let col_value = ::std::mem::take(&mut row_columns[#col_ix]);
```
wprzytula, line 24: `expect` (a method that may panic) is replaced with indexing (which may panic as well). This defeats the purpose of `expect_used` lint, which is to keep the code free of possible panics.

wprzytula (changes requested):

scylla-macros/src/from_row.rs
Mr-Leshiy: I've used an array instead of iterator `let mut row_columns: [_; #fields_count]`, which allows for Rust compiler to detect such `index out of range` errors during the compile time.
So technically speaking if you have not disabled unconditional panic lint (which you definitely not), this code is safe.
And if in this code a bug will be appears it will be `99.9%` (`0.1%` stand when unconditional panic will be disabled, and I dont why someone really want to do that, its like to disable a borrow checker) catches by the compiler, with the detailed explanation what was incorrect.

E.g. I've modified this line of the code to
```Rust
let col_value = ::std::mem::take(&mut row_columns[#col_ix + 1]);
```
And rustc immediately complains
```
error: this operation will panic at runtime
    --> scylla-cql/src/frame/response/cql_to_rust.rs:1014:43
     |
1014 |         struct MyRow(i32, Option<String>, Option<Vec<i32>>);
     |                                           ^^^^^^ index out of bounds: the length is 3 but the index is 3
     |
     = note: `#[deny(unconditional_panic)]` on by default

error: could not compile `scylla-cql` (lib test) due to 1 previous error
```

https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/builtin/static.UNCONDITIONAL_PANIC.html

Mr-Leshiy: So to sum up, this code allows to catch an error on the compile time, using `.expect` does not allow to do that and will return it during the runtime only

wprzytula: You're right, it seems I have missed that you made `row_columns` an array. In such case indexing can be indeed checked by the compiler, with no panic risk in runtime if no warning is emitted about that.
Could you please write in comments why the array is used (to preserve length info in the type, to avoid panics) and why indexing won't panic (because array has a compile-time known length, and the index is a compile-time const, too)?

wprzytula:
@Mr-Leshiy Note that the new deserialization framework is going to soon replace the old one, with `DeserializeRow/Value` replacing `FromRow/CqlVal`. Derive macros for the former ones are going to need your refactor, too; I encourage you to try to get rid of packing there, too, in a follow-up.

scylla-macros/src/from_row.rs
Mr-Leshiy: @wprzytula done

Mr-Leshiy:
@wprzytula Yea, would be happy to help there as well.
But I would do that in the next PR if you dont mind

scylla-cql/src/frame/response/cql_to_rust.rs
```diff
@@ -941,11 +941,6 @@ mod tests {
         );
     }
 
-    // Enabling `expect_used` clippy lint,
-    // validates that `derive(FromRow)` macro definition does do not violates such rule under the hood.
-    // Could be removed after such rule will be applied for the whole crate.
-    // <https://rust-lang.github.io/rust-clippy/master/index.html#/expect_used>
-    #[deny(clippy::expect_used)]
```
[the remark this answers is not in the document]
Mr-Leshiy: Following our discussion, I agreed with the value of such possible linting errors which could appear under the macro definition.
So as we dont want to prevent using `expect` or `unwrap` under the macro definition, this line of code just does not make sense.

wprzytula (approved):

Lorak-mmk (approved):

PR #1082, opened 2024-10-02
make `ResultMetadata` lifetime-generic

wprzytula:
Ref: #462 

This is a necessary step towards non-allocating (in fact less-allocating) result metadata deserialization.

# What's done

- `ColumnType` is made ownership-generic; there will be two distinct functions to deserialize owned and borrowed `ColumnType`;
- `ColumnSpec` is made ownership-generic; in a follow-up PR, there will be a distinct function to deserialize borrowed `Vec<ColumnSpec<'frame>>`.
- `ResultMetadata` is made ownership-generic;
- `deser_table_spec` no longer allocates; it's up to `deser_col_specs` to allocate or not, depending on desired ownership of deserialized metadata;
- `deser_col_specs` verifies that table specs are the same for all columns. We assume it's true, but the CQL protocol does not enforce it. With that check, we will be able to hold only one table spec per query, not one per column; this will reduce memory footprint. TODO in a follow-up.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -424,9 +424,38 @@ impl CqlValue {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ColumnSpec {
-    pub table_spec: TableSpec<'static>,
-    pub name: String,
-    pub typ: ColumnType,
+    pub(crate) table_spec: TableSpec<'static>,
+    pub(crate) name: String,
+    pub(crate) typ: ColumnType,
+}
+
+impl ColumnSpec {
+    #[inline]
+    pub fn table_spec(&self) -> &TableSpec<'static> {
+        &self.table_spec
+    }
+
+    #[inline]
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    #[inline]
+    pub fn typ(&self) -> &ColumnType {
+        &self.typ
+    }
+}
+
+// Test utils for scylla crate.
+impl ColumnSpec {
+    #[doc(hidden)]
+    pub fn new_for_test(table_spec: TableSpec<'static>, name: String, typ: ColumnType) -> Self {
+        Self {
+            table_spec,
+            name,
+            typ,
+        }
+    }
 }
 
 #[derive(Debug, Clone)]
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -603,16 +603,6 @@ pub enum Result {
     SchemaChange(SchemaChange),
 }
 
-fn deser_table_spec(buf: &mut &[u8]) -> StdResult<TableSpec<'static>, TableSpecParseError> {
-    let ks_name = types::read_string(buf)
-        .map_err(TableSpecParseError::MalformedKeyspaceName)?
-        .to_owned();
-    let table_name = types::read_string(buf)
-        .map_err(TableSpecParseError::MalformedTableName)?
-        .to_owned();
-    Ok(TableSpec::owned(ks_name, table_name))
-}
-
 macro_rules! generate_deser_type {
     ($deser_type: ident, $l: lifetime, $read_string: expr) => {
         fn $deser_type<'frame>(
@@ -707,6 +697,16 @@ generate_deser_type!(deser_type_owned, 'static, |buf| types::read_string(buf).ma
 
 // This is going to be used for deserializing borrowed result metadata.
 generate_deser_type!(_deser_type_borrowed, 'frame, types::read_string);
+
+fn deser_table_spec(buf: &mut &[u8]) -> StdResult<TableSpec<'static>, TableSpecParseError> {
+    let ks_name = types::read_string(buf)
+        .map_err(TableSpecParseError::MalformedKeyspaceName)?
+        .to_owned();
+    let table_name = types::read_string(buf)
+        .map_err(TableSpecParseError::MalformedTableName)?
+        .to_owned();
+    Ok(TableSpec::owned(ks_name, table_name))
+}
 
 fn mk_col_spec_parse_error(
     col_idx: usize,
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -45,8 +45,8 @@ pub struct TableSpec<'a> {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub enum ColumnType {
-    Custom(String),
+pub enum ColumnType<'frame> {
+    Custom(Cow<'frame, str>),
     Ascii,
     Boolean,
     Blob,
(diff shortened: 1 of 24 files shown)


scylla/src/statement/prepared_statement.rs
```diff
@@ -620,7 +621,7 @@ mod tests {
         let col_specs: Vec<_> = cols
             .into_iter()
             .enumerate()
-            .map(|(i, typ)| ColumnSpec::new_for_test(table_spec.clone(), format!("col_{}", i), typ))
+            .map(|(i, typ)| ColumnSpec::owned(format!("col_{}", i), typ, table_spec.clone()))
```
muzarski, line 624: Why we decided not to use `new_for_test` here anymore? I actually like `ColumnSpec::owned` here more, since we are explicit about creating an owned type.

muzarski: I think we should remove `new_for_test` - I checked out and don't see any usages of this function anymore (replaced with `CS::owned` and `CS::borrowed` calls in tests).

scylla-cql/src/frame/response/result.rs
```diff
@@ -536,12 +536,12 @@ impl<'a> ColumnSpec<'a> {
 }
 
 #[derive(Debug, Clone)]
-pub struct ResultMetadata {
+pub struct ResultMetadata<'a> {
     col_count: usize,
-    pub col_specs: Vec<ColumnSpec<'static>>,
```
muzarski: nit: `pub` removal should belong to previous commit

scylla-cql/src/frame/response/result.rs
```diff
@@ -619,14 +619,30 @@ pub enum Result {
...
-fn deser_table_spec(buf: &mut &[u8]) -> StdResult<TableSpec<'static>, TableSpecParseError> {
-    let ks_name = types::read_string(buf)
-        .map_err(TableSpecParseError::MalformedKeyspaceName)?
-        .to_owned();
-    let table_name = types::read_string(buf)
-        .map_err(TableSpecParseError::MalformedTableName)?
-        .to_owned();
-    Ok(TableSpec::owned(ks_name, table_name))
+// Only allocates a new `TableSpec` if one is not yet given.
+fn deser_table_spec<'frame: 'spec, 'spec>(
```
muzarski, line 623: Can we close https://github.com/scylladb/scylla-rust-driver/pull/956 then?

muzarski: If we are going to store a borrowed `TableSpec` in each `ColumnSpec`, then I don't think we need to remove it from `ColumnSpec`.

scylla-cql/src/frame/response/result.rs
```diff
@@ -730,21 +746,30 @@ fn mk_col_spec_parse_error(
...
+        }
         let name = types::read_string(buf)
             .map_err(|err| mk_col_spec_parse_error(col_idx, err))?
             .to_owned();
         let typ = deser_type_owned(buf).map_err(|err| mk_col_spec_parse_error(col_idx, err))?;
-        col_specs.push(ColumnSpec::owned(name, typ, table_spec));
+        col_specs.push(ColumnSpec::owned(
+            name,
+            typ,
+            table_spec.as_ref().unwrap().clone(),
```
muzarski, line 771: nit: justify unwrap

muzarski (changes requested):

scylla-cql/src/frame/response/result.rs
muzarski: commit `result: make ResultMetadata lifetime generic` - `ResultMetadata` struct

scylla-cql/src/frame/response/result.rs
wprzytula: This is not the final state. In a follow-up of the deserialization refactor, I'm going to decouple `TableSpec` from `ColumnSpec`. The reason for doing this in a follow-up is to prove that such change won't break the public API, so an opposite change won't, too, which is what @Lorak-mmk believes is important in case Scylla/Cassandra once introduce joins to CQL.

muzarski: Ok,, makes sense

scylla-cql/src/frame/response/result.rs
```diff
@@ -542,87 +595,95 @@ fn deser_table_spec(buf: &mut &[u8]) -> StdResult<TableSpec<'static>, TableSpecP
...
+    ($deser_type: ident, $l: lifetime, $read_string: expr) => {
+        fn $deser_type<'frame>(
+            buf: &mut &'frame [u8],
+        ) -> StdResult<ColumnType<$l>, CqlTypeParseError> {
+            use ColumnType::*;
+            let id = types::read_short(buf)
+                .map_err(|err| CqlTypeParseError::TypeIdParseError(err.into()))?;
+            Ok(match id {
+                0x0000 => {
+                    let type_str = types::read_string(buf)
```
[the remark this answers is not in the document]
wprzytula: This is done on purpose. Chances are the underlying string is ...DurationType, in which case we don't need to allocate it at all. Only for Custom types (which we don't support anyway) do we need to allocate. OTOH, the macro argument function deserializes borrowed OR owned string; here we want to always deserialize borrowed string.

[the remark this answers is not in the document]
muzarski: > This is done on purpose. Chances are the underlying string is ...DurationType,

Right, this makes sense to me now

scylla/src/statement/prepared_statement.rs
```diff
@@ -513,13 +513,14 @@ impl From<SerializationError> for PartitionKeyError {
...
 }
 
-pub(crate) type PartitionKeyValue<'ps> = (&'ps [u8], &'ps ColumnSpec);
+pub(crate) type PartitionKeyValue<'ps, 'spec> = (&'ps [u8], &'ps ColumnSpec<'spec>);
 
-pub(crate) struct PartitionKey<'ps> {
-    pk_values: SmallVec<[Option<PartitionKeyValue<'ps>>; PartitionKey::SMALLVEC_ON_STACK_SIZE]>,
+pub(crate) struct PartitionKey<'ps, 'spec: 'ps> {
+    pk_values:
+        SmallVec<[Option<PartitionKeyValue<'ps, 'spec>>; PartitionKey::SMALLVEC_ON_STACK_SIZE]>,
```
Lorak-mmk, line 520: Is there a reason for having 2 lifetimes? I think this type is covariant over both 'ps and 'spec, so it should be fine to unify those lifetimes - but maybe I'm just talking gibberish, idk. @piodul ?

scylla-cql/src/frame/response/result.rs
```diff
@@ -558,6 +558,16 @@ impl ResultMetadata {
...
+
+    #[inline]
+    pub fn col_count(&self) -> usize {
+        self.col_count
+    }
+
+    #[inline]
+    pub fn col_specs(&self) -> &[ColumnSpec] {
+        &self.col_specs
+    }
```
Lorak-mmk, line 570: Could you use `ColumnSpec<'_>` to signify that lifetime is not static?

scylla-cql/src/frame/response/result.rs
Lorak-mmk: +1, it's not obvious to me that `col_count` must be greater than 0.

Lorak-mmk (changes requested):

scylla-cql/src/frame/response/result.rs
```diff
@@ -61,23 +61,72 @@ pub enum ColumnType {
...
+            ColumnType::Time => ColumnType::Time,
+            ColumnType::Timeuuid => ColumnType::Timeuuid,
+            ColumnType::Tuple(vec) => {
+                ColumnType::Tuple(vec.into_iter().map(ColumnType::into_owned).collect())
+            }
+            ColumnType::Uuid => ColumnType::Uuid,
+            ColumnType::Varint => ColumnType::Varint,
+        }
+    }
+}
```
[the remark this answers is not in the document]
wprzytula: AFAIK this function is only going to be used in error conditions, to create owned `ColumnType` for use in error types, so the waste does not matter.

scylla-cql/src/frame/value_tests.rs
```diff
@@ -1079,7 +1079,7 @@ fn vec_value_list() {
     );
 }
 
-fn col_spec(name: &str, typ: ColumnType) -> ColumnSpec {
+fn col_spec(name: &str, typ: ColumnType<'static>) -> ColumnSpec {
     ColumnSpec {
         table_spec: TableSpec::owned("ks".to_string(), "tbl".to_string()),
```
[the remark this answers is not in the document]
wprzytula: This could be done with any lifetime-generic type that we have in our codebase, and there are quite a number of them. Should we introduce such an alias for each of them?
I'm a bit reluctant about type aliases, because they tend to hide that the underlying type is the same **yet** differs wrt ownership. The thing is, we want to expose that fact.

scylla-cql/src/frame/response/result.rs
```diff
@@ -698,7 +722,7 @@ fn deser_col_specs(
     buf: &mut &[u8],
     global_table_spec: &Option<TableSpec<'static>>,
     col_count: usize,
-) -> StdResult<Vec<ColumnSpec>, ColumnSpecParseError> {
+) -> StdResult<Vec<ColumnSpec<'static>>, ColumnSpecParseError> {
     let mut col_specs = Vec::with_capacity(col_count);
     for col_idx in 0..col_count {
         let table_spec = if let Some(spec) = global_table_spec {
```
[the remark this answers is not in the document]
wprzytula: (already answered elsewhere)

scylla-cql/src/frame/response/result.rs
wprzytula: Actually, what lifetime does the current syntax imply? How does it differ from `'_` ?

Lorak-mmk: Current syntax does imply `'_` lifetime - I did not ask for a change in semantics, but for highlighting current semantics.

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -1146,16 +1147,16 @@ impl_tuple_multiple!(
...
 /// - `Some(Some(...))` - non-null, present value
 pub struct UdtIterator<'frame> {
-    all_fields: &'frame [(String, ColumnType)],
+    all_fields: &'frame [(Cow<'frame, str>, ColumnType<'frame>)],
     type_name: &'frame str,
     keyspace: &'frame str,
-    remaining_fields: &'frame [(String, ColumnType)],
+    remaining_fields: &'frame [(Cow<'frame, str>, ColumnType<'frame>)],
     raw_iter: BytesSequenceIterator<'frame>,
 }
```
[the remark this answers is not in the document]
wprzytula: This is a valid question. `ColumnType::UserDefinedType` contains `Vec<(Cow<'frame, str>, ColumnType<'frame>)>`, so when borrowing from that Vec, we get `&'frame [(Cow<'frame, str>, ColumnType<'frame>)]`, which can't be cast to `[(&'frame str, ColumnType<'frame>)]. So, the answer is that `Cow`s have to be propagated.

scylla-cql/src/frame/response/result.rs
wprzytula: Good that you noticed that bad practice. I refactored the code much.

scylla-cql/src/frame/response/result.rs
```diff
@@ -424,9 +424,38 @@ impl CqlValue {
...
+    pub(crate) name: String,
+    pub(crate) typ: ColumnType,
+}
+
+impl ColumnSpec {
+    #[inline]
+    pub fn table_spec(&self) -> &TableSpec {
+        &self.table_spec
+    }
+
```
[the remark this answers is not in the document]
wprzytula: This is a bug. Thanks for noticing that. It should be `TableSpec<'static>`.

wprzytula:
v2.0:
- refactored `deser_table_spec`/`deser_col_specs` tandem again to remove `unwrap()` and for more sense in responsibility division,
- addressed comments.

wprzytula:
In the PR description, I listed out the changes introduced.

Lorak-mmk (approved):

muzarski (approved):

PR #1083, opened 2024-10-04
Internal error matches no wildcard

muzarski:
ref: https://github.com/scylladb/scylla-rust-driver/issues/519

Since last time, during error refactor I introduced a silent
bug to the code (https://github.com/scylladb/scylla-rust-driver/pull/1075),
I'd like to prevent that from happening in the future. This is why I replace `_` matches
with explicit error variants.
This way, if some developer introduces a new variant to `QueryError`,
the compiler will complain about it in multiple places, forcing the developer to adjust `match` expressions there.

Notice that `LatencyAwareness::reliable_latency_measure` matched errors this way even before this PR.

Discussion:
I think it's a great place to discuss what decision should be made based on the received error in following modules/places:
- `LatencyAwareness::reliable_latency_measure`
- receiving `USE KEYSPACE` result from a single connection/node
- `DefaultRetrySession::decide_should_retry` (same for `DowngradingConsistencyRetrySession`)
- deciding whether the error received after speculative execution should be ignored

I did not change the logic at all yet, simply transformed the code to use explicit match expressions.
My first suspect is `QueryError::TimeoutError` in `can_be_ignored` in `speculative_execution` module. I believe we should 
return true for this error only iff we return true for `QueryError::RequestTimeout`.
  


scylla/src/transport/downgrading_consistency_retry_policy.rs
```diff
@@ -171,7 +171,18 @@ impl RetrySession for DowngradingConsistencyRetrySession {
...
+            | QueryError::CqlRequestSerialization(_)
+            | QueryError::BodyExtensionsParseError(_)
+            | QueryError::EmptyPlan
+            | QueryError::CqlResultParseError(_)
+            | QueryError::CqlErrorParseError(_)
+            | QueryError::ProtocolError(_)
+            | QueryError::TimeoutError
+            | QueryError::RequestTimeout(_)
+            // Remaining db errors
+            | QueryError::DbError(_, _) => RetryDecision::DontRetry,
```
wprzytula, line 185: I think DB errors should not be matched with a wildcard, too.

scylla/src/transport/downgrading_consistency_retry_policy.rs
```diff
@@ -171,7 +171,18 @@ impl RetrySession for DowngradingConsistencyRetrySession {
...
-            _ => RetryDecision::DontRetry,
+            QueryError::BadQuery(_)
+            | QueryError::MetadataError(_)
+            | QueryError::CqlRequestSerialization(_)
+            | QueryError::BodyExtensionsParseError(_)
+            | QueryError::EmptyPlan
+            | QueryError::CqlResultParseError(_)
+            | QueryError::CqlErrorParseError(_)
+            | QueryError::ProtocolError(_)
+            | QueryError::TimeoutError
```
wprzytula, line 182: Should we really propagate `TimeoutError` instead of trying another node? Well, it's the weird error about `use_keyspace`...

scylla/src/transport/retry_policy.rs
```diff
@@ -208,7 +208,18 @@ impl RetrySession for DefaultRetrySession {
...
+            | QueryError::CqlRequestSerialization(_)
+            | QueryError::BodyExtensionsParseError(_)
+            | QueryError::EmptyPlan
+            | QueryError::CqlResultParseError(_)
+            | QueryError::CqlErrorParseError(_)
+            | QueryError::ProtocolError(_)
+            | QueryError::TimeoutError
+            | QueryError::RequestTimeout(_)
+            // Remaining db errors
+            | QueryError::DbError(_, _) => RetryDecision::DontRetry,
```
wprzytula, line 222: ditto

scylla/src/transport/errors.rs
```diff
@@ -100,6 +100,29 @@ pub enum QueryError {
     RequestTimeout(String),
 }
 
+impl QueryError {
+    pub(crate) fn is_connection_broken(&self) -> bool {
+        match self {
+            // Error variants that imply that some connection error appeared before/during exeuction.
```
wprzytula, line 106: typo: exeuction -> execution

Lorak-mmk:
> ref: #519
> 
> Since last time, during error refactor I introduced a silent bug to the code (#1075), I'd like to prevent that from happening in the future. This is why I replace `_` matches with explicit error variants. This way, if some developer introduces a new variant to `QueryError`, the compiler will complain about it in multiple places, forcing the developer to adjust `match` expressions there.

Idea: places where explicit match should be used could use a comment about it, and perhaps this lint: https://rust-lang.github.io/rust-clippy/master/index.html#/wildcard_enum_match_arm
That way a future contributor won't simplify the code by using wildcard.

> 
> Notice that `LatencyAwareness::reliable_latency_measure` matched errors this way even before this PR.
> ## Discussion:
> 
> I think it's a great place to discuss what decision should be made based on the received error in following modules/places:
> 

The most important thing is to estabilish a rule for each of the following places, so that contributors know how to categorize possible future variants - and perhaps fix bugs in categorization that we do.
General question: are we sure that if connection is broken then we always receive ConnectionBroken variant? It's not possible to receive some deserialization error because of stream ending too early, right?
I will assume that we always receive ConnectionBroken in that case - and also in the case that we can't write to socket because it was closed.

>     * `LatencyAwareness::reliable_latency_measure`
> 

The rule in the comments is imo reasonable, but a bit imprecise.
Some errors are not always fast or slow - and the comments don't specify where they should go.
I suspect that "slow" errors should be only those that are always "slow" - otherwise they can't be a reliable latency measure.

Another issue: what does it mean for the query to be "slow" here?
I don't get why "DbError::IsBootstrapping", "DbError::Unprepared" or "DbError::Unavailable" are fast, but "DbError::SyntaxError" or "DbError::Invalid" (and some others) are "slow".

Note: "DbError" in the slow branch should also not use wildcard.

>     * receiving `USE KEYSPACE` result from a single connection/node (see `use_keyspace_result` from cluster.rs)
> 

The rule in the comment of that function looks reasonable to me, and the current filtering matches this rule.
The invariant is that after the function returns all connections are either broken or will have new keyspace set.
If this invariant was kept in the upper layers then the behavior would be good: after `Session.use_keyspace` returns all new queries will be issued on connections with the new keyspace.
However it looks like this invariant is not kept in the upper layers - `PoolRefiller.use_keyspace` doesn't close connections that returned an error.

>     * `DefaultRetrySession::decide_should_retry` (same for `DowngradingConsistencyRetrySession`)
> 

Those are complicated - I don't think there can exist a general rule. I'll look at them later.

>     * deciding whether the error received after speculative execution should be ignored
> 

The rule here seems to be the we can ignore errors if the presence of the error doesn't mean other speculative tries will fail too.
As I mention below I think it is not possible to receive `QueryError::TimeoutError` or `QueryError::RequestTimeout` here - so we should not ignore them so they are propagated to the user if the assumption turns out to be wrong.

`QueryError::DbError`:
    - `SyntaxError` - can't ignore
    - `Invalid` - I'm not sure. When can it happen?
    - `AlreadyExists` - can't ignore
    - `FunctionFailure` - not sure if UDF failure is transient. @wmitros ?
    - `AuthenticationError` - Can this happen? I think not. We can't ignore it anyway
    - `Unauthorized` - Can't ignore
    - `ConfigError` - not sure. When can this happen?
    - `Unavailable` - may be a split brain situation etc. I think we can ignore.
    - `Overloaded` - We can ignore
    - `IsBootstrapping` - can ignore
    - `TruncateError` - not sure. When can this happen?
    - `ReadTimeout` - can ignore
    - `WriteTimeout` - can ignore
    - `ReadFailure` - I'm not sure when can this happen. Probably can ignore
    - `WriteFailure` - as above
    - `Unprepared` - hmm... this can happen if we tried to reprepare, but it is still unprepared. I think we can ignore - it may be preparable on another node
    - `ServerError` - I think we can ignore, but I'm not sure
    - `ProtocolError` - why is there protocol error here too? The one in QueryError is not enough? Anyway, we probably can't ignore
    - `RateLimitReached` - Can ignore
    - `Other` - this shouldn't happen, right? I think we should not ignore
`QueryError::BadQuery` - we can't ignore because query is bad always or never
`QueryError::CqlRequestSerialization` - same as above
`QueryError::BodyExtensionsParseError` - if we get parse / deserialization error then something went terribly wrong, no need to retry on another node
`QueryError::CqlResultParseError` - same as above
`QueryError::CqlErrorParseError` - same as above
`QueryError::ConnectionPoolError` - we can ignore, we just don't have connections to given node
`QueryError::ProtocolError` - we can't ignore, same as with serialization / deserialization error
`QueryError::InvalidMessage` - I don't know when it happens. Assuming it's similar to ProtocolError we can't ignore it.
`QueryError::TimeoutError` - shouldn't happen, so we can't ignore
`QueryError::BrokenConnection` - we can ignore
`QueryError::UnableToAllocStreamId` - we can ignore
`QueryError::RequestTimeout` - shouldn't happen, so we can't ignore


> 
> I did not change the logic at all yet, simply transformed the code to use explicit match expressions. My first suspect is `QueryError::TimeoutError` in `can_be_ignored` in `speculative_execution` module. I believe we should return true for this error only iff we return true for `QueryError::RequestTimeout`.
> 

I'm fairly sure that `RequestTimeout` is applied to the whole execution, not individual speculative fibers, so it is not possible to receive it there.

muzarski:
> It's not possible to receive some deserialization error because of stream ending too early, right?

That's right. We always read the whole frame, and only then we start to deserialize the data. The only exception is frame header deserialization (see `FrameHeaderParseError`), but if such error appears, we close the connection anyway (i.e. it's included in `BrokenConnectionError`).

> The rule in the comments is imo reasonable, but a bit imprecise. Some errors are not always fast or slow - and the comments don't specify where they should go. I suspect that "slow" errors should be only those that are always "slow" - otherwise they can't be a reliable latency measure.
> 
> Another issue: what does it mean for the query to be "slow" here? I don't get why "DbError::IsBootstrapping", "DbError::Unprepared" or "DbError::Unavailable" are fast, but "DbError::SyntaxError" or "DbError::Invalid" (and some others) are "slow".

When I first looked at this code, my first was thought that "fast" errors are the errors that appear on driver's side, before sending the request to the server. I'm not sure why the `DbErrors` you mentioned are "fast" as well.


> - `Invalid` - I'm not sure. When can it happen? 

An example of this error that appears in cpp-rust-driver integration tests:
```
Invalid query: Database returned an error: 
The query is syntactically correct but invalid, 
Error message: No keyspace has been specified. 
USE a keyspace, or explicitly specify 
keyspace.tablename
```

I believe this should not be ignored - it will fail on other nodes as well.

> - `ProtocolError` - why is there protocol error here too? The one in QueryError is not enough? Anyway, we probably can't ignore 

This is a protocol error that was recognized by the server. The `QueryError::ProtocolError` are the protocol errors recognized by the driver. Anyway, I agree that we should not ignore it.

> - `QueryError::InvalidMessage` - I don't know when it happens. Assuming it's similar to ProtocolError we can't ignore it.
 
I removed this variant during protocol error refactor.

Lorak-mmk:
> When I first looked at this code, my first was thought that "fast" errors are the errors that appear on driver's side, before sending the request to the server. I'm not sure why the `DbErrors` you mentioned are "fast" as well.
> 

My first thought was that "fast" errors are errors on the driver side and error on server side that happen before it tries to contact other nodes (so can reply quickly, without actually performing the query).
I'm not sure this is correct.

> > * `Invalid` - I'm not sure. When can it happen?
> 
> An example of this error that appears in cpp-rust-driver integration tests:
> 
> ```
> Invalid query: Database returned an error: 
> The query is syntactically correct but invalid, 
> Error message: No keyspace has been specified. 
> USE a keyspace, or explicitly specify 
> keyspace.tablename
> ```
> 
> I believe this should not be ignored - it will fail on other nodes as well.
> 

I agree

muzarski:
v2:
- rebased on main (I apologize for that - should be done in a separate push, but I forgot that I already pulled some changes locally)
- enabled `wildcard_enum_match_arm` lint in some places (including latency awareness)
- using explicit matches for `DbErrors` (including latency awareness)
- fixed `can_be_ignored` function in speculative_execution module

scylla/src/transport/downgrading_consistency_retry_policy.rs
muzarski: Funny thing is that this error variant can't even appear there. AFAIU, `use keyspace` requests go through different path than regular requests. They are not a subject to retries, since we send `use keyspace` requests to all nodes simultaneously.

wprzytula (dismissed):

wprzytula (approved):

Issue #1085, opened 2024-10-09
Possible panic in speculative exection


Lorak-mmk:
There is a bug in `scylla::transport::speculative_execution::execute`
It is possible for all speculative fibers to finish without returning from the function.
Consider the scenario where all executions return `BrokenConnection`: we will only assign to `last_error` and never return.

There are more elaborate scenarios too:
- `retries_remaining` starts at 1, so we have 1 main execution and 1 additional speculative execution to spawn.
- Main execution does some retries and exhausts execution plan this way.
- It sends the last attempt to DB and waits for response.
- Now the speculative execution starts.
- The plan is empty, so it immediately finishes and enters `None` block of the `match`.
- `async_tasks` is not empty, because main execution is still ongoing, so nothing happens.
- Main execution finishes with ignorable error, e.g. broken connection.
- `can_be_ignored` returns true, so we don't return and only assign to `last_error`
- Now `async_tasks` is empty, so `async_tasks.select_next_some()` branch of `select!` will never be executed again.
- Only `&mut sleep` will be executed (once) and do nothing because `retries_remaining == 0`.
- That means we are hanging forever - or until client timeout happens if one was configured.

When all fibers finish and not return, then `select!` macro will panic because all its futures are completed.

Lorak-mmk:
Update: `select!` with only complete futures panics, so this would be a panic not a hang.
And now that I think of it there is simpler scenario where it happens: if all requests return BrokenConnection or another ignorable error.

PR #1086, opened 2024-10-09
Fix panic in speculative execution

Lorak-mmk:
This PR fixes linked bug by simplifying the logic inside the function.
The new logic always checks if the resolved task is the last one, and if so it returns, so it's not possible for the tasks to be exhausted without finishing the `select!`.

This PR also adds regression test for the issue. I verified that the test actually panics without the fix.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1085

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/retries.rs
+++ b/scylla/tests/integration/retries.rs
@@ -164,3 +164,71 @@ async fn retries_occur() {
         Err(err) => panic!("{}", err),
     }
 }
+
+// See https://github.com/scylladb/scylla-rust-driver/issues/1085
+#[tokio::test]
+#[ntest::timeout(30000)]
+#[cfg(not(scylla_cloud_tests))]
+async fn speculative_execution_panic_regression_test() {
+    use scylla_proxy::RunningProxy;
+
+    setup_tracing();
+    let test = |proxy_uris: [String; 3], translation_map, mut running_proxy: RunningProxy| async move {
+        let se = SimpleSpeculativeExecutionPolicy {
+            max_retry_count: 2,
+            retry_interval: Duration::from_millis(1),
+        };
+        let profile = ExecutionProfile::builder()
+            .speculative_execution_policy(Some(Arc::new(se)))
+            .retry_policy(Box::new(FallthroughRetryPolicy))
+            .build();
+        // DB preparation phase
+        let session: Session = SessionBuilder::new()
+            .known_node(proxy_uris[0].as_str())
+            .address_translator(Arc::new(translation_map))
+            .default_execution_profile_handle(profile.into_handle())
+            .build()
+            .await
+            .unwrap();
+
+        let ks = unique_keyspace_name();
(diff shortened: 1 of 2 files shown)


Lorak-mmk:
Oh, it looks like passing only completed futures to select! without complete branch causes panic instead of deadlock - in that case it may be possible to write a test.

wprzytula:
> Oh, it looks like passing only completed futures to select! without complete branch causes panic instead of deadlock - in that case it may be possible to write a test.

Good that they designed `select!` this way. Deadlocks are sad and much harder to debug.

Lorak-mmk:
I renamed the test, because the bug is a panic not a deadlock. PR is ready for review.

scylla/tests/integration/retries.rs
```diff
@@ -164,3 +164,70 @@ async fn retries_occur() {
...
+        );
+
+        running_proxy.running_nodes[0]
+            .change_request_rules(Some(vec![drop_connection_on_execute.clone()]));
+        running_proxy.running_nodes[1]
+            .change_request_rules(Some(vec![drop_connection_on_execute.clone()]));
+        running_proxy.running_nodes[2]
+            .change_request_rules(Some(vec![drop_connection_on_execute.clone()]));
+
+        let _result = session.execute_unpaged(&q, (2,)).await;
```
muzarski, line 222: Does it make sense to assert that we received an error (i.e. `unwrap_err()`)?

muzarski (approved):

scylla/tests/integration/retries.rs
```diff
@@ -164,3 +164,70 @@ async fn retries_occur() {
...
+        session
+            .query_unpaged("CREATE TABLE t (a int primary key)", &[])
+            .await
+            .unwrap();
+
+        let mut q = session
+            .prepare("INSERT INTO t (a) VALUES (?)")
+            .await
+            .unwrap();
+        q.set_is_idempotent(true); // this is to allow retry to fire
```
wprzytula, line 205: `retry` -> `speculative execution`

scylla/tests/integration/retries.rs
```diff
@@ -164,3 +164,70 @@ async fn retries_occur() {
...
+
+    setup_tracing();
+    let test = |proxy_uris: [String; 3], translation_map, mut running_proxy: RunningProxy| async move {
+        let se = SimpleSpeculativeExecutionPolicy {
+            max_retry_count: 2,
+            retry_interval: Duration::from_millis(1),
+        };
+        let profile = ExecutionProfile::builder()
+            .speculative_execution_policy(Some(Arc::new(se)))
+            .retry_policy(Box::new(FallthroughRetryPolicy));
```
wprzytula, line 183: Technically, this is a profile builder, not a profile. Maybe let's `build()` here, not later?

scylla/tests/integration/retries.rs
wprzytula: Yes, because otherwise this test gives us weaker guarantees about correctness of underlying mechanisms.

wprzytula (changes requested):

Lorak-mmk:
Addressed review comments

wprzytula (approved):
Awesome that you noticed the bug! Congrats!

PR #1087, opened 2024-10-10
errors: expose pub getters for type-erased errors

muzarski:
## Motivation

Sometimes, user might want to downcast `Arc<dyn Error>` to specific error type.

Actually, there is a use case for this in cpp-rust-driver, where we need to define custom `SerializeRow` implementation. It returns a custom error type. We would like to match against it during rust-to-C error conversion - downcasting is thus required.

See: https://github.com/scylladb/cpp-rust-driver/pull/187

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types/deserialize/mod.rs
+++ b/scylla-cql/src/types/deserialize/mod.rs
@@ -179,7 +179,6 @@ pub use row::DeserializeRow;
 pub use value::DeserializeValue;
 
 use std::error::Error;
-use std::fmt::Display;
 use std::sync::Arc;
 
 use thiserror::Error;
@@ -202,7 +201,7 @@ use thiserror::Error;
 ///   It won't be returned by the `Session` directly, but it might be nested
 ///   in the [`row::BuiltinTypeCheckError`].
 #[derive(Debug, Clone, Error)]
-#[error(transparent)]
+#[error("TypeCheckError: {0}")]
 pub struct TypeCheckError(pub(crate) Arc<dyn std::error::Error + Send + Sync>);
 
 impl TypeCheckError {
@@ -233,6 +232,7 @@ impl TypeCheckError {
 ///   It won't be returned by the `Session` directly, but it might be nested
 ///   in the [`row::BuiltinDeserializationError`].
 #[derive(Debug, Clone, Error)]
+#[error("DeserializationError: {0}")]
 pub struct DeserializationError(Arc<dyn Error + Send + Sync>);
 
 impl DeserializationError {
@@ -245,12 +245,6 @@ impl DeserializationError {
     /// Retrieve an error reason by downcasting to specific type.
     pub fn downcast_ref<T: Error + 'static>(&self) -> Option<&T> {
         self.0.downcast_ref()
-    }
-}
-
-impl Display for DeserializationError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "DeserializationError: {}", self.0)
     }
 }
 
--- a/scylla-cql/src/types/deserialize/mod.rs
+++ b/scylla-cql/src/types/deserialize/mod.rs
@@ -211,6 +211,11 @@ impl TypeCheckError {
     pub fn new(err: impl std::error::Error + Send + Sync + 'static) -> Self {
         Self(Arc::new(err))
     }
+
+    /// Retrieve an error reason by downcasting to specific type.
+    pub fn downcast_ref<T: std::error::Error + 'static>(&self) -> Option<&T> {
+        self.0.downcast_ref()
+    }
 }
 
 /// An error indicating that a failure happened during deserialization.
@@ -235,6 +240,11 @@ impl DeserializationError {
     #[inline]
     pub fn new(err: impl Error + Send + Sync + 'static) -> Self {
         Self(Arc::new(err))
+    }
+
+    /// Retrieve an error reason by downcasting to specific type.
+    pub fn downcast_ref<T: Error + 'static>(&self) -> Option<&T> {
+        self.0.downcast_ref()
     }
 }
 
--- a/scylla/src/transport/errors.rs
+++ b/scylla/src/transport/errors.rs
@@ -728,15 +728,15 @@ impl ConnectionSetupRequestError {
 pub struct BrokenConnectionError(Arc<dyn Error + Sync + Send>);
 
 impl BrokenConnectionError {
-    /// Retrieve an error reason.
-    pub fn get_reason(&self) -> &Arc<dyn Error + Sync + Send> {
-        &self.0
+    /// Retrieve an error reason by downcasting to specific type.
+    pub fn downcast_ref<T: Error + 'static>(&self) -> Option<&T> {
+        self.0.downcast_ref()
     }
 }
 
 /// A reason why connection was broken.
 ///
-/// See [`BrokenConnectionError::get_reason()`].
+/// See [`BrokenConnectionError::downcast_ref()`].
 /// You can retrieve the actual type by downcasting `Arc<dyn Error>`.
 #[derive(Error, Debug)]
 #[non_exhaustive]
--- a/scylla-cql/src/types/serialize/mod.rs
+++ b/scylla-cql/src/types/serialize/mod.rs
@@ -41,6 +41,11 @@ impl SerializationError {
     pub fn new(err: impl Error + Send + Sync + 'static) -> SerializationError {
         SerializationError(Arc::new(err))
     }
+
+    /// Retrieve an error reason by downcasting to specific type.
+    pub fn downcast_ref<T: Error + 'static>(&self) -> Option<&T> {
+        self.0.downcast_ref()
+    }
 }
 
 impl Display for SerializationError {
--- a/scylla-cql/src/types/serialize/mod.rs
+++ b/scylla-cql/src/types/serialize/mod.rs
@@ -2,7 +2,7 @@
 
 //! Types and traits related to serialization of values to the CQL format.
 
-use std::{error::Error, fmt::Display, sync::Arc};
+use std::{error::Error, sync::Arc};
 
 use thiserror::Error;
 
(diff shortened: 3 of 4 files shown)


scylla-cql/src/types/deserialize/mod.rs
```diff
@@ -211,6 +211,11 @@ impl TypeCheckError {
     pub fn new(err: impl std::error::Error + Send + Sync + 'static) -> Self {
         Self(Arc::new(err))
     }
+
+    /// Retrieve an error reason.
+    pub fn get_reason(&self) -> &Arc<dyn std::error::Error + Send + Sync> {
+        &self.0
+    }
```
wprzytula, line 218: Do we prefer to expose `&Arc<dyn...>` or `&dyn...`? In the former case, it's possible to clone the underlying error; no idea if it's better or not.

Lorak-mmk: With `&dyn` it is still possible to clone the outer error so I don't see the difference.
Downcast methods on `Error` are not defined for `Arc` anyway, only for `Box`, so user will have to use methods on refs anyway.

wprzytula: Let's expose bare `&dyn` then.

muzarski: `downcast_ref` method requires `T: 'static`. See: https://doc.rust-lang.org/std/error/trait.Error.html#method.downcast_ref-1. If we return `&dyn Error`, then result will not satisfy `'static` bound. I'm curious why does it work for `Arc<dyn Error>` though - I'd be glad for an explanation.

muzarski: I tried to change `BrokenConnectionError::get_reason` accordingly, but received a compile error (because of the reasons mentioned above).

wprzytula: `Arc` is `'static`, because its lifetime is independent of any external data. It manages its data on its own, so it's `'static` just like `Vec` or `String`.

muzarski: > I tried to change `BrokenConnectionError::get_reason` accordingly, but received a compile error (because of the reasons mentioned above).

Forgot to mention that this method is used in one test - compiler error appears there

Lorak-mmk: > `downcast_ref` method requires `T: 'static`. See: https://doc.rust-lang.org/std/error/trait.Error.html#method.downcast_ref-1. If we return `&dyn Error`, then result will not satisfy `'static` bound. I'm curious why it works for `Arc<dyn Error>` though - I'd be glad for an explanation.

This is a bound on T, not on the reference. So if you downcast `&dyn Error ` to `SomeError` (so the method returns `Option<&SomeError>` then `SomeError` must be `Send + 'static` - which it typically is. What is the code and error you are getting?

muzarski: > > `downcast_ref` method requires `T: 'static`. See: https://doc.rust-lang.org/std/error/trait.Error.html#method.downcast_ref-1. If we return `&dyn Error`, then result will not satisfy `'static` bound. I'm curious why it works for `Arc<dyn Error>` though - I'd be glad for an explanation.
> 
> This is a bound on T, not on the reference.

Yeah, I just noticed that `T` is a return type. There are no lifetime bounds on arguments

muzarski: Here is a snippet from test:
```rust
        // Wait until keepaliver gots impatient and terminates router.
        // Then, the error from keepaliver will be propagated to the error receiver.
        let err = error_receiver.await.unwrap();
        let err_inner: &BrokenConnectionErrorKind = match err {
            crate::transport::connection::ConnectionError::BrokenConnection(ref e) => {
                e.get_reason().downcast_ref().unwrap()
            }
            _ => panic!("Bad error type. Expected keepalive timeout."),
        };
        assert_matches!(err_inner, BrokenConnectionErrorKind::KeepaliveTimeout(_));
```

It works when `BrokenConnectionError` returns `Arc<dyn ...>`, but when I change the return type to `&dyn Error`, I get the following error:
```
error[E0597]: `err.0` does not live long enough
    --> scylla/src/transport/connection.rs:2782:77
     |
2782 |             crate::transport::connection::ConnectionError::BrokenConnection(ref e) => {
     |                                                                             ^^^^^ borrowed value does not live long enough
2783 |                 e.get_reason().downcast_ref().unwrap()
     |                 -------------- argument requires that `err.0` is borrowed for `'static`
...
2796 |     }
     |     - `err.0` dropped here while still borrowed
```

Lorak-mmk: What does it say if you add `drop(err);` after `assert_matches`?

muzarski: > What does it say if you add `drop(err);` after `assert_matches`?

"cannot move out of `err` because it is borrowed"

Lorak-mmk: Can you try to assign `e` to a variable first? So the snippet would look like this:
```
        // Wait until keepaliver gots impatient and terminates router.
        // Then, the error from keepaliver will be propagated to the error receiver.
        let err = error_receiver.await.unwrap();
        let err_inner = match err {
            crate::transport::connection::ConnectionError::BrokenConnection(ref e) => {
                e
            }
            _ => panic!("Bad error type. Expected keepalive timeout."),
        };
        let err_typed: &BrokenConnectionErrorKind = err_inner.get_reason().downcast_ref().unwrap();
        assert_matches!(err_typed, BrokenConnectionErrorKind::KeepaliveTimeout(_));
```

muzarski: The same error:
```
error[E0597]: `err.0` does not live long enough
    --> scylla/src/transport/connection.rs:2782:77
     |
2782 |             crate::transport::connection::ConnectionError::BrokenConnection(ref e) =...
     |                                                                             ^^^^^ borrowed value does not live long enough
...
2785 |         let err_typed: &BrokenConnectionErrorKind = err_inner.get_reason().downcast_...
     |                                                     ---------------------- argument requires that `err.0` is borrowed for `'static`
...
2795 |     }
     |     - `err.0` dropped here while still borrowed
```

Lorak-mmk: Wtf. 
What docs.rs show:
image
What is really there when you click "source":

Lorak-mmk: Wtf. 
What docs.rs show:
image
What is really there when you click "source":
image

Lorak-mmk: In that case, I think the best solution is to create our own downcast_ref method on TypeCheckError etc - that should be possible because inside we have access to Arc, right?

muzarski: I'll try it out. But I still don't really get it why it works for `Arc<dyn Error>`, while it does not for `&dyn Error`. AFAIU, what happens when we call `downcast_ref` on `Arc<dyn Error>` is value inside `Arc` is implicitly dereferenced first (i.e. `&dyn Error` is retrieved automatically via `Deref` trait). Then, `downcast_ref` is called on obtained `&dyn Error` reference. It seems almost identical to me.

muzarski:
v2: replaced public getters with custom `downcast_ref` implementations for type-erased error types.

wprzytula (approved):

scylla-cql/src/types/deserialize/mod.rs
Lorak-mmk: We should understand it and decide on how to best fix that before merging. I'll try to look into this today or during the weekend.

Lorak-mmk (approved):
Maybe not very related: Error impls / deriving seem to be a bit inconsistent:
- SerializationError / DeserializationError has derive(Error) + impl Display
- BrokenConnectionError additionally has `#[error("Connection broken, reason: {0}")]` and no manual Display impl
- TypeCheckError has `#[error(transparent)]` and no manual Display impl

I wonder if there is a good reason for those differences.

wprzytula:
> Maybe not very related: Error impls / deriving seem to be a bit inconsistent:
> 
>     * SerializationError / DeserializationError has derive(Error) + impl Display
> 
>     * BrokenConnectionError additionally has `#[error("Connection broken, reason: {0}")]` and no manual Display impl
> 
>     * TypeCheckError has `#[error(transparent)]` and no manual Display impl
> 
> 
> I wonder if there is a good reason for those differences.

`TypeCheckError` should be consistent with `DeserializationError`. If it's not, let's fix this.

Lorak-mmk (dismissed):

Lorak-mmk:
Sorry, one more thing: `TypeCheckError` is the only one of those 4 types that has its inside `pub(crate) ` instead of private. Can we change that?

muzarski:
> Sorry, one more thing: `TypeCheckError` is the only one of those 4 types that has its inside `pub(crate) ` instead of private. Can we change that? I doubt that with `downcast_ref` there is need to use this member directly by other places in the code.

This member is actually used directly in `scylla-cql/src/frame/response/result.rs` (on line 1052)

scylla/src/transport/errors.rs
```diff
@@ -734,6 +733,12 @@ impl BrokenConnectionError {
     }
 }
 
+impl std::fmt::Display for BrokenConnectionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Connection broken, reason: {}", self.0)
+    }
+}
+
```
wprzytula, line 741: Hmm, why this change from auto-generated Display to boilerplate'ish Display impl? Why not the other way round, so that all three error types use `#[error()]` shorthand?

muzarski: True.

wprzytula (approved):

muzarski:
Flaky test `transport::connection::tests::test_coalescing` failed yet again...

wprzytula:
> Flaky test transport::connection::tests::test_coalescing failed yet again...

Another time in a row.

Lorak-mmk:
Well, flakiness is one more reason to change this test

Lorak-mmk (approved):

PR #1088, opened 2024-10-11
New deserialization API - yet another bunch of preparations

wprzytula:
Ref: #462

In order to make the main deserialization PR at least a bit smaller, I extracted another bunch of preparatory changes.

## Pre-review checklist


scylla-cql/src/frame/response/result.rs
```diff
@@ -835,7 +853,7 @@ fn deser_result_metadata(
             .then(|| deser_table_spec(buf))
             .transpose()?;
 
-        deser_col_specs(buf, global_table_spec, col_count)?
+        deser_col_specs_owned(&mut *buf, global_table_spec, col_count)?
```
muzarski, line 856: nit: Why `buf` argument transformed to `&mut *buf`? I believe it could remain as `buf`. (`deser_result_metadata` function)

muzarski (approved):

scylla-cql/src/frame/response/result.rs
wprzytula: This is a leftover.

Lorak-mmk (approved):

PR #1090, opened 2024-10-14
scylla/lib.rs: Remove stars in serialization frameworks imports

Lorak-mmk:
It is easier to see what is imported, and decide if it should be, if one can see what exactly is imported. This is hard to do with stars, so, as part of API stabilization effort, those are removed here.

This PR also removes raw_batch re-export. This module should not be needed by users.

I verified using semver-checks that the first commit doesn't introduce any breaking changes.

## Pre-review checklist

- [x] ~~All commits compile, pass static checks and pass test.~~

### DIFF
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -150,15 +150,6 @@ pub mod serialize {
         };
     }
 
-    /// Contains the [RawBatchValues][raw_batch::RawBatchValues] and [RawBatchValuesIterator][raw_batch::RawBatchValuesIterator]
-    /// trait and their implementations.
-    pub mod raw_batch {
-        pub use scylla_cql::types::serialize::raw_batch::{
-            RawBatchValues, RawBatchValuesAdapter, RawBatchValuesIterator,
-            RawBatchValuesIteratorAdapter,
-        };
-    }
-
     /// Contains the [SerializeRow][row::SerializeRow] trait and its implementations.
     pub mod row {
         // Main types
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -63,8 +63,7 @@
 //! # Ok(())
 //! # }
 //! ```
-//! But the driver will accept anything implementing the trait [SerializeRow]
-//! (crate::serialize::row::SerializeRow)
+//! But the driver will accept anything implementing the trait [SerializeRow].
 //!
 //! ### Receiving results
 //! The easiest way to read rows returned by a query is to cast each row to a tuple of values:
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -131,8 +131,84 @@ pub mod frame {
 }
 
 /// Serializing bound values of a query to be sent to the DB.
+// Note: When editing comment on submodules here edit corresponding comments
+// on scylla-cql modules too.
 pub mod serialize {
-    pub use scylla_cql::types::serialize::*;
+    pub use scylla_cql::types::serialize::SerializationError;
+    /// Contains the [BatchValues][batch::BatchValues] and [BatchValuesIterator][batch::BatchValuesIterator] trait and their
+    /// implementations.
+    pub mod batch {
+        // Main types
+        pub use scylla_cql::types::serialize::batch::{
+            BatchValues, BatchValuesFromIterator, BatchValuesIterator,
+            BatchValuesIteratorFromIterator, TupleValuesIter,
+        };
+
+        // Legacy migration types - to be removed when removing legacy framework
+        pub use scylla_cql::types::serialize::batch::{
+            LegacyBatchValuesAdapter, LegacyBatchValuesIteratorAdapter,
+        };
+    }
+
+    /// Contains the [RawBatchValues][raw_batch::RawBatchValues] and [RawBatchValuesIterator][raw_batch::RawBatchValuesIterator]
+    /// trait and their implementations.
+    pub mod raw_batch {
+        pub use scylla_cql::types::serialize::raw_batch::{
+            RawBatchValues, RawBatchValuesAdapter, RawBatchValuesIterator,
+            RawBatchValuesIteratorAdapter,
+        };
+    }
+
+    /// Contains the [SerializeRow][row::SerializeRow] trait and its implementations.
+    pub mod row {
+        // Main types
+        pub use scylla_cql::types::serialize::row::{RowSerializationContext, SerializeRow};
+
+        // Errors
+        pub use scylla_cql::types::serialize::row::{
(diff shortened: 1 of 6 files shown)


scylla/src/lib.rs
```diff
@@ -131,8 +131,89 @@ pub mod frame {
 }
 
 /// Serializing bound values of a query to be sent to the DB.
+// I don't know any other way to tell rustfmt to not reorder
+// imports here.
+#[rustfmt::skip]
```
muzarski, line 136: I was only able to come up with something like this (we can't achieve the type per line formatting this way, though):
```rust
/// Serializing bound values of a query to be sent to the DB.
pub mod serialize {
    pub use scylla_cql::types::serialize::SerializationError;
    pub mod batch {
        // Main types
        pub use scylla_cql::types::serialize::batch::{
            BatchValues, BatchValuesFromIterator, BatchValuesIterator,
            BatchValuesIteratorFromIterator, TupleValuesIter,
        };

        // Legacy migration types - to be removed when removing legacy framework
        pub use scylla_cql::types::serialize::batch::{
            LegacyBatchValuesAdapter, LegacyBatchValuesIteratorAdapter,
        };
    }

    pub mod row {
        // Main types
        pub use scylla_cql::types::serialize::row::{RowSerializationContext, SerializeRow};

        // Errors
        pub use scylla_cql::types::serialize::row::{
            BuiltinTypeCheckError, BuiltinTypeCheckErrorKind,
        };

        // Legacy migration types - to be removed when removing legacy framework
        pub use scylla_cql::types::serialize::row::{
            // Legacy migration types - to be removed when removing legacy framework
            serialize_legacy_row,
            BuiltinSerializationError,
            BuiltinSerializationErrorKind,
        };

        // Not part of the old framework, but something that we should
        // still aim to remove.
        pub use scylla_cql::types::serialize::row::{
            SerializedValues, SerializedValuesIterator, ValueListAdapter,
            ValueListToSerializeRowAdapterError,
        };
    }

    pub mod value {
        // Main types
        pub use scylla_cql::types::serialize::value::SerializeValue;

        // Errors
        pub use scylla_cql::types::serialize::value::{
            BuiltinSerializationError, BuiltinSerializationErrorKind, BuiltinTypeCheckError,
            BuiltinTypeCheckErrorKind, MapSerializationErrorKind, MapTypeCheckErrorKind,
            SetOrListSerializationErrorKind, SetOrListTypeCheckErrorKind,
            TupleSerializationErrorKind, TupleTypeCheckErrorKind, UdtSerializationErrorKind,
            UdtTypeCheckErrorKind,
        };

        // Legacy migration types - to be removed when removing legacy framework
        pub use scylla_cql::types::serialize::value::{
            serialize_legacy_value, ValueAdapter, ValueToSerializeValueAdapterError,
        };
    }

    pub mod writers {
        pub use scylla_cql::types::serialize::writers::{
            CellOverflowError, CellValueBuilder, CellWriter, RowWriter, WrittenCellProof,
        };
    }
}
```

Lorak-mmk: I don't love the need to repeat `pub use scylla_cql::types::serialize::something` so many times, but it is probably the best solution for now. 

I'm starting to get really frustrated with rustfmt lately...

Lorak-mmk: I applied the sugestion.
Thenks to it I found that earlier reorders by rustfmt caused some types to land in the wrong category - I fixed that too.

scylla/src/lib.rs
```diff
@@ -130,9 +130,77 @@ pub mod frame {
     }
 }
 
+/// Serializing bound values of a query to be sent to the DB.
```
muzarski, line 133: Duplicated comment

muzarski (approved):

muzarski (approved):

scylla/src/lib.rs
```diff
@@ -131,8 +131,84 @@ pub mod frame {
 }
 
 /// Serializing bound values of a query to be sent to the DB.
+// Note: When editing comment on submodules here edit corresponding comments
+// on scylla-cql modules too.
 pub mod serialize {
-    pub use scylla_cql::types::serialize::*;
+    pub use scylla_cql::types::serialize::SerializationError;
+    /// Contains the [BatchValues][batch::BatchValues] and [BatchValuesIterator][batch::BatchValuesIterator] trait and their
+    /// implementations.
```
wprzytula, line 138: Oh, so `[][]` works as well? I only knew about `[]()`.

wprzytula (approved):

scylla/src/lib.rs
muzarski: > Actually you will notice that my third commit in this PR removes an occurence of `[]()` because it was rendered incorrectly.

It was not working because `[]` and `()` were on a separate line (separated by `//!`):
```rust
//! But the driver will accept anything implementing the trait [SerializeRow]
//! (crate::serialize::row::SerializeRow)
```

I even checked out locally to see, because at first glance, this looked correct to me.

Lorak-mmk: That makes sense. The version after my fix works too, right? At least it did when I checked locally.

muzarski: Yes, for me it renders correctly as well

Lorak-mmk: @muzarski do you know why 
```
[batch::BatchValues]
[`batch::BatchValues`]
```
don't work? This of course applies to all the other submodules in this place.

muzarski: I'm not really sure. But actually, I see that `[BatchValues][batch::BatchValues]` does not work as well...

image

Lorak-mmk: The colors in vscode are not working great. Something can be in brighter green and still not render correctly (or being in darker green and render correctly).
Check using cargo doc.

muzarski: Oh, wasn't aware of that.

muzarski: So it looks like things that are highlighted in vscode, do not render correctly for `cargo doc` and vice-versa...
image

wprzytula: `[]()` should work everywhere, so let's just stick to it. It's the ordinary Markdown format.

Lorak-mmk: > `[]()` should work everywhere, so let's just stick to it. It's the ordinary Markdown format.

I'll do it. Still it's sad that this is so difficult to get right.

Lorak-mmk: Hmm.. I can change it but both `[BatchValues](batch::BatchValues)` and `[BatchValues][batch::BatchValues]` are not highlighted in VSCode (and both render corrrectly), so there is no difference.

PR #1092, opened 2024-10-16
keepalive: trace in case of success

muzarski:
cpp-rust-driver tests sometimes depend on logs emitted by the driver. HeartbeatTests are on of them. Tests look for a following log: `logger_.add_critera("Heartbeat completed on host " + ccm_->get_ip_prefix());`.

Before this commit, rust-driver was not emitting any logs in case of keepalive request success. I think it's not a bad idea to add a trace log here, considering that keepalive intervals are rather long (default is 30s).

Obviously, the logger criteria will need to
be adjusted in `cpp-rust-driver` tests as well.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/connection.rs
+++ b/scylla/src/transport/connection.rs
@@ -1839,6 +1839,11 @@ impl Connection {
                     );
                     return Err(err);
                 }
+
+                trace!(
+                    "Keepalive request successful on connection to node {}",
+                    node_address
+                );
             }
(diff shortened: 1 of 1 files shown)


wprzytula (approved):

muzarski:
`Failed to apply group 0 change due to concurrent modification` for `retries::speculative_execution_panic_regression_test` ...

PR #1094, opened 2024-10-17
Fix unnamable types

Lorak-mmk:
This PR enables the `unnameable_types` lint. This lint catches types that are
- `pub`
- Used in some public API
- Not exported - and thus unusable by the user of the driver.

The only situation when this is desired that I know of is creating a sealed trait.
This PR fixes all occurrences of unnameable types and enables the lint so that
we don't introduce more in the future.

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -27,7 +27,7 @@ use tracing::{debug, warn};
 use uuid::Uuid;
 
 use super::locator::tablets::{RawTablet, Tablet, TabletsInfo};
-use super::node::{KnownNode, NodeAddr};
+use super::node::{InternalKnownNode, NodeAddr};
 use super::NodeRef;
 
 use super::locator::ReplicaLocator;
@@ -145,7 +145,7 @@ struct UseKeyspaceRequest {
 
 impl Cluster {
     pub(crate) async fn new(
-        known_nodes: Vec<KnownNode>,
+        known_nodes: Vec<InternalKnownNode>,
         pool_config: PoolConfig,
         keyspaces_to_fetch: Vec<String>,
         fetch_schema_metadata: bool,
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -59,12 +59,6 @@ impl<'a> std::fmt::Debug for ClusterNeatDebug<'a> {
     }
 }
 
-#[derive(Clone, Debug)]
-pub struct Datacenter {
-    pub nodes: Vec<Arc<Node>>,
-    pub rack_count: usize,
-}
-
 #[derive(Clone)]
 pub struct ClusterData {
     pub(crate) known_peers: HashMap<Uuid, Arc<Node>>, // Invariant: nonempty after Cluster::new()
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -257,18 +257,6 @@ impl Cluster {
 }
 
 impl ClusterData {
-    // Updates information about rack count in each datacenter
-    fn update_rack_count(datacenters: &mut HashMap<String, Datacenter>) {
-        for datacenter in datacenters.values_mut() {
-            datacenter.rack_count = datacenter
-                .nodes
-                .iter()
-                .filter_map(|node| node.rack.as_ref())
-                .unique()
-                .count();
-        }
-    }
-
     pub(crate) async fn wait_until_all_pools_are_initialized(&self) {
         for node in self.locator.unique_nodes_in_global_ring().iter() {
             node.wait_until_pool_initialized().await;
@@ -289,7 +277,6 @@ impl ClusterData {
         let mut new_known_peers: HashMap<Uuid, Arc<Node>> =
             HashMap::with_capacity(metadata.peers.len());
         let mut ring: Vec<(Token, Arc<Node>)> = Vec::new();
-        let mut datacenters: HashMap<String, Datacenter> = HashMap::new();
 
         for peer in metadata.peers {
             // Take existing Arc<Node> if possible, otherwise create new one
@@ -324,19 +311,6 @@ impl ClusterData {
             };
 
             new_known_peers.insert(peer_host_id, node.clone());
-
-            if let Some(dc) = &node.datacenter {
-                match datacenters.get_mut(dc) {
-                    Some(v) => v.nodes.push(node.clone()),
-                    None => {
-                        let v = Datacenter {
-                            nodes: vec![node.clone()],
-                            rack_count: 0,
-                        };
-                        datacenters.insert(dc.clone(), v);
-                    }
-                }
-            }
 
             for token in peer_tokens {
                 ring.push((token, node.clone()));
@@ -383,8 +357,6 @@ impl ClusterData {
                 &recreated_nodes,
             )
         }
-
-        Self::update_rack_count(&mut datacenters);
 
         let keyspaces = metadata.keyspaces;
         let (locator, keyspaces) = tokio::task::spawn_blocking(move || {
--- a/scylla/src/transport/cluster.rs
+++ b/scylla/src/transport/cluster.rs
@@ -381,25 +381,6 @@ impl ClusterData {
         &self.keyspaces
     }
 
-    /// Access datacenter details collected by the driver
-    /// Returned `HashMap` is indexed by names of datacenters
-    pub fn get_datacenters_info(&self) -> HashMap<String, Datacenter> {
-        self.locator
-            .datacenter_names()
-            .iter()
-            .map(|dc_name| {
-                let nodes = self
-                    .locator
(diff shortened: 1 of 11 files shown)


Lorak-mmk:
Fixed compilation errors when building without `cloud` feature.

scylla/src/transport/node.rs
```diff
@@ -242,8 +242,7 @@ pub struct CloudEndpoint {
 
 /// Describes a database server known on Session startup, with already resolved address.
 #[derive(Debug, Clone)]
-#[non_exhaustive]
-pub struct ResolvedContactPoint {
+pub(crate) struct ResolvedContactPoint {
     pub address: SocketAddr,
     pub datacenter: Option<String>,
 }
```
wprzytula, line 266: Let's not have any pub fields in non-pub structs.

scylla/src/transport/node.rs
```diff
@@ -227,15 +227,33 @@ impl Hash for Node {
 pub enum KnownNode {
     Hostname(String),
     Address(SocketAddr),
+}
+
+/// Describes a database server known on `Session` startup.
+/// It is similar to [KnownNode] but include also `CloudEndpoint` variant,
```
wprzytula, line 233: include -> includes

scylla/src/transport/node.rs
```diff
@@ -227,15 +227,33 @@ impl Hash for Node {
...
+
 /// Describes a database server in the serverless Scylla Cloud.
 #[cfg(feature = "cloud")]
-#[non_exhaustive]
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
-pub struct CloudEndpoint {
+pub(crate) struct CloudEndpoint {
     pub hostname: String,
     pub datacenter: String,
 }
```
wprzytula, line 259: ditto: make fields `pub(crate)`.

scylla/src/transport/session.rs
```diff
@@ -489,23 +489,35 @@ impl Session {
...
+        let cloud_known_nodes: Option<Vec<InternalKnownNode>> = None;
+
+        let known_nodes = cloud_known_nodes.unwrap_or_else(|| {
             known_nodes
-        };
+                .into_iter()
+                .map(|node| match node {
+                    KnownNode::Address(a) => InternalKnownNode::Address(a),
+                    KnownNode::Hostname(a) => InternalKnownNode::Hostname(a),
+                })
```
wprzytula, line 518: You perform manual conversion, even though you implemented `From`. Why not use it then?

scylla/src/transport/cluster.rs
```diff
@@ -257,18 +257,6 @@ impl Cluster {
...
-            datacenter.rack_count = datacenter
-                .nodes
-                .iter()
-                .filter_map(|node| node.rack.as_ref())
-                .unique()
-                .count();
-        }
-    }
-
     pub(crate) async fn wait_until_all_pools_are_initialized(&self) {
```
wprzytula, line 254: wow.

scylla/src/transport/cluster.rs
```diff
@@ -325,19 +312,6 @@ impl ClusterData {
...
-                        let v = Datacenter {
-                            nodes: vec![node.clone()],
-                            rack_count: 0,
-                        };
-                        datacenters.insert(dc.clone(), v);
-                    }
-                }
-            }
-
             for token in peer_tokens {
```
wprzytula, line 309: but...

scylla/src/transport/cluster.rs
```diff
@@ -384,8 +358,6 @@ impl ClusterData {
             )
         }
 
-        Self::update_rack_count(&mut datacenters);
-
         let keyspaces = metadata.keyspaces;
```
wprzytula, line 355: ...was that really pointless?! I can't believe it has always been... Could you please locate the commit that made `datacenters` unused?

scylla/src/transport/cluster.rs
```diff
@@ -381,25 +381,6 @@ impl ClusterData {
...
-                    .unwrap()
-                    .to_vec();
-                let rack_count = nodes.iter().map(|node| node.rack.as_ref()).unique().count();
-
-                (dc_name.clone(), Datacenter { nodes, rack_count })
-            })
-            .collect()
-    }
-
     /// Access details about nodes known to the driver
```
wprzytula, line 378: This is part of the "cluster examination" API. I agree with removing this particular method, because it's extremely wasteful, but I think you should include something similar (in the aspect of being a convenient way to examine the cluster) in the metadata refactor.

scylla/src/transport/cluster.rs
```diff
@@ -381,25 +381,6 @@ impl ClusterData {
         &self.keyspaces
     }
```
wprzytula, line 376: `ClusterData: remove get_datacenters_info()` commit message:
`trough` -> `through`

scylla/src/transport/session_builder.rs
```diff
@@ -29,6 +29,9 @@ use openssl::ssl::SslContext;
 use tracing::warn;
 
 mod sealed {
+    // This is a sealed trait - it's whole purpose is to be unnameable.
```
wprzytula, line 32: `it's` -> `its`

wprzytula (changes requested):

scylla/src/transport/session.rs
Lorak-mmk: Opps, I'll fix it.

scylla/src/transport/cluster.rs
Lorak-mmk: This code was introduced in this commit: https://github.com/scylladb/scylla-rust-driver/commit/8d82b043e1bc154330de78c183ea1d311ce4b2b9#diff-fc8ab4fd7742b17e4767ade6844fd697e5a7b0050bbadb414f52d19f6245bbac (PR: https://github.com/scylladb/scylla-rust-driver/pull/172 )
You can see that the result is saved in ClusterData.

Then the field in ClusterData was removed by the same author 2 years later: https://github.com/scylladb/scylla-rust-driver/commit/0db57c160924330cd1d92c418887e4feb4d4fd03 - but the calculation in ClusterData::new remained, introducing uselessness. PR: https://github.com/scylladb/scylla-rust-driver/pull/612

Lorak-mmk: I'm pretty sure that even thought there was no useless calculation before #612 this API was still useless because `Datacenter` struct was never truly public (= pub and exported)

scylla/src/transport/node.rs
Lorak-mmk: I wonder if there is a lint for that, let me check

Lorak-mmk: It looks like `unreachable_pub` should warn abut it. I wonder why it doesn't.

scylla/src/transport/cluster.rs
Lorak-mmk: I will think of a proper API and implementation during refactor.

Please note that this doesn't remove cluster examination API, only 1 (imo not very useful) method of it. The "primary" method, `get_nodes_info`, is still present.

scylla/src/transport/session_test.rs
```diff
@@ -1874,10 +1873,11 @@ async fn test_turning_off_schema_fetching() {
...
 
-    let datacenters: HashMap<String, Datacenter> = cluster_data.get_datacenters_info();
-    let datacenter_repfactors: HashMap<String, usize> = datacenters
-        .into_keys()
-        .map(|dc_name| (dc_name, 1))
+    let datacenter_repfactors: HashMap<String, usize> = cluster_data
+        .replica_locator()
+        .datacenter_names()
+        .iter()
+        .map(|dc_name| (dc_name.to_owned(), 1))
```
muzarski, line 1880: nit (commit `session_test: Don't use ClusterData::get_datacenters_info`): I believe commit message should be adjusted, as there is no call to `get_nodes_info()` and `unique()`.

muzarski (approved):

scylla/src/transport/session_test.rs
Lorak-mmk: Fixed.

scylla/src/transport/session_builder.rs
```diff
@@ -29,6 +29,10 @@ use openssl::ssl::SslContext;
 use tracing::warn;
 
 mod sealed {
+    // This is a sealed trait - its whole purpose is to be unnameable.
+    // This means we need to disable the check.
+    #[allow(unknown_lints)] // Rust 1.66 doesn't know this lint
```
wprzytula, line 34: We no longer support Rust 1.66. Does 1.70 know this lint?

Lorak-mmk: Unfortunately not - it was introduced in 1.79, so we would need to bump MSRV much higher for that.

wprzytula (approved):

PR #1098, opened 2024-10-21
Bump msrv to 1.70

Lorak-mmk:
Checking MSRV locally is frustrating, because it requires switching toolchain and renaming files, but it is even more frustrating because updating cargo index before 1.70 is so sloooooow.
1.70 switched to sparse cargo index protocol by default, which makes checking the crate much faster. Let's bump MSRV to improve developer experience.

I also added "rust-version" field to Cargo.toml files, because this is the official place to specify MSRV and in the future will be used by Cargo during dependency resolving.

1.70 is over a year old (released on 1st June 2023), so I doubt this bump will cause any problems.
Related: we will need to decide on MSRV policy past 1.0. I propose to settle on 6 month policy (= our MSRV must always be at least 6 months old). 
Rust ecosystem seems to not be very conservative on this:
- Tokio has the exact same MSRV policy that I propose here: https://github.com/tokio-rs/tokio
- time crate guarantess support of current stable and 2 previous versions (= minimum 4 months )
- chrono is a bit more conservative, with 1.61 MSRV: https://docs.rs/chrono/latest/chrono/

## Pre-review checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -14,7 +14,7 @@ env:
   CARGO_TERM_COLOR: always
   RUSTFLAGS: -Dwarnings
   RUST_BACKTRACE: full
-  rust_min: 1.66.0 # <- Update this when bumping up MSRV
+  rust_min: 1.70.0 # <- Update this when bumping up MSRV
 
 jobs:
   build:
--- a/README.md
+++ b/README.md
@@ -66,7 +66,12 @@ Ongoing efforts:
 Please join the `#rust-driver` channel on [ScyllaDB Slack] to discuss any issues or questions you might have.
 
 ## Supported Rust Versions
-Our driver's minimum supported Rust version (MSRV) is 1.70.0. Any changes will be explicitly published and will only happen during major releases.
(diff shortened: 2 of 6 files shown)


muzarski:
> I propose to settle on 6 month policy (= we only support Rust versions released during last 6 months)

Then we will not support versions that were released more than 6 months ago. You probably meant that MSRV should be a version that is at least 6 months old (AFAIU, this is the tokio's policy), correct? If so, then I agree that such policy makes sense.

Lorak-mmk:
> > I propose to settle on 6 month policy (= we only support Rust versions released during last 6 months)
> 
> Then we will not support versions that were released more than 6 months ago. You probably meant that MSRV should be a version that is at least 6 months old (AFAIU, this is the tokio's policy), correct? If so, then I agree that such policy makes sense.

Yes, you're right, and that is what I'd like to propose.

And just like Tokio, this wouldn't be automatic: if we don't see the benefit in bumping MSRV we can stay on older one for much longer.

muzarski (approved):

Lorak-mmk:
@muzarski I updated the description regarding the 6 months policy.

wprzytula (approved):

PR #1101, opened 2024-10-21
Parametrize `Deserialize{Value,Row}` with two lifetimes: `'frame` and `'metadata` separately

wprzytula:
Refs: #462

## Tl;dr
A second lifetime parameter is introduced to entities comprising the bottom part of the new deserialization framework: `'metadata`.

## Deserialization traits - recap

`DeserializeValue` requires two types of data in order to perform deserialization:
1) a reference to the CQL frame (a `FrameSlice`),
2) the type of the column being deserialized, being part of the result metadata.

Similarly, `DeserializeRow` requires two types of data in order to perform deserialization:
1) a reference to the CQL frame (a `FrameSlice`),
2) a slice of specifications of all columns in the row, being part of the result metadata.

### Observation

When deserializing owned types, both the frame and the metadata can have any lifetime and it's not important. When deserializing borrowed types, however, they borrow from the frame, so their lifetime must necessarily be bound by the lifetime of the frame. Metadata is only needed for the deserialization, so its lifetime does not abstractly bound the deserialized value.

### Problem

Up to this PR, `Deserialize{Value, Row}` were only parametrized by one lifetime: `'frame`, which bounded both the frame slice and the metadata. This was done that way due to an assumption that both the metadata and the frame (shared using `Bytes`) will be owned by the same entity. However, the new idea of deserializing result metadata to a borrowed form (to save allocations) makes result metadata's lifetime shorter than the frame's lifetime.

### Implemented solution

Not to unnecessarily shorten the deserialized values' lifetime, a separate lifetime parameter is introduced for result metadata: `'metadata`.

The PR is large, but the changes are mostly mechanical, by adding
the second lifetime parameter.

## Pre-review checklist

- [ ] I have adjusted the documentation in `./docs/source/`. <-- _not yet, will be done at the end of the deserialization refactor_

### DIFF
--- a/scylla-cql/src/types/deserialize/value.rs
+++ b/scylla-cql/src/types/deserialize/value.rs
@@ -31,7 +31,7 @@ use crate::frame::value::{
 /// The crate also provides a derive macro which allows to automatically
 /// implement the trait for a custom type. For more details on what the macro
 /// is capable of, see its documentation.
-pub trait DeserializeValue<'frame>
+pub trait DeserializeValue<'frame, 'metadata>
 where
     Self: Sized,
 {
@@ -45,19 +45,19 @@ where
     /// so it should not use the assumption about `type_check` being called
     /// as an excuse to run `unsafe` code.
     fn deserialize(
-        typ: &'frame ColumnType,
+        typ: &'metadata ColumnType<'metadata>,
         v: Option<FrameSlice<'frame>>,
     ) -> Result<Self, DeserializationError>;
 }
 
-impl<'frame> DeserializeValue<'frame> for CqlValue {
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for CqlValue {
     fn type_check(_typ: &ColumnType) -> Result<(), TypeCheckError> {
         // CqlValue accepts all possible CQL types
         Ok(())
     }
 
     fn deserialize(
-        typ: &'frame ColumnType,
+        typ: &'metadata ColumnType<'metadata>,
         v: Option<FrameSlice<'frame>>,
     ) -> Result<Self, DeserializationError> {
         let mut val = ensure_not_null_slice::<Self>(typ, v)?;
@@ -69,16 +69,16 @@ impl<'frame> DeserializeValue<'frame> for CqlValue {
 // Option represents nullability of CQL values:
 // None corresponds to null,
 // Some(val) to non-null values.
-impl<'frame, T> DeserializeValue<'frame> for Option<T>
+impl<'frame, 'metadata, T> DeserializeValue<'frame, 'metadata> for Option<T>
 where
-    T: DeserializeValue<'frame>,
+    T: DeserializeValue<'frame, 'metadata>,
 {
     fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
         T::type_check(typ)
     }
 
     fn deserialize(
-        typ: &'frame ColumnType,
+        typ: &'metadata ColumnType<'metadata>,
         v: Option<FrameSlice<'frame>>,
     ) -> Result<Self, DeserializationError> {
         v.map(|_| T::deserialize(typ, v)).transpose()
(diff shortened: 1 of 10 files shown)


wprzytula:
> `cargo semver-checks` found no API-breaking changes in this PR! 🎉🥳 Checked commit: [4406db5](https://github.com/scylladb/scylla-rust-driver/commit/4406db55710c161a35188629bfe348e38e9eafa6)

???

This looks like a crude bug... Adding a new lifetime parameter to a trait necessarily breaks the API, in the simplest possible way: syntactically.

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -216,34 +216,36 @@ impl_fixed_numeric_type!(f64, Double);
...
-    Ok(num_bigint_03::BigInt::from_signed_bytes_be(val))
-});
+impl_emptiable_strict_type!(
+    num_bigint_03::BigInt,
+    Varint,
+    |typ: &'metadata ColumnType<'metadata>, v: Option<FrameSlice<'frame>>| {
+        let val = ensure_not_null_slice::<Self>(typ, v)?;
+        Ok(num_bigint_03::BigInt::from_signed_bytes_be(val))
+    }
+);
```
muzarski, line 233: The old odd formatting got fixed. Nice!

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -436,7 +438,9 @@ fn get_days_since_epoch_from_date_column<T>(
 }
 
 #[cfg(feature = "chrono-04")]
-impl_emptiable_strict_type!(chrono_04::NaiveDate, Date, |typ: &'frame ColumnType,
+impl_emptiable_strict_type!(chrono_04::NaiveDate, Date, |typ: &'metadata ColumnType<
+    'metadata,
+>,
```
muzarski, line 443: The formatter going crazy again

muzarski (approved):
LGTM. Great explanation of why we need to introduce second lifetime parameter, so the review went smoothly

Lorak-mmk:
> Refs: #462
> ## Tl;dr
> 
> A second lifetime parameter is introduced to entities comprising the bottom part of the new deserialization framework: `'metadata`.
> ## Deserialization traits - recap
> 
> `DeserializeValue` requires two types of data in order to perform deserialization:
> 
>     1. a reference to the CQL frame (a `FrameSlice`),
> 
>     2. the type of the column being deserialized, being part of the result metadata.
> 
> 
> Similarly, `DeserializeRow` requires two types of data in order to perform deserialization:
> 
>     1. a reference to the CQL frame (a `FrameSlice`),
> 
>     2. a slice of specifications of all columns in the row, being part of the result metadata.
> 
> 
> ### Observation
> 
> When deserializing owned types, both the frame and the metadata can have any lifetime and it's not important. When deserializing borrowed types, however, they borrow from the frame, so their lifetime must necessarily be bound by the lifetime of the frame. Metadata is only needed for the deserialization, so its lifetime does not abstractly bound the deserialized value.

If that were the case then there would be no need to introduce second lifetime to those traits, you could just use anonymous lifetime / have a lifetime-generic method:
```
    fn deserialize(
        typ: &ColumnType,
        v: Option<FrameSlice<'frame>>,
    ) -> Result<Self, DeserializationError>;
```

The reason that seems not possible it precisely that deserialized values are bound by metadata lifetime, for example a `ListlikeIterator`.

This sentence confused me so much that I had to go and try to do this change (and learn why it is not possible) in order to understand why this PR introduces the second lifetime. 
Please fix the sentence.

> ### Problem
> 
> Up to this PR, `Deserialize{Value, Row}` were only parametrized by one lifetime: `'frame`, which bounded both the frame slice and the metadata. This was done that way due to an assumption that both the metadata and the frame (shared using `Bytes`) will be owned by the same entity. However, the new idea of deserializing result metadata to a borrowed form (to save allocations) makes result metadata's lifetime shorter than the frame's lifetime.
> ### Implemented solution
> 
> Not to unnecessarily shorten the deserialized values' lifetime, a separate lifetime parameter is introduced for result metadata: `'metadata`.
> 
> The PR is large, but the changes are mostly mechanical, by adding the second lifetime parameter.

Lorak-mmk (changes requested):
Suggestion: such change should have new tests, right? Those should showcase usage with 2 different lifetimes.

wprzytula:
v1.1: addressed comments.

Lorak-mmk (approved):

Issue #1048, opened 2024-07-31
Change the way tablet support is checked in tests


Lorak-mmk:
It is possible for Scylla to support tablets, but for them to be disabled in config.
In that case our current check:
```rust
pub async fn scylla_supports_tablets(session: &Session) -&gt; bool {
    let result = session
        .query(
            "select column_name from system_schema.columns where 
                keyspace_name = 'system_schema'
                and table_name = 'scylla_keyspaces'
                and column_name = 'initial_tablets'",
            &[],
        )
        .await
        .unwrap();
    result.single_row().is_ok()
}
```
will succeed, but it will not be possible to create a tablets table. It will fail with message like this:
```
Unique name: test_rust_1721129687_11
thread '<unnamed>' panicked at scylla/tests/integration/tablets.rs:264:10:
called `Result::unwrap()` on an `Err` value: DbError(ConfigError, "Tablet replication is not enabled")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tablets::test_default_policy_is_tablet_aware' panicked at scylla/tests/integration/tablets.rs:288:1:
explicit panic
```

We should check this differently. @michoecho suggested checking `TABLETS` cluster feature.
We have a `supports_feature` function for this.

Issue #1097, opened 2024-10-21
Changing topology and custom clusters in integration tests


Lorak-mmk:
Currently all our integration tests are executed on the same 3-node cluster.
That means we are not testing topology changes at all.

Side note: I think that we are also not adequately testing schema changes - there are 5 occurences of `ALTER` statement and 6 occurences of `DROP` statement in tests.

There are ideas like https://github.com/scylladb/scylla-rust-driver/issues/1078 but it doesn't fully address the issue:
- Any SCT run takes a loot of time and resources and is not easy to set up. That means it can't be used locally by developers - tests to run locally must finish in reasonable time and be easy to run.
- It uses `cql-stress`: it won't t thoroughly test drivers APIs correctness like in-repo tests can.
- Being out of repo means it is harder to edit / fix: you need to makes changes here, in cql stress and in sct. That's a lot of friction.

For those reasons I mostly view SCT tests as benchmarks (which is something we severely miss!) and as a guard against some regressions.

What I propose to have in-repo topology tests, in steps:
1. Move all tests that require running Scylla/C* from unit tests to integration tests.
1.5 Step 1 may require exposing more private stuff in test_utils module. To avoid having it compiled in user code (and to prevent users from accidentaly modifying it) we should move it behind a feature - or even better, a `cfg`, like Tokio does: https://docs.rs/tokio/latest/tokio/#unstable-features
2. Create an API for tests to choose the required cluster configuration and other parameters - more on that later.
For that I'd see a macro that describes the cluster. It would look a bit like this:
```
#[scylla::integration_test(cluster=Cluster(nodes=3, tablets=off), ks=( parameters for created keyspace), table=(parameters for created table), session = (parameters for created session))]
test_fn(session, ks_name, table_name) {

}
```
3. Add a test runner that would collect the test, decide on execution order (to minimize cluster changes) and execute the tests.
It would use ccm underneath for cluster provisioning. It would also create a keyspace(s) and table(s) required by the test and clean them up after.
Relevant links:
https://www.infinyon.com/blog/2021/04/rust-custom-test-harness/
https://crates.io/crates/inventory
https://github.com/infinyon/fluvio/tree/master/crates/fluvio-test/src
https://github.com/infinyon/fluvio/blob/6635b8aae20df796604d60d2706289984bcc8da4/tests/runner/src/fluvio-integration-derive/src/lib.rs

Lorak-mmk:
@dkropachev I updated the description.

PR #1102, opened 2024-10-22
Move some tests to integration

Lorak-mmk:
This is the first step towards improved integration tests.
Tests that require a running cluster should be in integration tests.

This PR moves some of such tests. Not all - for that there are a few other changes needed, but I wanted to merge what I already have.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/1097
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1048

## Pre-review checklist

### DIFF
--- a/scylla/src/lib.rs
+++ b/scylla/src/lib.rs
@@ -257,10 +257,8 @@ pub mod transport;
 
 pub(crate) mod utils;
 
-/// This module is NOT part of the public API (it is `pub` only for internal use of integration tests).
-/// Future minor releases are free to introduce breaking API changes inside it.
-#[doc(hidden)]
-pub use utils::test_utils;
+#[cfg(test)]
+pub(crate) use utils::test_utils;
 
 pub use statement::batch;
 pub use statement::prepared_statement;
--- a/scylla/tests/integration/utils.rs
+++ b/scylla/tests/integration/utils.rs
@@ -1,4 +1,5 @@
 use futures::Future;
+use scylla::deserialize::DeserializeValue;
 use scylla::frame::response::result::Row;
 use scylla::transport::session_builder::{GenericSessionBuilder, SessionBuilderKind};
 use scylla::Session;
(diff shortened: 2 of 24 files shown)


scylla/tests/integration/utils.rs
```diff
@@ -14,6 +19,22 @@ pub(crate) fn setup_tracing() {
...
+        "test_rust_{}_{}",
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_secs(),
+        cnt
+    );
+    println!("Unique name: {}", name);
+    name
+}
```
wprzytula, line 36: The counter is not shared between `scylla` crate and the integration tests, yet it's a separate counter. Won't there be a name slash between unit tests still using Scylla and integration tests?
Maybe a better solution would be to have this counter exposed in `test_utils` in `scylla` and shared with integration tests?

Lorak-mmk: Moving to test_utils in scylla doesn't address this issue at all. Unit tests and integration tests are still 2 separate crates, and don't share memory - so the counters will be separate.

This is however not an issue: those binaries are run sequentially afaik, and current timestamp is included in the name, so there will be no clash.

In the future, the only counter will be in integration tests because unit tests won't connect to database.

Lorak-mmk:
1st push: rebased on main
2nd push: Fixed errors, fixed minor related issue (https://github.com/scylladb/scylla-rust-driver/issues/1048)

Lorak-mmk:
Fixed errors in serverless tests

scylla/src/lib.rs
```diff
@@ -246,8 +246,8 @@ pub(crate) mod utils;
 
 /// This module is NOT part of the public API (it is `pub` only for internal use of integration tests).
 /// Future minor releases are free to introduce breaking API changes inside it.
-#[doc(hidden)]
-pub use utils::test_utils;
+#[cfg(test)]
+pub(crate) use utils::test_utils;
```
muzarski, line 250: The docstring is outdated - it mentions that module is `pub`.

muzarski (approved):

scylla/src/lib.rs
Lorak-mmk: I removed the comment.

wprzytula (dismissed):
Please don't merge it before #1057 is merged - it's going to introduce a lot of painful conflicts for me.

muzarski (approved):

PR #1103, opened 2024-10-24
retry_policy: replace all usages of `Box<dyn RetryPolicy>` with `Arc<...>`

muzarski:
This way, when user wants to reuse a retry policy in multiple places, there is no need for an additional allocation for Box (`RetryPolicy::clone_boxed`).

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/execution_profile.rs
+++ b/scylla/src/transport/execution_profile.rs
@@ -191,8 +191,8 @@ pub(crate) mod defaults {
     pub(crate) fn load_balancing_policy() -> Arc<dyn LoadBalancingPolicy> {
         Arc::new(load_balancing::DefaultPolicy::default())
     }
-    pub(crate) fn retry_policy() -> Box<dyn RetryPolicy> {
-        Box::new(DefaultRetryPolicy::new())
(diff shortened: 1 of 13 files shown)


scylla/src/transport/execution_profile.rs
```diff
@@ -191,8 +191,8 @@ pub(crate) mod defaults {
     pub(crate) fn load_balancing_policy() -> Arc<dyn LoadBalancingPolicy> {
         Arc::new(load_balancing::DefaultPolicy::default())
     }
-    pub(crate) fn retry_policy() -> Box<dyn RetryPolicy> {
-        Box::new(DefaultRetryPolicy::new())
+    pub(crate) fn retry_policy() -> Arc<dyn RetryPolicy> {
+        Arc::new(DefaultRetryPolicy::new())
```
wprzytula, line 195: We do lose _something_. Namely, `Box` has an optimisation which makes it not allocate anything for ZSTs, and `Arc` allocates always (its control block). However, I don't know if the optimisation is viable for `Box<dyn Trait>` with ZST inside.

wprzytula (approved):

Lorak-mmk (approved):

PR #1107, opened 2024-10-25
Remove redundant macros in response/result.rs

Lorak-mmk:
`generate_deser_type` and `generate_deser_col_specs` are not necessary: generic function is enough.

To ease future development this PR removes those macros.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -780,58 +780,73 @@ fn deser_table_spec_for_col_spec<'frame>(
     Ok(table_spec)
 }
 
-macro_rules! generate_deser_col_specs {
-    ($deser_col_specs: ident, $l: lifetime, $deser_type: ident, $make_col_spec: expr $(,)?) => {
-        /// Deserializes col specs (part of ResultMetadata or PreparedMetadata)
-        /// in the form mentioned by its name.
-        ///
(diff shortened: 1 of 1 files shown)


scylla-cql/src/frame/response/result.rs
```diff
@@ -603,99 +603,104 @@ pub enum Result {
...
+    let id =
+        types::read_short(buf).map_err(|err| CqlTypeParseError::TypeIdParseError(err.into()))?;
+    Ok(match id {
+        0x0000 => {
+            // We use types::read_string instead of $read_string here on purpose.
+            // Chances are the underlying string is `...DurationType`, in which case
+            // we don't need to allocate it at all. Only for Custom types
+            // (which we don't support anyway) do we need to allocate.
+            // OTOH, the macro argument function deserializes borrowed OR owned string;
+            // here we want to always deserialize borrowed string.
```
wprzytula, line 620: Please update the comment, as it still mentions the macro.

wprzytula (approved):
Thank you very much for noticing that it's not necessary to use macros here!

scylla-cql/src/frame/response/result.rs
Lorak-mmk: Fixed

wprzytula (approved):

PR #1109, opened 2024-10-28
result: fix wrong implied lifetime in `col_specs()`

wprzytula:
As `ColumnSpec` is now parametrised by a lifetime, methods that return it implicitly bound its lifetime with lifetime of `&self`. This bug would lead to inefficiencies.
As a fix, an explicit lifetime is added in every method returning `ColumnSpec`.

Ref: https://github.com/scylladb/cpp-rust-driver/pull/200#discussion_r1819118480

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -549,7 +549,7 @@ impl<'a> ResultMetadata<'a> {
     }
 
     #[inline]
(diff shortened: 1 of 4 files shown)


muzarski:
Following should be adjusted as well:
- `TypedRowIterator::get_column_specs`
- `PreparedStatement::get_variable_col_specs`
- `PreparedStatement::get_result_set_col_specs`

Issue #1033, opened 2024-07-09
Configurable DNS resolve timeout


wprzytula:
cpp-driver [exposes API](https://github.com/scylladb/cpp-driver/blob/fa0f27069a625057984d1fa58f434ea99b86c83f/include/cassandra.h#L2047) to configure [DNS resolution timeout](https://github.com/scylladb/cpp-driver/blob/fa0f27069a625057984d1fa58f434ea99b86c83f/src/resolver.hpp#L85). Let's do the same.

PR #1113, opened 2024-10-30
Introduce configurable DNS resolution timeout

muzarski:
fix: https://github.com/scylladb/scylla-rust-driver/issues/1033

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/node.rs
+++ b/scylla/src/transport/node.rs
@@ -1,5 +1,6 @@
 use itertools::Itertools;
-use tokio::net::lookup_host;
+use thiserror::Error;
+use tokio::net::{lookup_host, ToSocketAddrs};
 use tracing::warn;
 use uuid::Uuid;
 
@@ -13,6 +14,7 @@ use crate::transport::errors::{ConnectionPoolError, QueryError};
 use std::fmt::Display;
 use std::io;
 use std::net::IpAddr;
+use std::time::Duration;
 use std::{
     hash::{Hash, Hasher},
     net::SocketAddr,
@@ -267,27 +269,53 @@ pub(crate) struct ResolvedContactPoint {
     pub(crate) datacenter: Option<String>,
 }
 
+#[derive(Error, Debug)]
+pub(crate) enum DnsLookupError {
+    #[error("Failed to perform DNS lookup within {0}ms")]
+    Timeout(u128),
+    #[error("Empty address list returned by DNS for {0}")]
+    EmptyAddressListForHost(String),
+    #[error(transparent)]
+    IoError(#[from] io::Error),
+}
+
+/// Performs a DNS lookup with provided optional timeout.
+async fn lookup_host_with_timeout<T: ToSocketAddrs>(
+    host: T,
+    hostname_resolution_timeout: Option<Duration>,
+) -> Result<impl Iterator<Item = SocketAddr>, DnsLookupError> {
+    if let Some(timeout) = hostname_resolution_timeout {
+        match tokio::time::timeout(timeout, lookup_host(host)).await {
+            Ok(res) => res.map_err(Into::into),
+            // Elapsed error from tokio library does not provide any context.
+            Err(_) => Err(DnsLookupError::Timeout(timeout.as_millis())),
+        }
+    } else {
+        lookup_host(host).await.map_err(Into::into)
+    }
+}
+
 // Resolve the given hostname using a DNS lookup if necessary.
 // The resolution may return multiple IPs and the function returns one of them.
 // It prefers to return IPv4s first, and only if there are none, IPv6s.
-pub(crate) async fn resolve_hostname(hostname: &str) -> Result<SocketAddr, io::Error> {
-    let addrs = match lookup_host(hostname).await {
+pub(crate) async fn resolve_hostname(
+    hostname: &str,
+    hostname_resolution_timeout: Option<Duration>,
+) -> Result<SocketAddr, DnsLookupError> {
+    let addrs = match lookup_host_with_timeout(hostname, hostname_resolution_timeout).await {
         Ok(addrs) => itertools::Either::Left(addrs),
         // Use a default port in case of error, but propagate the original error on failure
         Err(e) => {
-            let addrs = lookup_host((hostname, 9042)).await.or(Err(e))?;
+            let addrs = lookup_host_with_timeout((hostname, 9042), hostname_resolution_timeout)
+                .await
+                .or(Err(e))?;
             itertools::Either::Right(addrs)
         }
     };
 
     addrs
         .find_or_last(|addr| matches!(addr, SocketAddr::V4(_)))
-        .ok_or_else(|| {
-            io::Error::new(
-                io::ErrorKind::Other,
-                format!("Empty address list returned by DNS for {}", hostname),
-            )
-        })
+        .ok_or_else(|| DnsLookupError::EmptyAddressListForHost(hostname.to_owned()))
 }
 
 /// Transforms the given [`InternalKnownNode`]s into [`ContactPoint`]s.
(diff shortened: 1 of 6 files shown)


scylla/src/transport/session_builder.rs
```diff
@@ -762,6 +762,27 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
+    ///     .build() // Turns SessionBuilder into Session
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn hostname_resolution_timeout(mut self, duration: Duration) -> Self {
+        self.config.hostname_resolution_timeout = Some(duration);
+        self
+    }
+
```
wprzytula, line 785: I believe this should accept `Option<Duration>`, especially that the default is `Some`. Now it's impossible to set no timeout with `SessionBuilder`'s API.

scylla/src/transport/node.rs
```diff
@@ -267,27 +269,53 @@ pub(crate) struct ResolvedContactPoint {
...
+    EmptyAddressListForHost(String),
+    #[error(transparent)]
+    IoError(#[from] io::Error),
+}
+
+/// Performs a DNS lookup with provided optional timeout.
+async fn lookup_host_with_timeout<T: ToSocketAddrs>(
+    host: T,
+    hostname_resolution_timeout: Option<Duration>,
+) -> Result<impl Iterator<Item = SocketAddr>, DnsLookupError> {
```
wprzytula, line 286: I'd use `impl ToSocketAddrs`, as it's more concise and we get nothing with explicit type parameter here.

scylla/src/transport/node.rs
```diff
@@ -267,27 +269,53 @@ pub(crate) struct ResolvedContactPoint {
...
+    let addrs = match lookup_host_with_timeout(hostname, hostname_resolution_timeout).await {
         Ok(addrs) => itertools::Either::Left(addrs),
         // Use a default port in case of error, but propagate the original error on failure
         Err(e) => {
-            let addrs = lookup_host((hostname, 9042)).await.or(Err(e))?;
+            let addrs = lookup_host_with_timeout((hostname, 9042), hostname_resolution_timeout)
+                .await
+                .or(Err(e))?;
             itertools::Either::Right(addrs)
         }
```
wprzytula, line 313: This means that the timeout is effectively twice the value provided in the config...
I'm not convinced it's correct. This must be at least noted in documentation.

wprzytula (changes requested):

scylla/src/transport/session_builder.rs
muzarski: I decided to be consistent with other optional timeout options. See for example `SessionBuilder::keepalive_interval()` It accepts `Duration`, and sets the corresponding option to `Some(duration)` as well.

If user wants to set any of these to `None`, it can do so by direct access to the field (fields are public):
```rust
sb.config.hostname_resolution_timeout = None;
```

scylla/src/transport/node.rs
muzarski: Hmm... you are right. I wonder whether we should leave it as is, or rather recompute the timeout for the second try in case first failed.

```rust
    let deadline = hostname_resolution_timeout.map(|dur| tokio::time::Instant::now() + dur);
    let addrs = match lookup_host_with_timeout(hostname, hostname_resolution_timeout).await {
        Ok(addrs) => itertools::Either::Left(addrs),
        // Use a default port in case of error, but propagate the original error on failure
        Err(e) => {
            let new_timeout =
                deadline.map(|deadline| deadline.duration_since(tokio::time::Instant::now()));
            let addrs = lookup_host_with_timeout((hostname, 9042), new_timeout)
                .await
                .or(Err(e))?;
            itertools::Either::Right(addrs)
        }
    };
```

scylla/src/transport/session_builder.rs
wprzytula: Hmm, I dislike the idea of `SessionBuilder` exposing only subset of possible configuration options using methods, with some being available only through fields.
WDYT @Lorak-mmk ?

scylla/src/transport/node.rs
wprzytula: If the first attempt failed from reasons different that a timeout, then such recomputation makes perfect sense. However, if it failed due to a timeout (because the wrong port somehow caused a timeout - idk how likely this scenario is), then we won't be able to do the second attempt.

wprzytula: Any ideas @Lorak-mmk?

muzarski: We could also execute `resolve_hostname` function with a given timeout (i.e. `tokio::time::timeout(t, resolve_hostname)`). It's much easier and cleaner than recomputing the timeout. However, in such case it's still true that if first attempt takes too long, we won't be able to do a second attempt.

scylla/src/transport/session_builder.rs
Lorak-mmk: We could expose another methods (`disable_hostname_resolution_timeout`, `disable_keepalive`) which would be a bit more descriptive.

scylla/src/transport/node.rs
Lorak-mmk: Wow, I'm suprised that such a trait exists

scylla/src/transport/node.rs
Lorak-mmk: When can it happen that the port is not specified and what is the expected behavior of `lookup_host` in this case?

If it always fails immediately with a specific error, then we can check this error and perform lookup with default port with the same timeout.

If it can timeout, or take some time in general, then I'd perform another lookup with the same timeout and document it.

muzarski:
Rebased on main

scylla/src/transport/node.rs
muzarski: From my investigation, it looks like port is not used in the actual DNS lookup (which makes sense).

What's the difference between `String` and `(String, u16)`, then?
When `String` is provided, both `tokio` and `std::net` expect that it is of form `"host:port"`. If it's not the case, the error is returned immidiately. Otherwise, `host` part is used in DNS lookup which may timeout.

When `(String, u16)` is provided, the DNS lookup is performed based on the string - port is ignored during lookup (same as above).

This means that:
- if hostname is of form `"addr:port"`, then the first lookup timeouts **iff** the second lookup (with default port) timeouts
- if hostname is not of this form, then the first lookup will fail immidiately. We can try with default port.

In that case, I think that we can match the error returned from `lookup_host_with_timeout` If it's something other than `DnsLookupError::Timeout(_)`, then we can try again with the same timeout.

Lorak-mmk: Ok. Let's describe this behavior in the docs.

Lorak-mmk:
Rebased on top of https://github.com/scylladb/scylla-rust-driver/pull/1487 (which required some changes to the PR), addressed some review comments (but not all yet).
One thing left to address it the one with big discussion, about timeout being doubled,

scylla/src/cluster/node.rs
```diff
@@ -267,27 +269,53 @@ pub(crate) struct ResolvedContactPoint {
...
+    let addrs = match lookup_host_with_timeout(hostname, hostname_resolution_timeout).await {
         Ok(addrs) => itertools::Either::Left(addrs),
         // Use a default port in case of error, but propagate the original error on failure
         Err(e) => {
-            let addrs = lookup_host((hostname, 9042)).await.or(Err(e))?;
+            let addrs = lookup_host_with_timeout((hostname, 9042), hostname_resolution_timeout)
+                .await
+                .or(Err(e))?;
             itertools::Either::Right(addrs)
         }
```
Lorak-mmk: Changed the code to not perform second resolution if first one ended in timeout.

scylla/src/errors.rs
```diff
@@ -576,6 +576,24 @@ impl ConnectionError {
     }
 }
 
+/// Error returned from custom implementations of [AddressTranslator](crate::policies::address_translator::AddressTranslator);
```
wprzytula: :pick:

Lorak-mmk:
Rebased on main, addressed review comment

wprzytula (approved):

PR #1115, opened 2024-11-06
feat: implment FromCqlVal and SerializeValue for PrimitiveDatetime

dracarys18:
## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/cql_to_rust.rs
+++ b/scylla-cql/src/frame/response/cql_to_rust.rs
@@ -228,6 +228,15 @@ impl FromCqlVal<CqlValue> for time_03::OffsetDateTime {
     }
 }
 
+#[cfg(feature = "time-03")]
+impl FromCqlVal<CqlValue> for time_03::PrimitiveDateTime {
+    fn from_cql(cql_val: CqlValue) -> Result<Self, FromCqlValError> {
+        let datetime = time_03::OffsetDateTime::from_cql(cql_val)?;
+
+        Ok(Self::new(datetime.date(), datetime.time()))
+    }
+}
+
 #[cfg(feature = "secrecy-08")]
 impl<V: FromCqlVal<CqlValue> + secrecy_08::Zeroize> FromCqlVal<CqlValue> for secrecy_08::Secret<V> {
     fn from_cql(cql_val: CqlValue) -> Result<Self, FromCqlValError> {
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -609,6 +609,13 @@ impl From<time_03::OffsetDateTime> for CqlTimestamp {
 }
 
 #[cfg(feature = "time-03")]
+impl From<time_03::PrimitiveDateTime> for CqlTimestamp {
+    fn from(value: time_03::PrimitiveDateTime) -> Self {
+        value.assume_utc().into()
+    }
+}
+
+#[cfg(feature = "time-03")]
 impl TryInto<time_03::OffsetDateTime> for CqlTimestamp {
     type Error = ValueOverflow;
 
--- a/scylla-cql/src/types/serialize/value.rs
+++ b/scylla-cql/src/types/serialize/value.rs
@@ -196,6 +196,15 @@ impl SerializeValue for time_03::OffsetDateTime {
         <CqlTimestamp as SerializeValue>::serialize(&(*me).into(), typ, writer)?
     });
 }
+
+#[cfg(feature = "time-03")]
+impl SerializeValue for time_03::PrimitiveDateTime {
+    impl_serialize_via_writer!(|me, typ, writer| {
+        exact_type_check!(typ, Timestamp);
+        <CqlTimestamp as SerializeValue>::serialize(&(*me).into(), typ, writer)?
+    });
+}
+
 #[cfg(feature = "time-03")]
 impl SerializeValue for time_03::Time {
     impl_serialize_via_writer!(|me, typ, writer| {


wprzytula:
Please note that `FromCqlVal` is going to be deprecated beginning with the next release. Please implement `DeserializeValue`, too.

dracarys18:
Done @wprzytula

wprzytula (dismissed):

wprzytula:
Could you add relevant unit tests, too? Just to be sure the implementation is correct.

dracarys18:
> Could you add relevant unit tests, too? Just to be sure the implementation is correct.

Do we need test cases for it though? It uses already existing implementation of Serialization and deserialization

wprzytula:
> > Could you add relevant unit tests, too? Just to be sure the implementation is correct.
> 
> Do we need test cases for it though? It uses already existing implementation of Serialization and deserialization

I agree that the ser/de part of this PR is minimal. However, we are still adding a new type that wasn't tested in any way before and it's a good habit to have at least one test: ser/de equivalence, on one concrete value of the type.

wprzytula (approved):

muzarski (approved):

Lorak-mmk:
Looks good to me, but I'm don't feel that confident when working with date / time.

@piodul Do you know if there is any reason to not have such implementation?
Also pinging @Anfid because you implemented https://github.com/scylladb/scylla-rust-driver/pull/745

dracarys18:
I can use the `UtcOffset::current_local_offset()` to get the current system time through syscall, let me know if you expect this behavior I will change it in the PR

Anfid:
I'm not a frequent contributor, I just happened to make that PR, so please correct me if I'm wrong. The timezone on a coordinating ScyllaDB node may not match the timezone on the machine where this current code is executed, so the `UtcOffset::current_local_offset()` won't match the documented behavior.

Unfortunately, I myself am not sure how to address this problem and if it's a major problem at all, or just a niche case that should be documented. I'd prefer some input from other contributors on that matter.

dracarys18:
I don't think it's ever possible for a client to know the server's system time. I think when they say local time they must mean the client local time. Because if the client is providing the CQLTimestamp input to the database the client must be providing the client's local time.

I am also not sure how the server would get the local time, Let's say if I have a table and I have default value set on my table schema for a field with type Timestamp. Every time I insert a data the server would take local time and if the nodes are in different regions this local time would vary everywhere. So I am not entirely sure.

Anfid:
@dracarys18 This is reasonable for unix timestamp, as it's specifically defined to be in UTC, and then you get local time by adding timezone offset information to it.

dracarys18:
@Anfid Correct me if I am wrong, that's the same behavior followed in this PR right?

Anfid:
@dracarys18 not really, as `PrimitiveDateTime` specifically describes time completely stripped of any timezone information. It could be some local time in any timezone, not just UTC. Unix timestamp, however, is defined to always be relative to UTC regardless of your local timezone. For that same reason there aren't any `impl From<PrimitiveDateTime> for OffsetDateTime` in the `time` crate, because just blindly assuming UTC may lead to subtle issues.

Anfid:
In general, you can break timestamps into the following categories:
1) Describing the exact moment in time
These timestamps describe the moment, but not the way it should be displayed. CQL time and unix timestamps fall into this category. Both are always recorded in UTC timezone.
2) Describing the exact moment in time + how it should be displayed to the user
These timestamps contain both timestamp information and the timezone offset and they're intended to be displayed to the user in some way. Examples are `OffsetDateTime` in `time`. Since `time` doesn't have any type to fit category 1, `OffsetDateTime` is also often used to describe the exact moment in time.
3) Describing the local time
These ones have no information about the moment when something happened and are only useful to display something to the user or for more flexible interactions with other programs that don't have timezone support. For example, some software developed for local markets may only accept local time in the timezone of that country, using UTC there would lead to bugs. Examples of these types are `PrimitiveDateTime` in `time` and, if I'm not mistaken, SQL timestamps.

dracarys18:
@Anfid I agree that PrimitiveDateTime wont have the timezone context, Here for this use case we are converting a `PrimitiveDateTime` to a Unix Timestamp which will always be UTC. And while converting it from Timestamp to PrimitiveDatetime we will have to convert it into `OffsetDateTime` and then PrimitiveDateTime

Anfid:
The point of my message above is precisely about how I don't consider implicit conversion of `PrimitiveDateTime` to unix timestamp, CQL timestamp or `OffsetDateTime` a good idea. Even more so if this behavior is not explicitly documented.

To illustrate a potential example of problems that it might cause, let's think about some hypothetical calendar or reminders app. When user creates an event, it contains date and time info, but no timezone. If the developer forgets to add timezone info manually and writes a `PrimitiveDateTime` to the DB, after this PR it will not result in compile error. The event would be recorded, but it would have implicit UTC timezone. If the developer then fetches events as `OffsetDateTime` and compares them to `OffsetDateTime::now_utc()`, this may result in sending events at an incorrect time.

An even more problematic example might occur of there's a mix of events serialized from `PrimitiveDateTime` and `OffsetDateTime` in one table, where they would just be impossible to compare, work with and fix after the fact.

Ideally I personally prefer types to support the development process as much as possible. However if others have different opinions on that matter, I wouldn't start a war over it :)

dracarys18:
@wprzytula @Lorak-mmk Can we please merge this?

Lorak-mmk:
> When working on my implementation, I referenced this doc a lot https://opensource.docs.scylladb.com/stable/cql/types.html#timestamps
> 
> What it says about timestamps without timezone is that they're assumed to be in the local timezone of the coordinating ScyllaDB node, while the implementation in this PR uses UTC. My hesitation is that it may result in confusing behavior if some other library is used for a different server, for example. The doc also recommends always explicitly adding timestamps to avoid this exact potential ambiguity in implementations, as far as I understand.

The doc describes textual format, used e.g. when querying manually through clqsh. Binary cql format uses unix timestamp, and thus utc:
https://github.com/apache/cassandra/blob/f5793af48f3381234f48619d2568f6b64cefbf14/doc/native_protocol_v4.spec#L946-L950


> I'm not a frequent contributor, I just happened to make that PR, so please correct me if I'm wrong. The timezone on a coordinating ScyllaDB node may not match the timezone on the machine where this current code is executed, so the `UtcOffset::current_local_offset()` won't match the documented behavior.

This is true, but not really relevant because of what I wrote above.



> I don't think it's ever possible for a client to know the server's system time. I think when they say local time they must mean the client local time. Because if the client is providing the CQLTimestamp input to the database the client must be providing the client's local time.
> 
> I am also not sure how the server would get the local time, Let's say if I have a table and I have default value set on my table schema for a field with type Timestamp. Every time I insert a data the server would take local time and if the nodes are in different regions this local time would vary everywhere. So I am not entirely sure.

The doc do mean coordinator local time, and not client time. In the case you are discussing, server get a string with time, without timezone. Server doesn't know client's timezone, so it uses its own. Why use coordinator local time and to utc? No idea, probably C* did it that way so Scylla had to follow.
Again, this is totally irrelevant for this PR because this is textual CQL syntax and not binary protocol.





> @wprzytula @Lorak-mmk Can we please merge this?

The comments from @Anfid ( https://github.com/scylladb/scylla-rust-driver/pull/1115#issuecomment-2475763778 https://github.com/scylladb/scylla-rust-driver/pull/1115#issuecomment-2475795794 https://github.com/scylladb/scylla-rust-driver/pull/1115#issuecomment-2476048769 ) convinced me that we should not add this implementation. 
PrimitiveDateTime has no information about timezone, so it may have any timezone.
Inserting it into DB requires using UTC as its timezone (`.assume_utc()` call in the PR), but there is no way to know if this assumption is correct.
Selecting it from DB, while not strictly incorrect like inserting, loses information (because from UTC time in DB you got no-timezone time), so is a potential for bugs.

PR #1117, opened 2024-11-08
errors: replace `RowsParseError` and narrow the return types after deserialization refactor

muzarski:
## Motivation

After deserialization refactor, the `RowsParseError` was widely abused in multiple places, namely:
- `deser_rows()` - initial deserialization of RESULT:Rows response
- `QueryResult::into_rows_result()` - lazy deserialization of metadata
- `QueryResult::into_legacy_result()` - lazy metadata deserialization, rows deserialization etc.
- everytime we needed to convert `Typecheck/DeserializationError`s to `QueryError` - widely abused in iterator API

This PR fixes that by narrowing the return types of functions originally depending on `RowsParseError`. The `RowsParseError` is removed at the end of this PR.

## `QueryResult::into_rows_result()` return type

The last two commits adjust the API of `QueryResult`:
1. Flattened the return type of `QueryResult::into_rows_result()` to `Result<_, IntoRowsResultError>`. I believe that this API is cleaner than returning `Option<Result<...>>`.
2. ***[postponed]*** Introduced `QueryResult::into_rows_result_with_recovery()` - it does the same as `into_rows_result()`, but recovers `self` in case error occurred.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -1204,12 +1204,6 @@ impl RawMetadataAndRawRows {
                     col_specs,
                 }
             };
-            if server_metadata.col_count() != server_metadata.col_specs().len() {
-                return Err(RowsParseError::ColumnCountMismatch {
-                    col_count: server_metadata.col_count(),
-                    col_specs_count: server_metadata.col_specs().len(),
-                });
-            }
             Ok(server_metadata)
         }
     }
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -2,8 +2,9 @@ use crate::cql_to_rust::{FromRow, FromRowError};
 use crate::frame::frame_errors::{
     ColumnSpecParseError, ColumnSpecParseErrorKind, CqlResultParseError, CqlTypeParseError,
     LowLevelDeserializationError, PreparedMetadataParseError, PreparedParseError,
-    RawRowsAndPagingStateResponseParseError, ResultMetadataParseError, RowsParseError,
-    SchemaChangeEventParseError, SetKeyspaceParseError, TableSpecParseError,
+    RawRowsAndPagingStateResponseParseError, ResultMetadataAndRowsCountParseError,
+    ResultMetadataParseError, SchemaChangeEventParseError, SetKeyspaceParseError,
+    TableSpecParseError,
 };
 use crate::frame::request::query::PagingStateResponse;
 use crate::frame::response::event::SchemaChangeEvent;
@@ -1187,17 +1188,16 @@ impl RawMetadataAndRawRows {
     fn metadata_deserializer(
         col_count: usize,
         global_tables_spec: bool,
-    ) -> impl for<'frame> FnOnce(&mut &'frame [u8]) -> StdResult<ResultMetadata<'frame>, RowsParseError>
-    {
+    ) -> impl for<'frame> FnOnce(
+        &mut &'frame [u8],
+    ) -> StdResult<ResultMetadata<'frame>, ResultMetadataParseError> {
         move |buf| {
             let server_metadata = {
                 let global_table_spec = global_tables_spec
                     .then(|| deser_table_spec(buf))
-                    .transpose()
-                    .map_err(ResultMetadataParseError::from)?;
+                    .transpose()?;
 
-                let col_specs = deser_col_specs_borrowed(buf, global_table_spec, col_count)
-                    .map_err(ResultMetadataParseError::from)?;
+                let col_specs = deser_col_specs_borrowed(buf, global_table_spec, col_count)?;
 
                 ResultMetadata {
                     col_count,
@@ -1212,7 +1212,9 @@ impl RawMetadataAndRawRows {
     ///
     /// If metadata is cached (in the PreparedStatement), it is reused (shared) from cache
     /// instead of deserializing.
-    pub fn deserialize_metadata(self) -> StdResult<DeserializedMetadataAndRawRows, RowsParseError> {
+    pub fn deserialize_metadata(
+        self,
+    ) -> StdResult<DeserializedMetadataAndRawRows, ResultMetadataAndRowsCountParseError> {
         let (metadata_deserialized, row_count_and_raw_rows) = match self.cached_metadata {
             Some(cached) if self.no_metadata => {
                 // Server sent no metadata, but we have metadata cached. This means that we asked the server
@@ -1257,7 +1259,7 @@ impl RawMetadataAndRawRows {
         let mut frame_slice = FrameSlice::new(&row_count_and_raw_rows);
 
         let rows_count: usize = types::read_int_length(frame_slice.as_slice_mut())
-            .map_err(RowsParseError::RowsCountParseError)?;
+            .map_err(ResultMetadataAndRowsCountParseError::RowsCountParseError)?;
 
         Ok(DeserializedMetadataAndRawRows {
             metadata: metadata_deserialized,
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -1,8 +1,9 @@
 use crate::cql_to_rust::{FromRow, FromRowError};
 use crate::frame::frame_errors::{
     ColumnSpecParseError, ColumnSpecParseErrorKind, CqlResultParseError, CqlTypeParseError,
-    LowLevelDeserializationError, PreparedParseError, ResultMetadataParseError, RowsParseError,
-    SchemaChangeEventParseError, SetKeyspaceParseError, TableSpecParseError,
+    LowLevelDeserializationError, PreparedMetadataParseError, PreparedParseError,
+    ResultMetadataParseError, RowsParseError, SchemaChangeEventParseError, SetKeyspaceParseError,
+    TableSpecParseError,
 };
 use crate::frame::request::query::PagingStateResponse;
 use crate::frame::response::event::SchemaChangeEvent;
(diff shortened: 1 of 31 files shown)


muzarski:
Rebased on `main`. Adjusted commit messages and PR description

scylla/src/transport/iterator.rs
```diff
@@ -622,7 +622,14 @@ impl QueryPager {
         let mut s = self.as_mut();
 
         let received_page = ready_some_ok!(Pin::new(&mut s.page_receiver).poll_recv(cx));
-        let raw_rows_with_deserialized_metadata = received_page.rows.deserialize_metadata()?;
+
+        // TODO: see my other comment next to QueryError::NextPageError
```
wprzytula, line 626: It should be: `QueryError::NextRowError`.

scylla/src/transport/query_result.rs
```diff
@@ -237,6 +239,52 @@ impl QueryResult {
...
+    ///     }
+    /// }
+    ///
+    /// Ok(())
+    /// # }
+    ///
+    /// ```
+    pub fn into_rows_result_with_recovery(
+        self,
+    ) -> Result<QueryRowsResult, (Self, IntoRowsResultError)> {
```
wprzytula, line 267: I'm not convinced by this method: neither by its naming nor its working.

1. The "recovery" thing sounds strange and unintuitive. Did you invent it yourself or borrowed from some codebase?
1. What I initially wanted to see is: `IntoRowsResultError::NotRows(QueryResult)`. No weird naming, no need to choose between two methods.
1. I see that you `clone` some types when recovery is enabled. Can't we do without it? After all, we can backpropagate all owned types in their corresponding error variants, because IIUC if an error is returned, then the owned type (i.e., `Arc` or `Bytes` were not consumed (by being moved to a newly constructed type, e.g. `QueryRowsResult`).

wprzytula: I suggest:
- either removing this commit and delaying this part of the refactor post 0.15,
- or quickly changing (if it's possible) so that:
  - there is only `into_rows_result()` method,
  - `IntoRowsResultError::NotRows` contain the `QueryResult`,
  - no cloning is necessary at any point.

wprzytula (changes requested):

wprzytula:
Apart from nits and comments about the last commit, looks good!

scylla/src/transport/query_result.rs
muzarski: > I'm not convinced by this method: neither by its naming nor its working.
> 
>     1. The "recovery" thing sounds strange and unintuitive. Did you invent it yourself or borrowed from some codebase?

I just came up with this name. We recover `self`, which otherwise would be consumed.

>     2. What I initially wanted to see is: `IntoRowsResultError::NotRows(QueryResult)`.

Oh, this makes much more sense.

>     3. I see that you `clone` some types when recovery is enabled. Can't we do without it? After all, we can backpropagate all owned types in their corresponding error variants, because IIUC if an error is returned, then the owned type (i.e., `Arc` or `Bytes` were not consumed (by being moved to a newly constructed type, e.g. `QueryRowsResult`).

True, cloning is not necessary. However, the lower layer API would need to be adjusted. I started doing it in the first place, but abandoned it since it required much more changes than I expected. For example we can't pass owned objects to `map_err` closure, because then compiler complains when corresponding object is accessed later - even though the object is consumed only when error occurred and we return prematurely using `?` operator.

I introduced this commit as a small PoC and a suggestion. I think we can abandon it for now and introduce it after 0.15 since it requires rewriting multiple functions (if we don't want to clone).

muzarski:
Rebased on `main` - haven't addressed comments yet.

muzarski:
v2: Addressed @wprzytula comments. Dropped the last commit that introduces `into_rows_result_with_recovery`. The commit can be found here https://github.com/muzarski/scylla-rust-driver/commit/217437866e08d1307c20bf06a526e49509248c03 (cc: @Lorak-mmk)

muzarski:
v2.1: adjusted book to new `into_rows_result()` API.

muzarski:
v2.2: Adjusted book again..

wprzytula (approved):

scylla-cql/src/frame/response/result.rs
```diff
@@ -1187,28 +1188,32 @@ impl RawMetadataAndRawRows {
...
         col_count: usize,
         global_tables_spec: bool,
-    ) -> impl for<'frame> FnOnce(&mut &'frame [u8]) -> StdResult<ResultMetadata<'frame>, RowsParseError>
-    {
+    ) -> impl for<'frame> FnOnce(
+        &mut &'frame [u8],
+    ) -> StdResult<
+        ResultMetadata<'frame>,
+        ResultMetadataLazyDeserializationError,
+    > {
```
Lorak-mmk, line 1196: I don't understand why we need another error type here. Why not return ResultMetadataParseError? Why do we even need ResultMetadataLazyDeserializationError instead of just using ResultMetadataParseError?

muzarski: `ResultMetadataParseError` is a broader type that is used to parse result prepared metadata eagerly (after PREPARE request).

Lorak-mmk: It is used to parse result metadata, and here we also parse result metadata, just in a different way. Error structure doesn't have to reflect our internal code structure - I'd say its undesirable.

muzarski: So do you suggest to reuse `ResultMetadataParseError` and return a broader error type in such case?

Lorak-mmk: I'm not insisting on it, but I don't really see the point of introducing second error for basically the same operation. @wprzytula ?

muzarski: Notice that `ResultMetadataParseError` is not really a superset of new error. In lazy deserialization we also try to deserialize rows count, which is not applicable for prepared statement's result metadata (in response to PREPARE request).

muzarski: And we also have a column count check in new error:
```rust
            if server_metadata.col_count() != server_metadata.col_specs().len() {
                return Err(
                    ResultMetadataLazyDeserializationError::ColumnCountMismatch {
                        col_count: server_metadata.col_count(),
                        col_specs_count: server_metadata.col_specs().len(),
                    },
                );
            }
```

But this could apply to the result metadata in `PREPARED` response as well. Not sure why we don't perform the same check there.

Lorak-mmk: The better question is why do we have this check here. 
From protocol spec:
```
    - <metadata> is composed of:
        <flags><columns_count>[<paging_state>][<global_table_spec>?<col_spec_1>...<col_spec_n>]
```

deser_col_specs_generic takes col_count  and deserializes column specs that many times. The only possibility for this check to be triggered is a bug in the driver - not invalid bytes from the network.

Lorak-mmk: I don't see the point of having user-facing error type that can only indicate a serious bug in the driver.

muzarski: True.

muzarski: Before refactor (0.13) it looked like this: https://github.com/scylladb/scylla-rust-driver/blob/v0.13.0/scylla-cql/src/frame/response/result.rs#L927-L940

muzarski: And I think your claim is true for previous version as well. `deser_col_specs` would take a `column_count` and  deserialize col specs `column_count` times: https://github.com/scylladb/scylla-rust-driver/blob/v0.13.0/scylla-cql/src/frame/response/result.rs#L554-L575

muzarski:
v3:
- returning `self` if response is non-Rows in `QueryResult::into_rows_result()` (thanks @Lorak-mmk)
- removed `ResultMetadataLazyDeserializationError` error. Introduced `ResultMetadataAndRowsCountParseError` in its place. This error is described below
- removed the column count mismtatch check during metadata deserialization (did not make much sense)

## ResultMetadataAndRowsCountParseError
It's a simple wrapper over existing `ResultMetadataParseError` with an additional variant (`RowsCountParseError`). I introduced it only because I really didn't want to include `RowsCountParseError` in `ResultMetadataParseError`. Parsing rows count has nothing to do with parsing result metadata.

wprzytula (approved):

Lorak-mmk (approved):

muzarski:
Let me quickly rebased on main before merge. There are no conflicts with https://github.com/scylladb/scylla-rust-driver/pull/1124, but I think the code won't compile (since there is an additional variant to `QueryError` which is not matched in `speculative_execution::can_be_ignored`

Lorak-mmk:
> Let me quickly rebase on main before merge. There are no conflicts with #1124, but I think the code won't compile (since there is an additional variant to `QueryError` which is not matched in `speculative_execution::can_be_ignored`

Please check. I thought Github runs CI on generated merge commit, so if it didn't compile with current main then CI would fail. Is that not the case?

muzarski:
> > Let me quickly rebase on main before merge. There are no conflicts with #1124, but I think the code won't compile (since there is an additional variant to `QueryError` which is not matched in `speculative_execution::can_be_ignored`
> 
> Please check. I thought Github runs CI on generated merge commit, so if it didn't compile with current main then CI would fail. Is that not the case?

I checked. It does not compile if I rebase without any additional changes.

Lorak-mmk:
> > > Let me quickly rebase on main before merge. There are no conflicts with #1124, but I think the code won't compile (since there is an additional variant to `QueryError` which is not matched in `speculative_execution::can_be_ignored`
> > 
> > 
> > Please check. I thought Github runs CI on generated merge commit, so if it didn't compile with current main then CI would fail. Is that not the case?
> 
> I checked (locally). It does not compile if I rebase without any additional changes.

Weird

muzarski:
Rebased on main (running clippy checks on each commit during rebase). Adjusted the `speculative_execution::can_be_ignored` accordingly.

wprzytula (approved):

Lorak-mmk (approved):

PR #1119, opened 2024-11-11
Unify macro attributes between serialization and deserialization derive macros

wprzytula:
## Motivation
When deserialization macros were introduced in #1024, their attributes differed in naming and usage from those introduced in #851 for serialization derive macros.

This PR aims to unify attributes between `SerializeValue` and `DeserializeValue`, and between `SerializeRow` and `DeserializeRow`, so that both traits can be derived for a struct without a clash in attribute parsing.

Fixes: #462
FINALLY!

## What's done

1. In `SerializeValue`: `force_exact_match` -> `forbid_excess_udt_fields` (attribute renamed for consistency with `DeserializeValue`);
2. `Flavor` is shared between serialization and deserialization derive macros, and both have the same syntax for choosing the flavor: `flavor = "enforce_order"`;
3. `allow_missing` is supported for `SerializeValue`, as it is in `DeserializeValue`;
4. `default_when_null` is parsed and ignored in `SerializeValue`, as it's a no-op for serialization (it's only useful for deserialization, as it says: _"If I got NULL for a non-Option Rust field, let's fill it with Default::default()"_; currently, it would cause a parse error in `SerializeValue`.
5. Doctests are added (analogous to those for `Deserialize{Value,Row}`) for sanity verification of provided attributes.
6. Tests are added for Ser/De macros integration. Those make sure that those macros don't conflict and their attributes' syntax and behaviour are consistent.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types/deserialize/mod.rs
+++ b/scylla-cql/src/types/deserialize/mod.rs
@@ -323,7 +323,7 @@ macro_rules! make_error_replace_rust_name {
 use make_error_replace_rust_name;
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use bytes::{Bytes, BytesMut};
 
     use crate::frame::response::result::{ColumnSpec, ColumnType, TableSpec};
@@ -342,7 +342,7 @@ mod tests {
         bytes.freeze()
     }
 
-    pub(super) const fn spec<'a>(name: &'a str, typ: ColumnType<'a>) -> ColumnSpec<'a> {
+    pub(crate) const fn spec<'a>(name: &'a str, typ: ColumnType<'a>) -> ColumnSpec<'a> {
         ColumnSpec::borrowed(name, typ, TableSpec::borrowed("ks", "tbl"))
     }
 }
--- a/scylla-cql/src/types/serialize/value.rs
+++ b/scylla-cql/src/types/serialize/value.rs
@@ -1496,6 +1496,52 @@ pub enum ValueToSerializeValueAdapterError {
     },
 }
 
+mod doctests {
+    /// ```compile_fail
+    ///
+    /// #[derive(scylla_macros::SerializeValue)]
+    /// #[scylla(crate = scylla_cql, skip_name_checks)]
+    /// struct TestUdt {}
+    /// ```
+    fn _test_udt_bad_attributes_skip_name_check_requires_enforce_order() {}
+
+    /// ```compile_fail
+    ///
+    /// #[derive(scylla_macros::SerializeValue)]
+    /// #[scylla(crate = scylla_cql, flavor = "enforce_order", skip_name_checks)]
+    /// struct TestUdt {
+    ///     #[scylla(rename = "b")]
+    ///     a: i32,
+    /// }
+    /// ```
+    fn _test_udt_bad_attributes_skip_name_check_conflicts_with_rename() {}
+
+    /// ```compile_fail
+    ///
+    /// #[derive(scylla_macros::SerializeValue)]
+    /// #[scylla(crate = scylla_cql)]
+    /// struct TestUdt {
+    ///     #[scylla(rename = "b")]
+    ///     a: i32,
+    ///     b: String,
+    /// }
+    /// ```
+    fn _test_udt_bad_attributes_rename_collision_with_field() {}
+
+    /// ```compile_fail
+    ///
+    /// #[derive(scylla_macros::SerializeValue)]
+    /// #[scylla(crate = scylla_cql)]
+    /// struct TestUdt {
+    ///     #[scylla(rename = "c")]
+    ///     a: i32,
+    ///     #[scylla(rename = "c")]
+    ///     b: String,
+    /// }
+    /// ```
+    fn _test_udt_bad_attributes_rename_collision_with_another_rename() {}
+}
+
 #[cfg(test)]
 mod tests {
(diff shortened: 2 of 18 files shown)


scylla-cql/src/types/serialize/value.rs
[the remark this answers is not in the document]
wprzytula: - `forbid_excess_udt_fields` concerns having some fields in the DB definition of the UDT that are not present in the Rust struct.
- `allow_missing` concerns having some fields in the Rust struct that are not not present in the DB definition of the UDT.

These are different things.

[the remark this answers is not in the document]
Lorak-mmk: So allow missing will skip the field during serialization if it is not present in DB definition and default-initialize it during deserialization if it is not present?

wprzytula:
Rebased on main after #1117 was merged.

scylla-macros/src/serialize/value.rs
```diff
@@ -125,6 +137,28 @@ impl Context {
...
+            // `allow_missing` and `skip` is allowed to be after any field
+            // with `allow_missing`.
+            let invalid_default_when_missing_field = self
+                .fields
+                .iter()
+                .rev()
+                // Skip the whole suffix of <allow_missing> and <skip>.
+                .skip_while(|field| !field.attrs.skip)
+                // skip_while finished either because the iterator is empty or it found a field without both <allow_missing> and <skip>.
+                // In either case, there aren't allowed to be any more fields with `allow_missing`.
```
muzarski, line 156: Is this true? I see that we only check for `skip` and not `allow_missing` (i.e. `fields.attr.ignore_missing`).

muzarski: Is there a test for this?

muzarski: Also, I see the following in `derive_serialize_value`:
```rust
        // Filter the fields now instead of at the places that use them later
        // as it's less error prone - we just filter in one place instead of N places.
        .filter(|f| f.as_ref().map(|f| !f.attrs.skip).unwrap_or(true))
```

Doesn't this mean that there will be no fields with `skip = true` in your validation code?

muzarski (changes requested):
I stopped review at: `macros: SerializeValue supports #[allow_missing]`. Will continue later.

Also a small request: could you please add missing comments for fields of `Attributes` and `FieldAttributes`? I remembered what each attribute is responsible for when I reviewed your PR for deserialization macros. I managed to forget it since then, thus such comments would be helpful during review.

scylla-macros/src/serialize/value.rs
wprzytula: Eh, it's again the same issue as once in the past in #1024...
The `compile_fail` tests fail to compile due to another reason and thus pass, even though the check is ill-formed, as you cleverly noticed.

wprzytula:
Rebased on main.

scylla-macros/src/serialize/value.rs
wprzytula: Actually, the condition in the current state is too rigid. It rejects structs as e.g.:
```rust
struct Udt {
  a: i32,
  #[allow_missing]
  b: i32,
  #[allow_missing]
  c: i32,
}
```
because `c` doesn't have `#[skip]`.

wprzytula: I fixed the bug. Thank you!

wprzytula:
Addressed @muzarski comments:
- fixed the bug in `#[allow_missing] validation,
- fixed the bug in validation doctests,
- added more comments.

scylla-cql/src/types/deserialize/mod.rs
Lorak-mmk, line 1: Commit: "macros: add tests for {Ser,De}Row macros integration "

Instead of making test modules pub(crate), can you create a test_utils module (with cfg(test) of course) and move functions needed by more than 1 test module there?

Lorak-mmk (changes requested):

scylla-macros/src/serialize/value.rs
```diff
@@ -60,6 +66,12 @@ struct FieldAttributes {
 
     #[darling(default)]
     skip: bool,
+
+    // If true, then - if this field is missing from the UDT fields metadata
+    // - it will be ignored during serialization.
+    #[darling(default)]
+    #[darling(rename = "allow_missing")]
+    ignore_missing: bool,
 }
```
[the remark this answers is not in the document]
Lorak-mmk: Ok

muzarski (approved):

scylla-cql/src/types/deserialize/mod.rs
wprzytula: I'm against this, because those entities that I export are specific to the place where they currently are. It's OK that `deserialize` function that works for `DeserializeValue` is defined in `value.rs`/`value_tests.rs`, and `deserialize` for `DeserializeRow` is in `row.rs`/`row_tests.rs`.

wprzytula: It's normal to export things. Why `pub(crate)` on test modules would hurt? Especially that majority of the test modules' contents are still module-private.

Lorak-mmk: ok

Lorak-mmk (approved):

Issue #965, opened 2024-03-19
Deserialization refactor: adjust docs and provide a migration guide


wprzytula:
Sub-task of #961.

PR #1121, opened 2024-11-12
Aid in migration to the new deserialization API

wprzytula:
# Motivation

Let's reduce friction when migrating to the new deserialization API.

# What's done

### Deprecations

Entities (traits and structs) belonging to the old deserialization API, which we plan to remove soon, are labeled with `#[deprecated]` attribute. This way they will trigger warnings when used, urging users to consider migration.

### Convenient switching between the legacy & the new Session's API

There already are `Session` and `LegacySession`. However, a viable migration strategy involves incremental migration from the old API to the new API. In such case, it makes sense to have cheap conversions between Session APIs, using the same data underneath. This is done: `LegacySession` is convertible to `Session` and vice versa (and they share the same `Cluster` through an `Arc`).
 
### Migration guide

A migration guide is added. It briefly describes the old and the new deserialization traits and compares them. Then, it summarises what changes are needed to adjust to the new deserialization framework entities.

Fixes:  #965

## Pre-review checklist

### DIFF
--- a/docs/source/migration-guides/0.15-deserialization.md
+++ b/docs/source/migration-guides/0.15-deserialization.md
@@ -1,0 +1,295 @@
+# Adjusting code to changes in deserialization API introduced in 0.15
+
+In 0.15, a new deserialization API has been introduced. The new API improves type safety and performance of the old one, so it is highly recommended to switch to it. However, deserialization is an area of the API that users frequently interact with: deserialization traits appear in generic code and custom implementations have been written. In order to make migration easier, the driver still offers the old API, which - while opt-in - can be very easily switched to after version upgrade. Furthermore, a number of facilities have been introduced which help migrate the user code to the new API piece-by-piece.
+
+The old API and migration facilities will be removed in a future major release.
+
+## Introduction
+
+### Old traits
+
+The legacy API works by deserializing rows in the query response to a sequence of `Row`s. The `Row` is just a `Vec<Option<CqlValue>>`, where `CqlValue` is an enum that is able to represent any CQL value.
+
+The user can request this type-erased representation to be converted into something useful. There are two traits that power this:
+
+__`FromRow`__
+
+```rust
+# extern crate scylla;
+# use scylla::frame::response::cql_to_rust::FromRowError;
+# use scylla::frame::response::result::Row;
+pub trait FromRow: Sized {
+    fn from_row(row: Row) -> Result<Self, FromRowError>;
+}
+```
+
+__`FromCqlVal`__
+
+```rust
+# extern crate scylla;
+# use scylla::frame::response::cql_to_rust::FromCqlValError;
+// The `T` parameter is supposed to be either `CqlValue` or `Option<CqlValue>`
+pub trait FromCqlVal<T>: Sized {
+    fn from_cql(cql_val: T) -> Result<Self, FromCqlValError>;
+}
+```
+
+These traits are implemented for some common types:
+
+- `FromRow` is implemented for tuples up to 16 elements,
+- `FromCqlVal` is implemented for a bunch of types, and each CQL type can be converted to one of them.
+
+While it's possible to implement those manually, the driver provides procedural macros for automatic derivation in some cases:
+
+- `FromRow` - implements `FromRow` for a struct.
+- `FromUserType` - generated an implementation of `FromCqlVal` for the struct, trying to parse the CQL value as a UDT.
+
+Note: the macros above have a default behavior that is different than what `FromRow` and `FromUserType` do.
+
+### New traits
+
+The new API introduce two analogous traits that, instead of consuming pre-parsed `Vec<Option<CqlValue>>`, are given raw, serialized data with full information about its type. This leads to better performance and allows for better type safety.
+
+The new traits are:
+
+__`DeserializeRow<'frame, 'metadata>`__
+
+```rust
+# extern crate scylla;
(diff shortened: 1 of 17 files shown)


docs/source/migration-guides/0.15-deserialization.md
```diff
@@ -0,0 +1,292 @@
+# Adjusting code to changes in deserialization API introduced in 0.15
+
+In 0.15, a new deserialization API has been introduced. The new API improves type safety and performance of the old one, so it is highly recommended to switch to it. However, deserialization is an area of the API that users frequently interact with: deserialization traits appear in generic code and custom implementations have been written. In order to make migration easier, the driver still offers the old API, which - while opt-in - can be very easily switched to after version upgrade. Furthermore, a number of facilities have been introduced which help migrate the user code to the new API piece-by-piece.
+
+The old API and migration facilities will be removed in the next major release (0.16).
+
```
Lorak-mmk, line 6: Please change this to "The old API and migration facilities will be removed in the future major release."

If our plans change and we make a new release without removing this (as it happened with serialization) it is too easy to have outdated docs.

I've forwarded you email from @annastuchlik that discourages such practices and provides examples from Scylla docs of such issues.

docs/source/migration-guides/0.15-deserialization.md
```diff
@@ -0,0 +1,292 @@
...
+# async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
+let mut rows_stream = session
+    .query_iter("SELECT name, age FROM my_keyspace.people", &[])
+    .await?
+    .rows_stream::<(String, i32)>()?;
+
+while let Some(next_row_res) = rows_stream.next().await {
+    let (a, b): (String, i32) = next_row_res?;
+    println!("a, b: {}, {}", a, b);
+}
```
Lorak-mmk, line 197: Can you add a comment about stream API not supporting borrowed types? Here, or even better above this snippet.

Question: Why is type annotation needed in `let (a, b): (String, i32) = next_row_res?;`? You already provided the type in `rows_stream` call. Is it to make the code easier to read or is it actually required for some reason?

Lorak-mmk (changes requested):

docs/source/migration-guides/0.15-deserialization.md
Lorak-mmk: I don't, I just wanted to know why it is there. Type annotation for clarity are good.

wprzytula: I removed it, and added a comment about possible use of turbofish.

Lorak-mmk (dismissed):

scylla/src/transport/session.rs
```diff
@@ -64,6 +65,7 @@ use crate::transport::cluster::{Cluster, ClusterData, ClusterNeatDebug};
 use crate::transport::connection::{Connection, ConnectionConfig, VerifiedKeyspaceName};
 use crate::transport::connection_pool::PoolConfig;
 use crate::transport::host_filter::HostFilter;
+#[allow(deprecated)]
 use crate::transport::iterator::{LegacyRowIterator, PreparedIteratorConfig};
```
muzarski, line 69: Is it ok? I'm looking at `PreparedIteratorConfig`. The analogous import:
```rust
use crate::{LegacyQueryResult, QueryResult};
```

was decoupled into:
```rust
#[allow(deprecated)]
use crate::LegacyQueryResult;
use crate::QueryResult;
```

scylla/src/transport/iterator.rs
```diff
@@ -1190,4 +1206,5 @@ mod legacy {
     // LegacyTypedRowIterator can be moved freely for any RowT so it's Unpin
     impl<RowT> Unpin for LegacyTypedRowIterator<RowT> {}
 }
+#[allow(deprecated)]
 pub use legacy::{LegacyRowIterator, LegacyTypedRowIterator, NextRowError};
```
muzarski, line 1210: ditto: `NextRowError` is not going to be deprecated

scylla/src/lib.rs
```diff
@@ -269,13 +269,16 @@ pub use statement::batch;
...
+pub use frame::response::cql_to_rust::{self, FromRow};
 
+#[allow(deprecated)]
 pub use transport::caching_session::{CachingSession, GenericCachingSession, LegacyCachingSession};
 pub use transport::execution_profile::ExecutionProfile;
+#[allow(deprecated)]
 pub use transport::legacy_query_result::LegacyQueryResult;
 pub use transport::query_result::{QueryResult, QueryRowsResult};
+#[allow(deprecated)]
 pub use transport::session::{IntoTypedRows, LegacySession, Session, SessionConfig};
```
muzarski, line 282: ditto: `CachingSession`, `Session`, `SessionConfig`

scylla/src/transport/session.rs
wprzytula: It does not hurt. We allow deprecations in a bit more places than we would like, but it's benign.

wprzytula: I would left this as-is.

muzarski: Ok, then ignore my remaining comments

muzarski (approved):

wprzytula:
Rebased on main after #1117 was merged.

Lorak-mmk (dismissed):

Lorak-mmk (approved):

Issue #1126, opened 2024-11-20
CI: "group 0 concurrent modification"


Lorak-mmk:
New versions of Scylla use Raft for schema / topology operations.
Raft serializes those operations internally. Concurrent ddl operations are done similarly to optimistic locking: 
they basically work like "if raft state is X, change schema that way (to state Y)".
If 2 such operations execute at the same time, then one of them will fail (because the state is no longer X).
Scylla server will restart such operations in that case, but the amount of restarts is limited to 10.
If there are many concurrent DDL queries, then its possible to hit this limit, and it happens very often in our tests,
which makes CI unstable.

There are several solutions:
- Limit tests concurrency to 1. This makes the tests multiple times slower, so we should avoid it.
- Have a global mutex for DDL changes
- Send DDL changes always to the same node. This works because Scylla handles this case internally.

Third solution should be best performance-wise because it doesn't introduce any additional locking and just lets Scylla handle this case.

The way to do this is to introduce a helper function in test utils that sends a DDL query.
This function would either:
- take a mutex (second solution)
- set a load balancing policy to send the query to correct node (third solution).

muzarski:
The third option could also use https://github.com/scylladb/scylla-rust-driver/issues/1031 once it's implemented.

Lorak-mmk:
> The third option could also use [#1031](https://github.com/scylladb/scylla-rust-driver/issues/1031) once it's implemented.

I'm currently working on this issue. The way I see the LBP is this:

```rust
// This LBP produces a predictable query plan - it order the nodes
// by position in the ring.
// This is to make sure that all DDL queries land on the same node,
// to prevent errors from concurrent DDL queries executed on different nodes.
#[derive(Debug)]
struct SchemaQueriesLBP {}

impl LoadBalancingPolicy for SchemaQueriesLBP {
    fn pick<'a>(
        &'a self,
        _query: &'a scylla::load_balancing::RoutingInfo,
        cluster: &'a scylla::transport::ClusterData,
    ) -> Option<(
        scylla::transport::NodeRef<'a>,
        Option<scylla::routing::Shard>,
    )> {
        // I'm not sure if Scylla can handle concurrent DDL queries to different shard,
        // in other words if its local lock is per-node or per shard.
        // Just to be safe, let's use explicit shard.
        cluster.get_nodes_info().first().map(|node| (node, Some(0)))
    }

    fn fallback<'a>(
        &'a self,
[...]

PR #1127, opened 2024-11-20
Fix concurrent DDLs

Lorak-mmk:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1126

This PR implements the third solution from the ones described in the issue.

## Pre-review checklist


scylla/src/utils/test_utils.rs
```diff
@@ -100,3 +107,81 @@ pub(crate) fn setup_tracing() {
         .with_writer(tracing_subscriber::fmt::TestWriter::new())
         .try_init();
 }
+
+// This LBP produces a predictable query plan - it order the nodes
+// by position in the ring.
+// This is to make sure that all DDL queries land on the same node,
+// to prevent errors from concurrent DDL queries executed on different nodes.
+#[derive(Debug)]
+struct SchemaQueriesLBP {}
```
muzarski, line 116: nit: This can be a unit struct

muzarski (commented):
Looks good. Just a small question: I saw that you adjusted the tests marked with `#[cfg(cassandra_tests)]` as well. Is it necessary? And this implies more general question: should we use this LBP when running tests against Cassandra?

I'm not against that, as it should not cause any problems. WDYT?

Lorak-mmk:
> Looks good. Just a small question: I saw that you adjusted the tests marked with `#[cfg(cassandra_tests)]` as well. Is it necessary? And this implies more general question: should we use this LBP when running tests against Cassandra?
> 
> I'm not against that, as it should not cause any problems. WDYT?

Using the `ddl` method for ddl queries everywhere is imo beneficial, because it allows us to easily find all DDL queries. I had to find them manually to do this PR and it was not a fun task.

What the ddl method does is up for discussion. We can make it not use the LBP in cassandra tests. Do you think there is any benefit in that?

I just had an idea: maybe I should add info to CONTRIBUTING.md that in DDL queries in tests should use the new method?

muzarski:
> Using the `ddl` method for ddl queries everywhere is imo beneficial, because it allows us to easily find all DDL queries. I had to find them manually to do this PR and it was not a fun task.

That's for sure.

> What the ddl method does is up for discussion. We can make it not use the LBP in cassandra tests. Do you think there is any benefit in that?

This is what I had in mind - to adjust the `ddl` method logic based on whether the tests are run against Scylla or Cassandra. But as I said, I'm not against using the custom LBP for Cassandra as well. Personally, I don't see any benefit in not doing this. We can leave it as is.

> I just had an idea: maybe I should add info to CONTRIBUTING.md that in DDL queries in tests should use the new method?

Definitely, let's do that.

Lorak-mmk:
> > I just had an idea: maybe I should add info to CONTRIBUTING.md that in DDL queries in tests should use the new method?
> 
> Definitely, let's do that.

Done. I also changed the struct to be unit struct.

muzarski (approved):

PR #1133, opened 2024-11-29
rows_result: implement into_inner behind `cpp_rust_unstable` cfg

muzarski:
This is convenience introduced only for cpp-rust-driver purposes.

We will use DeserializedMetadataAndRawRows directly to implement lazy deserialization in cpp-rust-driver

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -98,4 +98,8 @@ harness = false
 [lints.rust]
 unnameable_types = "warn"
 unreachable_pub = "warn"
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(scylla_cloud_tests)', 'cfg(cassandra_tests)'] }
+unexpected_cfgs = { level = "warn", check-cfg = [
+    'cfg(scylla_cloud_tests)',
+    'cfg(cassandra_tests)',
+    'cfg(cpp_rust_unstable)',
+] }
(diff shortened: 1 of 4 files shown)


scylla/Cargo.toml
```diff
@@ -39,6 +39,7 @@ full-serialization = [
     "num-bigint-04",
     "bigdecimal-04",
 ]
+cpp-rust = []
 
 [dependencies]
 scylla-macros = { version = "0.7.0", path = "../scylla-macros" }
```
Lorak-mmk, line 45: 1. Can you make it a cfg instead of a feature? This is to make it harder to accidentaly enable, e.g. by some dependency. Tokio does it the same way: https://docs.rs/tokio/latest/tokio/#unstable-features
2. Please prepend "unstable-" to the name - this is a popular convention to mark that contents of this feature / cfg do not adhere to semver.

Lorak-mmk (changes requested):

scylla/Cargo.toml
muzarski: Great suggestion. I applied it (but I appended the `unstable` suffix :)).

muzarski:
v2: Changes `cpp-rust` feature to `cpp_rust_unstable` cfg option.

Lorak-mmk (approved):

muzarski:
v2.1: rebased on main (CI fixes) and adjusted PR title

Lorak-mmk (approved):

Issue #1134, opened 2024-12-01
TableSpecDiffersAcrossColumns when preparing a batch


Lorak-mmk:
Batch statement can contain multiple statements, and they can possible target different tables. Those statements can contain bind markers. An example of such statement:
```
BEGIN BATCH
    INSERT INTO friends (account_id, friend_id) VALUES (?, ?);
    DELETE FROM enemies WHERE account_id = ? AND enemy_id = ?;
APPLY BATCH;
```

If such statement is prepared, the bind markers will have different table specs, which 0.15 driver doesn't consider a possibility, so it throws an error:

```
called `Result::unwrap()` on an `Err` value: CqlResultParseError(PreparedParseError(PreparedMetadataParseError(ColumnSpecParseError(ColumnSpecParseError { column_index: 2, kind: TableSpecDiffersAcrossColumns(TableSpec { ks_name: "ks", table_name: "friends" }, TableSpec { ks_name: "ks", table_name: "enemies" }) }))))
```

Credits to Johhny Bike from User Slack for reporting this!

wprzytula:
This is a serious regression. I suggest releasing 0.15.1 containing the fix. WDYT @Lorak-mmk @dkropachev?

Lorak-mmk:
I agree

PR #1135, opened 2024-12-02
frame/result: allow differing TableSpecs in PreparedMetadata

wprzytula:
As noted in #1134, when preparing batches containing requests to multiple to different tables, the PreparedMetadata received upon preparation contains differing TableSpecs (i.e., TableSpecs with more than table mentioned).
This fails the check that we introduced with ResultMetadata in mind: we've been assuming that all TableSpecs are the same, because ScyllaDB/ Cassandra has no support for JOINs.
Not to fail preparation of cross-table batches, the check is now only turned on for ResultMetadata and turned off for PreparedMetadata.

Fixes: #1134

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session_test.rs
+++ b/scylla/src/transport/session_test.rs
@@ -457,7 +457,7 @@ async fn test_batch() {
         .await
         .unwrap();
 
-    // TODO: Add API, that supports binding values to statements in batch creation process,
+    // TODO: Add API that supports binding values to statements in batch creation process,
     // to avoid problem of statements/values count mismatch
     use crate::batch::Batch;
     let mut batch: Batch = Default::default();
@@ -535,6 +535,42 @@ async fn test_batch() {
         .unwrap();
 
     assert_eq!(results, vec![(4, 20, String::from("foobar"))]);
+}
+
+// This is a regression test for #1134.
+#[tokio::test]
+async fn test_batch_to_multiple_tables() {
+    setup_tracing();
+    let session = create_new_session_builder().build().await.unwrap();
+    let ks = unique_keyspace_name();
+
+    session.ddl(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks)).await.unwrap();
+    session.use_keyspace(&ks, true).await.unwrap();
+    session
+        .ddl("CREATE TABLE IF NOT EXISTS t_batch1 (a int, b int, c text, primary key (a, b))")
+        .await
+        .unwrap();
+    session
+        .ddl("CREATE TABLE IF NOT EXISTS t_batch2 (a int, b int, c text, primary key (a, b))")
+        .await
+        .unwrap();
+
+    let prepared_statement = session
+        .prepare(
+            "
+            BEGIN BATCH
+                INSERT INTO t_batch1 (a, b, c) VALUES (?, ?, ?);
+                INSERT INTO t_batch2 (a, b, c) VALUES (?, ?, ?);
+            APPLY BATCH;
+            ",
+        )
+        .await
+        .unwrap();
+
+    session
+        .execute_unpaged(&prepared_statement, (1, 2, "ala", 4, 5, "ma"))
+        .await
+        .unwrap();
 }
 
 #[tokio::test]


muzarski (commented):
Looks good, but can we add a regression test for the scenario mentioned in issue (i.e., where we prepare a batch statement with bind markers)?

wprzytula:
Rebased on main.

scylla/src/statement/batch.rs
```diff
@@ -178,6 +178,12 @@ impl From<&str> for BatchStatement {
     }
 }
 
+impl From<String> for BatchStatement {
+    fn from(s: String) -> Self {
+        BatchStatement::Query(Query::from(s))
+    }
+}
+
```
muzarski, line 186: Was this change required to implement the test (it's in the same commit)?

scylla/src/transport/session_test.rs
```diff
@@ -537,6 +537,42 @@ async fn test_batch() {
     assert_eq!(results, vec![(4, 20, String::from("foobar"))]);
 }
 
+// This is a regression test for #1134.
+#[tokio::test]
+async fn test_batch_to_multiple_tables() {
+    setup_tracing();
+    let session = Arc::new(create_new_session_builder().build().await.unwrap());
```
muzarski, line 544: nit: is Arc required here?

muzarski (approved):

scylla/src/statement/batch.rs
wprzytula: Good catch. I included this by accident, after I noticed such implementation is missing.

scylla/src/transport/session_test.rs
wprzytula: Not at all, indeed.

scylla/src/statement/batch.rs
wprzytula: I'm move this implementation out of this PR.

Lorak-mmk (changes requested):
Now that the metadata supports zero-copy deserialization, what is the benefit of having this check at all? If the only benefit is that some objects are a bit smaller, then this is imo not enough to justify additional complexity.
Diverging from CQL protocol have just hurt us, so we should avoid letting it hurt us again in the future.

wprzytula:
> Now that the metadata supports zero-copy deserialization, what is the benefit of having this check at all? If the only benefit is that some objects are a bit smaller, then this is imo not enough to justify additional complexity. Diverging from CQL protocol have just hurt us, so we should avoid letting it hurt us again in the future.

OK, I'll remove the check at all.

Lorak-mmk:
Ok. If there are other considerations here (e.g. simpler API) then it may be worth keeping the check. So the question remains - is it only a performance opt or something more?

wprzytula:
> Ok. If there are other considerations here (e.g. simpler API) then it may be worth keeping the check. So the question remains - is it only a performance opt or something more?

What is the "it" that you are referring to? What do you consider to be a performance optimisation? Because the check itself is for sure not an optimisation.

wprzytula:
v2:
- addressed @muzarski's comments,
- deleted the check (and the corresponding error variant) completely. All things considered, it's no use having it, especially once we established that we are not going to decouple TableSpec from ColumnSpec. We also no longer have strong motivation for it, because the ResultMetadata is now always deserialized in a borrowed form, which is much more lightweight.

A step that should actually be performed in that regard is pushing https://github.com/scylladb/scylladb/issues/17788 forwards, because the current way it's done is a dramatic waste.

Lorak-mmk:
> > Ok. If there are other considerations here (e.g. simpler API) then it may be worth keeping the check. So the question remains - is it only a performance opt or something more?
> 
> What is the "it" that you are referring to? What do you consider to be a performance optimisation? Because the check itself is for sure not an optimisation.

Sorry, I was writing the comment quickly and hastily.
The thing is I don't remember all the discussion and API changes around this.

What I meant is that if assuming the same table spec for columns gives us some API advantages, like being able to expose a table spec directly on query response (compared to expising it on the column metadata of the response), then it may be worth considering.

Getting rid of the assumption also lets us get rid of some types (like ColumnSpecView / ColumnSpecs), right? Then imo it is argument for that because it removes and simplifies code.

wprzytula:
What I like about `ColumnSpecs` is that it contains logic for `get_by_name` retrieval, which is absent on plain `&[ColumnSpec]`. I'd like to keep `ColumnSpecs` as such in our codebase. We could indeed remove `TableSpecView` and `ColumnSpecView`, and also provide an `inner` method for accessing `&[ColumnSpec]` from `ColumnSpecs`.

Lorak-mmk:
> What I like about `ColumnSpecs` is that it contains logic for `get_by_name` retrieval, which is absent on plain `&[ColumnSpec]`. I'd like to keep `ColumnSpecs` as such in our codebase. We could indeed remove `TableSpecView` and `ColumnSpecView`, and also provide an `inner` method for accessing `&[ColumnSpec]` from `ColumnSpecs`.
> 
> cc @muzarski

Ok, makes sense.

wprzytula:
This is API-breaking due to an error variant being removed. I suggest leaving the error variant for now under a TODO and removing it in the next major release. cc @Lorak-mmk

Lorak-mmk:
Yes, lets do that so we can release 0.15.1

wprzytula:
> This is API-breaking due to an error variant being removed. I suggest leaving the error variant for now under a TODO and removing it in the next major release. FYI @Lorak-mmk

Done.

Lorak-mmk (approved):

muzarski (approved):

muzarski:
Actually - one more nit: aren't docstrings/comment related to this function outdated now?

wprzytula:
> Actually - one more nit: aren't docstrings/comment related to this function outdated now?

Yees, good catch!

wprzytula:
> Actually - one more nit: aren't docstrings/comment related to this function outdated now?

Fixed.

Lorak-mmk (dismissed):

scylla-cql/src/frame/response/result.rs
```diff
@@ -977,55 +977,24 @@ fn mk_col_spec_parse_error(
...
     for col_idx in 0..col_count {
-        let table_spec = deser_table_spec_for_col_spec(
-            buf,
-            global_table_spec_provided,
-            &mut known_table_spec,
-            col_idx,
-        )?;
+        let table_spec = match global_table_spec {
+            // If global table spec was provided, we simply clone it to each column spec.
+            Some(ref known_spec) if global_table_spec.is_some() => known_spec.clone(),
```
muzarski, line 991: nit: we match against `global_table_spec` and have a match arm: `Some(_) if [...].is_some()`

muzarski: I'm surprised clippy did not catch it

Lorak-mmk (approved):

PR #1137, opened 2024-12-03
Allow serialization of "columns" that are not valid rust identifiers

nrxus:
## Fixes

When trying to serialize a row that isn't a valid rust identifier, the macro fails to expand due to the invalid rust identifier being used when creating variables for the derived implementation. This PR fixes that such that those variables are driven by the field identifier rather than the column name. 

In particular, this occurs when trying to serialize a dynamic TTL that would be using in a query such as:

```cql
INSERT INTO my_table (/* snip cols */) VALUES((/* snip cols */) USING TTL ?
```

The struct desired for serialization is:
```rust
struct InsertRow {
    #[scylla(rename = "[ttl]"]
    ttl: i32,
    /* snip cols */
}
```

However, in the current version of the driver this fails because it tries to create a variable called `visited_flag_[ttl]` which is not a valid rust identifier. As far as I could tell there is no reason for this variable to be based on the column name rather than the field name so I switched it to be based off the field name since that already has to be a valid identifier. 

I added a test to verify that the serialization works but I didn't add any integration tests to make sure that the TTL insert actually works. I have tested it locally however. 

For context: I am using this dynamic TTL to allow moving of rows to different "buckets" (which drive partitions in my schema). I want the re-bucketing to not affect the existing TTL so I am reading the existing TTL in one query and then applying it to the new row in the new bucket in an insert. 

I also fixed a doc typo that was missing an ending square bracket. 

## Pre-review checklist

- [X] I have adjusted the documentation in `./docs/source/`.

### DIFF
--- a/scylla-cql/src/types/serialize/row.rs
+++ b/scylla-cql/src/types/serialize/row.rs
@@ -1568,6 +1568,23 @@ pub(crate) mod tests {
         assert_eq!(reference, row);
     }
 
+    #[test]
+    fn test_row_serialization_with_not_rust_idents() {
+        #[derive(SerializeRow, Debug)]
+        #[scylla(crate = crate)]
+        struct RowWithTTL {
+            #[scylla(rename = "[ttl]")]
+            ttl: i32,
+        }
+
+        let spec = [col("[ttl]", ColumnType::Int)];
+
+        let reference = do_serialize((42i32,), &spec);
+        let row = do_serialize(RowWithTTL { ttl: 42 }, &spec);
+
+        assert_eq!(reference, row);
+    }
+
     #[derive(SerializeRow, Debug)]
     #[scylla(crate = crate)]
     struct TestRowWithSkippedFields {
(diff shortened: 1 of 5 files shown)


Lorak-mmk (dismissed):
@piodul I don't see anything wrong with this change, but you wrote those macros so it would be good if you took a look - maybe I missed something. 
@wprzytula Do we need the same change for deserialization?

Lorak-mmk:
> Looks good, but I believe we have exactly the same issue in SerializeValue as well. @nrxus could you fix it in this PR, too? The fix should be very similar to what you did in the first commit.

Is it possible to have UDT field names that are not correct Rust identifiers?

wprzytula:
> @wprzytula Do we need the same change for deserialization?

This is how `Deserialize{Row,Value}` builds a visited flag variable's ident:
```rust
    fn visited_flag_variable(field: &Field) -> syn::Ident {
        quote::format_ident!("visited_{}", field.ident.as_ref().unwrap().unraw())
    }
```
So it should be free of the mentioned limitation.

wprzytula (approved):
:100: Thank you for this contribution, @nrxus!

Lorak-mmk (approved):

Lorak-mmk:
Thanks for the contribution! We plan to release 0.15.1 this or next week, it will contain your patch.

PR #1138, opened 2024-12-04
Displayer for QueryRowsResult

### LINKED ISSUE #999: Implement Display for `QueryResult`
A very useful QoL feature would be to have Display implementation on QueryResult that printed a table like cqlsh.
It is a frequent scenario to want to just see what results were returned by a query, but Debug will never really be easily human-readable (and it is not its purpose).

Michu1596:
This PR adds Displayer for QueryRowsResult trying to mimic `cqlsh`  way of displaying tables, but there are some differences:
 - I implement the display for the duration type differently, because I think `cqlsh` displays this type in an unclear way,for more details look in the comments.
 -  When displaying collections (`list`, `map`, `set`, `tuple`, `udt`), I use quotes only for `text` as I think putting other types in quotes might be confusing.

Fixes: #999 

## Pre-review checklist

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -20,6 +20,7 @@ scylla = { path = "../scylla", features = [
     "num-bigint-04",
     "bigdecimal-04",
     "metrics",
+    "result-displayer",
 ] }
 tokio = { version = "1.34", features = ["full"] }
 tracing = { version = "0.1.25", features = ["log"] }
@@ -137,3 +138,8 @@ path = "tls-rustls.rs"
 [[example]]
 name = "execution_profile"
 path = "execution_profile.rs"
+
+
+[[example]]
+name = "displayer"
+path = "displayer.rs"
--- a/examples/displayer.rs
+++ b/examples/displayer.rs
@@ -1,0 +1,341 @@
+use anyhow::Result;
+use scylla::{
+    client::{session::Session, session_builder::SessionBuilder},
+    response::rows_displayer::ByteDisplaying,
+};
+use std::env;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let uri = env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+
+    println!("Connecting to {} ...", uri);
+
+    // prepare the session
+    let session: Session = SessionBuilder::new().known_node(uri).build().await?;
+
+    session.query_unpaged("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await?;
+
+    session
+        .query_unpaged(
+            "CREATE TABLE IF NOT EXISTS examples_ks.basic (a int, b int, c text, primary key (a, b))",
+            &[],
+        )
+        .await?;
+
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.basic (a, b, c) VALUES (?, ?, ?)",
+            (3, 4, "lorem Ipsum jest tekstem stosowanym jako przykładowy wypełniacz w przemyśle poligraficznym. Został po raz pierwszy użyty w XV w. przez nieznanego drukarza do wypełnienia tekstem próbnej książki. Pięć wieków później zaczął być używany przemyśle elektronicznym,"),
+        )
+        .await?;
+
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.basic (a, b, c) VALUES (1, 2, 'abc')",
+            &[],
+        )
+        .await?;
+
+    let prepared = session
+        .prepare("INSERT INTO examples_ks.basic (a, b, c) VALUES (?, 7, ?)")
+        .await?;
+    session
+        .execute_unpaged(&prepared, (42_i32, "I'm prepared!"))
+        .await?;
+    session
+        .execute_unpaged(&prepared, (43_i32, "I'm prepared 2!"))
+        .await?;
+    session
+        .execute_unpaged(&prepared, (44_i32, "I'm prepared 3!"))
+        .await?;
+
+    // example 1 - basic table
+
+    // fetch the data
+
+    // in real life scenario you should always fetch data in pages for maximum performance
+    let result = session
+        .query_unpaged("SELECT a, b, c FROM examples_ks.basic", &[])
+        .await?
+        .into_rows_result()?;
+
+    let displayer = result.rows_displayer();
+    println!("\nlong text and special characters:");
+    println!("{}", displayer);
+
+    // example 2 - blob, double, float, time, timestamp
+    session
+    .query_unpaged(
+        "CREATE TABLE IF NOT EXISTS examples_ks.basic4 (a int, b int, c text, d int, timest timestamp, bytes blob, fl float, db double, time1 time, primary key (a, c))",
+        &[],
+    )
+    .await?;
+
+    session
+    .query_unpaged(
+        "INSERT INTO examples_ks.basic4 
+        (a, b, c, d, timest, bytes, fl, db, time1) 
+        VALUES 
+        (1, 10, 'example text', 3, toTimestamp(now()), textAsBlob('sample bytes'), 3.14, 2.718281, '14:30:00');",
+        &[],
+    )
+    .await?;
+
+    // fetch the data
+
+    // in real life scenario you should always fetch data in pages for maximum performance
+    let result = session
+        .query_unpaged("SELECT * FROM examples_ks.basic4", &[])
+        .await?
+        .into_rows_result()?;
+
+    let mut displayer = result.rows_displayer();
+    displayer.set_blob_displaying(ByteDisplaying::Ascii);
+    println!("\nblob, double, float, time, timestamp:");
+    println!("{}", displayer);
+
+    // example 3 - date, duration, ip address, timeuuid
+
+    session
+    .query_unpaged(
+        "CREATE TABLE IF NOT EXISTS examples_ks.basic6 (a int, timud timeuuid, date1 date, ipaddr inet, dur duration, primary key (a))",
+        &[],
+    )
+    .await?;
+
+    // insert some data
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.basic6 
+        (a, timud, date1, ipaddr, dur) 
+        VALUES 
+        (1, now(), '2021-01-01', '3.14.15.9', 1h);",
+            &[],
+        )
+        .await?;
+
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.basic6 
+        (a, timud, date1, ipaddr, dur) 
+        VALUES
+        (3, NOW(), '2024-01-15', '128.0.0.1', 89h4m48s137ms);", // cqlsh prints this as 89.08003805555556h4.8022833333333335m48.137s137.0ms
+            &[],
+        )
+        .await?;
+
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.basic6 
+        (a, timud, date1, ipaddr, dur) 
+        VALUES
+        (4, NOW(), '2024-01-15', '192.168.0.14', 13y2w89h4m48s137ms);",
+            &[],
+        )
+        .await?;
+
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.basic6 (a, timud, date1, ipaddr, dur) 
+        VALUES (2, NOW(), '2024-02-20', '2001:0db8:0:0::1428:57ab', 5d2h);",
+            &[],
+        )
+        .await?;
+
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.basic6 (a, timud, date1, ipaddr, dur) 
+        VALUES (5, NOW(), '-250000-02-20', '2001:db8::1428:57ab', 1y1mo1w1d1h1m1s700ms);",
+            &[],
+        )
+        .await?;
+
+    // fetch the data
+
+    // in real life scenario you should always fetch data in pages for maximum performance
+    let result = session
+        .query_unpaged("SELECT * FROM examples_ks.basic6", &[])
+        .await?
+        .into_rows_result()?;
+
+    let mut displayer = result.rows_displayer();
+    println!("\ndate, duration, ip address, timeuuid:");
+    println!("{}", displayer);
+
+    displayer.set_terminal_width(80);
+    displayer.use_color(false);
+    println!("\nno color, width = 80:");
+    println!("{}", displayer);
+
+    // example 4 - List
+    // Create a table with a list of text
+    session
+    .query_unpaged(
+        "CREATE TABLE IF NOT EXISTS examples_ks.upcoming_calendar ( year int, month int, events list<text>, PRIMARY KEY ( year, month) )",
+        &[],
+    )
+    .await?;
+
+    // Insert some data
+    session
+    .query_unpaged(
+        "INSERT INTO examples_ks.upcoming_calendar(year, month, events) VALUES (2015, 6, ['e1', 'e2', 'e3'])",
+        &[],
+    )
+    .await?;
+
+    // fetch the data
+
+    // in real life scenario you should always fetch data in pages for maximum performance
+    let result = session
+        .query_unpaged("SELECT * FROM examples_ks.upcoming_calendar", &[])
+        .await?
+        .into_rows_result()?;
+
+    let displayer = result.rows_displayer();
+    println!("\nList:");
+    println!("{}", displayer);
+
+    // example 5 - map
+    // Create a table with a list of text
+    session
+    .query_unpaged(
+        "CREATE TABLE IF NOT EXISTS examples_ks.cyclist_teams ( id UUID PRIMARY KEY, lastname text, firstname text, teams map<int,text> )",
+        &[],
+    )
+    .await?;
+
+    // Insert some data
+    session
+    .query_unpaged(
+        "INSERT INTO examples_ks.cyclist_teams (id, lastname, firstname, teams) 
+        VALUES (
+        5b6962dd-3f90-4c93-8f61-eabfa4a803e2,
+        'VOS', 
+        'Marianne', 
+        {2015 : 'Rabobank-Liv Woman Cycling Team', 2014 : 'Rabobank-Liv Woman Cycling Team', 2013 : 'Rabobank-Liv Giant', 
+            2012 : 'Rabobank Women Team', 2011 : 'Nederland bloeit' })",
+                &[],
+            )
+    .await?;
+
+    // fetch the data
+
+    // in real life scenario you should always fetch data in pages for maximum performance
+    let result = session
+        .query_unpaged("SELECT * FROM examples_ks.cyclist_teams", &[])
+        .await?
+        .into_rows_result()?;
+
+    let displayer = result.rows_displayer();
+    println!("\nMap:");
+    println!("{}", displayer);
+
+    // example 6 - set
+    session
+    .query_unpaged(
+        "CREATE TABLE IF NOT EXISTS examples_ks.cyclist_career_teams ( id UUID PRIMARY KEY, lastname text, teams set<text> );",
+        &[],
+    )
+    .await?;
+
+    // Insert some data
+    session
+    .query_unpaged(
+        "INSERT INTO examples_ks.cyclist_career_teams (id,lastname,teams) 
+        VALUES (5b6962dd-3f90-4c93-8f61-eabfa4a803e2, 'VOS', 
+        { 'Rabobank-Liv Woman Cycling Team','Rabobank-Liv Giant','Rabobank Women Team','Nederland bloeit' } )",
+                &[],
+            )
+    .await?;
+
+    // fetch the data
+
+    // in real life scenario you should always fetch data in pages for maximum performance
+    let result = session
+        .query_unpaged("SELECT * FROM examples_ks.cyclist_career_teams", &[])
+        .await?
+        .into_rows_result()?;
+
+    let displayer = result.rows_displayer();
+    println!("\nSet:");
+    println!("{}", displayer);
+
+    // example 7 - user defined type
+    session
+        .query_unpaged(
+            "CREATE TYPE IF NOT EXISTS examples_ks.basic_info (
+        birthday timestamp,
+        nationality text,
+        weight text,
+        height text
+        )",
+            &[],
+        )
+        .await?;
+
+    // make table
+    session
+    .query_unpaged(
+        "CREATE TABLE IF NOT EXISTS examples_ks.cyclist_stats ( id uuid PRIMARY KEY, lastname text, basics FROZEN<basic_info>)",
+        &[],
+    )
+    .await?;
+
+    // Insert some data
+    session
+        .query_unpaged(
+            "INSERT INTO examples_ks.cyclist_stats (id, lastname, basics) VALUES (
+        e7ae5cf3-d358-4d99-b900-85902fda9bb0, 
+        'FRAME', 
+        { birthday : '1993-06-18', nationality : 'New Zealand', weight : null, height : null }
+        )",
+            &[],
+        )
+        .await?;
+
+    // fetch the data
+
+    // in real life scenario you should always fetch data in pages for maximum performance
+    let result = session
+        .query_unpaged("SELECT * FROM examples_ks.cyclist_stats", &[])
+        .await?
+        .into_rows_result()?;
+
+    let displayer = result.rows_displayer();
+    println!("\nUser defined type:");
+    println!("{}", displayer);
+
+    // example 8 - tuples
+
+    // make table
+    session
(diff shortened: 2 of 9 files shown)


scylla/src/transport/query_result.rs
```diff
@@ -444,10 +446,43 @@ impl QueryRowsResult {
...
+/// ```rust
+/// # use scylla::transport::query_result::{QueryResult, QueryRowsResult, RowsDisplayer};
+/// # fn example(query_result: QueryResult) -> Result<(), Box<dyn std::error::Error>> {
+/// let rows_result = query_result.into_rows_result()?;
+/// let mut displayer = RowsDisplayer::new(&rows_result);
+/// displayer.set_terminal_width(80);
+/// displayer.use_color(true);
+/// println!("{}", displayer);
+/// # Ok(())
+/// # }
```
wprzytula, line 466: :wrench: Missing indentation in function's body.

Michu1596:
obraz
screenshot shows results of command `SCYLLA_URI="127.0.0.1:9042" cargo run --example displayer`

note the difference between the way `cqlsh` is displaying `duration` 
obraz

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1190 @@ impl QueryRowsResult {
...
+    pub int: &'static str,
+    pub float: &'static str,
+    pub decimal: &'static str,
+    pub inet: &'static str,
+    pub boolean: &'static str,
+    pub uuid: &'static str,
+    pub duration: &'static str,
+    pub collection: &'static str,
+    pub reset: &'static str,
+}
```
wprzytula, line 820: :question: Should this struct really be `pub`? It looks like it should be module-private.

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1190 @@ impl QueryRowsResult {
...
+        }
+    };
+}
+
+#[derive(PartialEq)]
+pub enum ByteDisplaying {
+    Ascii,
+    Hex,
+    Dec,
+}
```
wprzytula, line 866: :question: ditto: shouldn't this be module-private?

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1190 @@ impl QueryRowsResult {
...
+
+// timestamp
+impl fmt::Display for WrapperDisplay<'_, CqlTimestamp> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // example of formating timestamp 14:30:00.000000000
+        let seconds_from_epoch = self.value.0;
+        let datetime = Utc
+            .timestamp_millis_opt(seconds_from_epoch)
+            .single()
+            .expect("Invalid timestamp");
```
wprzytula, line 1183: :wrench: This `expect` worries me. If the value might be invalid, then we definitely should not panic.

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1190 @@ impl QueryRowsResult {
...
+        let nanos = nanoseconds % 1_000_000_000;
+
+        // Create NaiveTime with the calculated components
+        let time = NaiveTime::from_hms_nano_opt(
+            hours as u32,
+            minutes as u32,
+            seconds as u32,
+            nanos as u32,
+        )
+        .expect("Invalid time");
```
wprzytula, line 1214: :wrench: ditto: `expect`

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
...
+                    self.settings.print_in_color,
+                    self.settings.colors.inet,
+                    "{}",
+                    ipv6
+                )
+            }
+        }
+    }
+}
+
```
wprzytula, line 1357: :question: :wrench: Can't we simply display the enum?

wprzytula:
```suggestion
// inet
impl fmt::Display for WrapperDisplay<'_, std::net::IpAddr> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        let ip = self.value;
        write_colored!(
            f,
            self.settings.print_in_color,
            self.settings.colors.inet,
            "{}",
            ip
        )
    }
}

```

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
...
+            )?;
+        } else {
+            // it is a set
+            write_colored!(
+                f,
+                self.settings.print_in_color,
+                self.settings.colors.collection,
+                "{{"
+            )?;
+        }
```
wprzytula, line 1419: :wrench: Let's make sure that it is indeed a set.

wprzytula:
```suggestion
        if matches!(self.cql_value, CqlValue::List(_))  {
            write_colored!(
                f,
                self.settings.print_in_color,
                self.settings.colors.collection,
                "["
            )?;
        } else if matches!(self.cql_value, CqlValue::Set(_)) {
            // it is a set
            write_colored!(
                f,
                self.settings.print_in_color,
                self.settings.colors.collection,
                "{{"
            )?;
        } else {
            panic!("BUG: Misuse of..."); // Describe the reason why it might happen.
        }
```

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
...
+                    self.settings.colors.collection,
+                    ", "
+                )?;
+            }
+            first_ture = false;
+
+            // if item is text, display it in colored quotes, otherwise display it as is (e.g. number)
+            // to not confuse with text
+            if let CqlValue::Text(_item) = item {
+                display_in_quotes_and_colored(f, &*item_wrapper, self.settings)?;
```
wprzytula, line 1441: :pick: Let's use `matches!`:

wprzytula:
```suggestion
            if matches!(item, CqlValue::Text(_) {
                display_in_quotes_and_colored(f, &*item_wrapper, self.settings)?;
```

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
...
+            // it is a set
+            write_colored!(
+                f,
+                self.settings.print_in_color,
+                self.settings.colors.collection,
+                "}}"
+            )
+        }
+    }
+}
```
wprzytula, line 1465: :wrench: ditto: let's explicitly check that it's indeed a set, and use `matches!`.

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
...
+// map
+impl fmt::Display for WrapperDisplay<'_, Vec<(CqlValue, CqlValue)>> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write_colored!(
+            f,
+            self.settings.print_in_color,
+            self.settings.colors.collection,
+            "{{"
+        )?;
+        let mut first_ture = true;
```
wprzytula, line 1476: :question: ditto: what's `ture`?

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
...
+            write_colored!(
+                f,
+                self.settings.print_in_color,
+                self.settings.colors.collection,
+                "{{"
+            )?;
+        }
+
+        // print elements
+        let mut first_ture = true;
```
wprzytula, line 1422: :question: What's `ture`?

docs/source/queries/displaying-results.md
```diff
@@ -0,0 +1,125 @@
...
+
+```rust
+let result: QueryRowsResult = session
+        .query_unpaged("SELECT * FROM examples_ks.basic1", &[])
+        .await?
+        .into_rows_result()?;
+
+    let displayer = result.rows_displayer();
+    println!("{}", displayer);
+```
```
wprzytula, line 17: :wrench: bad indent

```suggestion
```rust
let result: QueryRowsResult = session
    .query_unpaged("SELECT * FROM examples_ks.basic1", &[])
    .await?
    .into_rows_result()?;

let displayer = result.rows_displayer();
println!("{}", displayer);
```
```

File: docs/source/queries/displaying-results.md [hunk 13]
```diff
@@ -0,0 +1,125 @@
+# Displaying Query Results
+
+This guide explains how to display query results from the database using the `RowsDisplayer` utility. The `RowsDisplayer` provides a flexible way to format and visualize query results as tables with various customization options, it tries to copy the behavior of the `cqlsh` utility.
```
wprzytula, line 3: :100: This is a great overview of the new functionality! Good job!

docs/source/queries/displaying-results.md
```diff
@@ -0,0 +1,125 @@
...
+3. **Binary Data**
+   - Choose appropriate blob display format based on data content
+   - Use Hex for general binary data
+   - Use ASCII when data is known to be text
+   - Use Dec for byte-oriented analysis
+
+4. **Number Formatting**
+   - Adjust floating point precision based on data requirements
+   - Enable scientific notation for very large/small numbers
+   - Consider locale-specific formatting needs
```
wprzytula, line 117: :100: Yeah! Best Practices described!

wprzytula: Was this generated by ChatGPT BTW? :smile:

examples/displayer.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    // Or as custom structs that derive DeserializeRow
+    #[allow(unused)]
+    #[derive(Debug, DeserializeRow)]
+    struct RowData {
+        a: i32,
+        b: Option<i32>,
+        c: String,
+    }
+
```
wprzytula, line 61: :wrench: It looks to me that it's unused, so let's remove it.

examples/displayer.rs
```diff
@@ -0,0 +1,324 @@
...
+    // example 1 - basic table
+    let result: QueryRowsResult = session
+        .query_unpaged("SELECT a, b, c FROM examples_ks.basic", &[])
+        .await?
+        .into_rows_result()?;
```
wprzytula, line 66: :wrench:  It's an anti-pattern to perform `SELECT`s with `*_unpaged` methods.

:thought_balloon: :pushpin:  Unfortunately, only `*_single_page` methods fulfil the following requirement:
1. they are recommended for `SELECTs`, **and**
2. they return `QueryResult`.

`*_iter` methods return `QueryPager`, which is a distinct type that `QueryResult`.
I think it would be beneficial to support displaying of rows coming through both `QueryResult` and `QueryPager`. Extending the framework for `QueryPager`, however, can be left for a follow-up.

:wrench: For now, let's use `query_single_page` in a loop here.

examples/displayer.rs
```diff
@@ -0,0 +1,324 @@
...
+    .query_unpaged(
+        "INSERT INTO examples_ks.upcoming_calendar(year, month, events) VALUES (2015, 6, ['e1', 'e2', 'e3'])",
+        &[],
+    )
+    .await?;
+
+    let result4: QueryRowsResult = session
+        .query_unpaged("SELECT * FROM examples_ks.upcoming_calendar", &[])
+        .await?
+        .into_rows_result()?;
```
wprzytula, line 189: :wrench: :recycle: Numbering of the `result` variables is messy. There are multiple `result4`s. Instead, let's just use `result` for each subexample, and let's put all code related to a single subexample (e.g., a list) in a separate code block (`{}`).

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
 
         (raw_rows_with_metadata, tracing_id, warnings)
     }
+
+    pub fn rows_displayer(&self) -> RowsDisplayer<'_> {
+        RowsDisplayer::new(self)
+    }
+}
+
+/// A utility struct for displaying rows received from the database in a [`QueryRowsResult`].
```
wprzytula, line 467: :pushpin: As I noted elsewhere, we should consider making this work both with `QueryRowsResult` and with `TypedRowStream`.

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1180 @@ impl QueryRowsResult {
...
+/// as a table for easier visualization. It supports various display settings such as
+/// terminal width, color usage, and formatting of different CQL value types.
+///
+/// # Example
+///
+/// ```rust
+/// # use scylla::response::query_result::{QueryResult, QueryRowsResult, RowsDisplayer};
+/// # fn example(query_result: QueryResult) -> Result<(), Box<dyn std::error::Error>> {
+///     let rows_result = query_result.into_rows_result()?;
+///     let mut displayer = RowsDisplayer::new(&rows_result);
```
wprzytula, line 479: :wrench: Let's suggest `rows_displayer` as the default way to create a `RowsDisplayer`.
```suggestion
///     let rows_result = query_result.into_rows_result()?;
///     let mut displayer = rows_result.rows_displayer();
```

wprzytula (changes requested):
:100: This is a really great job done! In the aspect of usability, I'm very satisfied.

:wrench: There are a few items to be addressed, though, mainly about:
- efficiency: I believe we can get rid of String allocations on the way to display,
- more idiomatic code,
- division into modules.

scylla/src/response/query_result.rs
Michu1596: It is used in the `set_blob_displaying` method, I'm not sure what is the best way to set the way of displaying blob, I could go with a bunch of methods like `set_blob_ascii`, `set_blob_hex`, and so on, but providing public enum just makes more sense for me, what do you think?

scylla/src/response/query_result.rs
```diff
@@ -444,6 +458,1190 @@ impl QueryRowsResult {
...
+struct WrapperDisplay<'a, T: 'a> {
+    value: &'a T,
+    settings: &'a RowsDisplayerSettings,
+    cql_value: &'a CqlValue,
+}
+
+/// A trait for types that can be converted to a `String`.
+///
+/// Trait bound to ensure From<WrapperDisplay<T>> for String is implemented
+trait StringConvertible<'a>: 'a + fmt::Display {}
```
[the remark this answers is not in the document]
Michu1596: This trait is needed by table builder.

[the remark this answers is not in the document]
wprzytula: You mean the external library, `tabled`?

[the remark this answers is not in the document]
Michu1596: Yes

examples/Cargo.toml
```diff
@@ -131,4 +132,5 @@ path = "execution_profile.rs"
 
 [[example]]
 name = "displayer"
-path = "displayer.rs"
\ No newline at end of file
+path = "displayer.rs"
+futures = ["result-displayer"]
```
wprzytula, line 136: :wrench: `futures` -> `features`

scylla/Cargo.toml
```diff
@@ -39,6 +39,7 @@ full-serialization = [
     "num-bigint-04",
     "bigdecimal-04",
 ]
+result-displayer = ["dep:tabled", "dep:inline_colorization"]
```
wprzytula, line 42: :wrench: This list should contain `num-bigint-04` and `bigdecimal-04`, too.

scylla/Cargo.toml
```diff
@@ -77,11 +78,13 @@ base64 = { version = "0.22.1", optional = true }
 rand_pcg = "0.3.1"
 socket2 = { version = "0.5.3", features = ["all"] }
 lazy_static = "1"
+tabled =  { version = "0.17.0", features = ["std", "ansi"], optional = true }
+inline_colorization = { version = "0.1.6", optional = true }
+num-bigint-04 = { package = "num-bigint", version = "0.4" }
+bigdecimal-04 = { package = "bigdecimal", version = "0.4" }
```
wprzytula, line 84: These should be made `optional = true` here and left as mandatory in dev-dependencies.

scylla/src/response/query_result.rs
```diff
@@ -444,6 +446,11 @@ impl QueryRowsResult {
 
         (raw_rows_with_metadata, tracing_id, warnings)
     }
+
+    #[cfg(feature = "result-displayer")]
+    pub fn rows_displayer(&self) -> RowsDisplayer<'_> {
+        RowsDisplayer::new(self)
+    }
```
wprzytula, line 453: :wrench: Missing docstring.

scylla/src/response/rows_displayer.rs
```diff
@@ -0,0 +1,1194 @@
...
+    pub time: &'static str,
+    pub int: &'static str,
+    pub float: &'static str,
+    pub decimal: &'static str,
+    pub inet: &'static str,
+    pub boolean: &'static str,
+    pub uuid: &'static str,
+    pub duration: &'static str,
+    pub collection: &'static str,
+}
```
wprzytula, line 346: :wrench: As the struct itself is mod-private, then its fields should not be more public than the struct itself. Please remove the `pub`s.

scylla/src/response/rows_displayer.rs
```diff
@@ -0,0 +1,1194 @@
...
+        }
+        Ok(())
+        // Format the time with 9 digits of nanoseconds
+        // write_colored!(
+        //     f,
+        //     self.settings.print_in_color,
+        //     self.settings.colors.duration,
+        //     "{}",
+        //     result
+        // )
```
wprzytula, line 859: :recycle: Leftover commented out code.

scylla/src/response/rows_displayer.rs
```diff
@@ -0,0 +1,1194 @@
...
+    }
+}
+
+// date
+impl fmt::Display for WrapperDisplay<'_, CqlDate> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // example of formating date 2021-12-31
+        let magic_constant = 2055453495; // it is number of days from -5877641-06-23 to -250000-01-01
+                                         // around -250000-01-01  is the limit of naive date
+
```
wprzytula, line 869: :question: Please explain in detail what the "magic constant" is and why it is correct in your opinion.

scylla/src/response/rows_displayer.rs
```diff
@@ -0,0 +1,1194 @@
...
+            // it is a set
+            write_colored!(
+                f,
+                self.settings.print_in_color,
+                self.settings.colors.collection,
+                "}}"
+            )
+        } else {
+            panic!("BUG") // Describe the reason why it might happen.
+        }
```
wprzytula, line 1022: :recycle: If matching has at least 3 branches (and this one indeed has), then it's best to use `match` expression.

wprzytula, line 1022: :wrench: This must be addressed before merging.

wprzytula (changes requested):

Michu1596:
rebased on main

Michu1596:
addressed comments

scylla/src/response/rows_displayer.rs
Michu1596: I believe it is correct because the dates printed by the displayer are the same as those inserted into the database, this constant allows shifting the date calculation to a valid range, thus preventing `from_ymd_opt()` from returning `None`. The date -5877641-06-23 is the reference point for `CqlDate`.

scylla/src/response/query_result.rs
wprzytula: I think your approach is correct.

examples/displayer.rs
```diff
@@ -51,10 +51,15 @@ async fn main() -> Result<()> {
...
-        .query_unpaged("SELECT a, b, c FROM examples_ks.basic", &[])
-        .await?
-        .into_rows_result()?;
+
+    // fetch the data
+    let paging_state: PagingState = PagingState::start();
+    let (res, _) = session
+        .query_single_page("SELECT a, b, c FROM examples_ks.basic", &[], paging_state)
+        .await?;
+    // all the data is in the first page, so there is no need to fetch more pages
```
wprzytula, line 60: :face_with_diagonal_mouth: It's not what I meant by migrating from `query_unpaged` to `query_single_page`... The reason was that there was an assumption that the returned data may be arbitrarily large, that's why `query_unpaged` is unsuitable. With such assumption, executing `query_single_page` only once is **incorrect**.

scylla/Cargo.toml
```diff
@@ -77,11 +78,15 @@ base64 = { version = "0.22.1", optional = true }
...
+tabled =  { version = "0.17.0", features = ["std", "ansi"], optional = true }
+inline_colorization = { version = "0.1.6", optional = true }
+num-bigint-04 = { package = "num-bigint", version = "0.4", optional = true }
+bigdecimal-04 = { package = "bigdecimal", version = "0.4", optional = true }
 
 [dev-dependencies]
-num-bigint-03 = { package = "num-bigint", version = "0.3" }
 num-bigint-04 = { package = "num-bigint", version = "0.4" }
 bigdecimal-04 = { package = "bigdecimal", version = "0.4" }
+num-bigint-03 = { package = "num-bigint", version = "0.3" }
```
wprzytula, line 89: :pick: This move is redundant and unjustified.

wprzytula (commented):
:100: Looks good!

Could you please squash the 3 latter commits into the respective former 2? So that nothing is introduced in a bad way in one commit only to be fixed in another.

scylla/Cargo.toml
Michu1596: This crate is still used in `scylla/tests/integration/cql_types.rs:136:17` `run_tests::<num_bigint_03::BigInt>(&tests, "varint").await;`

wprzytula: That's right; I mean merely moving this line from 3 lines down.

wprzytula:
:wrench: There are new merge conflicts; when squashing the commits I asked you for, please remember to rebase on main.

wprzytula (dismissed):
:100: :tada: Looks good to me! Thank you for the contribution!

scylla/src/response/rows_displayer.rs
```diff
@@ -0,0 +1,1191 @@
...
+                }
+                ByteDisplaying::Dec => {
+                    write_colored!(
+                        f,
+                        self.settings.print_in_color,
+                        self.settings.colors.blob,
+                        "{}",
+                        byte
+                    )?;
+                }
```
Lorak-mmk, line 685: This will write each byte as decimal, and concatenate this. 
This will result in a very unreadable output. Is that what cqlsh does?
Even if it is, we should do it in some actually usable way.

Lorak-mmk (changes requested):
This code desperately needs:
- Tests
- Proper commit messages and cover letter

I am sure writing this needed a lot of design decisions. Those design decisions and limitations should be all described in PR and commits. I managed to decipher some of the approaches, but not all, and I have no idea how correct I am. I also should not have to guess why the code is written the way it is - it should be explained.

I managed to get through some of the changes. I'll go through the rest after proper descriptions and tests are added.

scylla/src/response/query_result.rs
```diff
@@ -709,6 +722,25 @@ mod tests {
...
+            displayer.set_blob_displaying(ByteDisplaying::Hex);
+            displayer.use_color(true);
+            let _ = format!("{}", displayer);
+            let rows = qr.rows::<(&str, bool)>();
+
+            let mut rows_data = rows.unwrap();
+            let row = rows_data.next().unwrap().unwrap();
+
+            assert_eq!(row, ("MOCK", true));
+        }
```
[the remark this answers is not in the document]
wprzytula: > (btw @wprzytula is there a reason for this being a 300 lines test instead of few smaller ones? test cases in it seem independent at the first glance).

It was easier to be written as a single test, just that.

scylla/src/response/rows_displayer.rs
```diff
@@ -0,0 +1,1191 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write_colored!(
+            f,
+            self.settings.print_in_color,
+            self.settings.colors.error,
+            "null"
+        )
+    }
+}
+
```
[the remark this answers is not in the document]
Michu1596: It is a special case provided just to display `null`. We use this when we want to display `null`. It doesn't have to be an option with CqlValue it can be with anything else by design it should always be `None` there.

scylla/src/response/rows_displayer.rs
```diff
@@ -0,0 +1,1191 @@
...
+impl<'a, T> From<WrapperDisplay<'a, T>> for String
+where
+    T: 'a,
+    WrapperDisplay<'a, T>: fmt::Display,
+{
+    fn from(wrapper: WrapperDisplay<'a, T>) -> Self {
+        format!("{}", wrapper)
+    }
+}
```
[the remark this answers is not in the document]
Michu1596: trait `StringConvertible` follows the requirements of the `push_record` builder method from the tabled crate. Lifetime is needed to carry lifetime information between the `get_item_wrapper` function and `WrapperDisplay`. Probably there are better solutions to this problem, but this is the one I came up with.
@wprzytula suggested squashing commits, I will try to add more helpful comments.

PR #1148, opened 2024-12-11
value: `Cql[Varint/Decimal]Borrowed`

muzarski:
This PR introduces borrowed versions of CqlDecimal and CqlVarint. It implements SerializeValue and DeserializeValue for new types.

There is a use case for this in cpp-rust-driver - we need to set the pointer user provided, so it points to the underlying data. After lazy-deserialization refactor in cpp-rust-driver, CqlDecimal is not useful anymore. If we deserialize to CqlDecimal, it owns the bytes and is a stack variable - thus, we cannot set the pointer to its underlying data (it will be freed after function call).

Questions for reviewers:
- Should I put it under #[cfg(cpp_rust_unstable]? I think it may be useful for other users as well.
- I only implemente `From<Cql[Varint/Decimal]Borrowed> for Big[Int/Decimal]`. I didn't implement conversions in other direction, since new types borrow bytes (and From consumes). Do you think it's ok to make an exception here, and implement `From<&Big[Int/Decimal]>` (from reference)?

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -303,6 +303,32 @@ impl CqlVarint {
     }
 }
 
+/// Compares two [`CqlVarint`] values after normalization.
+///
+/// # Example
+///
+/// ```rust
+/// # use scylla_cql::frame::value::CqlVarint;
+/// let non_normalized_bytes = vec![0x00, 0x01];
+/// let normalized_bytes = vec![0x01];
+/// assert_eq!(
+///     CqlVarint::from_signed_bytes_be(non_normalized_bytes),
+///     CqlVarint::from_signed_bytes_be(normalized_bytes)
+/// );
+/// ```
+impl PartialEq for CqlVarint {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_normalized_slice() == other.as_normalized_slice()
+    }
+}
+
+/// Computes the hash of normalized [`CqlVarint`].
+impl std::hash::Hash for CqlVarint {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.as_normalized_slice().hash(state)
+    }
+}
+
 #[cfg(feature = "num-bigint-03")]
 impl From<num_bigint_03::BigInt> for CqlVarint {
     fn from(value: num_bigint_03::BigInt) -> Self {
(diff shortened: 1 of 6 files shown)


scylla-cql/src/frame/value.rs
```diff
@@ -233,6 +236,13 @@ impl std::hash::Hash for CqlTimeuuid {
 #[derive(Clone, Eq, Debug)]
 pub struct CqlVarint(Vec<u8>);
 
+/// A borrowed version of native CQL `varint` representation.
+///
+/// Refer to the documentation of [`CqlVarint`].
+/// Especially, see the disclaimer about [non-normalized values](CqlVarint#db-data-format).
+#[derive(Clone, Eq, Debug)]
+pub struct CqlVarintBorrowed<'b>(&'b [u8]);
+
```
wprzytula, line 245: Have you considered putting a `Cow<[u8]>` in `CqlVarint` instead and making it lifetime-generic? I'm not convinced this would be a better choice, but at least we could deduplicate code - it would be lifetime-generic, too, taking advantage of `<Cow as Deref>::Target` being `[u8]`.

wprzytula: This would make the borrowed variant more advertised more recognized, as it would have the same name as before and offer cheaper deserialization.

scylla-cql/src/frame/value.rs
```diff
@@ -267,9 +288,39 @@ impl CqlVarint {
...
+        self.as_signed_bytes_be_slice()
+    }
+}
+
+/// An internal utility trait used to implement [`PartialEq`] and [`std::hash::Hash`]
+/// for [`CqlVarint`] and [`CqlVarintBorrowed`].
+trait AsNormalizedVarintSlice {
+    fn as_normalized_slice(&self) -> &[u8];
+}
+impl<V: AsVarintSlice> AsNormalizedVarintSlice for V {
```
wprzytula, line 321: This traits wouldn't be needed if `CqlVaring` were implemented using `Cow` inside.

scylla-cql/src/frame/value.rs
muzarski: To prevent this, I think we would require additional generic parameter representing an ownership. But maybe you have some other idea on how to overcome this problem. If not, I'd prefer to stick to the current implementation (`CqlVarintBorrowed`)

wprzytula: I'd always deserialize to the borrowed version (i.e., with lifetime tied to the lifetime of the frame), only then allowing to make it owned with `to_owned()` (or rather `into_owned()`, just as in case of, e.g., `TableSpec`).

muzarski: Ok, makes sense then. I'll apply the suggestion

Lorak-mmk:
Now that I think of it: can we just use Bytes for CqlVarint?

The only downside I see is that it would make the whole frame alive as long as CqlVarint is alive, but it is also the case for CqlVarintBorrowed (as it borrows the frame) - the only difference is that the latter is enforced at compile time, but complicates the code.

This is a tradeoff, and I'm not sure which solution is better.

muzarski:
I applied @Lorak-mmk suggestion - replaced `Vec<u8>` with `bytes::Bytes` in `CqlVarint`.

muzarski:
Reverted the changes with `Bytes`. After a discussion, we decided that `CqlVarint` should not keep whole frame alive (and this is what it does when it holds `Bytes`).

wprzytula (approved):

Lorak-mmk (approved):

Issue #1139, opened 2024-12-04
Deprecate legacy APIs before we delete them in 1.0


wprzytula:
There are a number of API items that we consider legacy, but still haven't labeled with `#[deprecated]` attributes. Let's do that in 0.16, so that in 1.0 we can remove them.

This is going to be a list of such items:
- `LegacySerializedValues`,
- TODO

Lorak-mmk:
Actually let's do that in 0.15.1. It's possible we won't release 0.16 because of time constraints, especially if we want to release release candidate for 1.0

PR #1149, opened 2024-12-11
Remove uses of `SerializeValuesError` and deprecate it

wprzytula:
`SerializeValuesError` is part of the legacy serialization API, so it should be deprecated as well.

Refs: #1139

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/value.rs
+++ b/scylla-cql/src/frame/value.rs
@@ -13,29 +13,8 @@ use super::types::vint_encode;
 use super::types::RawValue;
 
 #[derive(Debug, Error, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-#[error("Value too big to be sent in a request - max 2GiB allowed")]
-#[deprecated(
-    since = "0.15.1",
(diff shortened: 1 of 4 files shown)


scylla-cql/src/frame/value.rs
```diff
@@ -25,6 +25,10 @@ pub struct ValueTooBig;
 pub struct ValueOverflow;
 
 #[allow(deprecated)]
+#[deprecated(
+    since = "0.15.1",
+    note = "Legacy serialization API is not type-safe and is going to be removed soon"
+)]
 #[derive(Debug, Error, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub enum SerializeValuesError {
     #[error("Too many values to add, max 65,535 values can be sent in a request")]
```
Lorak-mmk, line 34: I think this should be moved to legacy module, no? Same for `ValueTooBig`.

Lorak-mmk (changes requested):

Lorak-mmk (approved):

PR #1156, opened 2024-12-23
session: adjust naming in internal session functions

muzarski:
During error refactor, I found it hard to understand what which function does on a request execution path. Three different names were used:
- `run_query`
- `execute_query`
- `do_query`

I renamed them (and all internal structures/traits related to these functions). The `_query`  part is very misleading. It would suggest that we want to execute a CQL QUERY request, which is not true - these functions/closures are generic and allow to execute QUERY, EXECUTE and BATCH CQL requests.

Kind of related to https://github.com/scylladb/scylla-rust-driver/issues/713

## Pre-review checklist

### DIFF
--- a/scylla/src/transport/session.rs
+++ b/scylla/src/transport/session.rs
@@ -472,7 +472,7 @@ impl Default for SessionConfig {
     }
 }
 
-pub(crate) enum RunQueryResult<ResT> {
+pub(crate) enum RunRequestResult<ResT> {
     IgnoredWriteError,
     Completed(ResT),
 }
@@ -1203,8 +1203,8 @@ where
 
         let span = RequestSpan::new_query(&query.contents);
         let span_ref = &span;
-        let run_query_result = self
-            .run_query(
+        let run_request_result = self
+            .run_request(
                 statement_info,
                 &query.config,
                 execution_profile,
@@ -1257,13 +1257,13 @@ where
             .instrument(span.span().clone())
             .await?;
 
-        let response = match run_query_result {
-            RunQueryResult::IgnoredWriteError => NonErrorQueryResponse {
+        let response = match run_request_result {
+            RunRequestResult::IgnoredWriteError => NonErrorQueryResponse {
                 response: NonErrorResponse::Result(result::Result::Void),
                 tracing_id: None,
                 warnings: Vec::new(),
             },
-            RunQueryResult::Completed(response) => response,
+            RunRequestResult::Completed(response) => response,
         };
 
         self.handle_set_keyspace_response(&response).await?;
@@ -1526,8 +1526,8 @@ where
             }
         }
 
-        let run_query_result: RunQueryResult<NonErrorQueryResponse> = self
-            .run_query(
+        let run_request_result: RunRequestResult<NonErrorQueryResponse> = self
+            .run_request(
                 statement_info,
                 &prepared.config,
                 execution_profile,
@@ -1558,13 +1558,13 @@ where
             .instrument(span.span().clone())
             .await?;
 
-        let response = match run_query_result {
-            RunQueryResult::IgnoredWriteError => NonErrorQueryResponse {
+        let response = match run_request_result {
+            RunRequestResult::IgnoredWriteError => NonErrorQueryResponse {
                 response: NonErrorResponse::Result(result::Result::Void),
                 tracing_id: None,
                 warnings: Vec::new(),
             },
-            RunQueryResult::Completed(response) => response,
+            RunRequestResult::Completed(response) => response,
         };
 
         self.handle_set_keyspace_response(&response).await?;
@@ -1650,8 +1650,8 @@ where
 
         let span = RequestSpan::new_batch();
 
-        let run_query_result = self
-            .run_query(
+        let run_request_result = self
+            .run_request(
                 statement_info,
                 &batch.config,
                 execution_profile,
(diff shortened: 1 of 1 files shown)


Lorak-mmk (commented):
"Request" is a good word here, but you still use "execute" which is also a CQL command. Maybe "do" / "perform" / "run" are better words than "execute"?

muzarski:
> "Request" is a good word here, but you still use "execute" which is also a CQL command. Maybe "do" / "perform" / "run" are better words than "execute"?

That was one of my doubts. IMO, `to execute a request` fits the best, but I agree that it can suggest the `EXECUTE` CQL request. What I don't like about "run" is that we then would have a `run_request` function which then calls `run_request_with_retries` (or maybe I'm just overthinking, and that's totally fine). I don't really like "perform" that much. My picks would be "run_request_[with_retries/once]", "send_request_[with_retries/once]" or "do_request_[with_retries/once]" with a little bias towards "send".

Lorak-mmk:
I don't really like "send" because we are not just sending request - we receive the response, we perform speculative execution, retries etc.
"send" would be a good word for some method in `Connection` that does just that.


> What I don't like about "run" is that we then would have a `run_request` function which then calls `run_request_with_retries` (or maybe I'm just overthinking, and that's totally fine).

Now we have what? `run_request` that calls `execute_request_with_retries`?
I don't really see what is the difference between "running" a request and "executing" the request . Those names don't really say which layer is responsible for what, apart from "with_retries" telling us that the inner one will use retry policy.

I'm not sure what is the proper naming here tbh. Maybe this naming difficulty is a hint that the division of responsibilities between various functions should be different?

muzarski:
> I'm not sure what is the proper naming here tbh. Maybe this naming difficulty is a hint that the division of responsibilities between various functions should be different?

Current division makes sense to me. `run_request` (previously `run_query`) takes speculative execution into account, and prepares the plan (either exclusive or shared) based on whether speculative execution applies for the statement. It also constructs the `RetrySession` from statement config. Then it calls (directly for standard path, or indirectly in `speculative_execution::execute()`) `execute_request_with_retries` (previously `execute_query`) providing it the plan and retry session. `execute_request_with_retries` iterates over the plan, executing the request on each target sequentially wrt. plan and retry decisions.

When it comes to naming, I'll rename all of the functions/closures to `run_*`. As you said, `run` and `execute` do not distinguish between the execution layers anyway.

Lorak-mmk:
> > I'm not sure what is the proper naming here tbh. Maybe this naming difficulty is a hint that the division of responsibilities between various functions should be different?
> 
> Current division makes sense to me. `run_request` (previously `run_query`) takes speculative execution into account, and prepares the plan (either exclusive or shared) based on whether speculative execution applies for the statement. It also constructs the `RetrySession` from statement config. Then it calls (directly for standard path, or indirectly in `speculative_execution::execute()`) `execute_request_with_retries` (previously `execute_query`) providing it the plan and retry session. `execute_request_with_retries` iterates over the plan, executing the request on each target sequentially wrt. plan and retry decisions.

What do you think about utilizing "speculative fiber" terminology?
`run_request_speculative_fiber` does convey that speculative execution won't be utilized in the function and must be instead handled by the caller.

muzarski:
> What do you think about utilizing "speculative fiber" terminology? `run_request_speculative_fiber` does convey that speculative execution won't be utilized in the function and must be instead handled by the caller.

Sounds good. I like it.

muzarski:
Flaky `test_coalescing` test failed:
```
'transport::connection::tests::test_coalescing' panicked at scylla/src/transport/connection.rs:2609:78:
called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(1184), "called `Result::unwrap()` on an `Err` value: DbError(Unprepared { statement_id: b\"\\xb7/\\x95\\xb8\\xd0\\\"\\x9dq\\xcfp]\\x1a\\x9f\\xa3e<\" }, \"Prepared query with ID b72f95b8d0229d71cf705d1a9fa3653c not found (either the query was not prepared on this host (maybe the host has been restarted?) or you have prepared too many queries and it has been evicted from the internal cache)\")", ...)
```

Lorak-mmk:
> Flaky `test_coalescing` test failed:
> 
> ```
> 'transport::connection::tests::test_coalescing' panicked at scylla/src/transport/connection.rs:2609:78:
> called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(1184), "called `Result::unwrap()` on an `Err` value: DbError(Unprepared { statement_id: b\"\\xb7/\\x95\\xb8\\xd0\\\"\\x9dq\\xcfp]\\x1a\\x9f\\xa3e<\" }, \"Prepared query with ID b72f95b8d0229d71cf705d1a9fa3653c not found (either the query was not prepared on this host (maybe the host has been restarted?) or you have prepared too many queries and it has been evicted from the internal cache)\")", ...)
> ```

Restarted

Lorak-mmk (approved):

scylla/src/transport/session.rs
```diff
@@ -2107,7 +2107,7 @@ where
             .consistency_set_on_statement
             .unwrap_or(execution_profile.consistency);
 
-        'nodes_in_plan: for (node, shard) in query_plan {
+        'nodes_in_plan: for (node, shard) in request_plan {
             let span = trace_span!("Executing query", node = %node.address);
```
wprzytula, line 2111: :pick: Let's rename the trace message to `"Executing request"`, too.

wprzytula (dismissed):

muzarski:
Renamed remaining variables and logs which mentioned "query".

Lorak-mmk (approved):

wprzytula (approved):

Issue #864, opened 2023-11-21
transport::connection::tests::test_coalescing failed in CI


Lorak-mmk:
https://github.com/scylladb/scylla-rust-driver/pull/863#issuecomment-1820496090

> ```
> failures:
> 
> ---- transport::connection::tests::test_coalescing stdout ----
> Unique name: test_rust_1700493980_11
> thread 'transport::connection::tests::test_coalescing' panicked at scylla/src/transport/connection.rs:2034:74:
> called `Result::unwrap()` on an `Err` value: DbError(Unprepared { statement_id: b"i\xb60\x01/\xdc\x02\xb6\xec;\x18\x8e\xaa\xf38\xcf" }, "Prepared query with ID 69b630012fdc02b6ec3b188eaaf338cf not found (either the query was not prepared on this host (maybe the host has been restarted?) or you have prepared too many queries and it has been evicted from the internal cache)")
> ```
> 
> Strange; I know that we now sometimes prepare statements in case of `query` as well, but it falls back to `execute` which automatically reprepares the statement (but just once).

This test failed in CI, but the failure is unrelated to the PR it failed it. Needs to be investigated.

piodul:
Fortunately, it's pretty easy to reproduce locally:

1. Modify the docker-compose command to use the `test/cluster/cassandra/docker-compose.yml` file,
2. `make up` to start the cluster
3. `SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test -- test_coalescing`

and it reproduces pretty much every time. What's interesting is that it's not necessary to run all the tests, running just this one test alone is sufficient to reproduce.

roydahan:
Is this test still failing? flakey? or it was removed from CI?

wprzytula:
It has recently failed:
https://github.com/scylladb/scylla-rust-driver/actions/runs/9693468038/job/26748974433#step:6:1652

muzarski:
Failed again:
https://github.com/scylladb/scylla-rust-driver/actions/runs/10178379297/job/28151882606?pr=1047#step:6:1234

muzarski:
Failed https://github.com/scylladb/scylla-rust-driver/actions/runs/12476813645/job/34822159920?pr=1156#step:6:1411

muzarski:
[...]

PR #1157, opened 2024-12-23
session: internal request execution API error types refactor

muzarski:

## run_request_once closure
To narrow the error return types of some important API functions (e.g. `RS::decide_should_retry` and `LBP::on_query_failure`), the error type of `execute_request_once` (previously known as `do_query`) needs to be adjusted as well. A couple of commits at the beginning of this PR do just that. What we achieve, is that we narrow the return error type of the closure from `QueryError` to `UserRequestError`.

## UserRequestError
This error is renamed to `RequestAttemptError` - as the name suggests, it represents a failure of a single attempt of execution.

## LBP and RetrySession
Both of these traits accept some error and do something based on the error. Before this PR, they would accept a `QueryError`, resulting in matches against variants that could never even be constructed before the call. In this PR, I pub-ified the `RequestAttemptError`, and made `LBP::on_query_failure` and `RetrySession::decide_should_retry` accept it instead of `QueryError`.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -30,7 +30,7 @@ use crate::cluster::{ClusterState, NodeRef};
 use crate::cql_to_rust::{FromRow, FromRowError};
 use crate::deserialize::DeserializeOwnedRow;
 use crate::errors::ProtocolError;
-use crate::errors::{QueryError, UserRequestError};
+use crate::errors::{QueryError, RequestAttemptError};
 use crate::frame::response::result;
 use crate::network::Connection;
 use crate::observability::driver_tracing::RequestSpan;
@@ -135,7 +135,7 @@ struct PagerWorker<'a, QueryFunc, SpanCreatorFunc> {
     sender: ProvingSender<Result<ReceivedPage, QueryError>>,
 
     // Closure used to perform a single page query
-    // AsyncFn(Arc<Connection>, Option<Arc<[u8]>>) -> Result<QueryResponse, UserRequestError>
+    // AsyncFn(Arc<Connection>, Option<Arc<[u8]>>) -> Result<QueryResponse, RequestAttemptError>
     page_query: QueryFunc,
 
     statement_info: RoutingInfo<'a>,
@@ -158,7 +158,7 @@ struct PagerWorker<'a, QueryFunc, SpanCreatorFunc> {
 impl<QueryFunc, QueryFut, SpanCreator> PagerWorker<'_, QueryFunc, SpanCreator>
 where
     QueryFunc: Fn(Arc<Connection>, Consistency, PagingState) -> QueryFut,
-    QueryFut: Future<Output = Result<QueryResponse, UserRequestError>>,
+    QueryFut: Future<Output = Result<QueryResponse, RequestAttemptError>>,
     SpanCreator: Fn() -> RequestSpan,
 {
     // Contract: this function MUST send at least one item through self.sender
@@ -482,7 +482,7 @@ struct SingleConnectionPagerWorker<Fetcher> {
 impl<Fetcher, FetchFut> SingleConnectionPagerWorker<Fetcher>
 where
     Fetcher: Fn(PagingState) -> FetchFut + Send + Sync,
-    FetchFut: Future<Output = Result<QueryResponse, UserRequestError>> + Send,
+    FetchFut: Future<Output = Result<QueryResponse, RequestAttemptError>> + Send,
 {
     async fn work(mut self) -> PageSendAttemptedProof {
         match self.do_work().await {
@@ -499,10 +499,10 @@ where
         loop {
             let result = (self.fetcher)(paging_state)
                 .await
-                .map_err(UserRequestError::into_query_error)?;
+                .map_err(RequestAttemptError::into_query_error)?;
             let response = result
                 .into_non_error_query_response()
-                .map_err(UserRequestError::into_query_error)?;
+                .map_err(RequestAttemptError::into_query_error)?;
             match response.response {
                 NonErrorResponse::Result(result::Result::Rows((rows, paging_state_response))) => {
                     let (proof, send_result) = self
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -37,7 +37,7 @@ use crate::observability::driver_tracing::RequestSpan;
 use crate::observability::history::{self, HistoryListener};
 use crate::observability::metrics::Metrics;
 use crate::policies::load_balancing::{self, RoutingInfo};
-use crate::policies::retry::{QueryInfo, RetryDecision, RetrySession};
+use crate::policies::retry::{RequestInfo, RetryDecision, RetrySession};
 use crate::response::query_result::ColumnSpecs;
 use crate::response::{NonErrorQueryResponse, QueryResponse};
 use crate::statement::{prepared_statement::PreparedStatement, query::Query};
@@ -223,7 +223,7 @@ where
                 };
 
                 // Use retry policy to decide what to do next
-                let query_info = QueryInfo {
+                let query_info = RequestInfo {
                     error: &request_error,
                     is_idempotent: self.query_is_idempotent,
                     consistency: self.query_consistency,
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -329,7 +329,7 @@ where
                 self.log_query_success();
                 self.execution_profile
                     .load_balancing_policy
-                    .on_query_success(&self.statement_info, elapsed, node);
+                    .on_request_success(&self.statement_info, elapsed, node);
 
                 request_span.record_raw_rows_fields(&rows);
 
@@ -362,7 +362,7 @@ where
                 self.metrics.inc_failed_paged_queries();
                 self.execution_profile
                     .load_balancing_policy
-                    .on_query_failure(&self.statement_info, elapsed, node, &err);
+                    .on_request_failure(&self.statement_info, elapsed, node, &err);
                 Err(err.into_query_error())
             }
             Ok(NonErrorQueryResponse {
@@ -383,7 +383,7 @@ where
                     RequestAttemptError::UnexpectedResponse(response.response.to_response_kind());
                 self.execution_profile
                     .load_balancing_policy
-                    .on_query_failure(&self.statement_info, elapsed, node, &err);
+                    .on_request_failure(&self.statement_info, elapsed, node, &err);
                 Err(err.into_query_error())
             }
         }
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -206,7 +206,7 @@ where
 
                 let request_error: RequestAttemptError = match queries_result {
                     Ok(proof) => {
-                        trace!(parent: &span, "Query succeeded");
+                        trace!(parent: &span, "Request succeeded");
                         // query_pages returned Ok, so we are guaranteed
                         // that it attempted to send at least one page
                         // through self.sender and we can safely return now.
@@ -216,7 +216,7 @@ where
                         trace!(
                             parent: &span,
                             error = %error,
-                            "Query failed"
+                            "Request failed"
                         );
                         error
                     }
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -204,7 +204,7 @@ where
                     .instrument(span.clone())
                     .await;
 
-                last_error = match queries_result {
+                let request_error: RequestAttemptError = match queries_result {
                     Ok(proof) => {
                         trace!(parent: &span, "Query succeeded");
                         // query_pages returned Ok, so we are guaranteed
@@ -218,13 +218,13 @@ where
                             error = %error,
                             "Query failed"
                         );
-                        error.into_query_error()
+                        error
                     }
                 };
 
                 // Use retry policy to decide what to do next
                 let query_info = QueryInfo {
-                    error: &last_error,
+                    error: &request_error,
                     is_idempotent: self.query_is_idempotent,
                     consistency: self.query_consistency,
                 };
@@ -234,7 +234,10 @@ where
                     parent: &span,
                     retry_decision = format!("{:?}", retry_decision).as_str()
                 );
+
+                last_error = request_error.into_query_error();
                 self.log_attempt_error(&last_error, &retry_decision);
+
                 match retry_decision {
                     RetryDecision::RetrySameNode(cl) => {
                         self.metrics.inc_retries_num();
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -359,12 +359,11 @@ where
                 Ok(ControlFlow::Continue(()))
             }
             Err(err) => {
-                let err = err.into_query_error();
                 self.metrics.inc_failed_paged_queries();
                 self.execution_profile
                     .load_balancing_policy
                     .on_query_failure(&self.statement_info, elapsed, node, &err);
-                Err(err)
+                Err(err.into_query_error())
             }
             Ok(NonErrorQueryResponse {
                 response: NonErrorResponse::Result(_),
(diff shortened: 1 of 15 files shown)


scylla/src/transport/connection.rs
```diff
@@ -1296,7 +1296,7 @@ impl Connection {
         &self,
         init_batch: &'b Batch,
         values: impl BatchValues,
-    ) -> Result<Cow<'b, Batch>, QueryError> {
+    ) -> Result<Cow<'b, Batch>, UserRequestError> {
         let mut to_prepare = HashSet::<&str>::new();
 
         {
```
Lorak-mmk, line 1303: Commit: "batch: narrow the return error type of prepare_batch "

🤔 It is a bit surprising that this change does not require any other changes. This is due to some conversions done by "?", right? I wonder if we should maybe limit such conversion.
If error A has a variant with type B, then conversion from B to A is imo fine. But this is not the case here, right? This is a conversion between two separate errors. Maybe it should be explicit?

scylla/src/transport/connection.rs
```diff
@@ -1284,7 +1284,7 @@ impl Connection {
                     _ => Err(err.into()),
                 },
                 Response::Result(_) => Ok(query_response.into_query_result()?),
-                _ => Err(ProtocolError::UnexpectedResponse(
+                _ => Err(UserRequestError::UnexpectedResponse(
                     query_response.response.to_response_kind(),
                 )
                 .into()),
```
Lorak-mmk: It took me a few seconds to notice the "into()" call, and now I'm more convinced that such conversions are a bad idea. Why not return the variant that we want explicitly instead of relying on conversions?

scylla/src/transport/session.rs
```diff
@@ -1650,7 +1650,7 @@ where
 
         let span = RequestSpan::new_batch();
 
-        let run_request_result = self
+        let run_query_result: RunRequestResult<NonErrorQueryResponse> = self
             .run_request(
                 statement_info,
                 &batch.config,
```
Lorak-mmk: 🔧 Previous PR (which is included in this one change the name from using "query " to "request" but here you are changing it back. It looks like a mistake.

Lorak-mmk (changes requested):

scylla/src/transport/session.rs
muzarski: Yeah, I wrote this code a while ago. The previous PR was created just today. I had to miss it during a rebase (or there had to be no conflict in this place, so I did not notice). Good catch

scylla/src/transport/connection.rs
muzarski: This `into()` call is later removed anyway (commit `batch: return UserRequestError from batch_with_consistency`). But I agree with the suggestion, and will start applying it in the future.

scylla/src/transport/connection.rs
muzarski: `UserRequestError` has 11 variants. I think that `From<UserRequestError` for QueryError` impl comes in handy. I'm not a big fan of matching against all variants each time (assuming this is what you mean by explicit conversion).

Lorak-mmk: That is not what I meant. I'm talking specifically about `From` trait, because this is what seems to be used by `?` operator and `into` calls. 
A method like `into_query_error` would be totally fine.

muzarski: Okay, this makes sense. And I totally agree. Using `?` may lead to some unwanted implicit conversions.

muzarski:
v1.1: Rebased on latest https://github.com/scylladb/scylla-rust-driver/pull/1156 and addressed @Lorak-mmk comments. Most important change is that we replaced `From<UserRequestError> for QueryError` with explicit conversion method `UserRequestError::into_query_error()`.

Lorak-mmk (approved):
Looks good.

A suggestion for the future (because this PR is probably not a good place for this): could you re-order variants in UserRequestError so that they match the order in QueryError? Changing the QueryError instead would be fine too - I just think that the variants present in both should appear in the same order in both.

scylla/src/transport/errors.rs
```diff
@@ -908,6 +908,8 @@ pub(crate) enum UserRequestError {
     },
     #[error("Reprepared statement's id does not exist in the batch.")]
     RepreparedIdMissingInBatch,
+    #[error("Failed to serialize query paramters: {0}")]
+    SerializationError(#[from] SerializationError),
```
wprzytula: :pick: typo: `paramters`

scylla/src/transport/errors.rs
```diff
@@ -871,32 +871,52 @@ pub enum CqlEventHandlingError {
...
+
+    /// Failed to deserialize frame body extensions.
     #[error(transparent)]
     BodyExtensionsParseError(#[from] FrameBodyExtensionsParseError),
+
+    /// Driver was unable to allocate a stream id to execute a query on.
     #[error("Unable to allocate stream id")]
     UnableToAllocStreamId,
+
+    /// Prepared statement id changed after repreparation.
```
wprzytula, line 919: :camping:  Could we reorder variants of this enum so that they are grouped and sorted according to the moment in the request execution they can arise at?
I.e., serialization errors would come before deserialization errors, and `BodyExtensionsParseError` would go before `DbError` and `CqlResultParseError`.

This way it is easier to follow the possible errors along the request execution path.

scylla/src/transport/load_balancing/mod.rs
```diff
@@ -33,31 +33,31 @@ pub struct RoutingInfo<'a> {
...
-/// When a query is prepared to be sent to ScyllaDB/Cassandra, a `LoadBalancingPolicy`
+/// When a request is prepared to be sent to ScyllaDB/Cassandra, a `LoadBalancingPolicy`
 /// implementation constructs a load balancing plan. That plan is a list of
 /// targets (target is a node + an optional shard) to which
-/// the driver will try to send the query. The first elements of the plan are the targets which are
+/// the driver will try to send the request. The first elements of the plan are the targets which are
 /// the best to contact (e.g. they might have the lowest latency).
 ///
-/// Most queries are sent on the first try, so the query execution layer rarely needs to know more
+/// Most queries are sent on the first try, so the request execution layer rarely needs to know more
```
wprzytula, line 58: :wrench: Please replace the plural version, too:
`queries` -> `requests`

scylla/src/transport/load_balancing/mod.rs
```diff
@@ -33,31 +33,31 @@ pub struct RoutingInfo<'a> {
...
-/// Policy that decides which nodes and shards to contact for each query.
+/// Policy that decides which nodes and shards to contact for each request.
 ///
-/// When a query is prepared to be sent to ScyllaDB/Cassandra, a `LoadBalancingPolicy`
+/// When a request is prepared to be sent to ScyllaDB/Cassandra, a `LoadBalancingPolicy`
 /// implementation constructs a load balancing plan. That plan is a list of
 /// targets (target is a node + an optional shard) to which
-/// the driver will try to send the query. The first elements of the plan are the targets which are
+/// the driver will try to send the request. The first elements of the plan are the targets which are
 /// the best to contact (e.g. they might have the lowest latency).
```
wprzytula, line 56: :100: Thanks for this naming clarification!

scylla/src/transport/session.rs
```diff
@@ -2151,7 +2151,7 @@ where
                         trace!(parent: &span, "Query succeeded");
```
wprzytula, line 2151: :wrench: `Query` -> `Request`

scylla/src/transport/session.rs
```diff
@@ -2165,7 +2165,7 @@ where
                             "Query failed"
```
wprzytula, line 2165: :wrench: `Query` -> `Request`

scylla/src/transport/iterator.rs
```diff
@@ -217,7 +217,7 @@ where
                             error = %error,
                             "Query failed"
                         );
```
wprzytula: :wrench: :thought_balloon:  Here I'm not sure if we need the name change or not. After all, `QueryPager`, as the name suggests, should be mainly used for queries, so we can assume such usage in the naming.
OTOH Session is no longer going to log `Query failed`, but rather `Request failed`.

scylla/src/transport/retry_policy.rs
```diff
@@ -3,12 +3,12 @@
 //! the `RetryPolicy` trait
 
 use crate::frame::types::Consistency;
-use crate::transport::errors::{DbError, QueryError, WriteType};
+use crate::transport::errors::{DbError, UserRequestError, WriteType};
 
-/// Information about a failed query
-pub struct QueryInfo<'a> {
+/// Information about a failed request
+pub struct RequestInfo<'a> {
```
wprzytula, line 9: :pick: I'd appreciate this name change in a separate commit, because now it interferes with (rather small) logic changes due to `QueryError` -> `UserRequestError` change.

wprzytula: (no need to extract the commit now, it's more of a note for the future)

wprzytula (changes requested):

muzarski:
Addressed @wprzytula comments

scylla/src/transport/iterator.rs
muzarski: I decided to change it (to be consistent with session.rs).

wprzytula (approved):

muzarski:
Rebased on main

wprzytula (approved):

scylla/src/transport/errors.rs
```diff
@@ -908,6 +908,8 @@ pub(crate) enum UserRequestError {
     },
     #[error("Reprepared statement's id does not exist in the batch.")]
     RepreparedIdMissingInBatch,
+    #[error("Failed to serialize query parameters: {0}")]
+    SerializationError(#[from] SerializationError),
 }
 
 impl UserRequestError {
```
Lorak-mmk: Commit: "errors: Add serialization error to UserRequestError "

🤔 At this point I'd like comment on `QueryError` and `UserRequestError` that explain what they are for and what is the difference. I'm not 100% sure what is their intended meaning, which makes the changes a bit more difficult to understand.
My current understanding is that "UserRequestError" is an error that happened during a single try of a request execution (single try = no retries or speculative executions). Is that correct?

If it is correct, could you explain why you chose "**User**RequestError"? 
Who is the "user"?
User of the library? Probably not, because our internal queries in topology use that code too, right?
The caller? Sure, but then why "QueryError" does not have "User" in  the name?

If my understanding is correct, maybe "RequestAttemptError" would be a better name?

scylla/src/transport/errors.rs
```diff
@@ -871,32 +871,52 @@ pub enum CqlEventHandlingError {
...
+/// - `EXECUTE`
+/// - `BATCH`
 ///
-/// requests.
+/// requests. This error can appear during a single try
+/// of request execution. The retry decision is made based
+/// on this error.
 #[derive(Error, Debug)]
-pub(crate) enum UserRequestError {
+#[non_exhaustive]
```
Lorak-mmk: Commit: "errors: pubify UserRequestError"

🤔 Okay, so this comment confirms my understanding - and I'm even more sure that the name should be changed.

scylla/src/transport/downgrading_consistency_retry_policy.rs
```diff
@@ -49,11 +49,11 @@ impl Default for DowngradingConsistencyRetrySession {
...
 impl RetrySession for DowngradingConsistencyRetrySession {
-    fn decide_should_retry(&mut self, query_info: QueryInfo) -> RetryDecision {
+    fn decide_should_retry(&mut self, query_info: RequestInfo) -> RetryDecision {
         let cl = match query_info.consistency {
             Consistency::Serial | Consistency::LocalSerial => {
                 return match query_info.error {
-                    QueryError::DbError(DbError::Unavailable { .. }, _) => {
+                    UserRequestError::DbError(DbError::Unavailable { .. }, _) => {
                         // JAVA-764: if the requested consistency level is serial, it means that the operation failed at
                         // the paxos phase of a LWT.
```
Lorak-mmk: Commit: "retry_policy: decision based on UserRequestError"

🔧 There seem to be 2 changes combined in this commit:
- Narrowing the type, as the description says
- Renaming "QueryInfo" to "RequestInfo". Can those be separated?

scylla/src/transport/downgrading_consistency_retry_policy.rs
```diff
@@ -49,11 +49,11 @@ impl Default for DowngradingConsistencyRetrySession {
...
 
 impl RetrySession for DowngradingConsistencyRetrySession {
-    fn decide_should_retry(&mut self, query_info: QueryInfo) -> RetryDecision {
+    fn decide_should_retry(&mut self, query_info: RequestInfo) -> RetryDecision {
         let cl = match query_info.consistency {
             Consistency::Serial | Consistency::LocalSerial => {
                 return match query_info.error {
-                    QueryError::DbError(DbError::Unavailable { .. }, _) => {
+                    UserRequestError::DbError(DbError::Unavailable { .. }, _) => {
                         // JAVA-764: if the requested consistency level is serial, it means that the operation failed at
```
Lorak-mmk: 🔧 You changed the type of the argument, but not the name. Could you change the name too (here and in other places)?

scylla/src/transport/downgrading_consistency_retry_policy.rs
```diff
@@ -49,10 +49,10 @@ impl Default for DowngradingConsistencyRetrySession {
...
 impl RetrySession for DowngradingConsistencyRetrySession {
-    fn decide_should_retry(&mut self, query_info: RequestInfo) -> RetryDecision {
-        let cl = match query_info.consistency {
+    fn decide_should_retry(&mut self, request_info: RequestInfo) -> RetryDecision {
+        let cl = match request_info.consistency {
             Consistency::Serial | Consistency::LocalSerial => {
-                return match query_info.error {
+                return match request_info.error {
                     UserRequestError::DbError(DbError::Unavailable { .. }, _) => {
                         // JAVA-764: if the requested consistency level is serial, it means that the operation failed at
```
Lorak-mmk: Ah, you changed the name in this commit - this makes my previous commit obsolete.
However, I still think that "QueryInfo" -> "RequestInfo" rename should be moved into this commit (or into its own commit).

Lorak-mmk (changes requested):

scylla/src/transport/errors.rs
muzarski: >  My current understanding is that "UserRequestError" is an error that happened during a single try of a request execution (single try = no retries or speculative executions). Is that correct?

Correct.

> If it is correct, could you explain why you chose "**User**RequestError"? Who is the "user"? User of the library? Probably not, because our internal queries in topology use that code too, right? The caller? Sure, but then why "QueryError" does not have "User" in the name?

User of the library. I know that this name is not perfect. Probably the best choice would be to ditch the `User` prefix. However, we would need to rename the internal `RequestError` enum which represents a failure of a single attempt of ANY CQL request (including the ones sent during connection initialization - see `ConnectionSetupRequestError`). But since `RequestError` is currently a `pub(crate)` error type (and I do not plan to change that), I think we can safely rename it (to something like `RequestErrorInternal`)  if we don't like the **User** prefix in the public error type.

> If my understanding is correct, maybe "RequestAttemptError" would be a better name?

This (adding `Attempt` to error name) is actually done in https://github.com/scylladb/scylla-rust-driver/pull/1159/commits/1aab889cc9f08522aab811e01c76910a7a39f456 of follow-up [PR](https://github.com/scylladb/scylla-rust-driver/pull/1159)

scylla/src/transport/downgrading_consistency_retry_policy.rs
muzarski: Moved to `retry_policy: replace query mentions with request` commit

muzarski:
Rebased on main (modules refactor)

muzarski:
Flaky `test_coalescing` failed

mykaul:
> Flaky `test_coalescing` failed

Is there an open issue about it? Reference it please (or open one)

muzarski:
> > Flaky `test_coalescing` failed
> 
> Is there an open issue about it? Reference it please (or open one)

https://github.com/scylladb/scylla-rust-driver/issues/864

muzarski:
Rebased on main

muzarski:
v1.2: Renamed `UserRequestError` -> `RequestAttemptError` (commit taken from follow-up PR)

wprzytula (approved):

PR #1160, opened 2024-12-27
errors: pager API errors refactor

muzarski:

## Motivation
There are three main objectives:
### Purge pager.rs module of QueryError
It's yet another module where usage of `QueryError` was abused. It's taken care of in this PR

### Narrow return error type of `[poll_]next()` methods on iterators/streams
They now return a self-contained `NextRow` error. It's independent of `QueryError`.

### Step forward narrowing error type of metadata related functions
Currently all top-level metadata related functions/methods (e.g. `Session::refresh_metadata`) return `QueryError`. This is because low-level methods use `QueryPager` API under the hood. Before this PR, the pager API would return `QueryError`. In the follow-up PR we will continue on purging the metadata methods of `QueryError`. It will be replaced with an error type designated for these operations - namely `MetadataError`.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -29,8 +29,7 @@ use crate::cluster::{ClusterState, NodeRef};
 #[allow(deprecated)]
 use crate::cql_to_rust::{FromRow, FromRowError};
 use crate::deserialize::DeserializeOwnedRow;
-use crate::errors::{ProtocolError, RequestError};
-use crate::errors::{QueryError, RequestAttemptError};
+use crate::errors::{QueryError, RequestAttemptError, RequestError};
 use crate::frame::response::result;
 use crate::network::Connection;
 use crate::observability::driver_tracing::RequestSpan;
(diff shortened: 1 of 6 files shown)


wprzytula (approved):

scylla/src/client/pager.rs
```diff
@@ -1053,12 +1057,17 @@ where
...
+    /// Failed to run a request responsible for fetching new page.
+    #[error(transparent)]
+    RequestFailure(#[from] RequestError),
+
     /// Failed to deserialize result metadata associated with next page response.
     #[error("Failed to deserialize result metadata associated with next page response: {0}")]
     ResultMetadataParseError(#[from] ResultMetadataAndRowsCountParseError),
-    // TODO: This should also include a variant representing an error that occurred during
-    // query that fetches the next page. However, as of now, it would require that we include QueryError here.
-    // This would introduce a cyclic dependency: QueryError -> NextRowError -> NextPageError -> QueryError.
```
Lorak-mmk, line 1065: Commit: "iterator: narrow error type of internal items "
❓ You removed this comment, and included "RequestFailure". This does not introduce a cycle because you used "RequestError" instead of "QueryError", right?

scylla/src/cluster/metadata.rs
```diff
@@ -1881,7 +1881,15 @@ async fn query_table_partitioners(
...
-        Err(QueryError::DbError(DbError::Invalid, _)) => Ok(HashMap::new()),
+        // FIXME 2: The specific error we expect here should appear in QueryError::NextRowError. Currently
+        // leaving match against both variants. This will be fixed, once `MetadataError` is further adjusted
+        // in a follow-up PR. The goal is to return MetadataError from all functions related to metadata fetch.
+        Err(QueryError::DbError(DbError::Invalid, _))
+        | Err(QueryError::NextRowError(NextRowError::NextPageError(
+            NextPageError::RequestFailure(RequestError::LastAttemptError(
+                RequestAttemptError::DbError(DbError::Invalid, _),
+            )),
+        ))) => Ok(HashMap::new()),
```
Lorak-mmk, line 1892: ❓ Why can't you just move this commit before the previous one? Then all the commits would compile and pass tests, right?

scylla/src/client/pager.rs
muzarski: Correct.

scylla/src/cluster/metadata.rs
muzarski: True, I'll do that.

muzarski:
v1.1:
- Rebased on `main`
- Reordered commits so tests pass for all of them

scylla/src/cluster/metadata.rs
muzarski: Done, but I lied in the PR cover letter. I mentioned that penultimate commit breaks things. I moved the commit with the fix, before `iterator: narrow error type of internal items`.

wprzytula (approved):

Lorak-mmk (approved):

PR #1161, opened 2024-12-28
Add working `Debug` impl for caching session

athre0z:
The previous derived implementation was not actually usable because it had `Debug` bounds on the type parameters. The `RandomState` default value for the `S` argument does not impl `Debug`, so this didn't work.

This commit switches to a manual `Debug` impl that doesn't have these unnecessary bounds.

## Pre-review checklist


wprzytula (approved):
Thank you very much for the contribution, @athre0z !

Lorak-mmk (approved):

Issue #559, opened 2022-09-28
Prefer the local DC for control connections


piodul:
Currently, if the node which handled the control connection goes down, the driver will walk over the list of remaining nodes, in random order, sequentially trying to re-restablish the control connection until it manages to do it. However, this the random order does not consider DCs at all. While the latency of requests on control connection isn't too important, it would be nice if the schema/topology information didn't have to involve cross-DC traffic.

While we don't have a concept of a local DC in the Session (yet?), we can approximate it by assuming that the contact nodes are from the local DC. After establishing the first control connection, we should always prefer nodes from the DC of the first node.

mykaul:
@avelanarius - need to get higher priority (than medium) to complete the gap on topology awareness

mykaul:
> @avelanarius - need to get higher priority (than medium) to complete the gap on topology awareness

@avelanarius - ping - let's ensure this get a priority - I see it may have missed 1.0.0?

mykaul:
@avelanarius , @roydahan - ping - what's the latest here?

roydahan:
With the current plan it will miss 1.0.0.
If it's important enough (why? can you please give a background?), we can reconsider the plan or have it as part of 1.0.x.

mykaul:
> With the current plan it will miss 1.0.0. If it's important enough (why? can you please give a background?), we can reconsider the plan or have it as part of 1.0.x.

Cost, but more importantly, it confuses customers when they see those connections.

wprzytula:
IIUC, such heuristics would be satisfactory:
- we assume that that the contact nodes are from the local DC;
- the first control connection is opened to one of the contact nodes;
- after establishing the first control connection, the nodes from the DC of the first contacted node are always preferred for control connections.

WDYT? @Lorak-mmk @avelanarius @mykaul

wprzytula:
After discussion with @Lorak-mmk, the conclusion is that the above heuristics is a reasonable default, **but** we have to provide a way to override the DC preferred for control connections with a user-provided one. For that, a `SessionConfig`'s parameter should be added.

wprzytula:
Yet another observation:
The notion of _"the local DC"_ and _"the local rack"_ are constant for the driver instance; i.e., we could enable the user to set it only once, upon `Session` construction. Then, there would be configurable options for:
- control connection re-establishing,
- load balancing policy's query plan generation,
whether to prefer the local DC and whether to prefer the local rack or not.

[...]

Issue #184, opened 2021-02-28
Reconnect Policy


cvybhu:
Currently when a connection breaks we attempt to reconnect with a fixed 8 second cooldown.

Reconnection strategy should be configurable by the user. 
DataStax Java Driver has a configurable [reconnection policy](https://docs.datastax.com/en/developer/java-driver/4.10/manual/core/reconnection/), we should implement something similar in this driver.

`ReconnectPolicy` could be a trait and a `Box<dyn ReconnectPolicy>` could be included in `SessionConfig` and `ConnectionConfig`. Then `ConnectionKeeperWorker` would use it to decide when to shedule reconnect attemptls.

kiran-4444:
I'd like to take this up.

wprzytula:
> I'd like to take this up.

Go on! Anything you want to know before you begin?

kiran-4444:
> Go on! Anything you want to know before you begin?

Thanks! I'm currently playing out with the driver and seeing how it works. Will let you know if anything comes up.

Also, just a minor request: can you assign this to me? So that I can find all my assigned issues in one place and not miss anything. Thanks!

Lorak-mmk:
@wprzytula @muzarski We should probably discuss this at least a bit to provide a bit more specific requirements / guidance. Otherwise we'll end up doing this during billion review cycles.

## Scope of the policy 

There are 3 main scenarios where we need to establish connection:
1. Opening connections to new host. Either because new node joined the cluster, or because we are creating new session.
2. Working connection was broken, so we need to re-create it.
3. Opening a connection failed, we want to re-open it (possibly after some time).

Why 2 and 3 are different scenarios? Because they have different error types. 2 is described by `BrokenConnectionError`, but 3 needs some different error type (@muzarski correct me if I'm wrong).

**Re**connectionPolicy naturally covers scenarios 2 and 3. The question here is, should it also cover scenario 1?
- It is of course a new functionality, so it means more work
- Scenario 1 is not really about **re**-connecting, just connecting. We could change the name to ConnectionPolicy if that is a problem
- Alternative is to make yet another policy, so we would have both ReconnectPolicy and ConnectPolicy.

## API

How should the API look like?

My first intuition was to make something similar as we did in RetryPolicy. It means that ReconnectPolicy would have a method to create new ReconnectSession. This ReconnectSession would have method (/methods) that would be called after reconnection attempt and return decision about when to perform next attempt.

[...]

PR #1163, opened 2025-01-01
Restructure modules - purge `transport` module

wprzytula:
## Motivation
The division into modules used to be poor. The `transport` module has long been abused, as a bag full of everything, with no clear intention behind it.

## What's done

The `transport` module was dissolved. Its contents were grouped in other modules, many of them newly introduced. The new modules were documented with docstrings, so that it's clear what is the intended scope of each module.

Now, the module tree looks the following way (omitting the unchanged modules - `cloud`, `statement`, `utils`, and `authentication`):
```
- client
  - caching_session.rs
  - execution_profile.rs
  - pager.rs
  - self_identity.rs
  - session_builder.rs
  - session.rs
  - session_test.rs
- cluster
  - metadata.rs
  - node.rs
  - state.rs
  - worker.rs
- network
  - connection.rs
  - connection_pool.rs
- errors.rs (top-level module)
- policies
  - address_translator.rs
  - host_filter.rs
  - load_balancing
    - default.rs
    - plan.rs
  - retry
    - default.rs
    - downgrading_consistency.rs
    - fallthrough.rs
    - retry_policy.rs
  - speculative_execution.rs
- observability
  - driver_tracing.rs
  - history.rs
  - metrics.rs
  - tracing.rs
- response
  - legacy_query_result.rs
  - query_result.rs
  - request_response.rs
- routing
  - locator
    - (unchanged contents)
  - partitioner.rs
  - sharding.rs
```

## Pre-review checklist


scylla/src/cluster/mod.rs
```diff
@@ -1,3 +1,20 @@
...
+//!     - topology metadata,
+//!     - schema metadata,
+//!     - tablet metadata,
+//! - ClusterData, which is a snapshot of the cluster's state.
+//!   - ClusterData is replaced atomically upon a metadata refresh,
+//!     preventing any issues arising from mutability, including races.
+//! - HostFilter, which is a way to filter out some nodes and thus
+//!   not contact them at all on any condition.
+//!
+
```
Lorak-mmk, line 17: Commit: "cluster: add module docstring"

🔧 Again there are some internal details - please put them in a place that is not visible in a public documentation.

scylla/src/execution/mod.rs
```diff
@@ -1,3 +1,8 @@
 //! This module holds entities that control, customize, trace, and measure statement execution.
 //! This includes:
+//! - execution profiles,
 //! - TODO.
+
+pub mod execution_profile;
+
+pub use execution_profile::ExecutionProfile;
```
Lorak-mmk: Commit: "move execution_profile.rs to new execution module"

❓ Why re-export ExecutionProfile here? It belongs in the execution_profile module, right?
If the goal is to ease access to commonly used stuff, then imo we should create a prelude module instead of such duplications.

scylla/src/execution/retries/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+mod downgrading_consistency_rp;
+mod fallthrough_rp;
+mod retry_policy;
+
+pub use default_rp::{DefaultRetryPolicy, DefaultRetrySession};
+pub use downgrading_consistency_rp::{
+    DowngradingConsistencyRetryPolicy, DowngradingConsistencyRetrySession,
+};
+pub use fallthrough_rp::{FallthroughRetryPolicy, FallthroughRetrySession};
+pub use retry_policy::{QueryInfo, RetryDecision, RetryPolicy, RetrySession};
```
Lorak-mmk, line 11: Commit: "extract and restructure retries module"

🔧 I really don't like this "_rp" in module names.
Wdyt about the following structure:
```
- execution
    - retry_policy
        - mod.rs
        - retry_policy.rs
        - default.rs
        - fallthrough.rs
        - etc...
```

?

Lorak-mmk (changes requested):
Really great work, this is something that was really needed here!

I'm posting the review for ~first half of the PR so that you can address comments quicker.

A few thoughts on the structure after reading the description:
- `execution` module is quite big and contains various things that are not very related.
- First thing that comes to my mind is to split off "policies" module from "execution" module, containing load balancing, retries and speculative execution.
- I did not yet reach those modules in review, so I'm not 100% sure what they contain, but we could also split off something like "inspection" / "introspection" / "observability" (I'm not sure what is the best name) containing driver_tracing.rs, history.rs, metrics.rs and tracing.rs.
- errors.rs should imo be a top-level file as it is not specific to "execution" module.
- After those changes, "execution" is left with "execution_profile.rs" and "pager.rs". It can be left like this, or maybe merged with session? I'm not sure.

Apart from removing oversized "execution" module, I really believe those changes would simply improve the structure.

scylla/src/session/mod.rs
```diff
@@ -1,3 +1,14 @@
...
+//! driver session with the cluster.
+//! The following abstractions are involved:
+//! - Session - the main entity of the driver. It:
+//!     - contains and manages all driver configuration,
+//!     - launches and communicates with ClusterWorker (see cluster module for more info),
+//!     - enables executing CQL requests, taking all configuration into consideration.
+//! - SessionBuilder - just a convenient builder for a Session.
+//! - CachingSession - a wrapper over a Session that keeps and manages a cache
+//!   of prepared statements, so that a user can be free of such considerations.
+
```
Lorak-mmk, line 11: Commit: "execution: add module docstring"

🔧 Commit message seems to be wrong - commit adds doc comment for session, not for execution.

scylla/src/execution/mod.rs
```diff
@@ -6,12 +6,14 @@
...
+//! - driver metrics,
 //! - request execution history,
 //! - TODO.
 
 pub mod execution_profile;
 pub mod history;
 pub mod load_balancing;
+pub(crate) mod metrics;
 pub mod retries;
 pub mod speculative_execution;
```
Lorak-mmk: 🌱 Internal code refers to metrics using the full path (`crate::execution::metrics::Metrics`) which is not accessible for the user. I believe we should make this module pub, and the current `scylla::Metrics` reexport should become `scylla::prelude::Metrics`. But this is something for a follow-up.

scylla/src/response/mod.rs
Lorak-mmk, line 1: Commit "re-export PagingState[Response] in response mod"

❓ Why?

scylla/src/response/mod.rs
```diff
@@ -1,3 +1,11 @@
+//! This module holds entities that represent responses to requests
+//! sent to the cluster by the driver.
+//! The following abstractions are involved:
+//! - QueryResponse - a response to any kind of a CQL request.
+//! - QueryResult - a result of a CQL QUERY/EXECUTE/BATCH request.
+//! - QueryRowsResult - a result of CQL QUERY/EXECUTE/BATCH request that contains
+//!   some rows, which can be deserialized by the user.
+
```
Lorak-mmk, line 8: 🔧 Again there are internal details in external doc.

Lorak-mmk (changes requested):
Second half was much quicker to review :D

Lorak-mmk:
Oh wow, I understand now why some comments are marked as outdated despite there not being any changes since the review.
GitHub marks comments as outdated if the current final state of the commented code is different - e.g. if the next commits changed this code again.

Another example of GitHub being not an appropriate tool for reviews.

wprzytula:
> Really great work, this is something that was really needed here!

:tada: 

> I'm posting the review for ~first half of the PR so that you can address comments quicker.
> 
> A few thoughts on the structure after reading the description:
> 
>   * `execution` module is quite big and contains various things that are not very related.

It's quite big, that's true.
> 
>   * First thing that comes to my mind is to split off "policies" module from "execution" module, containing load balancing, retries and speculative execution.

Extracting a `policies` module makes sense to me, but I would still keep it in the `execution` module, because my intention was to hold all _entities that control and configure statement execution_ in the `execution` module.

> 
>     * I did not yet reach those modules in review, so I'm not 100% sure what they contain, but we could also split off something like "inspection" / "introspection" / "observability" (I'm not sure what is the best name) containing driver_tracing.rs, history.rs, metrics.rs and tracing.rs.

:100: , I'll extract those into a new top-level `observability` module.

>     * errors.rs should imo be a top-level file as it is not specific to "execution" module.

Agreed.
 
>     * After those changes, "execution" is left with "execution_profile.rs" and "pager.rs". It can be left like this, or maybe merged with session? I'm not sure.

To me, `execution_profile.rs`, together with policies, makes great sense to constitute the `execution` module. However, I've not been sure at all where to put `pager.rs`. If not in `execution`, `session` would be my second guess.
> 
> Apart from removing oversized "execution" module, I really believe those changes would simply improve the structure.

WDYT about my above alternative suggestions to improve the `execution` module?
 
> I see you did not mark "All commits compile, pass static checks and pass test.", but I see that the commits I reviewed until now are done in a way to maintains this property. I found one problem which I posted in a comment. It would be really nice to fix this and potentially other such issues and maintain the passing state of CI throughout the PR - it looks like it should not be a lot of effort now.

:tada:

scylla/src/session/session.rs
```diff
@@ -75,14 +75,12 @@ use crate::transport::query_result::RowsError;
...
+use crate::transport::NodeRef;
 use crate::{
     batch::{Batch, BatchStatement},
     statement::StatementConfig,
 };
 
-pub use crate::transport::connection_pool::PoolSize;
-
 // This re-export is to preserve backward compatibility.
 // Those items are no longer here not to clutter session.rs with legacy things.
```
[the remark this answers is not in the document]
wprzytula: Perhaps because it's the Session that is used to set up the pool size - from a user standpoint, it happens when SessionBuilder is fed with a given pool size. OTOH, `connection_pool` is not a `pub` module.

scylla/src/cluster/mod.rs
```diff
@@ -1,6 +1,9 @@
 #[allow(clippy::module_inception)]
 mod cluster;
 pub mod metadata;
+pub(crate) mod node;
 
 pub use cluster::ClusterData;
 pub(crate) use cluster::{use_keyspace_result, Cluster, ClusterNeatDebug};
+
+pub use node::{KnownNode, Node, NodeAddr, NodeRef};
```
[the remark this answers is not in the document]
wprzytula: I have no immediate preferred choice.

docs/source/load-balancing/load-balancing.md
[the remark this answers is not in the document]
wprzytula: Well, `HostFilter` controls which nodes will have their `NodeConnectionPool` created for in `cluster.rs` logic, so it's used in the cluster management layer.
OTOH, `AddressTranslator` is used on the connection layer, to translate address before opening a new connection.
This explains their different position in the module tree to me, what about you?

scylla/src/execution/mod.rs
wprzytula: Hmm, I've found this a usual thing to re-export the main entity of a module in its parent module.
This way we avoid to long, verbose paths, such as `b::a::A` - instead we have `b::A`.

scylla/src/execution/retries/mod.rs
wprzytula: It's OK to me. Do you agree with the name of module to be `retries` instead of `retry_policy`, as I did in the following commit?

scylla/src/response/mod.rs
wprzytula: Hmm, I wrongly believed that there had been such re-export in `transport`.
I can remove the re-export if you believe it's superfluous.

scylla/src/connection/connection.rs
```diff
@@ -2086,6 +2086,28 @@ pub(crate) async fn open_connection(
     Ok((connection, error_receiver))
 }
 
+pub(crate) async fn open_connection_to_shard_aware_port(
+    endpoint: UntranslatedEndpoint,
+    shard: Shard,
+    sharder: Sharder,
+    connection_config: &ConnectionConfig,
+) -> Result<(Connection, ErrorReceiver), ConnectionError> {
```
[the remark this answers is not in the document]
wprzytula: I've reduced visibility of multiple functions in `connection.rs` to `pub(self)` or `pub(super` - see a new commit: `connection: reduce visibility of some functions`.

scylla/src/connection/connection_pool.rs
```diff
@@ -1208,6 +1208,8 @@ mod tests {
     #[tokio::test]
     #[cfg(not(scylla_cloud_tests))]
     async fn many_connections() {
+        use crate::connection::connection::open_connection_to_shard_aware_port;
+
         setup_tracing();
         let connections_number = 512;
```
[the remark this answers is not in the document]
wprzytula: It's rust analyzer / rustfmt who did that...
Moved it out.

Lorak-mmk:
> 
> Extracting a `policies` module makes sense to me, but I would still keep it in the `execution` module, because my intention was to hold all _entities that control and configure statement execution_ in the `execution` module.
> 

I proposed it as a top-level module because of the possibility that in the future we add policies that are not related to statement execution. There is for example an issue about ReconnectPolicy: https://github.com/scylladb/scylla-rust-driver/issues/184 and  related https://github.com/scylladb/scylla-rust-driver/issues/559 . 

There is also some relevant work in gocql: https://github.com/scylladb/gocql/pull/231

If we keep `policies` inside `execution` then we have to either add such policies there (which is weird because it does not affect statement execution), or outside the policies module (which is weird for obvious reasons).

scylla/src/transport/mod.rs
```diff
@@ -1,7 +1,7 @@
 pub(crate) mod caching_session;
-mod cluster;
+pub(crate) mod cluster;
 pub(crate) mod connection;
-mod connection_pool;
+pub(crate) mod connection_pool;
 pub mod downgrading_consistency_retry_policy;
 pub mod errors;
 pub mod execution_profile;
```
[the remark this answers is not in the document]
wprzytula: Actually, it's the opposite. My commit make some modules `pub(crate)`, whereas they had used to be `pub(self)`.

wprzytula:
> > Extracting a `policies` module makes sense to me, but I would still keep it in the `execution` module, because my intention was to hold all _entities that control and configure statement execution_ in the `execution` module.
> 
> I proposed it as a top-level module because of the possibility that in the future we add policies that are not related to statement execution. There is for example an issue about ReconnectPolicy: #184 and related #559 .

I see no reason to group various policies in one module just because they are all called _policies_. To me, execution-related policies can be grouped together, but `ReconnectPolicy` should **not** be anywhere close to those.

scylla/src/connection/connection.rs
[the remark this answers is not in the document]
wprzytula: :memo: I'm going to tidy up the `connection` module even more after 1.0, because the structural mess there makes it much harder to grasp.

scylla/src/connection/mod.rs
```diff
@@ -1,3 +1,10 @@
+//! This module holds entities that represent connections to the cluster
+//! and management over those connections (connection pooling).
+//! This includes two main abstractions:
+//! - Connection - a single, possibly encrypted, connection to a Scylla node over CQL protocol,
+//! - NodeConnectionPool - a manager that keeps a desired number of connections opened to each shard.
+
```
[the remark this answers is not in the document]
wprzytula: > 🔧 Doc comments are part of externally visible documentation, made for users.

I agree that doc comments are **_mainly_** intended for the users.

However, **doc comments give one advantage to us as well**: when hovering a name of an entity, VSCode displays its doc comment, which facilitates quick recap of what is its purpose.

Also, when building docs with a proper flag: 
```bash
cargo doc --document-private-items
```
we end up with a nice documentation of private entities, too, which can be a great source of Big Picture knowledge of the driver for new contributors. 

**I agree** that the current state is bad**, because users will see mentions of private entities in the public documentation. It's a pity, though, that if I make this particular doc comment an ordinary comment, VSCode will no longer display it when hovered `connection` module, neither in `pub mod connection;` in `lib.rs` nor in e.g. `use scylla::connection::PoolConfig;`.

Any ideas how to reconcile these problems?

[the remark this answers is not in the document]
wprzytula: BTW, have I convinced you that it's worth keeping doc comments of private items?
cc @muzarski

Lorak-mmk:
> 
> I see no reason to group various policies in one module just because they are all called _policies_. To me, execution-related policies can be grouped together, but `ReconnectPolicy` should **not** be anywhere close to those.
> 
> It makes sense IMO to have `scylla::execution::policies` path, because it sounds good when read from left to right, showing clearly what's inside and how it's related.

The reason I see is that the policies (either execution-related or not) are the main (only?) pluggable points of the driver - user can influence the behavior of the driver by writing their own implementation of those policies.
It would be nice to have all such customization points in one place.

This is done similarly in other driver, so keeping the structure familiar for other people is an additional advantage. 

Java Driver 3.x has a policies folder: https://github.com/scylladb/java-driver/tree/scylla-3.x/driver-core/src/main/java/com/datastax/driver/core/policies
It includes not only LBP / Retry / SE, but also AddressTranslator, ReconnectionPolicy and HostFilter.

Python Driver has policies.py file: https://github.com/scylladb/python-driver/blob/master/cassandra/policies.py
It contains LBP, Retry and SE and also `AddressTranslator` and `ReconnectionPolicy`

GoCql has a policies.go file https://github.com/scylladb/gocql/blob/master/policies.go
It contans retry policies, LBP, SE and also reconnection policy and other stuff (`ConvictionPolicy`, `ReadyPolicy`)

scylla/src/execution/retries/mod.rs
Lorak-mmk: Assuming that the parent module is `policies`, and contain `load_balancing` and `speculative_execution` submodules, I think the name `retry` would fit better.

scylla/src/transport/mod.rs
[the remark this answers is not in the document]
Lorak-mmk: wow, I must have been sleepy when writing this comment, sorry.

wprzytula:
> The reason I see is that the policies (either execution-related or not) are the main (only?) pluggable points of the driver - user can influence the behavior of the driver by writing their own implementation of those policies. It would be nice to have all such customization points in one place.
> 
> This is done similarly in other driver, so keeping the structure familiar for other people is an additional advantage.

This convinced me. Let's make a similar module for all user-implementable traits that configure driver's behaviour.

wprzytula:
v2.0: addressed some of the comments (mostly the resolved ones).

wprzytula:
v2.1: addressed another bunch of comments.
- extracted `policies` module with LBP, SEP, RP, HostFilter and AddressTranslator;
- moved `pager.rs` and `execution_profile.rs` to `session` module, annihilated `execution` module.

wprzytula:
v2.1.1: fixed docs.

wprzytula:
The module tree should reflect all suggestions proposed by @Lorak-mmk. Please confirm that this is the case.

wprzytula:
v2.1.2: appeased Clippy.

scylla/src/execution/retries/mod.rs
Lorak-mmk: Marking as resolved .

scylla/src/execution/mod.rs
Lorak-mmk: After the changes (`execution_profile` is in `session` now) I think we should either:
- re-export both `ExecutionProfile` and `ExecutionProfileBuilder` and then make `execution_profile` mod private (it does not define any other public items as far as I can tell).
- re-export neither, keep the mod public.

Rationale: typical use case requires building the profile, and so importing the builder too. I don't see compelling reason to re-export one struct but not the other.

scylla/src/cluster/mod.rs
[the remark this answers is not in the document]
Lorak-mmk: Let's leave it as is.

scylla/src/connection/mod.rs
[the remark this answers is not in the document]
Lorak-mmk: The same can be done for `cluster` module.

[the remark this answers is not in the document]
wprzytula: > Remove `pub(crate)` re-export to force driver code to import using full path (`connection::connection`).

This is ugly. I'll give up and make those docstrings ordinary comments.

scylla/src/execution/mod.rs
wprzytula: > * re-export both `ExecutionProfile` and `ExecutionProfileBuilder` and then make `execution_profile` mod private (it does not define any other public items as far as I can tell).

This is my preferred solution.

wprzytula: We also need public access to `ExecutionProfileHandle`, so I'm re-exporting it, too.

scylla/src/cluster/mod.rs
wprzytula: As agreed in person, this particular docstring is OK.

scylla/src/connection/mod.rs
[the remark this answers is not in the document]
wprzytula: Eventually, I removed exclamation marks from comments referencing some nonpublic items, removing them from docs.

scylla/src/response/mod.rs
wprzytula: Actually, the following line indeed has been there in `transport/mod.rs` before:
`pub use scylla_cql::frame::request::query::{PagingState, PagingStateResponse};`
Do you believe I should just remove it and not re-export it in `response/mod.rs`?

wprzytula: I removed it.

scylla/src/execution/mod.rs
wprzytula: I made the module `pub` and removed the re-export.

wprzytula:
v2.2: addressed all remaining comments.

Lorak-mmk (approved):

piodul:
I only reviewed the final structure, and my only issue would be with the `cluster`, `connection` and `session` modules. Their names are nouns which actually correspond to some concrete types in the code - AFAIK we do have `Cluster(Data)`, `Connection` and `Session` types; their names would be better suited as names of leaf modules (and all of them actually have a sub-module of the same name), while inner modules usually have names that are less concrete or are plural nouns (`errors`, `policies`, `observability`, `response`).

Perhaps you can think about the names for those modules a bit more, unfortunately I don't have a good idea for an alternative. Not sure if it is a big deal if you can't come up with better names.

(This thought was prompted by the fact that I noticed you disabling the `module_inception` clippy lint)

wprzytula:
> I only reviewed the final structure, and my only issue would be with the `cluster`, `connection` and `session` modules. Their names are nouns which actually correspond to some concrete types in the code - AFAIK we do have `Cluster(Data)`, `Connection` and `Session` types; their names would be better suited as names of leaf modules (and all of them actually have a sub-module of the same name), while inner modules usually have names that are less concrete or are plural nouns (`errors`, `policies`, `observability`, `response`).
> 
> Perhaps you can think about the names for those modules a bit more, unfortunately I don't have a good idea for an alternative. Not sure if it is a big deal if you can't come up with better names.
> 
> (This thought was prompted by the fact that I noticed you disabling the `module_inception` clippy lint)

- One idea that has come into my mind was: `session` -> `client`.
- Another idea (which I actually forgot to include in this revived version of the PR, but the old one contained it) is to split `cluster.rs` into `cluster_data.rs` (with `ClusterData`) and fully-`pub(crate) `cluster_worker.rs` (with `ClusterWorker` and `Cluster`).
- Regarding `connection`, we could rename it to `network` and make it nonpublic. `SelfIdentify` and `PoolSize` could be moved to `client` and accessible from there.

WDYT @piodul?

piodul:
>     * One idea that has come into my mind was: `session` -> `client`.

Makes sense, "client session" sounds natural.

> 
>     * Another idea (which I actually forgot to include in this revived version of the PR, but the old one contained it) is to split `cluster.rs` into `cluster_data.rs` (with `ClusterData`) and fully-`pub(crate)` `cluster_worker.rs` (with `ClusterWorker` and `Cluster`).

What about the stuff in `node.rs`?

> 
>     * Regarding `connection`, we could rename it to `network` and make it nonpublic. `SelfIdentify` and `PoolSize` could be moved to `client` and accessible from there.

Also sounds good, from the user's perspective these types (`SelfIdentity` and `PoolSize`) are used when configuring the driver's session, so it makes sense to me to put them close to `Session`/`SessionConfig`.

wprzytula:
> What about the stuff in `node.rs`?

What's the question about it? It's there, and it's only accessible by the `pub` reexport in `cluster/mod.rs`.

wprzytula:
> I'm still not sure about the naming in this particular module. Having `cluster/cluster_data.rs` and `cluster/cluster_worker.rs` instead of `cluster/cluster.rs` will satisfy clippy, but I'm wondering whether the "cluster data" is a good name at all. We have `cluster/metadata`, `cluster/cluster_data`... while "metadata" does make some sense (it's data that describes how data looks like in the cluster), "cluster data" looks a bit weird next to it.

I recognise your doubts. How would you solve this?

wprzytula:
v2.3:
- addressed @piodul's comments:
  - toplevel `session` -> `client`,
  - toplevel `connection` -> `network`,
  - nested `cluster` -> `worker` + `state` (and renamed `ClusterData` to `ClusterState`).
- moved `SelfIdentity` to `client`,
- re-exported `Compression` in `client`,
- removed re-exports from crate root; items should be only accessible through one path + optionally the `prelude` mod (coming soon, in a follow-up),

wprzytula:
v3.2.2: fixed serverless tests.

wprzytula:
v2.3.3: fixed the book once again.

wprzytula:
v2.2.4 - fixed serverless again...

Lorak-mmk (approved):

piodul:
One more thing - it would probably be nice to mention in the description of the PR which types were renamed, for posterity (take the contents of https://github.com/scylladb/scylla-rust-driver/pull/1163#issuecomment-2581038543 and put them in the description). Otherwise, LGTM.

Lorak-mmk:
> One more thing - it would probably be nice to mention in the description of the PR which types were renamed, for posterity (take the contents of [#1163 (comment)](https://github.com/scylladb/scylla-rust-driver/pull/1163#issuecomment-2581038543) and put them in the description). Otherwise, LGTM.

We'll definitely need to put  description of those changes in the release notes.

PR #1164, opened 2025-01-02
Custom timestamps as a per-execution setting

wprzytula:
**Note:** this is based on #1163 not to incur the rebasing effort when #1163 gets merged.
Please start review with the commit `session: pass custom timestamp in request execution APIs`.

# Motivation

As noted in #262, a timestamp is something related to a single execution of a statement, not to the statement itself. Also, setting a timestamp on a statement inside a Batch would not set a timestamp on the Batch itself, which [has shown to be a pitfall](https://github.com/scylladb/scylla-rust-driver/issues/262#issuecomment-1140966595).

# What's done
This PR introduces a new per-execution timestamp APIs on `Session` and `CachingSession`, by duplicating all execution methods and making the second version accept a custom timeout. For example, we had `Session::query_single_page(query, values)`, so now we have `Session::query_single_page_with_timestamp(query, values, timestamp)`, too.

The old API involving having timestamp on statements is removed. Statements can no longer hold a timestamp.

### Non-viable alternatives
[The original idea of @Lorak-mmk](https://github.com/scylladb/scylla-rust-driver/issues/262#issuecomment-2545604361) to leverage the `IntoFuture` trait to avoid API duplication showed to not be an option for us, because `IntoFuture` requires specifying (i.e., naming) the type of the future, which in our case is an anonymous type (being a future driving the `async` block of executing the statement).

Fixes: #262

## Pre-review checklist


wprzytula:
To be discussed after 1.0.

Lorak-mmk:
Closed as too ugly

PR #1165, opened 2025-01-06
CQL Vector support

### LINKED ISSUE #1014: CQL Vector type

smoczy123:
This PR adds serialization and deserialization of CQL Vector (as implemented in Cassandra) therefore achieving compatibility with Cassandra's Vector type. It's important to note that Cassandra implements Vector serialization and deserialization in a way that
contradicts the CQL protocol, using [unsigned vint] instead of [int] as the element size encoding for variable type length vectors.

Fixes #1014 

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/deserialize/value.rs
+++ b/scylla-cql/src/deserialize/value.rs
@@ -826,18 +826,40 @@ where
     T: DeserializeValue<'frame, 'metadata>,
 {
     fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
-        // It makes sense for both Set and List to deserialize to Vec.
-        ListlikeIterator::<'frame, 'metadata, T>::type_check(typ)
-            .map_err(typck_error_replace_rust_name::<Self>)
+        // It makes sense for Set, List and Vector to deserialize to Vec.
+        match typ {
+            ColumnType::Collection {
+                typ: CollectionType::List(_) | CollectionType::Set(_),
+                ..
+            } => ListlikeIterator::<'frame, 'metadata, T>::type_check(typ)
+                .map_err(typck_error_replace_rust_name::<Self>),
+            ColumnType::Vector { .. } => VectorIterator::<'frame, 'metadata, T>::type_check(typ)
+                .map_err(typck_error_replace_rust_name::<Self>),
+            _ => Err(mk_typck_err::<Self>(
+                typ,
+                BuiltinTypeCheckErrorKind::NotDeserializableToVec,
+            )),
+        }
     }
 
     fn deserialize(
         typ: &'metadata ColumnType<'metadata>,
         v: Option<FrameSlice<'frame>>,
     ) -> Result<Self, DeserializationError> {
-        ListlikeIterator::<'frame, 'metadata, T>::deserialize(typ, v)
-            .and_then(|it| it.collect::<Result<_, DeserializationError>>())
-            .map_err(deser_error_replace_rust_name::<Self>)
+        match typ {
+            ColumnType::Collection {
+                typ: CollectionType::List(_) | CollectionType::Set(_),
+                ..
+            } => ListlikeIterator::<'frame, 'metadata, T>::deserialize(typ, v)
+                .and_then(|it| it.collect::<Result<_, DeserializationError>>())
+                .map_err(deser_error_replace_rust_name::<Self>),
+            ColumnType::Vector { .. } => {
+                VectorIterator::<'frame, 'metadata, T>::deserialize(typ, v)
+                    .and_then(|it| it.collect::<Result<_, DeserializationError>>())
+                    .map_err(deser_error_replace_rust_name::<Self>)
+            }
+            _ => unreachable!("Should be prevented by typecheck"),
+        }
     }
 }
 
@@ -899,6 +921,182 @@ where
         ListlikeIterator::<'frame, 'metadata, T>::deserialize(typ, v)
             .and_then(|it| it.collect::<Result<_, DeserializationError>>())
             .map_err(deser_error_replace_rust_name::<Self>)
+    }
+}
+
+/// A deserialization iterator over a CQL vector.
+///
+/// Deserialization of a vector is done in two ways, depending on the element type:
+/// 1. If the element type is of a fixed-length, the iterator reads
+///    a fixed, known number of bytes for each element and deserializes
+///    according to the element type. The element count and element size
+///    are not encoded as the count is known from the vector type
+///    and the size is known from the element type.
+///
+/// 2. If the element type is of a variable-length, the iterator reads an vint
+///    (variable-length integer) for each element, which indicates the number of bytes
+///    to read for that element. Then, the iterator reads the specified number of bytes
+///    and deserializes according to the element type. This, however, is in contradiction
+///    with the CQL protocol, which specifies that the element count is encoded as a 4 byte int!
+///    Alas, Cassandra does not respect the protocol, so ScyllaDB and this driver have to follow.
+///
+/// It would be nice to have a rule to determine if the element type is fixed-length or not,
+/// however, we only have a heuristic. There are a few types that should, for all intents and purposes,
+/// be considered fixed-length, but are not, e.g TinyInt. See ColumnType::type_size() for the list.
+pub struct VectorIterator<'frame, 'metadata, T> {
+    collection_type: &'metadata ColumnType<'metadata>,
+    element_type: &'metadata ColumnType<'metadata>,
+    remaining: usize,
+    element_length: Option<usize>,
+    slice: FrameSlice<'frame>,
+    phantom_data: std::marker::PhantomData<T>,
+}
+
+impl<'frame, 'metadata, T> VectorIterator<'frame, 'metadata, T> {
+    fn new(
+        collection_type: &'metadata ColumnType<'metadata>,
+        element_type: &'metadata ColumnType<'metadata>,
+        count: usize,
+        element_length: Option<usize>,
+        slice: FrameSlice<'frame>,
+    ) -> Self {
+        Self {
+            collection_type,
+            element_type,
+            remaining: count,
+            element_length,
+            slice,
+            phantom_data: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<'frame, 'metadata, T> DeserializeValue<'frame, 'metadata>
+    for VectorIterator<'frame, 'metadata, T>
+where
+    T: DeserializeValue<'frame, 'metadata>,
+{
+    fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
+        match typ {
+            ColumnType::Vector { typ: t, .. } => {
+                <T as DeserializeValue<'frame, 'metadata>>::type_check(t).map_err(|err| {
+                    mk_typck_err::<Self>(typ, VectorTypeCheckErrorKind::ElementTypeCheckFailed(err))
+                })?;
+                Ok(())
+            }
+            _ => Err(mk_typck_err::<Self>(
+                typ,
+                VectorTypeCheckErrorKind::NotVector,
+            )),
+        }
+    }
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        let (element_type, dimensions) = match typ {
+            ColumnType::Vector {
+                typ: element_type,
+                dimensions,
+            } => (element_type, dimensions),
+            _ => {
+                unreachable!("Typecheck should have prevented this scenario!")
+            }
+        };
+
+        let v = ensure_not_null_frame_slice::<Self>(typ, v)?;
+
+        Ok(Self::new(
+            typ,
+            element_type,
+            *dimensions as usize,
+            element_type.type_size(),
+            v,
+        ))
+    }
+}
+
+impl<'frame, 'metadata, T> VectorIterator<'frame, 'metadata, T>
+where
+    T: DeserializeValue<'frame, 'metadata>,
+{
+    fn next_constant_length_elem(
+        &mut self,
+        element_length: usize,
+    ) -> Option<<Self as Iterator>::Item> {
+        self.remaining = self.remaining.checked_sub(1)?;
+        let raw = self.slice.read_n_bytes(element_length).map_err(|err| {
+            mk_deser_err::<Self>(
+                self.collection_type,
+                BuiltinDeserializationErrorKind::RawCqlBytesReadError(err),
+            )
+        });
+        Some(raw.and_then(|raw| {
+            T::deserialize(self.element_type, raw).map_err(|err| {
+                mk_deser_err::<Self>(
+                    self.collection_type,
+                    VectorDeserializationErrorKind::ElementDeserializationFailed(err),
+                )
+            })
+        }))
+    }
+
+    fn next_variable_length_elem(&mut self) -> Option<<Self as Iterator>::Item> {
+        self.remaining = self.remaining.checked_sub(1)?;
+        let size = types::unsigned_vint_decode(self.slice.as_slice_mut()).map_err(|err| {
+            mk_deser_err::<Self>(
+                self.collection_type,
+                BuiltinDeserializationErrorKind::RawCqlBytesReadError(
+                    LowLevelDeserializationError::IoError(Arc::new(err)),
+                ),
+            )
+        });
+        let raw = size
+            .and_then(|size| {
+                size.try_into().map_err(|_| {
+                    mk_deser_err::<Self>(
+                        self.collection_type,
+                        BuiltinDeserializationErrorKind::ValueOverflow,
+                    )
+                })
+            })
+            .and_then(|size: usize| {
+                self.slice.read_n_bytes(size).map_err(|err| {
+                    mk_deser_err::<Self>(
+                        self.collection_type,
+                        BuiltinDeserializationErrorKind::RawCqlBytesReadError(err),
+                    )
+                })
+            });
+
+        Some(raw.and_then(|raw| {
+            T::deserialize(self.element_type, raw).map_err(|err| {
+                mk_deser_err::<Self>(
+                    self.element_type,
+                    VectorDeserializationErrorKind::ElementDeserializationFailed(err),
+                )
+            })
+        }))
+    }
+}
+
+impl<'frame, 'metadata, T> Iterator for VectorIterator<'frame, 'metadata, T>
+where
+    T: DeserializeValue<'frame, 'metadata>,
+{
+    type Item = Result<T, DeserializationError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.element_length {
+            Some(element_length) => self.next_constant_length_elem(element_length),
+            None => self.next_variable_length_elem(),
+        }
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.remaining, Some(self.remaining))
     }
 }
 
@@ -1507,6 +1705,9 @@ pub enum BuiltinTypeCheckErrorKind {
     /// A type check failure specific to a CQL set or list.
     SetOrListError(SetOrListTypeCheckErrorKind),
 
+    /// A type check failure specific to a CQL vector.
+    VectorError(VectorTypeCheckErrorKind),
+
     /// A type check failure specific to a CQL map.
     MapError(MapTypeCheckErrorKind),
 
@@ -1515,12 +1716,22 @@ pub enum BuiltinTypeCheckErrorKind {
 
     /// A type check failure specific to a CQL UDT.
     UdtError(UdtTypeCheckErrorKind),
+
+    /// A type check detected type not deserializable to a vector.
+    NotDeserializableToVec,
 }
 
 impl From<SetOrListTypeCheckErrorKind> for BuiltinTypeCheckErrorKind {
     #[inline]
     fn from(value: SetOrListTypeCheckErrorKind) -> Self {
         BuiltinTypeCheckErrorKind::SetOrListError(value)
+    }
+}
+
+impl From<VectorTypeCheckErrorKind> for BuiltinTypeCheckErrorKind {
+    #[inline]
+    fn from(value: VectorTypeCheckErrorKind) -> Self {
+        BuiltinTypeCheckErrorKind::VectorError(value)
     }
 }
 
@@ -1552,9 +1763,13 @@ impl Display for BuiltinTypeCheckErrorKind {
                 write!(f, "expected one of the CQL types: {expected:?}")
             }
             BuiltinTypeCheckErrorKind::SetOrListError(err) => err.fmt(f),
+            BuiltinTypeCheckErrorKind::VectorError(err) => err.fmt(f),
             BuiltinTypeCheckErrorKind::MapError(err) => err.fmt(f),
             BuiltinTypeCheckErrorKind::TupleError(err) => err.fmt(f),
             BuiltinTypeCheckErrorKind::UdtError(err) => err.fmt(f),
+            BuiltinTypeCheckErrorKind::NotDeserializableToVec => {
+                f.write_str("the CQL type is not deserializable to a vector")
+            }
         }
     }
 }
@@ -1585,6 +1800,20 @@ impl Display for SetOrListTypeCheckErrorKind {
             }
         }
     }
+}
+
+/// Describes why type checking a vector type failed.
+#[derive(Error, Debug, Clone)]
+#[non_exhaustive]
+pub enum VectorTypeCheckErrorKind {
+    /// The CQL type is not a vector.
+    #[error(
+        "the CQL type the Rust type was attempted to be type checked against was not a vector"
+    )]
+    NotVector,
+    /// Incompatible element types.
+    #[error("the vector element types between the CQL type and the Rust type failed to type check against each other: {0}")]
+    ElementTypeCheckFailed(TypeCheckError),
 }
 
 /// Describes why type checking of a map type failed.
@@ -1833,6 +2062,9 @@ pub enum BuiltinDeserializationErrorKind {
     /// A deserialization failure specific to a CQL set or list.
     SetOrListError(SetOrListDeserializationErrorKind),
 
+    /// A deserialization failure specific to a CQL vector.
+    VectorError(VectorDeserializationErrorKind),
+
     /// A deserialization failure specific to a CQL map.
     MapError(MapDeserializationErrorKind),
 
@@ -1874,6 +2106,7 @@ impl Display for BuiltinDeserializationErrorKind {
                 "the length of read value in bytes ({got}) is not suitable for IP address; expected 4 or 16"
             ),
             BuiltinDeserializationErrorKind::SetOrListError(err) => err.fmt(f),
+            BuiltinDeserializationErrorKind::VectorError(err) => err.fmt(f),
             BuiltinDeserializationErrorKind::MapError(err) => err.fmt(f),
             BuiltinDeserializationErrorKind::TupleError(err) => err.fmt(f),
             BuiltinDeserializationErrorKind::UdtError(err) => err.fmt(f),
@@ -1912,6 +2145,22 @@ impl From<SetOrListDeserializationErrorKind> for BuiltinDeserializationErrorKind
     #[inline]
     fn from(err: SetOrListDeserializationErrorKind) -> Self {
         Self::SetOrListError(err)
+    }
+}
+
+/// Describes why deserialization of a vector type failed.
+#[derive(Error, Debug)]
+#[non_exhaustive]
+pub enum VectorDeserializationErrorKind {
+    /// One of the elements of the vector failed to deserialize.
+    #[error("failed to deserialize one of the elements: {0}")]
+    ElementDeserializationFailed(DeserializationError),
+}
+
+impl From<VectorDeserializationErrorKind> for BuiltinDeserializationErrorKind {
+    #[inline]
+    fn from(err: VectorDeserializationErrorKind) -> Self {
+        Self::VectorError(err)
     }
 }
 
--- a/scylla-cql/src/deserialize/value_tests.rs
+++ b/scylla-cql/src/deserialize/value_tests.rs
@@ -199,6 +199,308 @@ fn test_deserialize_bytes() {
 }
 
 #[test]
+fn test_deserialize_vector() {
+    // ser/de identity
+
+    // native types
+
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Ascii)),
+            dimensions: 3,
+        },
+        &vec!["ala", "ma", "kota"],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Boolean)),
+            dimensions: 2,
+        },
+        &vec![true, false],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Blob)),
+            dimensions: 2,
+        },
+        &vec![vec![1_u8, 2_u8], vec![3_u8, 4_u8]],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Counter)),
+            dimensions: 2,
+        },
+        &vec![Counter(1234), Counter(5678)],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Date)),
+            dimensions: 2,
+        },
+        &vec![CqlDate(1234), CqlDate(5678)],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Decimal)),
+            dimensions: 2,
+        },
+        &vec![
+            CqlDecimal::from_signed_be_bytes_slice_and_exponent(b"123", 42),
+            CqlDecimal::from_signed_be_bytes_slice_and_exponent(b"123", 42),
+        ],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Double)),
+            dimensions: 2,
+        },
+        &vec![0.1234, 0.5678],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Duration)),
+            dimensions: 2,
+        },
+        &vec![
+            CqlDuration {
+                months: 1,
+                days: 2,
+                nanoseconds: 3,
+            },
+            CqlDuration {
+                months: 1,
+                days: 2,
+                nanoseconds: 3,
+            },
+        ],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Float)),
+            dimensions: 2,
+        },
+        &vec![0.1234_f32, 0.5678_f32],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Int)),
+            dimensions: 2,
+        },
+        &vec![1, 2],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::BigInt)),
+            dimensions: 2,
+        },
+        &vec![1_i64, 2_i64],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Text)),
+            dimensions: 2,
+        },
+        &vec!["ala", "ma"],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Timestamp)),
+            dimensions: 2,
+        },
+        &vec![CqlTimestamp(1234), CqlTimestamp(5678)],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Inet)),
+            dimensions: 2,
+        },
+        &vec![
+            IpAddr::V4(Ipv4Addr::BROADCAST),
+            IpAddr::V6(Ipv6Addr::LOCALHOST),
+        ],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::SmallInt)),
+            dimensions: 2,
+        },
+        &vec![1_i16, 2_i16],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::TinyInt)),
+            dimensions: 2,
+        },
+        &vec![1_i8, 2_i8],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Time)),
+            dimensions: 2,
+        },
+        &vec![CqlTime(1234), CqlTime(5678)],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Timeuuid)),
+            dimensions: 2,
+        },
+        &vec![CqlTimeuuid::from_u128(123), CqlTimeuuid::from_u128(456)],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Uuid)),
+            dimensions: 2,
+        },
+        &vec![Uuid::from_u128(123), Uuid::from_u128(456)],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Varint)),
+            dimensions: 2,
+        },
+        &vec![
+            CqlVarint::from_signed_bytes_be(vec![1, 2]),
+            CqlVarint::from_signed_bytes_be(vec![3, 4]),
+        ],
+        &mut Bytes::new(),
+    );
+
+    // collection types
+
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Collection {
+                frozen: false,
+                typ: CollectionType::List(Box::new(ColumnType::Native(NativeType::Int))),
+            }),
+            dimensions: 2,
+        },
+        &vec![vec![1, 2], vec![3, 4]],
+        &mut Bytes::new(),
+    );
+
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Collection {
+                frozen: false,
+                typ: CollectionType::Set(Box::new(ColumnType::Native(NativeType::Int))),
+            }),
+            dimensions: 2,
+        },
+        &vec![vec![1, 2], vec![3, 4]],
+        &mut Bytes::new(),
+    );
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Collection {
+                frozen: false,
+                typ: CollectionType::Map(
+                    Box::new(ColumnType::Native(NativeType::Int)),
+                    Box::new(ColumnType::Native(NativeType::Int)),
+                ),
+            }),
+            dimensions: 2,
+        },
+        &vec![
+            BTreeMap::from_iter(vec![(1, 2), (3, 4)]),
+            BTreeMap::from_iter(vec![(5, 6), (7, 8)]),
+        ],
+        &mut Bytes::new(),
+    );
+
+    // tuple types
+
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Tuple(vec![
+                ColumnType::Native(NativeType::Int),
+                ColumnType::Native(NativeType::Int),
+            ])),
+            dimensions: 2,
+        },
+        &vec![(1, 2), (3, 4)],
+        &mut Bytes::new(),
+    );
+
+    // nested vector
+
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Vector {
+                typ: Box::new(ColumnType::Native(NativeType::Int)),
+                dimensions: 2,
+            }),
+            dimensions: 2,
+        },
+        &vec![vec![1, 2], vec![3, 4]],
+        &mut Bytes::new(),
+    );
+
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Collection {
+                frozen: false,
+                typ: CollectionType::List(Box::new(ColumnType::Vector {
+                    typ: Box::new(ColumnType::Native(NativeType::Int)),
+                    dimensions: 2,
+                })),
+            }),
+            dimensions: 2,
+        },
+        &vec![vec![vec![1, 2], vec![3, 4]], vec![vec![5, 6], vec![7, 8]]],
+        &mut Bytes::new(),
+    );
+
+    //empty vector
+
+    let vec: Vec<bool> = vec![];
+    assert_ser_de_identity(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(NativeType::Boolean)),
+            dimensions: 0,
+        },
+        &vec,
+        &mut Bytes::new(),
+    );
+
+    // deser_cql_value
+
+    let buf: Vec<u8> = vec![0, 0, 0, 1, 0, 0, 0, 2];
+    let decoded_vec = super::deser_cql_value(
+        &ColumnType::Vector {
+            typ: Box::new(ColumnType::Native(Int)),
+            dimensions: 2,
+        },
+        &mut buf.as_slice(),
+    )
+    .unwrap();
+    assert_eq!(
+        decoded_vec,
+        CqlValue::Vector(vec![CqlValue::Int(1), CqlValue::Int(2)])
+    );
+}
+
+#[test]
 fn test_deserialize_ascii() {
     const ASCII_TEXT: &str = "The quick brown fox jumps over the lazy dog";
 
--- a/scylla-cql/src/deserialize/value_tests.rs
+++ b/scylla-cql/src/deserialize/value_tests.rs
@@ -1553,7 +1553,7 @@ fn test_set_or_list_errors() {
             &Bytes::new(),
             Vec<i64>,
             ColumnType::Native(NativeType::Float),
-            BuiltinTypeCheckErrorKind::SetOrListError(SetOrListTypeCheckErrorKind::NotSetOrList)
+            BuiltinTypeCheckErrorKind::NotDeserializableToVec
         );
 
         // Type check of Rust set against CQL list must fail, because it would be lossy.
--- a/scylla-cql/src/deserialize/value_tests.rs
+++ b/scylla-cql/src/deserialize/value_tests.rs
@@ -12,6 +12,8 @@ use std::sync::Arc;
 
 use crate::deserialize::value::{TupleDeserializationErrorKind, TupleTypeCheckErrorKind};
 use crate::deserialize::{DeserializationError, FrameSlice, TypeCheckError};
+use crate::frame::frame_errors::CustomTypeParseError;
+use crate::frame::response::custom_type_parser::CustomTypeParser;
 use crate::frame::response::result::NativeType::*;
 use crate::frame::response::result::{CollectionType, ColumnType, NativeType, UserDefinedType};
 use crate::serialize::value::SerializeValue;
@@ -28,6 +30,126 @@ use super::{
     SetOrListDeserializationErrorKind, SetOrListTypeCheckErrorKind, UdtDeserializationErrorKind,
     UdtTypeCheckErrorKind,
 };
+
+#[test]
+fn test_custom_cassandra_type_parser() {
+    let tests = vec![
+    (  "636f6c756d6e:org.apache.cassandra.db.marshal.ListType(org.apache.cassandra.db.marshal.Int32Type)", 
+        ColumnType::Collection {
+            frozen: false,
+            typ: CollectionType::List(Box::new(ColumnType::Native(NativeType::Int))),
+        }
+    ),
+    (
+        "org.apache.cassandra.db.marshal.SetType(org.apache.cassandra.db.marshal.Int32Type)",
+        ColumnType::Collection {
+            frozen: false,
+            typ: CollectionType::Set(Box::new(ColumnType::Native(NativeType::Int))),
+        },
+    ),
+    (
+        "org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.Int32Type)",
+        ColumnType::Collection {
+            frozen: false,
+            typ: CollectionType::Map(
+                Box::new(ColumnType::Native(NativeType::Int)),
+                Box::new(ColumnType::Native(NativeType::Int)),
+            ),
+        },
+    ),
+    (
+        "org.apache.cassandra.db.marshal.DurationType",
+        ColumnType::Native(NativeType::Duration)
+    ),
+    (
+        "org.apache.cassandra.db.marshal.TupleType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.Int32Type)",
+        ColumnType::Tuple(vec![
+            ColumnType::Native(NativeType::Int),
+            ColumnType::Native(NativeType::Int)
+        ]),
+    ),
+    (
+        "org.apache.cassandra.db.marshal.UserType(keyspace1,61646472657373,737472656574:org.apache.cassandra.db.marshal.UTF8Type,63697479:org.apache.cassandra.db.marshal.UTF8Type,7a6970:org.apache.cassandra.db.marshal.Int32Type)",
+        ColumnType::UserDefinedType {
+            frozen: false,
+            definition: Arc::new(UserDefinedType {
+                name: "address".into(),
+                keyspace: "keyspace1".into(),
+                field_types: vec![
+                    ("street".into(), ColumnType::Native(NativeType::Text)),
+                    ("city".into(), ColumnType::Native(NativeType::Text)),
+                    ("zip".into(), ColumnType::Native(NativeType::Int))
+                ]
+            }),
+        }
+    ),
+    ( "org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.Int32Type, org.apache.cassandra.db.marshal.TupleType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.Int32Type))",
+        ColumnType::Collection {
+            frozen: false,
+            typ: CollectionType::Map(
+                Box::new(ColumnType::Native(NativeType::Int)),
+                Box::new(ColumnType::Tuple(vec![
+                    ColumnType::Native(NativeType::Int),
+                    ColumnType::Native(NativeType::Int)
+                ]))
+            )
+        }
+    )
+    ];
+
+    for (input, expected) in tests {
+        assert_eq!(CustomTypeParser::parse(input), Ok(expected));
+    }
+}
+
+#[test]
+fn test_custom_cassandra_type_parser_errors() {
+    assert_eq!(
+        CustomTypeParser::parse("org.apache.cassandra.db.marshal.RandomType"),
+        Err(CustomTypeParseError::UnknownSimpleCustomTypeName(
+            "RandomType".to_string()
+        ))
+    );
+
+    assert_eq!(
+        CustomTypeParser::parse("org.apache.cassandra.db.marshal.RandomType()"),
+        Err(CustomTypeParseError::UnknownComplexCustomTypeName(
+            "RandomType".to_string()
+        ))
+    );
+
+    assert_eq!(
+        CustomTypeParser::parse(
+            "org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.Int32Type)"
+        ),
+        Err(CustomTypeParseError::InvalidParameterCount {
+            actual: 1,
+            expected: 2
+        })
+    );
+
+    assert_eq!(
+        CustomTypeParser::parse("org.apache.cassandra.db.marshal.UserType(keyspace1,gg,org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.Int32Type)"),
+        Err(CustomTypeParseError::BadHexString("gg".to_string()))
+    );
+
+    assert_eq!(
+        CustomTypeParser::parse("org.apache.cassandra.db.marshal.UserType(keyspace1,ff,org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.UTF8Type,org.apache.cassandra.db.marshal.Int32Type)"),
+        Err(CustomTypeParseError::InvalidUtf8(vec![0xff]))
+    );
+
+    assert_eq!(
+        CustomTypeParser::parse(
+            "org.apache.cassandra.db.marshal.TupleType(org.apache.cassandra.db.marshal.Int32Type"
+        ),
+        Err(CustomTypeParseError::UnexpectedEndOfInput)
+    );
+
+    assert_eq!(
+        CustomTypeParser::parse("org.apache.cassandra.db.marshal.UserType(keyspace1,61646472657373,737472656574#org.apache.cassandra.db.marshal.UTF8Type)"),
+        Err(CustomTypeParseError::UnexpectedCharacter('#', ':'))
+    );
+}
 
 #[test]
 fn test_deserialize_bytes() {
--- a/scylla-cql/src/deserialize/value_tests.rs
+++ b/scylla-cql/src/deserialize/value_tests.rs
@@ -18,6 +18,7 @@ use crate::frame::response::result::NativeType::*;
 use crate::frame::response::result::{CollectionType, ColumnType, NativeType, UserDefinedType};
 use crate::serialize::value::SerializeValue;
 use crate::serialize::CellWriter;
+use crate::utils::parse::ParseErrorCause;
 use crate::value::{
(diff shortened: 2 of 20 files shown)


smoczy123:
I'm not sure this is the correct way to split this PR into commits (I'm pretty sure it isn't, as the commits won't compile), however I can't think of a proper way.

muzarski (commented):
I only reviewed the first commit (introduction of `TypeParser`)

Some general comments:
1. The logic of `TypeParser` is quite complex. I suggest adding some docstrings next to the type definitions and methods. For example, I have no idea what `TypeParser::from_hex` does. Docstrings will also help a lot in the future in case some other developer touches this piece of code.
2. It's worth adding some comments next to the non-intuitive parts of the code. Example:
```rust

        if name.is_empty() {
            if !self.is_eos() {
                return Err(CqlTypeParseError::AbstractTypeParseError());
            }
            return Ok(ColumnType::Blob);
        }
```
It's not obvious why we return `Blob` if name is empty. A link to the corresponding part of original source code would be helpful.

3. Please, add some unit tests. I saw that there is some small test of `TypeParser` in a later commit. I think we should add more tests and try to handle as many parsing cases as we can. In addition, I think that in this case, unit tests should be added in the same commit (they help during review - it's easier to reason about the complex code when there are some use case examples one can look at)
4. This implementation is based on some existing (probably Java) implementation, correct? If so, please, provide the link to the source in the commit. Ideally, the link should be placed in the comments in code as well.

smoczy123:
Whole TypeParser logic was ripped straight out of ScyllaDB's vector implementation, however, as it still in development and probably won't be merged for a while, it will be hard to link directly. IIRC there is a lot of tests there for this functionality, so thay also can be borrowed.

muzarski:
> Whole TypeParser logic was ripped straight out of ScyllaDB's vector implementation, however, as it still in development and probably won't be merged for a while, it will be hard to link directly. IIRC there is a lot of tests there for this functionality, so thay also can be borrowed.

Ok, makes sense. And let's borrow the tests in such case :)

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -880,6 +914,206 @@ where
...
+        Some(raw.and_then(|raw| {
+            T::deserialize(self.elem_typ, raw).map_err(|err| {
+                mk_deser_err::<Self>(
+                    self.coll_typ,
+                    VectorDeserializationErrorKind::ElementDeserializationFailed(err),
+                )
+            })
+        }))
+    }
+}
```
wprzytula, line 1115: :question: Why did you choose to perform the whole deserialization process in `VariableLengthVectorIterator`, while at the same time `ConstLengthVectorIterator` delegates to `VectorBytesSequenceIterator`?

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -1554,6 +1831,38 @@ impl Display for SetOrListTypeCheckErrorKind {
     }
 }
 
+/// Describes why type checking a vector type failed.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub enum VectorTypeCheckErrorKind {
+    NotVector,
```
wprzytula, line 1838: Missing docstring for `NotVector` variant.

scylla-cql/src/types/serialize/value.rs
```diff
@@ -856,6 +871,43 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
...
+    for el in iter {
+        T::serialize(el, elt, builder.make_sub_writer()).map_err(|err| {
+            mk_ser_err_named(
+                rust_name,
+                typ,
+                VectorSerializationErrorKind::ElementSerializationFailed(err),
+            )
+        })?;
+    }
+    //return Err(mk_ser_err_named(rust_name,typ, BuiltinSerializationErrorKind::VectorError(VectorSerializationErrorKind::String(format!("{:x?}", builder.buf.to_owned())))));
```
wprzytula, line 905: :recycle: Leftover commented out code

scylla-cql/src/types/serialize/writers.rs
```diff
@@ -74,6 +76,8 @@ impl<'buf> RowWriter<'buf> {
 /// in nothing being written.
 pub struct CellWriter<'buf> {
     buf: &'buf mut Vec<u8>,
+    cell_len: Option<usize>,
+    size_as_uvarint: bool,
 }
```
wprzytula, line 81: @Lorak-mmk As you know the serialization framework quite well, could you please aid in review of this commit?

scylla-cql/src/types/serialize/writers.rs
```diff
@@ -135,37 +189,111 @@ impl<'buf> CellWriter<'buf> {
...
+    // Buffer for variable length vector cell.
+    variable_length_buffer: Option<Vec<u8>>,
 }
 
 impl<'buf> CellValueBuilder<'buf> {
     #[inline]
-    fn new(buf: &'buf mut Vec<u8>) -> Self {
+    fn new(buf: &'buf mut Vec<u8>, size_as_uvar_int: bool) -> Self {
         // "Length" of a [bytes] frame can either be a non-negative i32,
         // -1 (null) or -1 (not set). Push an invalid value here. It will be
```
wprzytula, line 211: :camping: This looks like a typo in the comment: `-1` is mentioned twice.
According to [the CQL specs](https://github.com/apache/cassandra/blob/57e5cdfb59485d70f6b340987957fa4301233e8c/doc/native_protocol_v4.spec#L229), _not set_ is represented using `-2`.

Could you please fix this as a bonus, @smoczy123?

scylla-cql/src/types/serialize/writers.rs
```diff
@@ -135,37 +189,111 @@ impl<'buf> CellWriter<'buf> {
...
+
+    // If writing to a fixed length type vector, the type length.
+    cell_len: Option<usize>,
+
+    //If serializing a variable length vector cell, the size is encoded as a varint.
+    is_variable_length: bool,
+
+    // Buffer for variable length vector cell.
+    variable_length_buffer: Option<Vec<u8>>,
 }
```
wprzytula, line 205: :thought_balloon: :wrench: I believe that with the new possible fields, `CellValueBuilder` should be made an enum, with distinct variants for non-Vector, const Vector and variable Vector cases (though I'm not sure about separating the latter two cases).

wprzytula: The same goes for `CellWriter` - I think it would benefit from being made an enum.

scylla-cql/src/types/serialize/writers.rs
```diff
@@ -174,11 +302,19 @@ impl<'buf> CellValueBuilder<'buf> {
...
+            types::unsigned_vint_encode(value_len as u64, &mut len);
+            self.buf.extend_from_slice(&len);
+            self.buf.extend_from_slice(&buffer);
+        } else {
+            let value_len: i32 = (self.buf.len() - self.starting_pos - 4)
+                .try_into()
+                .map_err(|_| CellOverflowError)?;
+            self.buf[self.starting_pos..self.starting_pos + 4]
+                .copy_from_slice(&value_len.to_be_bytes());
+        }
```
wprzytula, line 317: :thought_balloon: The logic around `Cell*` got much convoluted with these alterations. Let's think how we can make it more digestable. @Lorak-mmk @muzarski

scylla/src/client/session_test.rs
```diff
@@ -3169,6 +3169,8 @@ async fn test_vector_type_metadata() {
 #[cfg(cassandra_tests)]
 #[tokio::test]
 async fn test_vector_type_unprepared() {
+    use std::vec;
+
```
wprzytula, line 3173: :question: :recycle: Is this used anywhere?

wprzytula (changes requested):
:100: What a great piece of code! Thank you for the contribution!

There are quite many comments, though.
I think that the new parser module needs much more unit tests.
Also, tests for particular errors upon serialization and deserialization of Vector are missing.

scylla-cql/src/frame/response/result.rs
```diff
@@ -864,17 +913,12 @@ fn deser_type_generic<'frame, 'result, StrT: Into<Cow<'result, str>>>(
...
     Ok(match id {
         0x0000 => {
-            // We use types::read_string instead of read_string argument here on purpose.
-            // Chances are the underlying string is `...DurationType`, in which case
-            // we don't need to allocate it at all. Only for Custom types
-            // (which we don't support anyway) do we need to allocate.
-            // OTOH, the provided `read_string` function deserializes borrowed OR owned string;
-            // here we want to always deserialize borrowed string.
-            let type_str =
-                types::read_string(buf).map_err(CqlTypeParseError::CustomTypeNameParseError)?;
```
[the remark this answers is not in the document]
muzarski: I noticed it as well. However, when you simply revert it back to `types::read_string`, the borrow-checker complains. I believe this could be possibly solved by introducing additional lifetime generic to `TypeParser`'s methods, but I'm not quite sure, as I did not investigate further.

Another observation: why does `TypeParser` hold a string as `Cow<_, _>`. I think `&str` is enough here. Is there any scenario when we need an owned string?

scylla-cql/src/types/deserialize/frame_slice.rs
```diff
@@ -155,6 +155,35 @@ impl<'frame> FrameSlice<'frame> {
...
+    /// returning a subslice that encompasses that item.
+    ///
+    /// If this slice is empty, returns `Ok(None)`.
+    /// Otherwise, if the slice does not contain enough data, it returns `Err`.
+    /// If the operation fails then the slice remains unchanged.
+    #[inline]
+    pub(super) fn read_subslice(
+        &mut self,
+        count: usize,
+    ) -> Result<Option<FrameSlice<'frame>>, LowLevelDeserializationError> {
```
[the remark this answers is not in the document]
smoczy123: All other collection types use [bytes] as it's element type, that is a 4 byte integer representing the length of an element - n, followed by n bytes representing that element. All functions used by the driver to deserialize collections have that as an assumption. Fixed type length vector assumes that we know the size of an element and skips this integer. This breaks all of the functions used to read the frame slices. The same can be said for variable length vector which (**_contradicting the CQL protocol!_**) serializes it size as an unsigned varint.

scylla-cql/src/types/deserialize/value.rs
smoczy123: At first `VariableLengthVectorIterator` used `FixedLengthBytesSequenceIterator` to deserialize, until the discovery of the unsigned varint fiasco. `ConstLengthVectorIterator` used `VectorBytesSequenceIterator` not only to match the design, but also because the design of fixed type length vector (same number of bytes repeated) looks quite generalisable and could be reused in the future for some other types. I can't possibly imagine the same for the design of the variable type length vector.

scylla-cql/src/types/deserialize/frame_slice.rs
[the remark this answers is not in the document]
wprzytula: Thanks for the explanation. I believe it deserves being put:
- in the PR's cover letter,
- in the respective commit's description,
- in a comment, or even in multiple comments, as it's such a weird thing.

scylla-cql/src/types/serialize/writers.rs
wprzytula: This typo is present in more than one place in this file.

scylla-cql/src/types/serialize/writers.rs
smoczy123: I think it would be hard to split it into enums properly, as the behavior here depends not only on if we are serializing a vector, but also if we are serializing an element of a vector

smoczy123:
As of right now all changes requested above should be implemented, the only thing left is the lifetime issue with the new_type_parser in deser_type_generic. @wprzytula @Lorak-mmk

scylla-cql/src/frame/response/new_type_parser.rs
```diff
@@ -0,0 +1,308 @@
+use crate::frame::frame_errors::CustomTypeParseError;
+use std::{borrow::Cow, char, str::from_utf8};
+
+use super::result::NativeType;
+use super::result::{ColumnType, UserDefinedType};
+use std::sync::Arc;
+
+use super::result::CollectionType;
+use crate::utils::parse::ParserState;
+use either;
```
Lorak-mmk, line 10: Those imports are quite mixed up: you have multiple groups of imports, each has both crate imports and std/external imports.
We don't have an established import style in the project, mostly due to rustfmt not stabilizing import-related options for years. However, we should still try to do something reasonable.

Please either:
- Merge them into one group, and sort by std/crate/external. rust-analyzer should make that for you with the right settings
- Have a separate group for std / crate /external.

Apart from that, you have `use std::{borrow::Cow, char, str::from_utf8};`. Please avoid this, it is a pain when rebasing. Switch to `use` statement per module, or per item.

scylla-cql/src/frame/response/new_type_parser.rs
Lorak-mmk, line 1: Commit: `frame: added a parser for Java-class type names `

You introduced both `new_type_parser.rs` and `type_parser.rs`. 
- Both contain `UDTParameters` struct, and `TypeParser` struct with `pub(crate) parse` method. 
- There is no module-level comment on those files 
- No comments on the aforementioned structs and method.
- `type_parser` seems unused (Does `make ci` pass on this commit?).
- Commit message says nothing about those modules and structs.

Please tell me how I as a reviewer am supposed to understand the purpose of those modules and structures.

Lorak-mmk (changes requested):
I left just 2 comments. I'll continue the review after the second one is addressed.

scylla-cql/src/frame/response/new_type_parser.rs
smoczy123: Both of those do the same thing, parse the Java class type name, however the new_type_parser uses the parse.rs module that was already present in the project, however due to lifetime issues I was not able to make it work correctly. So right now there are two versions of the parser, once we either find a way to fix the lifetime issues or decide to go back to type_parser.rs (that would be quite sad, we would duplicate a lot of parsing code) the other version will be deleted. cc @wprzytula

swasik:
> I have decided to not touch the bigger issues with the type_parser.rs right now, as if we decide to throw it away it would be a waste of time right now.

If we decide not to fix it right now, could you create the issue with the whole connected covered so that we don't forget about it in the future?

scylla-cql/src/deserialize/value.rs
```diff
@@ -902,6 +935,220 @@ where
...
+            } => (t, dim),
+            _ => {
+                unreachable!("Typecheck should have prevented this scenario!")
+            }
+        };
+
+        let v = ensure_not_null_frame_slice::<Self>(typ, v)?;
+
+        Ok(Self::new(typ, t, *dim as usize, t.type_size().unwrap(), v))
+    }
```
Lorak-mmk, line 1009: `t` -> `elem_type` or `element_type`
`dim` -> `dimensions`

Don't be afraid of descriptive variable names.

scylla-cql/src/deserialize/value.rs
```diff
@@ -902,6 +935,220 @@ where
...
 }
 
+pub struct ConstLengthVectorIterator<'frame, 'metadata, T> {
+    coll_typ: &'metadata ColumnType<'metadata>,
+    elem_typ: &'metadata ColumnType<'metadata>,
+    remaining: usize,
+    elem_len: usize,
+    slice: FrameSlice<'frame>,
+    phantom_data: std::marker::PhantomData<T>,
+}
```
Lorak-mmk, line 945: You implemented 2 vector iterator types: one for vectors with know length and one for those without.
User needs to use the correct one depending on the type of vector element. 

As far as I can tell there is no difference in user-facing API between the two, they just do deserialization differently internally.

Can we have just one type, which will handle different vector types internally?

Lorak-mmk (changes requested):
I'm posting first part of review because there are already many comments and I don't want to prevent you from working on them while I continue the review.

I did not yet read 3 last commits.

scylla-cql/src/frame/response/new_type_parser.rs
```diff
@@ -0,0 +1,308 @@
...
+                return Err(CustomTypeParseError::UnknownComplexCustomTypeName(
+                    self.parser.s.into(),
+                ));
+            }
+            return Ok(ColumnType::Native(NativeType::Blob));
+        }
+
+        // Type can be prefixed by a hex number, which is ignored.
+        let result = self.parser.accept(":");
+        if let Ok(parser) = result {
```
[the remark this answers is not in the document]
smoczy123: https://github.com/scylladb/scylladb/commit/85be9f55e8d7692ef4d17458d21049e565ba3680

Here's the reason for the hex number.

scylla-cql/src/frame/response/new_type_parser.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    fn do_parse(&mut self) -> Result<ColumnType<'result>, CustomTypeParseError> {
+        self.skip_blank();
+
+        let mut name = self.read_next_identifier();
+
+        // Not sure why do we return blob, however this code is ripped from ScyllaDB codebase and they do it this way.
+        if name.is_empty() {
+            if !self.parser.is_at_eof() {
+                return Err(CustomTypeParseError::UnknownComplexCustomTypeName(
```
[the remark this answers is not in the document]
smoczy123: https://github.com/scylladb/scylladb/commit/f9b83e299e0c8f6bb5b8084afd6eb6c84e5a9b24

That's the commit that introduces this behavior in ScyllaDB's type parser, however it's not too informative about the reasoning so I'm not sure if linking it in code would be useful

scylla-cql/src/frame/response/new_type_parser.rs
```diff
@@ -0,0 +1,308 @@
...
+            let result = self.parser.accept(")");
+            match result {
+                Ok(parser) => {
+                    self.parser = parser;
+                    None
+                }
+                Err(_) => Some(self.do_parse()),
+            }
+        })))
+    }
```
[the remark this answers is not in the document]
smoczy123: This behavior is consistent with ScyllaDB's implementation of the type_parser.rs, I think it's a good idea to not stray away from it.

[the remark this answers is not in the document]
smoczy123: https://github.com/scylladb/scylladb/blob/f5125ffa18871ae74c08a4337c599009d5dd48a9/db/marshal/type_parser.cc#L290

scylla-cql/src/frame/response/new_type_parser.rs
```diff
@@ -134,6 +134,33 @@ impl<'result> TypeParser<'result> {
...
+    ) -> Result<(ColumnType<'result>, u16), CustomTypeParseError> {
+        self.parser = self
+            .parser
+            .accept("(")
+            .map_err(|_| CustomTypeParseError::BadCharacter)?;
+
+        self.skip_blank_and_comma();
+        if self.parser.accept(")").is_ok() {
+            return Err(CustomTypeParseError::BadCharacter);
+        }
```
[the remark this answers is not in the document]
smoczy123: https://github.com/scylladb/scylladb/blob/f5125ffa18871ae74c08a4337c599009d5dd48a9/db/marshal/type_parser.cc#L125

I think we should return CustomTypeParseError::InvalidParameterCount, as this situation represents lack of parameters for a vector, which would be a different case than do_parse erroring out on a bad character.

scylla-cql/src/frame/response/new_type_parser.rs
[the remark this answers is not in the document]
Lorak-mmk: Ok. Please describe this in a comment of `skip_blank_and_comma`

scylla-cql/src/frame/response/new_type_parser.rs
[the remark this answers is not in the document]
Lorak-mmk: Please link it anyway.

scylla-cql/src/frame/response/new_type_parser.rs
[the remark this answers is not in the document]
Lorak-mmk: Ok, explain in the comment and also put this link there.

scylla-cql/src/frame/response/new_type_parser.rs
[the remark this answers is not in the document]
Lorak-mmk: That makes sense. Please put this explanation in the comment.

scylla-cql/src/serialize/value.rs
Lorak-mmk, line 1: Commit: `types/serialize: added vector serialization `

Do we really need that much changes in the writers? I'd prefer to not modify them unless necessary: vector serialization impl is a better place for stuff specific to that.

## Variable-element-length vectors
They seem to be the biggest problem in serialization because we can't use a single buffer (because we don't know how many bytes we need for the length). This is absurd idea on Cassandra's side btw.
We need to serialize to a separate buffer, then serialize length varint to main buffer then copy the data to main buffer.
You implemented this in `CellWriter` / `CellValueBuilder`, but it seems to be possible to do in vector impl instead?
In pseudocode: 
```rust 
    for el in iter {
        let mut element_buffer = Vec::new();
        let inner_writer = CellWriter::new(&mut element_buffer);
        T::serialize(el, elt, inner_writer).map_err(|err| {
            mk_ser_err_named(
                rust_name,
                typ,
                VectorSerializationErrorKind::ElementSerializationFailed(err),
            )
        })?;
        // Serialize length of `element_buffer`, append with `builder.append_bytes()`
        // Then `builder.append_bytes(element_buffer.as_slice())`
    }
```

## Constant-length vectors

We could do a similar trick as above - serialize using another writer, then copy the bytes.
The problem is that in the previous case we inherently had to do this, while here we don't, so it would be a big waste. Here some changes in the writers seem warranted. It is probably enough to add another method like `make_sub_writer` (`make_sub_writer_no_length`? `make_raw_sub_writer`? I'd like to avoid Vector-specific terminology), and a bool field to `CellWriter`.


Am I missing anything? Is there some reason for all the changes you introduced?
This commit is a bit hard to decipher in that regard, and lacks any explanation of those decisions :(

scylla/tests/integration/session.rs
Lorak-mmk, line 1: Commit: `test: added tests for vector type `

Those tests are not nearly enough. Vector ser / deser is the most complicated of all types, we need to test it well.
There should be unit and integration tests for vector with all possible element types.
Are nested vectors possible? If so please add such test case too
Are indirectly nested vectors possible (something like `vector<list<vector<something>>>`)?

- Unit and integration tests for nested

Lorak-mmk (changes requested):
Reviewed last 3 commtis.

scylla-cql/src/frame/response/type_parser.rs
```diff
@@ -0,0 +1,316 @@
...
+                return Err(CustomTypeParseError::BadHexString);
+            }
+        }
+        let mut bytes = Vec::new();
+        for i in 0..s.len() / 2 {
+            let byte = u8::from_str_radix(&s[i * 2..i * 2 + 2], 16)
+                .map_err(|_| CustomTypeParseError::BadHexString)?;
+            bytes.push(byte);
+        }
+        Ok(bytes)
```
Lorak-mmk, line 188: I dislike that bracket indexing, but I don't have a good idea on how to change it. @wprzytula maybe you'll think of something?

Lorak-mmk (changes requested):
I re-reviewed first 4 commits. It looks much better now. I left only a few new comments, but there are points that were raised before by @muzarski and @wprzytula and are still relevant:
https://github.com/scylladb/scylla-rust-driver/pull/1165#pullrequestreview-2547063977 raises some unadressed issues
https://github.com/scylladb/scylla-rust-driver/pull/1165#discussion_r1913418578 still relevant
https://github.com/scylladb/scylla-rust-driver/pull/1165#discussion_r1913423375 not answered

https://github.com/scylladb/scylla-rust-driver/pull/1165#pullrequestreview-2552716338 I'm not the only one who thinks more tests are needed here, and tests for errors

scylla-cql/src/deserialize/value.rs
```diff
@@ -1507,6 +1686,9 @@ pub enum BuiltinTypeCheckErrorKind {
     /// A type check failure specific to a CQL set or list.
     SetOrListError(SetOrListTypeCheckErrorKind),
 
+    /// A type check failure specific to a CQL vector.
+    VectorError(#[from] VectorTypeCheckErrorKind),
+
```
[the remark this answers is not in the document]
smoczy123: This allows us to not implement the From trait explicitly, other variants have it implemented explicitly, however if we're using `thiserror` for `VectorError`, we can use it for the From trait too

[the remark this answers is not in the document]
smoczy123: I think it would be a good idea to switch to `#[from]` for all variants here, as IIRC their From implementation behaves identically

scylla-cql/src/deserialize/value_tests.rs
```diff
@@ -29,6 +30,91 @@ use super::{
...
+        "org.apache.cassandra.db.marshal.SetType(org.apache.cassandra.db.marshal.Int32Type)";
+    assert_eq!(
+        TypeParser::parse(test_set).unwrap(),
+        ColumnType::Collection {
+            frozen: false,
+            typ: CollectionType::Set(Box::new(ColumnType::Native(NativeType::Int))),
+        }
+    );
+    let test_map =
+        "org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.Int32Type)";
```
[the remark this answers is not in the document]
smoczy123: Hmmm, I have rewritten it in that way and it seems to me that the readability of the test cases got worse, so I'm not sure if we should change it.

scylla-cql/src/deserialize/value.rs
```diff
@@ -1495,7 +1674,7 @@ macro_rules! exact_type_check {
 use exact_type_check;
 
 /// Describes why type checking some of the built-in types failed.
-#[derive(Debug, Clone)]
+#[derive(Error, Debug, Clone)]
 #[non_exhaustive]
 pub enum BuiltinTypeCheckErrorKind {
```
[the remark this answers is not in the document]
Lorak-mmk: For now let's remove derive  Error and #[from]. I don't like the "Kind" being an error itself.

scylla-cql/src/deserialize/value_tests.rs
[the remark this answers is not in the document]
Lorak-mmk: Could you post the new version somewhere (e.g. create a gist)? I'd like to see it and maybe suggest how to improve it.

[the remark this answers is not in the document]
smoczy123: https://gist.github.com/smoczy123/d01ffe65289b1ada03c3f9f409cf6761

scylla-cql/src/frame/response/type_parser.rs
Lorak-mmk: I don't really see an advantage. Both versions contain a panic (that is verified to be impossible to trigger).

wprzytula: The advantage is in readability. It's much easier to parse (at least for me) such FP code, especially that it uses clear wording, than to parse those slice indexing arithmetics.

Lorak-mmk: Ok

Lorak-mmk (changes requested):
I left 2 comments, first already with ready solution.
Regarding the tests that you asked me about:
I thought we had some integration tests like this already, but I can'f find them.
What you can do is have a generic helper function , with T: SerializeValue + DeserializeValue + Eq.
It would accept string with type name S and value X of type T.
This function would create a table, with non-pk column of type S.
Prepare 2 statements: one to insert a value, one to select a value.
Execute them both, and verify that the resulting value is equal to the original.
At the end, drop the table.

This is quite similar to assert_ser_de_identity, but for integration tests.

scylla-cql/src/value.rs
ewienik, line 1: Deserialization of CqlValue::Vector is unsupported and because of this deserialization of Rows with Vectors is unsupported. Could `deser_cql_value` function be changed to support Vector values?

ewienik (changes requested):

scylla-cql/src/deserialize/value.rs
```diff
@@ -902,6 +924,163 @@ where
...
+                    mk_deser_err::<Self>(
+                        self.collection_type,
+                        BuiltinDeserializationErrorKind::RawCqlBytesReadError(
+                            LowLevelDeserializationError::IoError(Arc::new(err)),
+                        ),
+                    )
+                });
+                let raw = size.and_then(|size| {
+                    self.slice
+                        .read_n_bytes(size.try_into().unwrap())
```
wprzytula, line 1057: :wrench: `unwrap()` without comment nor justification.

scylla-cql/src/deserialize/value.rs
```diff
@@ -902,6 +924,163 @@ where
...
+                }))
+            }
+        }
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.remaining, Some(self.remaining))
+    }
+}
```
wprzytula, line 1082: :recycle: I suggest clearer decoupling of the constant length and variable length cases:

```rust
impl<'frame, 'metadata, T> VectorIterator<'frame, 'metadata, T>
where
    T: DeserializeValue<'frame, 'metadata>,
{
    fn next_constant_length_elem(
        &mut self,
        element_length: usize,
    ) -> Option<<Self as Iterator>::Item> {
        self.remaining = self.remaining.checked_sub(1)?;
        let raw = self.slice.read_n_bytes(element_length).map_err(|err| {
            mk_deser_err::<Self>(
                self.collection_type,
                BuiltinDeserializationErrorKind::RawCqlBytesReadError(err),
            )
        });
        Some(raw.and_then(|raw| {
            T::deserialize(self.element_type, raw).map_err(|err| {
                mk_deser_err::<Self>(
                    self.collection_type,
                    VectorDeserializationErrorKind::ElementDeserializationFailed(err),
                )
            })
        }))
    }

    fn next_variable_length_elem(&mut self) -> Option<<Self as Iterator>::Item> {
        self.remaining = self.remaining.checked_sub(1)?;
        let size = types::unsigned_vint_decode(self.slice.as_slice_mut()).map_err(|err| {
            mk_deser_err::<Self>(
                self.collection_type,
                BuiltinDeserializationErrorKind::RawCqlBytesReadError(
                    LowLevelDeserializationError::IoError(Arc::new(err)),
                ),
            )
        });
        let raw = size.and_then(|size| {
            self.slice
                .read_n_bytes(size.try_into().unwrap())
                .map_err(|err| {
                    mk_deser_err::<Self>(
                        self.collection_type,
                        BuiltinDeserializationErrorKind::RawCqlBytesReadError(err),
                    )
                })
        });

        Some(raw.and_then(|raw| {
            T::deserialize(self.element_type, raw).map_err(|err| {
                mk_deser_err::<Self>(
                    self.element_type,
                    VectorDeserializationErrorKind::ElementDeserializationFailed(err),
                )
            })
        }))
    }
}

impl<'frame, 'metadata, T> Iterator for VectorIterator<'frame, 'metadata, T>
where
    T: DeserializeValue<'frame, 'metadata>,
{
    type Item = Result<T, DeserializationError>;

    fn next(&mut self) -> Option<Self::Item> {
        match self.element_length {
            Some(element_length) => self.next_constant_length_elem(element_length),
            None => self.next_variable_length_elem(),
        }
    }

    #[inline]
    fn size_hint(&self) -> (usize, Option<usize>) {
        (self.remaining, Some(self.remaining))
    }
}
```

scylla-cql/src/serialize/writers.rs
```diff
@@ -139,20 +156,29 @@ pub struct CellValueBuilder<'buf> {
...
+    // Should we write the size of the value?
+    write_size: bool,
 }
 
 impl<'buf> CellValueBuilder<'buf> {
     #[inline]
-    fn new(buf: &'buf mut Vec<u8>) -> Self {
+    fn new(buf: &'buf mut Vec<u8>, write_size: bool) -> Self {
         // "Length" of a [bytes] frame can either be a non-negative i32,
         // -1 (null) or -1 (not set). Push an invalid value here. It will be
```
wprzytula, line 168: :camping: Let's fix the old typo. -1 is either null or not set, not both of them.

wprzytula: -2 corresponds to not set, according to the CQL specs.

scylla-cql/src/serialize/value.rs
```diff
@@ -471,13 +471,37 @@ impl<T: SerializeValue> SerializeValue for Vec<T> {
...
+                dimensions,
+                typ,
+                writer,
+            ),
+
+            _ => Err(mk_typck_err_named(
+                std::any::type_name::<Self>(),
+                typ,
+                SetOrListTypeCheckErrorKind::NotSetOrList,
+            )),
```
wprzytula, line 503: :thinking: The name is no longer correct. It should be `NotSetOrListOrVector`, but we cannot change the name - this would break the API. Perhaps we should introduce this new variant, leaving the old one unused? cc @Lorak-mmk

scylla-cql/src/serialize/value.rs
```diff
@@ -855,6 +879,66 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
         .map_err(|_| mk_ser_err_named(rust_name, typ, BuiltinSerializationErrorKind::SizeOverflow))
 }
 
+fn serialize_vector<'t, 'b, T: SerializeValue + 't>(
+    rust_name: &'static str,
+    len: usize,
+    iter: impl Iterator<Item = &'t T>,
+    element_type: &ColumnType,
+    dimensions: &u16,
```
wprzytula, line 887: :pick: Let's accept `u16` by copy, not by reference.

scylla-cql/src/serialize/value.rs
```diff
@@ -855,6 +879,66 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
...
+                    )
+                })?;
+                let mut element_length_buffer = Vec::new();
+                unsigned_vint_encode(
+                    element_buffer.len().try_into().unwrap(),
+                    &mut element_length_buffer,
+                );
+                builder.append_bytes(element_length_buffer.as_slice());
+                builder.append_bytes(element_buffer.as_slice());
+            }
```
wprzytula, line 934: :recycle: Similarly to deserialization, let's extract separate functions for the constant length and variable length element cases.

scylla-cql/src/serialize/value.rs
```diff
@@ -855,6 +879,66 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
...
+                        typ,
+                        VectorSerializationErrorKind::ElementSerializationFailed(err),
+                    )
+                })?;
+                let mut element_length_buffer = Vec::new();
+                unsigned_vint_encode(
+                    element_buffer.len().try_into().unwrap(),
+                    &mut element_length_buffer,
+                );
+                builder.append_bytes(element_length_buffer.as_slice());
```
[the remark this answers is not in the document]
wprzytula: Both solutions you proposed are not going to be API-breaking in any way, so let's not block this PR. Opening an issue is fine with me.

wprzytula (changes requested):

scylla-cql/src/deserialize/value_tests.rs
```diff
@@ -29,6 +31,119 @@ use super::{
...
 
+#[test]
+fn test_custom_cassandra_type_parser() {
+    let tests = vec![
+    (  "636f6c756d6e:org.apache.cassandra.db.marshal.ListType(org.apache.cassandra.db.marshal.Int32Type)", 
+        ColumnType::Collection {
+            frozen: false,
+            typ: CollectionType::List(Box::new(ColumnType::Native(NativeType::Int))),
+        }
+    ),
```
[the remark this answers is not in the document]
smoczy123: I've linked the ScyllaDB commit in the type parser, the hex has something to do with hashing the collection types.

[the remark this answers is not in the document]
smoczy123: And the rest of the format is just Java classes, I'm not sure if it requires further explanation

scylla-cql/src/deserialize/value_tests.rs
```diff
@@ -29,6 +31,119 @@ use super::{
...
+                    ColumnType::Native(NativeType::Int)
+                ]))
+            )
+        }
+    )
+    ];
+
+    for (input, expected) in tests {
+        assert_eq!(CustomTypeParser::parse(input).unwrap(), expected);
+    }
```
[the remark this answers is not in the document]
smoczy123: We would have to derive equality for the CustomTypeParseError, which is not needed anywhere else

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -430,6 +430,26 @@ pub enum ColumnSpecParseErrorKind {
...
+    #[error("Malformed complex custom type name: {0}")]
+    UnknownComplexCustomTypeName(String),
+    #[error("Bad character encountered")]
+    BadCharacter,
+    #[error("Unexpected end of input")]
+    UnexpectedEndOfInput,
+    #[error("Bad hex string")]
+    BadHexString,
+    #[error("Bytes do not represent a valid UTF-8 string")]
+    InvalidUtf8,
```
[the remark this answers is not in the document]
smoczy123: Here it would be quite difficult, as the String:from_utf8 function consumes the vector, so putting it in the error msg would require cloning it.

scylla-cql/src/frame/frame_errors.rs
```diff
@@ -430,6 +430,26 @@ pub enum ColumnSpecParseErrorKind {
...
+    #[error("Wrong number of parameters")]
+    InvalidParameterCount,
```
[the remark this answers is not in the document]
smoczy123: We can include the expected number, however we don't know the actual number of parameters as collect_array does not inform us about it

scylla-cql/src/utils/parse.rs
```diff
@@ -38,26 +40,27 @@ impl Display for ParseErrorCause {
...
 ///
 /// Basically, a wrapper over &str with nice methods that help with parsing.
 #[derive(Clone, Copy)]
 #[must_use]
-pub(crate) struct ParserState<'s> {
-    s: &'s str,
+#[non_exhaustive]
+pub struct ParserState<'s> {
+    pub(crate) s: &'s str,
 }
```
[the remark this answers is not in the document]
Lorak-mmk: Is it a problem to have it though? In general (so not talking about this struct specifically) it may be easy to miss in the review that a field we publish or remove is the last non-pub field, and thus  non_exhaustive becomes required.
Having non_exhausitve now would protect us from this.

scylla-cql/src/deserialize/value_tests.rs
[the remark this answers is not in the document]
Lorak-mmk: You can use `#[cfg_attr(test, derive(PartialEq, Eq))]` to only expose it to tests.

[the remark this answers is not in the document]
Lorak-mmk: What is the argument against implementing it always?

scylla-cql/src/serialize/value.rs
Lorak-mmk: I don't think the name is important enough to justify such change

scylla-cql/src/deserialize/value_tests.rs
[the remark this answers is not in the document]
smoczy123: I don't think there's one tbh

smoczy123:
I have added a lot of integration tests, which uncovered more inconsistencies in Cassandra type sizes. It seems like both Time and Date are treated as variable length even though they are always 8 and 4 bytes, respectively. Another quirk is that vectors are sometimes fixed size, that is when their element type is fixed size, they are too and their size is element_size * dimensions.

scylla-cql/src/utils/parse.rs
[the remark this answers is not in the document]
wprzytula: > Is it a problem to have it though? In general (so not talking about this struct specifically) it may be easy to miss in the review that a field we publish or remove is the last non-pub field, and thus non_exhaustive becomes required. Having non_exhausitve now would protect us from this.

That's valid reasoning. However, I'd like to prevent situation when an arbitrary subset of `pub` structs with some non-`pub` fields are marked with `#[non_exhaustive]` and the other subset are not. In general, inconsistency brings confusion. So either let's not leave here `#[non_exhaustive]` or let's decide that from now on we're adding `#[non_exhaustive]` eagerly, i.e., even for `pub` structs with some non-`pub` fields. In the latter case, we should probably add `#[non_exhaustive]` to existing such structs, too.

scylla-cql/src/frame/frame_errors.rs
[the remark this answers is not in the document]
wprzytula: See [`FromUtf8Error::into_bytes()`](https://doc.rust-lang.org/std/string/struct.FromUtf8Error.html#method.into_bytes). No cloning is needed.

scylla-cql/src/frame/frame_errors.rs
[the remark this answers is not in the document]
wprzytula: This is a limitation imposed by the current implementation using `collect_array`, not a limitation that is inherent to this problem. Please try to find possible workarounds.

scylla-cql/src/serialize/value.rs
wprzytula: OK, then let's update the docstring of that variant by mentioning Vector there.

scylla-cql/src/frame/frame_errors.rs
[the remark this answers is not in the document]
smoczy123: If itertools exposed their ArrayBuilder struct this would be easy, but unfortunately they don't. What we could do here is go back to the previous syntax, where we get the required argument manually one-by-one, split this error into two cases, `TooFewParameters` and `TooManyParameters` and add the actual number of parameters to `TooFewParameters`. In the case of `TooManyParameters` getting the actual number of parameters would require iterating through the whole iterator, I'm not a fan of this.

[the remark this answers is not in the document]
wprzytula: OK, this convinces me to give up.

scylla-cql/src/utils/parse.rs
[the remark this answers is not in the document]
Lorak-mmk: I think the latter solution (have too many non_exhaustive rather than too little) is better.

scylla/tests/integration/session.rs
```diff
@@ -3256,16 +3279,253 @@ async fn test_vector_type_prepared() {
...
+            .replace(",", "C")
+    );
+    let query = format!(
+        "CREATE TABLE {}.{} (a int PRIMARY KEY, b vector<{}, {}>)",
+        keyspace,
+        table_name,
+        type_name,
+        values.len()
+    );
+    session.ddl(query).await.unwrap();
```
Lorak-mmk, line 3337: It would be nice to drop the table after `assert_eq` passes successfully,

scylla/tests/integration/session.rs
```diff
@@ -3256,16 +3279,253 @@ async fn test_vector_type_prepared() {
...
+    let query_result = session.execute_unpaged(&prepred_select, &[]).await.unwrap();
+
+    let result = query_result.into_rows_result().unwrap();
+
+    let rows: Vec<(i32, Vec<T>)> = result
+        .rows::<(i32, Vec<T>)>()
+        .unwrap()
+        .map(|r| r.unwrap())
+        .collect();
+
```
Lorak-mmk, line 3364: I think there is a method like `single_row` which you can use instead of `rows` here to simplify code.

Lorak-mmk (approved):
I left 3 comments (one of them non-actionable) but I think those are not things that should block merging. If you can fix them, great. If not, I'm okay with merging as-is.

Thanks for writing the tests!

scylla-cql/src/serialize/value.rs
```diff
@@ -854,6 +879,99 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
...
+                serialize_next_variable_length_elem(
+                    rust_name,
+                    element_type,
+                    typ,
+                    &mut builder,
+                    element,
+                )?;
+            }
+        }
+    }
```
wprzytula, line 968:
```suggestion
    match element_type.type_size() {
        Some(_) => {
            for element in iter {
                serialize_next_constant_length_elem(
                    rust_name,
                    element_type,
                    typ,
                    &mut builder,
                    element,
                )?;
            }
        None => {
            for element in iter {
                serialize_next_variable_length_elem(
                    rust_name,
                    element_type,
                    typ,
                    &mut builder,
                    element,
                )?;
            }
        }
    }
```

wprzytula: :pick: We choose the `match` branch only once.

scylla/tests/integration/session.rs
```diff
@@ -3225,16 +3232,32 @@ async fn test_vector_type_unprepared() {
 
     session
         .query_unpaged(
-            format!(
-                "INSERT INTO {}.t (a, b, c) VALUES (1, [1, 2, 3, 4], ['foo', 'bar'])",
-                ks
-            ),
-            &[],
+            format!("INSERT INTO {}.t (a, b, c) VALUES (?, ?, ?)", ks),
+            &(1, vec![1, 2, 3, 4], vec!["foo", "bar"]),
```
wprzytula, line 3236: :wrench: I'd like to see an example with slices serialized as Vector, too.

wprzytula:
```rust
&(1, &[1, 2, 3, 4], &["foo", "bar"]),
```

scylla/tests/integration/session.rs
```diff
@@ -3225,16 +3232,32 @@ async fn test_vector_type_unprepared() {
...
         )
         .await
         .unwrap();
 
-    // TODO: Implement and test SELECT statements and bind values (`?`)
+    let query_result = session
+        .query_unpaged(format!("SELECT * FROM {}.t", ks), &[])
+        .await
+        .unwrap();
```
wprzytula, line 3244: :question: Is there any reason why we're using `SELECT *` queries instead of specifying column names?

scylla/tests/integration/session.rs
wprzytula, line 3337:
```suggestion
    let statement = format!(
        "CREATE TABLE {}.{} (a int PRIMARY KEY, b vector<{}, {}>)",
        keyspace,
        table_name,
        type_name,
        values.len()
    );
    session.ddl(statement).await.unwrap();
```

scylla/tests/integration/session.rs
```diff
@@ -3256,16 +3279,253 @@ async fn test_vector_type_prepared() {
...
     session
-        .execute_unpaged(&prepared_statement, &(2,))
+        .execute_unpaged(&prepared_insert, &(1, &values))
         .await
         .unwrap();
 
-    // TODO: Implement and test SELECT statements and bind values (`?`)
+    let query_result = session.execute_unpaged(&prepred_select, &[]).await.unwrap();
```
wprzytula, line 3355: :pick: prepred -> prepared

scylla/tests/integration/session.rs
```diff
@@ -3256,16 +3279,253 @@ async fn test_vector_type_prepared() {
...
+        .map(|r| r.unwrap())
+        .collect();
+
+    assert_eq!(rows[0], (1, values));
+}
+
+#[cfg_attr(not(cassandra_tests), ignore)]
+#[tokio::test]
+async fn test_vector_type_all_types() {
+    setup_tracing();
```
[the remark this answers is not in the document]
wprzytula: Let's add a TODO to make it harder to forget.

wprzytula (changes requested):

wprzytula:
:question: One question: could you explain this change? https://github.com/scylladb/scylla-rust-driver/compare/1eb80183b32bb74c440253534fd3ab1f57bad387..181c9f62624f8cb2974b24602b568d357e19735a

wprzytula (approved):
LGTM! :tada:

Thanks for patient polishing this PR up to the present point. :rocket:

Lorak-mmk (dismissed):

scylla-cql/src/frame/frame_errors.rs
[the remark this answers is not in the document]
piodul: > In the case of TooManyParameters getting the actual number of parameters would require iterating through the whole iterator, I'm not a fan of this.

Why is this bad? I would expect this error to be very rare, Scylla in general should send correct strings, so it is not a big deal that we spent some time on collecting diagnostic information. That said, because this is a rare argument one would say that this doesn't matter too much, so I won't insist on this.

By the way, a succint way to do it would be `.collect::<Vec<_>>().try_into().map_err(|v| CustomTypeParseError::InvalidParameterCount(v.len()))` (here, `into()` is supposed to convert into an array).

https://doc.rust-lang.org/std/primitive.array.html#impl-TryFrom%3CVec%3CT%3E%3E-for-Box%3C%5BT;+N%5D%3E

scylla-cql/src/value.rs
```diff
@@ -1180,6 +1181,11 @@ impl std::fmt::Display for CqlValue {
                     .fmt(f)?;
                 f.write_str("}")?;
             }
+            CqlValue::Vector(v) => {
+                f.write_str("[")?;
+                v.iter().format(",").fmt(f)?;
+                f.write_str("]")?;
+            }
```
piodul, line 1188: Nit: you could extend the existing List case instead of introducing a separate one, they have exactly the same syntax:

```rust
CqlValue::List(v) | CqlValue::Vector(v) => {
    // ...
}
```

scylla-cql/src/serialize/value.rs
```diff
@@ -855,6 +903,101 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
...
+            rust_name,
+            typ,
+            VectorSerializationErrorKind::ElementSerializationFailed(err),
+        )
+    })?;
+    let mut element_length_buffer = Vec::new();
+    unsigned_vint_encode(
+        element_buffer.len().try_into().unwrap(),
+        &mut element_length_buffer,
+    );
```
piodul, line 948: It seems a bit wasteful to allocate a Vec just for this. I suppose that vectors with variable length elements won't be too common, so I suppose this is not a huge issue.

piodul (dismissed):
LGTM, I left some nitpicks but I don't think they are critical (unless maintainers decide otherwise).

scylla-cql/src/serialize/value.rs
wprzytula: > It seems a bit wasteful to allocate a Vec just for this. I suppose that vectors with variable length elements won't be too common, so I suppose this is not a huge issue.

Won't they? There are quite a lot of CQL types that are considered variable-length due to god-know-why Cassandra choices...

Lorak-mmk: I raised it here before: https://github.com/scylladb/scylla-rust-driver/pull/1165#discussion_r2055690404
As you said, this is most likely not important, and if it is then we can fix it later.

scylla-cql/src/utils/parse.rs
```diff
@@ -38,26 +40,26 @@ impl Display for ParseErrorCause {
...
-pub(crate) struct ParserState<'s> {
-    s: &'s str,
+pub struct ParserState<'s> {
+    pub(crate) s: &'s str,
 }
 
 impl<'s> ParserState<'s> {
     /// Creates a new parser from given input string.
-    pub(crate) fn new(s: &'s str) -> Self {
+    pub fn new(s: &'s str) -> Self {
```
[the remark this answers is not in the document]
Lorak-mmk: It's fine with me, let's move map_string_to_cql_type and un-pub Parser-related stuff.

scylla-cql/src/serialize/value.rs
piodul: > Won't they?

For storing embeddings, I suppose that the default choice would be to use a vector of `float`s or `double`s.

scylla-cql/src/serialize/value.rs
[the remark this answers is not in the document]
piodul: I think this is fine to fix later. The most interesting use case for vector columns is using them in vector search indexes, and for those vectors of floats or doubles makes the most sense - and those are fixed size.

scylla-cql/src/utils/parse.rs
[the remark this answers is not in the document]
wprzytula: You made me realise that the code separation regarding schema metadata between `scylla` and `scylla-cql` crate is indeed weird.
1. `ColumnType` is defined in `scylla-cql`, yet `PreColumnType` is located in `scylla`...
2. CQL type name parser is now moved to `scylla-cql` (which makes sense, as it's used for Custom types deserialization, including Vector). But it's needed in `scylla` in `metadata.rs` (`map_string_to_cql_type()`).
3. If we move `map_string_to_cql_type()` to `scylla-cql`, then we need to move `PreColumnType` there as well. And related error types, too... (`InvalidCqlType`, `MissingUserDefinedType`)...

This is a horrible mess.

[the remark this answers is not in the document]
Lorak-mmk: Ah, I didn't notice that map_string_to_cql_type returns PreColumnType....

This "mess" imo is not caused that much by separation into scylla and scylla-cql, but rather by:
- Us (consciously) using `ColumnType` for both metadata, and prepared statement data. I still think this was a good decision.
- Cassandra creating this abomination that is Vector. Previously separation was clear: metadata needs to parse textual format (because they are located in schema tables), and protocol impl needs to parse binary format for prepared statements. Now the binary format contains the textual format, so it becomes a mess.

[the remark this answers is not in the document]
Lorak-mmk: > ColumnType is defined in scylla-cql, yet PreColumnType is located in scylla...

Because PreColumnType is an intermediate step in parsing metadata (= parsing schema tables), it has no relevance to binary protocol which resides in scylla-cql.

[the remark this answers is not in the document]
Lorak-mmk: I don't like moving PreColumnType. It is only ever useful for metadata and has nothing to do with protocol that scylla-cql implements, it makes 0 sense to me to put it there.
It is also extremely weird to move it to scylla-cql and making it useless (because of lack of into_cql_type). It makes no sense from the perspective of scylla-cql API.

Moving and publishing the parser may not be the most elegant, but I think it is better than moving PreColumnType.
I see such parser as just a low-level utility and I don't think there is anything inherently wrong with crates exposing such things. I also don't envision changes in Parser that this move would prevent.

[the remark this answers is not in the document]
wprzytula: It's just all bad. I guess that unless you agree to have the proxy depend on the `scylla` crate, we have to deal with this awful thing that `scylla-cql` is.

BTW: would merging `scylla-cql` with `scylla`:
1.  prevent usage of proxy in `scylla`'s unit tests, am I correct? But at the same time not in its integration tests?
2. block `scylla-udf` further development, because WASM does not support tokio, which `scylla` (but not `scylla-cql`) depends upon?

smoczy123:
> ❓ One question: could you explain this change? https://github.com/scylladb/scylla-rust-driver/compare/1eb80183b32bb74c440253534fd3ab1f57bad387..181c9f62624f8cb2974b24602b568d357e19735a

I have copied the test case and changed the expected error type, but forgot to change the test string, so the tests didn't pass, here I fix this

scylla-cql/src/utils/parse.rs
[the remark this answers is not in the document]
Lorak-mmk: > we have to deal with this awful thing that scylla-cql is.

And I am still convinced that the solution is to make it less bad by spending time on improving its API.
Many other drivers also have a separate package for protocol impl, and this separation makes sense to me. Even more than makes sense: I think such separation is an advantage, if done well.

> BTW: would merging scylla-cql with scylla:
Point 1 is correct, but we would have to re-export in `scylla` crate the stuff that proxy depends on.
Point 2: I see that scylla-cql depends on Tokio because of the following one usage:
```
pub async fn read_response_frame(
    reader: &mut (impl AsyncRead + Unpin),
```
I am pretty sure we could remove this dependency from scylla-cql in 2.0, by either introducing our own wrapper trait, or rewriting this piece of code in more sans-io way.

One other use case is when you really do want to access lower-level details like when I did when writing a reproducer for some Scylla issue: https://github.com/Lorak-mmk/udt_schema_change_reproducer
I would also happily use it when investigating https://github.com/scylladb/scylladb/security/advisories/GHSA-25pq-rrqm-6fmr instead of writing [CQL impl myself](https://gist.github.com/Lorak-mmk/c1702bf22c26020508acd837105c19af) if I knew about it at the time.

scylla-cql/src/deserialize/value_tests.rs
```diff
@@ -29,6 +31,123 @@ use super::{
...
+                keyspace: "keyspace1".into(),
+                field_types: vec![
+                    ("street".into(), ColumnType::Native(NativeType::Text)),
+                    ("city".into(), ColumnType::Native(NativeType::Text)),
+                    ("zip".into(), ColumnType::Native(NativeType::Int))
+                ]
+            }),
+        }
+    ),
+    ( "org.apache.cassandra.db.marshal.MapType(org.apache.cassandra.db.marshal.Int32Type, org.apache.cassandra.db.marshal.TupleType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.Int32Type))",
```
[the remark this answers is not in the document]
Lorak-mmk: Eh, I hate this bug in rustfmt. I think that it is fixed at least for long strings in 2024 edition, so we can leave it for now - it should get fixed when we bump the edition.

smoczy123:
@piodul (I'm not sure why, but I can't reply to your comment) The issue with invalid parameter count is that any elegant way of gathering that data that I could think of (and the one you propose here) would require allocating each time we parse a vector. The solution without allocations (getting the required arguments manually and iterating through the rest if needed) is quite ugly.

Lorak-mmk:
> @piodul (I'm not sure why, but I can't reply to your comment)

This is a weird behavior in Github UI. If, when doing a review, you respond to a comment thread belonging to some other review, then this new comment will show up in 2 places:
- As a comment that you can't respond to, in this new review
- In the original thread, where you can respond. In this case it is here: https://github.com/scylladb/scylla-rust-driver/pull/1165#discussion_r2077396085

scylla-cql/src/frame/frame_errors.rs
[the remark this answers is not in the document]
piodul: Replying to the comment here: https://github.com/scylladb/scylla-rust-driver/pull/1165#issuecomment-2858598954

> @piodul (I'm not sure why, but I can't reply to your comment) The issue with invalid parameter count is that any elegant way of gathering that data that I could think of (and the one you propose here) would require allocating each time we parse a vector. The solution without allocations (getting the required arguments manually and iterating through the rest if needed) is quite ugly.

Do we really create `CustomTypeParseError::InvalidParameterCount` every time we parse a vector, even if the vector type name is specified correctly?

I see that we indeed have `.ok_or(CustomTypeParseError::InvalidParameterCount)`, but we will have to change those occurrences so that `InvalidParameterCount` is only created when we have an invalid type string (otherwise you don't know what to put as the column count).

piodul:
> > @piodul (I'm not sure why, but I can't reply to your comment)
> 
> This is a weird behavior in Github UI. If, when doing a review, you respond to a comment thread belonging to some other review, then this new comment will show up in 2 places:
> 
> * As a comment that you can't respond to, in this new review
> * In the original thread, where you can respond. In this case it is here: [CQL Vector support #1165 (comment)](https://github.com/scylladb/scylla-rust-driver/pull/1165#discussion_r2077396085)

Replied here: https://github.com/scylladb/scylla-rust-driver/pull/1165#discussion_r2077698806

scylla-cql/src/frame/frame_errors.rs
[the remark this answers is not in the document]
smoczy123: We don't create `CustomTypeParseError::InvalidParameterCount`, but `.collect::<Vec<_>>()` allocates a Vec, even though it isn't really needed

scylla-cql/src/serialize/value.rs
Lorak-mmk: I'll mark that as resolved since we won't be addressing this now afaiu

scylla-cql/src/frame/frame_errors.rs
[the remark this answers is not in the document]
piodul: Yes, but my question is: how frequently this can occur? I claim (but correct me if I'm wrong) that this is a very rare error - basically, Scylla would have to send an improperly formatted type as a string or there would be some bug in the type parsing code in the driver. Both of those cases are bugs which would make it difficult to use Scylla with the driver, and any additional amount of context in the error can be helpful with debugging this later.

Unless this error represents a failure which is expected to happen sometimes, then it is completely fine to sacrifice some of the performance for the sake of the debugging experience. Because of this, it is also fine to do a similar sacrifice for the sake of code clarity. Of course, there is no need to make it slow on purpose, if you can implement it in a clear, understandable way without allocating the vector then it is fine, too.

scylla-cql/src/utils/parse.rs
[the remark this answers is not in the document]
piodul: I briefly discussed the matter with @Lorak-mmk and @wprzytula on Slack and we all agreed that there is no need to fix the issue in this PR, the parser can be kept exposed as public in scylla-cql.

wprzytula:
@smoczy123 Are you planning to address @piodul's nitpicks or shall we merge this as-is?

smoczy123:
They should be addressed now @wprzytula

scylla-cql/src/frame/response/custom_type_parser.rs
```diff
@@ -0,0 +1,357 @@
...
+            "ListType" => {
+                let [element_type_result] = self
+                    .get_type_parameters()?
+                    .collect::<Vec<_>>()
+                    .try_into()
+                    .map_err(
+                        |v: Vec<Result<ColumnType<'result>, CustomTypeParseError>>| {
+                            CustomTypeParseError::InvalidParameterCount(v.len(), 1)
+                        },
+                    )?;
```
wprzytula, line 252: :wrench: I don't think this is what @piodul had in mind. Now we `collect()` a `Vec` **unconditionally**, paying for allocation even when no error is involved.
How I imagined this change was to, upon occurrence of invalid parameter count (detected by `None` returned from `collect_array()`), re-create the iterator and only then  `collect::<Vec<_>>()`.

wprzytula (changes requested):

scylla-cql/src/frame/response/custom_type_parser.rs
smoczy123: Unfortunately, recreating that iterator seems quite difficult, as we mutate the parser state in the .next()

wprzytula: Can we clone (for backup) the parser state beforehand, and use that copy to recreate the iterator?

smoczy123: In that case I'd say we leave the error as is, without the parameter counts.

piodul: Not, I misunderstood - that's what I told to do but yeah, not what I intended. My bad.

wprzytula: `CustomTypeParser` stores only `ParserState`, which is `Copy` and very cheap to copy. So it seems trivial to do so.

smoczy123: > Can we clone (for backup) the parser state beforehand, and use that copy to recreate the iterator?

Hmmm, this could work

Lorak-mmk (changes requested):
`InvalidParameterCount(usize, usize),` - can we make that a struct variant (with actual and expected fields) instead of tuple variant?

```rust
                    .collect::<Vec<_>>()
                    .try_into()
                    .map_err(
                        |v: Vec<Result<ColumnType<'result>, CustomTypeParseError>>| {
                            CustomTypeParseError::InvalidParameterCount(v.len(), 1)
                        },
                    )?;
```

I think the intention was to only allocate in the case of an error. Now you made this function always allocate. One way to allocate only in case of error is to do something like this:
```rust
    fn get_complex_abstract_type(
        &mut self,
        mut name: &'result str,
    ) -> Result<ColumnType<'result>, CustomTypeParseError> {
        name = name
            .strip_prefix("org.apache.cassandra.db.marshal.")
            .unwrap_or(name);

        // Calculates the real number of parameters.
        // Can be called only after verifying that get_type_parameters() returned Ok,
        // because it panic in case of error.
        // It is declared in this admittedly weird way to make it FnOnce - calling it
        // more than once would obviously be a bug.
        let calc_params_count = {
            let self_clone = CustomTypeParser {
                // Clone here is cheap since parser is just one &str.
                parser: self.parser.clone(),
            };
            move || {
                // If we make `self_clone` mut and used it here,
                // the whole closure would be FnMut.
                let mut parser = self_clone;
                parser.get_type_parameters().unwrap().count()
            }
        };

        match name {
            "ListType" => {
                let [element_type_result] = self
                    .get_type_parameters()?
                    .collect_array::<1>()
                    .ok_or_else(move || {
                        CustomTypeParseError::InvalidParameterCount(calc_params_count(), 1)
                    })?;
                let element_type = element_type_result?;
                Ok(ColumnType::Collection {
                    frozen: false,
                    typ: CollectionType::List(Box::new(element_type)),
                })
            }
            "SetType" => {
                let [element_type_result] = self
                    .get_type_parameters()?
                    .collect_array::<1>()
                    .ok_or_else(move || {
                        CustomTypeParseError::InvalidParameterCount(calc_params_count(), 1)
                    })?;
                let element_type = element_type_result?;
                Ok(ColumnType::Collection {
                    frozen: false,
                    typ: CollectionType::Set(Box::new(element_type)),
                })
            }
            "MapType" => {
                let [key_type_result, value_type_result] = self
                    .get_type_parameters()?
                    .collect_array::<2>()
                    .ok_or_else(move || {
                        CustomTypeParseError::InvalidParameterCount(calc_params_count(), 2)
                    })?;
                let key_type = key_type_result?;
                let value_type = value_type_result?;
                Ok(ColumnType::Collection {
                    frozen: false,
                    typ: CollectionType::Map(Box::new(key_type), Box::new(value_type)),
                })
            }
            "TupleType" => {
                let params = self
                    .get_type_parameters()?
                    .collect::<Result<Vec<_>, CustomTypeParseError>>()?;
                if params.is_empty() {
                    return Err(CustomTypeParseError::InvalidParameterCount(0, 1));
                }
                Ok(ColumnType::Tuple(params))
            }
            "VectorType" => {
                let (typ, len) = self.get_vector_parameters()?;
                Ok(ColumnType::Vector {
                    typ: Box::new(typ),
                    dimensions: len,
                })
            }
            "UserType" => {
                let params = self.get_udt_parameters()?;
                Ok(ColumnType::UserDefinedType {
                    frozen: false,
                    definition: Arc::new(UserDefinedType {
                        name: params.type_name.into(),
                        keyspace: params.keyspace.into(),
                        field_types: params.field_types,
                    }),
                })
            }
            name => Err(CustomTypeParseError::UnknownComplexCustomTypeName(
                name.into(),
            )),
        }
    }
```

Before pushing this let's get @wprzytula or @piodul 's opinion on wheter this solution makes sense.

scylla-cql/src/frame/response/custom_type_parser.rs
Lorak-mmk: Oh, I actually wrote the possible solution (which involves copying the parser) before I noticed this discussion. See my comment below.

Lorak-mmk:
Oh, the `.clone()` can probably be ommited from my code since ParserState is Copy.

scylla-cql/src/frame/response/custom_type_parser.rs
wprzytula: Do you prefer to try implementing this yourself or leave it to me?

Lorak-mmk: > By the way: you could extract get_n_type_parameters<const n: usize>() method, which would wrap get_type_parameters() together with that parser state backup logic and parameter count mismatch handling.

That's a good idea, and should not be difficult, I encourage you to try @smoczy123

wprzytula: > Oh, I actually wrote the possible solution (which involves copying the parser) before I noticed this discussion. See my comment below.

I think I'll combine our ideas and create a new commit. Stay tuned.

wprzytula: @smoczy123 See https://github.com/wprzytula/scylla-rust-driver/commit/525ca1267288514348d3fb296f1393cc6aad9fd8

Lorak-mmk (approved):

Issue #691, opened 2023-04-05
Unify ColumnType and CqlType


piodul:
Currently, the driver has two complex but similar types: `ColumnType` and `CqlType`. They are both supposed to represent a CQL type, but have slight differences:

- `ColumnType` is used in metadata of prepared statements and responses to queries. It _does not_ include information about whether collections and UDTs are frozen. The type definitions are self contained - if it represents a complex type that contains a UDT somewhere inside it, it has the whole definition of the UDT inline.
- `CqlType` is used in ClusterMetadata - it is computed based on the information in the system tables. It _does_ include information about whether collections and UDTs are frozen. Until recently (changed in https://github.com/scylladb/scylla-rust-driver/pull/649) information about UDTs was not inlined, but now it is; however, if metadata fetch returns inconsistent metadata and it does not have a definition of a UDT that was needed inside some `CqlType`, then a placeholder `MissingUserDefinedType` will be put instead.

Those types have very similar names and function, so it might make sense to unify them. This will help write user logic that is supposed to operate on both representations. We will have to account for the differences:

- _Presence vs. absence of information about being frozen_: the newly introduced type could be generic over the "frozen flag" type. The "frozen flag" type could be `()` vs. `bool`, or two new enums isomorphic to those types that express the intent better.
- _Handling the `MissingUserDefinedType` case_:
	- The simplest way out would be to put `Result<_, MissingUserDefinedType>` in the unified type. However, this would force the users that need to operate on the existing `ColumnType` representation to handle the `Result`, even though it would never be an error, so that would introduce unnecessary complication and would be ugly.
	- An improvement upon the previous would be to convert `MissingUserDefinedType` into a generic parameter, like in the case of the "frozen flag" - in case of `ColumnType` it would be an uninhabited type. Slightly better, but the `Result` is still there.
	- (Preferred) Get rid of `MissingUserDefinedType` altogether! I insisted on handling the `MissingUserDefinedType` error like this to make the driver more robust; if we didn't have it then we would have to fail the whole metadata fetch and probably fail the creation of `Session`. However, if we move the cluster metadata module out of the driver like described in https://github.com/scylladb/scylla-rust-driver/issues/595, then it should be fine to fail the whole metadata fetch as it won't reduce availability. Therefore #595 is a requirement for this solution.

piodul:
[...]

PR #1166, opened 2025-01-09
Remove CqlType in favor of ColumnType

Lorak-mmk:
This PR removes `CqlType` (used in schema metadata) and replaces its usages with `ColumnType` (used in statement metadata).
Initially those two structures were quite different, so most of the commits are done to make them mostly identical, which made the unification possible.

## Layout 

CqlType was split into subtypes (native type, collection type, tuple, UDT) as defined by https://cassandra.apache.org/doc/stable/cassandra/cql/types.html
ColumnType was fully inlined - it had a variant for each possible type.

I decided to change ColumnType in this regard to follow CqlType, because CqlType's approach seems to be much more elegant.

## Frozen flag

CqlType contained frozen flag, because the schema contains information about whether the collection/UDT is frozen or not.
ColumnType did not contain such flag, because this information is not present in result metadata / arguments metadata.

There wasn't any real choice - we have to store the frozen flag to accurately represent the schema metadata.
For types used in statements, this flag is simply always set to false.

## UDT shared ownership

ColumnType stored the UDT data inline, this is how the variant looked:
```rust
    UserDefinedType {
        type_name: Cow<'frame, str>,
        keyspace: Cow<'frame, str>,
        field_types: Vec<(Cow<'frame, str>, ColumnType<'frame>)>,
    },
```

CqlType stored it in a separate struct, which was held under Arc:
```rust
/// Definition of a user-defined type
    UserDefinedType {
        frozen: bool,
        // Using Arc here in order not to have many copies of the same definition
        definition: Arc<UserDefinedType>,
    },
...

#[derive(Clone, Debug, PartialEq, Eq)]
pub struct UserDefinedType {
    pub name: String,
    pub keyspace: String,
    pub field_types: Vec<(String, CqlType)>,
}

```

This is because it may be stored in a few places:
- `Keyspace::user_defined_types`
- As a field in another UDT
- As a column in the table
- As a column in the view

This requires either cloning, or shared ownership.

## Custom type

ColumnType had a variant for Custom type, because it is defined by the protocol.
However, this type is rarely used, and we don't really have any working support for it, so I decided to remove it.

## Minor changes

I decided to change the size type of Vector to u16. It frees any users of this struct from dealing with negative numbers (which are disallowed by the protocol, and make no sense here).

Fixes https://github.com/scylladb/scylla-rust-driver/issues/691

## Pre-review checklist

- [ ] ~~I have adjusted the documentation in `./docs/source/`.~~ - Nothing to adjust

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -49,24 +49,47 @@ pub struct TableSpec<'a> {
     table_name: Cow<'a, str>,
 }
 
+/// A type of:
+/// - a column in schema metadata
+/// - a bind marker in a prepared statement
+/// - a column a in query result set
+///
+/// Some of the variants contain a `frozen` flag. This flag is only used
+/// in schema metadata. For prepared statement bind markers and query result
+/// types those fields will always be set to `false` (even if the DB column
+/// corresponding to given marker / result type is frozen).
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ColumnType<'frame> {
+    /// Types that are "simple" (non-recursive).
     Native(NativeType),
+
+    /// Collection types: Map, Set, and List. Those are composite types with
+    /// dynamic size but constant predefined element types.
     Collection {
         frozen: bool,
         typ: CollectionType<'frame>,
     },
+
+    /// A composite list-like type that has a defined size and all its elements
+    /// are of the same type. Intuitively, it can be viewed as a list with constant
+    /// predefined size, or as a tuple which has all elements of the same type.
     Vector {
         typ: Box<ColumnType<'frame>>,
         dimensions: u16,
     },
+
+    /// A C-struct-like type defined by the user.
     UserDefinedType {
         frozen: bool,
         definition: Arc<UserDefinedType<'frame>>,
     },
+
+    /// A composite type with a defined size and elements of possibly different,
+    /// but predefined, types.
     Tuple(Vec<ColumnType<'frame>>),
 }
 
+/// A [ColumnType] variants that are "simple" (non-recursive).
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum NativeType {
     Ascii,
@@ -91,6 +114,10 @@ pub enum NativeType {
     Varint,
 }
 
+/// Collection variants of [ColumnType]. A collection is a composite type that
+/// has dynamic size, so it is possible to add and remove values to/from it.
+///
+/// Tuple and vector are not collections because they have predefined size.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum CollectionType<'frame> {
     List(Box<ColumnType<'frame>>),
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -55,6 +55,10 @@ pub enum ColumnType<'frame> {
     Collection {
         type_: CollectionType<'frame>,
     },
+    Vector {
+        type_: Box<ColumnType<'frame>>,
+        dimensions: u16,
+    },
     UserDefinedType {
         definition: Arc<UserDefinedType<'frame>>,
     },
@@ -106,6 +110,10 @@ impl ColumnType<'_> {
             ColumnType::Native(b) => ColumnType::Native(b),
             ColumnType::Collection { type_: t } => ColumnType::Collection {
                 type_: t.into_owned(),
+            },
+            ColumnType::Vector { type_, dimensions } => ColumnType::Vector {
+                type_: Box::new(type_.into_owned()),
+                dimensions,
             },
             ColumnType::UserDefinedType { definition: udt } => {
                 let udt = Arc::try_unwrap(udt).unwrap_or_else(|e| e.as_ref().clone());
@@ -1391,6 +1399,12 @@ pub fn deser_cql_value(
             let s = Vec::<CqlValue>::deserialize(typ, v)?;
             CqlValue::Set(s)
         }
+        Vector { .. } => {
+            return Err(mk_deser_err::<CqlValue>(
+                typ,
+                BuiltinDeserializationErrorKind::Unsupported,
+            ))
+        }
         UserDefinedType { definition: udt } => {
             let iter = UdtIterator::deserialize(typ, v)?;
             let fields: Vec<(String, Option<CqlValue>)> = iter
@@ -1548,6 +1562,9 @@ mod test_utils {
                 Self::Collection {
                     type_: CollectionType::Set(_),
                 } => 0x0022,
+                Self::Vector { .. } => {
+                    unimplemented!();
+                }
                 Self::UserDefinedType { .. } => 0x0030,
                 Self::Tuple(_) => 0x0031,
             }
@@ -1581,6 +1598,9 @@ mod test_utils {
                     for typ in types.iter() {
                         typ.serialize(buf)?;
                     }
+                }
+                ColumnType::Vector { .. } => {
+                    unimplemented!()
                 }
                 ColumnType::UserDefinedType { definition: udt } => {
                     types::write_string(&udt.keyspace, buf)?;
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -56,9 +56,7 @@ pub enum ColumnType<'frame> {
         type_: CollectionType<'frame>,
     },
     UserDefinedType {
-        type_name: Cow<'frame, str>,
-        keyspace: Cow<'frame, str>,
-        field_types: Vec<(Cow<'frame, str>, ColumnType<'frame>)>,
+        definition: Arc<UserDefinedType<'frame>>,
     },
     Tuple(Vec<ColumnType<'frame>>),
 }
@@ -94,6 +92,14 @@ pub enum CollectionType<'frame> {
     Set(Box<ColumnType<'frame>>),
 }
 
+/// Definition of a user-defined type
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct UserDefinedType<'frame> {
+    pub name: Cow<'frame, str>,
+    pub keyspace: Cow<'frame, str>,
+    pub field_types: Vec<(Cow<'frame, str>, ColumnType<'frame>)>,
+}
+
 impl ColumnType<'_> {
     pub fn into_owned(self) -> ColumnType<'static> {
         match self {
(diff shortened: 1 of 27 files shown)


Lorak-mmk:
Rebased on main

Lorak-mmk:
Again known failure with "Failed to apply group 0 change due to concurrent modification"...

If Core is not willing to address this problem then maybe we should introduce client-side retries for those queries?

scylla/src/cluster/metadata.rs
```diff
@@ -216,7 +216,7 @@ enum PreCqlType {
         type_: Box<PreCqlType>,
         /// matches the datatype used by the java driver:
         /// <https://github.com/apache/cassandra-java-driver/blob/85bb4065098b887d2dda26eb14423ce4fc687045/core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java#L77>
-        dimensions: i32,
+        dimensions: u16,
```
muzarski, line 224: nit: You removed the comment in `CqlType` but not in `PreCqlType`.

scylla/src/cluster/metadata.rs
```diff
@@ -275,9 +275,7 @@ pub enum CqlType {
     Tuple(Vec<CqlType>),
     Vector {
         type_: Box<CqlType>,
-        /// matches the datatype used by the java driver:
-        /// <https://github.com/apache/cassandra-java-driver/blob/85bb4065098b887d2dda26eb14423ce4fc687045/core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java#L77>
-        dimensions: i32,
+        dimensions: u16,
```
muzarski: What about `(u16::MAX, i32::MAX]` values? I know that such values are impractical and probably will never appear there, but according the specification it is possible, right?

scylla/src/client/session_test.rs
```diff
@@ -6,7 +6,7 @@ use crate as scylla;
 use crate::batch::{Batch, BatchStatement};
```
muzarski, line 6: commit `Remove CqlType and related types` - I expected this commit to contain a lot of changes. To my surprise, there are very few of them. Great job of splitting this PR into the commits - thanks to that the final replacement was trivial.

Lorak-mmk:
Adrressed the comments

scylla/src/cluster/metadata.rs
Lorak-mmk: Which specification?

- CQLv5 binary protocol only mentions vectors once:
```
5.25 vector

  For a vector of n dimensions of a fixed-length type, a sequence of those n elements.
  For a vector with variable-length elements, the size of the elements will preced
  each element.  Each element is the [bytes] representing the serialized value.  The
  number of dimensions is not encoded, since it's part of the type definition.
```
There is no defined binary type encoding for vectors (which would give is the type to use for dimensions) - it is encoded as text (custom type) which is why https://github.com/scylladb/scylla-rust-driver/pull/1165 needs to do so many changes.

- Cassandra docs say that the maximum amount of elements in a vector is 8K (2^13) which fits in u16: https://cassandra.apache.org/doc/latest/cassandra/reference/vector-data-type.html

muzarski: >     * Cassandra docs say that the maximum amount of elements in a vector is 8K (2^13) which fits in u16: https://cassandra.apache.org/doc/latest/cassandra/reference/vector-data-type.html

Ok, this convinces me. I based my assumption on the code snippet (https://github.com/apache/cassandra-java-driver/blob/85bb4065098b887d2dda26eb14423ce4fc687045/core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java#L77). I forgot that Java does not have unsigned integral data types... But then, I believe that `short` data type in Java should be enough to store a value up to 8k. I wonder why they chose `int`.

Lorak-mmk: I don't think they pay much attention to such things. If they did then CQL protocol wouldn't have such obvious issues unfixed for so long is it does now.

muzarski (approved):

wprzytula:
> Again known failure with "Failed to apply group 0 change due to concurrent modification"...
> 
> If Core is not willing to address this problem then maybe we should introduce client-side retries for those queries?

This sounds like a good workaround.

scylla/src/utils/parse.rs
```diff
@@ -87,18 +87,18 @@ impl<'s> ParserState<'s> {
...
 
-    /// Parses a sequence of digits and '-' as an integer.
-    /// Consumes characters until it finds a character that is not a digit or '-'.
+    /// Parses a sequence of digits as an integer.
+    /// Consumes characters until it finds a character that is not a digit.
     ///
     /// An error is returned if:
-    /// * The first character is not a digit or '-'
+    /// * The first character is not a digit
     /// * The integer is larger than i32
```
wprzytula, line 95: :wrench: Outdated docstring: `i32` -> `u16`

scylla-cql/src/frame/response/result.rs
```diff
@@ -1391,6 +1399,12 @@ pub fn deser_cql_value(
             let s = Vec::<CqlValue>::deserialize(typ, v)?;
             CqlValue::Set(s)
         }
+        Vector { .. } => {
+            return Err(mk_deser_err::<CqlValue>(
+                typ,
+                BuiltinDeserializationErrorKind::Unuspported,
```
wprzytula, line 1419: :wrench: typo: `Unuspported` -> `Unsupported`

scylla-cql/src/frame/response/result.rs
```diff
@@ -1548,6 +1562,9 @@ mod test_utils {
                 Self::Collection {
                     type_: CollectionType::Set(_),
                 } => 0x0022,
+                Self::Vector { .. } => {
+                    panic!("Vector serialization unimplemented");
+                }
```
wprzytula, line 1586: :pick: `unimplemented!()` is designed specifically for this use case.

scylla-cql/src/frame/response/result.rs
```diff
@@ -1582,6 +1599,9 @@ mod test_utils {
                         typ.serialize(buf)?;
                     }
                 }
+                ColumnType::Vector { .. } => {
+                    panic!("Vector serialization unimplemented");
+                }
```
wprzytula, line 1626: :pick: ditto: `unimplemented!()`.

scylla-cql/src/types/deserialize/value.rs
```diff
@@ -1830,6 +1830,9 @@ pub enum BuiltinDeserializationErrorKind {
 
     /// A deserialization failure specific to a CQL UDT.
     UdtError(UdtDeserializationErrorKind),
+
+    /// Deserialization of this CQL type is not supported by the driver
+    Unuspported,
```
wprzytula, line 1846: :pick: Let's be consistent about dots at the end of docstrings here (and everywhere else). It matters at least for me, so perhaps for other readers too.

scylla-cql/src/frame/response/result.rs
```diff
@@ -52,9 +52,9 @@ pub struct TableSpec<'a> {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ColumnType<'frame> {
     Native(NativeType),
-    List(Box<ColumnType<'frame>>),
-    Map(Box<ColumnType<'frame>>, Box<ColumnType<'frame>>),
-    Set(Box<ColumnType<'frame>>),
+    Collection {
+        type_: CollectionType<'frame>,
+    },
```
wprzytula: :pick: `type_` looks awful. Let's use `typ` instead, as in many other places in the driver.

scylla-cql/src/frame/response/result.rs
```diff
@@ -53,13 +53,15 @@ pub struct TableSpec<'a> {
...
     },
     Vector {
         type_: Box<ColumnType<'frame>>,
         dimensions: u16,
     },
     UserDefinedType {
+        frozen: bool,
         definition: Arc<UserDefinedType<'frame>>,
     },
     Tuple(Vec<ColumnType<'frame>>),
```
wprzytula: :wrench: As this is a `pub` type, I suggest adding docstrings to all variants and their fields. Especially the `frozen` flag may require explanation, as it's only meaningful in the schema metadata case, which is not obvious.

wprzytula (changes requested):

scylla-cql/src/frame/response/result.rs
```diff
@@ -56,9 +56,7 @@ pub enum ColumnType<'frame> {
         type_: CollectionType<'frame>,
     },
     UserDefinedType {
-        type_name: Cow<'frame, str>,
-        keyspace: Cow<'frame, str>,
-        field_types: Vec<(Cow<'frame, str>, ColumnType<'frame>)>,
+        definition: Arc<UserDefinedType<'frame>>,
```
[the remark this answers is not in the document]
Lorak-mmk: That's true, but its worth noting that it's not a change from no allocations to an allocation, but from at least 1 allocation to at least 2 allocations - there was previously an allocation because of Vec.

scylla-cql/src/frame/response/result.rs
```diff
@@ -933,9 +943,11 @@ fn deser_type_generic<'frame, 'result, StrT: Into<Cow<'result, str>>>(
...
-                type_name: type_name.into(),
-                keyspace: keyspace_name.into(),
-                field_types,
+                definition: Arc::new(self::UserDefinedType {
+                    name: type_name.into(),
+                    keyspace: keyspace_name.into(),
+                    field_types,
+                }),
             }
         }
```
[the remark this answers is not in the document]
Lorak-mmk: Yes, that's how it has been before this PR too. This is also a pretty rare scenario I think.

scylla-cql/src/frame/response/result.rs
Lorak-mmk: Added documentation in a new commit.

Lorak-mmk:
Addressed all the review comments. I'll rebase on main in a separate push.

Lorak-mmk:
- Rebased on main
- Fixed an issue where the new types (NativeType) etc were not re-exported in `scylla` crate. Added a test for that in `hygiene.rs`

Lorak-mmk:
Fixed CI errors. I'll run `make  ci` again on  all commits to ensure they work.

scylla-cql/src/frame/response/result.rs
```diff
@@ -49,9 +49,50 @@ pub struct TableSpec<'a> {
...
+/// In schema metadata this represents simply a column type. In prepared
+/// statement this is used in two places: as a type of bind marker and as a type
+/// of returned value.
+///
+/// Some of the variants contain a `frozen` flag. This flag is only used
+/// in schema metadata. For prepared statement bind markers and result types
+/// those fields will always be set to `false` (even if the DB column
+/// corresponding to given marker / result type is frozen).
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ColumnType<'frame> {
```
wprzytula, line 63: :thinking: I believe the `ColumnType` is used in the result set of both prepared and unprepared statements. The docstring specifically mentions only prepared statements, which I'm afraid is misleading.

scylla-cql/src/frame/response/result.rs
```diff
@@ -49,9 +49,50 @@ pub struct TableSpec<'a> {
...
+
+    // A C-struct-like type defined by the user.
+    UserDefinedType {
+        frozen: bool,
+        definition: Arc<UserDefinedType<'frame>>,
+    },
+
+    // A composite type with a defined size and elements of possibly different,
+    // but predefined, types.
+    Tuple(Vec<ColumnType<'frame>>),
```
wprzytula, line 90: :+1: Very useful docstrings, good job!

:wrench: The lower three variants have `//` instead of `///`.

scylla-cql/src/frame/response/result.rs
```diff
@@ -66,68 +107,85 @@ pub enum ColumnType<'frame> {
...
     Timeuuid,
-    Tuple(Vec<ColumnType<'frame>>),
     Uuid,
     Varint,
 }
 
+/// Collection variants of [ColumnType]. A collection is a composite type that:
+/// has no predefined size, so it is possible to add and remove values to/from it.
+///
+/// Tuple and vector are not collections because they don't have dynamic size.
```
wprzytula, line 121: :pick: I'd remove negations from both sentences:

```suggestion
/// Collection variants of [ColumnType]. A collection is a composite type that:
/// has dynamic size, so it is possible to add and remove values to/from it.
///
/// Tuple and vector are not collections because they have predefined size.
```

scylla-cql/src/frame/response/result.rs
```diff
@@ -1269,109 +1342,116 @@ pub fn deser_cql_value(
     let v = Some(FrameSlice::new_borrowed(buf));
```
wprzytula, line 1342: :pushpin: When removing the legacy deserialization framework, please check if it's possible to remove the workaround in `FrameSlice`: the `new_borrowed` constructor.

wprzytula (changes requested):

wprzytula (approved):

PR #1170, opened 2025-01-16
errors: metadata errors refactor

muzarski:

## Motivation

The goal of this PR is to purge `metadata.rs` of `QueryError`, and replace it with `MetadataError`. Thanks to that metadata related functions are now independent of `QueryError`.

I tried to split this PR into commits, so each commit handles one function/method from metadata module. New error variants are introduced where necessary.

When it comes to public API - after this PR `Session::refresh_metadata()` will now return `MetadataError`.

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/metadata.rs
+++ b/scylla/src/cluster/metadata.rs
@@ -782,16 +782,16 @@ async fn query_peers(conn: &Arc<Connection>, connect_port: u16) -> Result<Vec<Pe
                     "system.peers or system.local has an invalid row, skipping it: {}",
                     err
                 );
-                Ok(None)
+                None
             }
         }
     });
 
     let peers = translated_peers_futures
         .buffer_unordered(256)
-        .try_filter_map(|x| std::future::ready(Ok(x)))
-        .try_collect::<Vec<_>>()
-        .await?;
+        .filter_map(std::future::ready)
+        .collect::<Vec<_>>()
+        .await;
     Ok(peers)
 }
 
@@ -799,7 +799,7 @@ async fn create_peer_from_row(
     source: NodeInfoSource,
     row: NodeInfoRow,
     local_address: SocketAddr,
-) -> Result<Option<Peer>, QueryError> {
+) -> Option<Peer> {
     let NodeInfoRow {
         host_id,
         untranslated_ip_addr,
@@ -812,7 +812,7 @@ async fn create_peer_from_row(
         Some(host_id) => host_id,
         None => {
             warn!("{} (untranslated ip: {}, dc: {:?}, rack: {:?}) has Host ID set to null; skipping node.", source.describe(), untranslated_ip_addr, datacenter, rack);
-            return Ok(None);
+            return None;
         }
     };
 
@@ -852,13 +852,13 @@ async fn create_peer_from_row(
         }
     };
 
-    Ok(Some(Peer {
+    Some(Peer {
         host_id,
         address: node_addr,
         tokens,
         datacenter,
         rack,
-    }))
+    })
 }
 
 fn query_filter_keyspace_name<'a, R>(
--- a/scylla/src/cluster/metadata.rs
+++ b/scylla/src/cluster/metadata.rs
@@ -19,7 +19,10 @@
 use crate::client::pager::{NextPageError, NextRowError, QueryPager};
 use crate::cluster::node::resolve_contact_points;
 use crate::deserialize::DeserializeOwnedRow;
-use crate::errors::{DbError, NewSessionError, QueryError, RequestAttemptError};
+use crate::errors::{
+    DbError, MetadataFetchError, MetadataFetchErrorKind, NewSessionError, QueryError,
+    RequestAttemptError,
+};
 use crate::frame::response::event::Event;
 use crate::network::{Connection, ConnectionConfig, NodeConnectionPool, PoolConfig, PoolSize};
 use crate::policies::host_filter::HostFilter;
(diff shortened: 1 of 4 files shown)


scylla/src/cluster/metadata.rs
```diff
@@ -1020,11 +1010,10 @@ where
                 .await
                 .map_err(E::convert_next_row_error)
         }
-    }
+    };
 
     let fut = async move {
-        let pager =
-            make_keyspace_filtered_query_pager::<E>(conn, query_str, keyspaces_to_fetch).await?;
+        let pager: QueryPager = pager_fut.await?;
```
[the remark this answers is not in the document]
muzarski: Done.

muzarski:
v1.1: addressed comments

muzarski:
Rebased on main

wprzytula (approved):

scylla/src/cluster/metadata.rs
```diff
@@ -961,30 +962,46 @@ async fn create_peer_from_row(
...
+/// Used by [`query_filter_keyspace_name`].
+trait MetadataErrorConverter {
+    type DestError;
+
+    fn convert_typecheck_error(err: TypeCheckError) -> Self::DestError;
+    fn convert_prepare_error(err: RequestAttemptError) -> Self::DestError;
+    fn convert_serialization_error(err: SerializationError) -> Self::DestError;
+    fn convert_next_row_error(err: NextRowError) -> Self::DestError;
+}
+
```
Lorak-mmk, line 971: 🤔
1. This trait introduces a bit weird inversion of responsibility. Typically the function itself defines the error it returns, and the caller decides what to do with the error. Now the function has to convert its errors to the one required by the caller.

Alternative solution that I see would be to make a private error type, returned by `query_filter_keyspace_name`. Other errors would then have constructors / `From` impls to convert from this error, and callers of `query_filter_keyspace_name` would perform the conversion.

This approach seems quite obvious to me, so if it has downsides that make the trait approach better then I think the rationale should be put in the comment and/or commit message,

2. If many errors share the same variants, why not make those variants their own type, and include this type in the other errors? What I mean is we can have `MetadataRequestError` with variants containing the 4 errors from the trait, and include this `MetadataRequestError` in `UdtMetadataError` and others. what is the benefit of inlining the variants?

3. I'm trying to think about a typical use case for matching on metadata errors. I think that a user will typically care more about why something failed, instead of which part of the process failed. What I mean is that it is more important that "some part of metadata fetch failed because server is overloaded" than "fetching column specs failed because <some reason>". 
Idea: maybe there should be `MetadataFetchError` or sth, containing:
- An enum with a reason of failing, including the 4 above (TypeCheckError etc) and possibly others
- [Optional] An enum describing which part of the fetch failed (tables, udts etc).

Of course there could be different errors for conditions that don't fit this one (UDT sorting wold be such a case, right?).

muzarski: > 🤔
> 
>     1. This trait introduces a bit weird inversion of responsibility. Typically the function itself defines the error it returns, and the caller decides what to do with the error. Now the function has to convert its errors to the one required by the caller.
> 
> 
> Alternative solution that I see would be to make a private error type, returned by `query_filter_keyspace_name`. Other errors would then have constructors / `From` impls to convert from this error, and callers of `query_filter_keyspace_name` would perform the conversion.
> 
> This approach seems quite obvious to me, so if it has downsides that make the trait approach better then I think the rationale should be put in the comment and/or commit message,

You are right. I was a bit blinded by a previous approach where we would accept a closure that converts the `TypeCheckError` to `MetadataError` - it made a lot of sense before, because we would return much broader type - `QueryError`.. But now that we can return narrower type (e.g. private `MetadataRequestError` you mentioned) , there is no point in accepting conversion closures/traits.

>     2. If many errors share the same variants, why not make those variants their own type, and include this type in the other errors? What I mean is we can have `MetadataRequestError` with variants containing the 4 errors from the trait, and include this `MetadataRequestError` in `UdtMetadataError` and others. what is the benefit of inlining the variants?
>
>     3. I'm trying to think about a typical use case for matching on metadata errors. I think that a user will typically care more about why something failed, instead of which part of the process failed. What I mean is that it is more important that "some part of metadata fetch failed because server is overloaded" than "fetching column specs failed because ".
>        Idea: maybe there should be `MetadataFetchError` or sth, containing:
> 
> 
>     * An enum with a reason of failing, including the 4 above (TypeCheckError etc) and possibly others
> 
>     * [Optional] An enum describing which part of the fetch failed (tables, udts etc).

Now that we have a big picture, these suggestions make a lot of sense. As you noticed, all enums share the 4 common variants (`TypeCheckError`, `NextRowError`, etc.).  It was not the case before this PR, and so it was hard for me to reason about it this way.


> Of course there could be different errors for conditions that don't fit this one (UDT sorting wold be such a case, right?).

Yes, such errors can stay in their corresponding variants that currently exist. They will contain logical errors (circular UDT dependency, bad keyspace replication strategy etc.) , rather than errors related to fetching metadata (sending requests, deserialization etc.)

I'll apply the suggestions. I really like the idea of extracting the `MetadataFetchError`. I wonder if I can do this change at the end of this PR (on top of existing commits), or think of a reasonable way to rebase and edit the existing commits.

Lorak-mmk: > I'll apply the suggestions. I really like the idea of extracting the `MetadataFetchError`. I wonder if I can do this change at the end of this PR (on top of existing commits), or think of a reasonable way to rebase and edit the existing commits.

Both approaches are fine with me.

muzarski:
v2: applied @Lorak-mmk suggestion. Introduced `MetadataFetchError`. I managed to edit existing commits.

wprzytula (approved):

muzarski:
Rebased on main

wprzytula (approved):

Lorak-mmk (approved):
I really like the new errors here!

Issue #1052, opened 2024-08-04
Decouple schema fetch queries from server-side timeouts


dkropachev:
### Details
If the user set server-side timeouts, or service level timeouts, this impacts schema fetches. The timeout can be too low to allow a schema fetch when the schema is very large.

Instead, qualify schema (and other metadata) fetches with USING TIMEOUT (with a high and configurable default) so schema fetches work regardless of timeout tuning.

Core issue: https://github.com/scylladb/scylladb/issues/18471

### Limitations

1. `USING TIMEOUT` is not supported by `cassandra` and support by all major `scylla` versions, so it is safe to assume it is supported by seeing scylla-specific metadata.
2. Minimal supported scale is `ms`, so statement suffix should look like this: `USING TIMEOUT {timeout_in_ms} ms` 

### Requirements

1. Timeout should be configurable and default should be `2s`
2. `0` signals that feature is disabled

Migrated to Jira: [DRIVER-192](https://scylladb.atlassian.net/browse/DRIVER-192)

[DRIVER-192]: https://scylladb.atlassian.net/browse/DRIVER-192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mykaul:
We've implemented it in all other drivers, we see customers hit by this in the field - I don't understand why it's a 1.1.0 content and not sooner.

CC @roydahan

Lorak-mmk:
We have a deadline for 1.0 and we won't be able to keep it if we add more stuff there.
In contrary to the changes that we will include in 1.0, this one can actually be implemented in 1.1.

mykaul:
> We have a deadline for 1.0 and we won't be able to keep it if we add more stuff there. In contrary to the changes that we will include in 1.0, this one can actually be implemented in 1.1.

Create a release before 1.0 with that fix. A mythical '1.0' cannot have a priority over real customer issues.

roydahan:
This is a new feature, not going to be merged into 0.15.
Nice to have for 1.0.0, only if it makes it in time.

PR #1171, opened 2025-01-17
Metadata queries with configurable server-side timeouts

wprzytula:
## Motivation
Some users tune their server-side timeouts so that they are tighter. For clusters with a large schema, however, this sometimes made schema queries time out. This PR excludes metadata queries from the default server-side timeout by overriding it with a custom one, to prevent timeouts upon querying metadata.

## What's done

**Adds `USING TIMEOUT` clause to metadata queries when applicable.**

By _metadata_ I mean schema + topology. I did it for both for consistency, even though only schema was required in the [issue](#1052)).

## Implementation

To ensure that no fetches are accidentally omitted from having the timeout added, both now and in the future, I created a new abstraction: `ControlConnection`, which is just a wrapper over `Arc<Connection>` that exposes some methods of `Arc<Connection>` (`query_iter`, `execute_iter`, `prepare`, and some minor getters), taking care of adding the timeout before execution if applicable.

A corresponding setting is added to `SessionConfig`, and `SessionBuilder` gets a new method for configuring it.

## Testing

I added two tests:
1. A unit test of `ControlConnection`, which asserts that for ScyllaDB the timeout is enforced (if set) and for Cassandra is always ignored, in the following cases:
    - when explicitly disabled (no custom timeout follows),
    - when explicitly set to some (only set for ScyllaDB).
2. An integration test for the driver, which asserts that for ScyllaDB the timeout is enforced (if set) and for Cassandra is always ignored, in the following cases:
    - when explicitly disabled (no custom timeout follows),
    - when explicitly set to some (only set for ScyllaDB),
    - when left as implicit default (only set to ScyllaDB).

Fixes: #1052

## Pre-review checklist


wprzytula:
As you can see, our SSL and Authenticate workflows fail, because the ScyllaDB run there is so old that it does not recognize `USING TIMEOUT` syntax...:
- [scylladb/scylla-passauth](https://hub.docker.com/r/scylladb/scylla-passauth)
  •Updated almost 4 years ago
- [scylladb/scylla-tls](https://hub.docker.com/r/scylladb/scylla-tls)
  •Updated almost 4 years ago

I guess we **urgently** need to create new images, to really test that the driver is compatible with those features in the new Scyllas, not 4-years-old ones...

dkropachev:
> As you can see, our SSL and Authenticate workflows fail, because the ScyllaDB run there is so old that it does not recognize `USING TIMEOUT` syntax...:
> 
> * [scylladb/scylla-passauth](https://hub.docker.com/r/scylladb/scylla-passauth)
>   •Updated almost 4 years ago
> * [scylladb/scylla-tls](https://hub.docker.com/r/scylladb/scylla-tls)
>   •Updated almost 4 years ago
> 
> I guess we **urgently** need to create new images, to really test that the driver is compatible with those features in the new Scyllas, not 4-years-old ones...
> 
> cc @dkropachev

From what I see these are regular images, based of scylladb/scylla/4.3.rc0 with some patches at `/etc/scylla` on top of it.
What I recomend you doing, create a configuration that you need to have at `/etc/scylla/` put it in repo and mount it at `/etc/scylla`,  and use scylladb/scylla as an image.

dkropachev:
> > As you can see, our SSL and Authenticate workflows fail, because the ScyllaDB run there is so old that it does not recognize `USING TIMEOUT` syntax...:
> > 
> > * [scylladb/scylla-passauth](https://hub.docker.com/r/scylladb/scylla-passauth)
> >   •Updated almost 4 years ago
> > * [scylladb/scylla-tls](https://hub.docker.com/r/scylladb/scylla-tls)
> >   •Updated almost 4 years ago
> > 
> > I guess we **urgently** need to create new images, to really test that the driver is compatible with those features in the new Scyllas, not 4-years-old ones...
> > cc @dkropachev
> 
> From what I see these are regular images, based of scylladb/scylla/4.3.rc0 with some patches at `/etc/scylla` on top of it. What I recomend you doing, create a configuration that you need to have at `/etc/scylla/` put it in repo and mount it at `/etc/scylla`, and use scylladb/scylla as an image.

Done up [here](https://github.com/scylladb/scylla-rust-driver/pull/1173)

muzarski (approved):
LGTM. Nice piece of code - especially the test cases

scylla/src/transport/connection.rs
```diff
@@ -2406,7 +2406,7 @@ mod tests {
     use std::time::Duration;
 
     // Just like resolve_hostname in session.rs
-    async fn resolve_hostname(hostname: &str) -> SocketAddr {
+    pub(crate) async fn resolve_hostname(hostname: &str) -> SocketAddr {
         match tokio::net::lookup_host(hostname).await {
             Ok(mut addrs) => addrs.next().unwrap(),
```
Lorak-mmk, line 2411: 🔧 Instead of pub(crate)'ing it please move it to test_utils - this is what this module is for.

scylla/src/transport/topology.rs
```diff
@@ -869,6 +870,181 @@ mod control_connection {
...
+        #[cfg(not(scylla_cloud_tests))]
+        #[tokio::test]
+        #[ntest::timeout(2000)]
+        async fn test_custom_timeouts() {
+            setup_tracing();
+
+            let proxy_addr = SocketAddr::new(scylla_proxy::get_exclusive_local_address(), 9042);
+            let uri = std::env::var("SCYLLA_URI").unwrap_or_else(|_| "127.0.0.1:9042".to_string());
+            let node_addr: SocketAddr = resolve_hostname(&uri).await;
+
```
Lorak-mmk, line 908: ❓ This test requires a running Scylla / C* cluster, right? Is it possible to avoid this?

scylla/src/transport/topology.rs
```diff
@@ -743,28 +743,30 @@ impl MetadataReader {
...
-    let keyspaces_query = query_keyspaces(conn, keyspace_to_fetch, fetch_schema);
+impl ControlConnection {
+    async fn query_metadata(
+        self,
+        connect_port: u16,
+        keyspace_to_fetch: &[String],
+        fetch_schema: bool,
+    ) -> Result<Metadata, QueryError> {
+        let peers_query = self.clone().query_peers(connect_port);
+        let keyspaces_query = self.query_keyspaces(keyspace_to_fetch, fetch_schema);
```
Lorak-mmk, line 758: Commit: "topology: make metadata fetchers methods on ControlConnection "

🔧 For me it is extremely weird and unintuitive to have multiple impl blocks for a struct (ControlConnection), scattered in various places of 2 modules (control_connection and topology).
Can we please put all the `impl ControlConnection` into the `control_connection` module, and preferably make them a single `impl` block?

scylla/src/transport/topology.rs
```diff
@@ -743,28 +743,30 @@ impl MetadataReader {
...
-    let peers_query = query_peers(conn, connect_port);
-    let keyspaces_query = query_keyspaces(conn, keyspace_to_fetch, fetch_schema);
+impl ControlConnection {
+    async fn query_metadata(
+        self,
+        connect_port: u16,
+        keyspace_to_fetch: &[String],
+        fetch_schema: bool,
+    ) -> Result<Metadata, QueryError> {
+        let peers_query = self.clone().query_peers(connect_port);
```
Lorak-mmk, line 757: 🌱 When porting the PR to the `main` branch, after applying my previous comment about ControlConnection impls, I think it would be a good idea to extract `control_connection` module to a separate file.

Lorak-mmk (changes requested):

scylla/src/transport/topology.rs
wprzytula: https://www.reddit.com/r/rust/comments/w5l320/is_having_multiple_impl_blocks_idiomatic/

It's a perfectly valid way of logical separation. In this particular case, we separate metadata-related functionalities (defined in the impl block in `metadata.rs` main module) from the lower-level functionalities (defined in the impl block in `control_connection` module). Also, what we get is that methods defined out of `control_connection` module **can't access** mod-private fields and methods of `ControlConnection`.

scylla/src/transport/topology.rs
wprzytula: I think `ControlConnection` could be extracted to a separate file, but with those higher-level methods (`query_peers` etc.) left in the `metadata.rs` file. The reason is that `ControlConnection` itself is a medium, which should be logically separate from the logic performing queries using it.

scylla/src/transport/topology.rs
wprzytula: It is possible to avoid this by using the proxy in the _dry mode_. We could mock the whole CQL handshake (`OPTIONS -> SUPPORTED -> STARTUP -> READY -> REGISTER -> READY`) and then intercept `QUERY`s and `PREPARE`s as before. An additional gain would be to test multiple scenarios at ones, i.e. sharded and non-sharded endpoints, independently of the actual cluster being deployed for tests.

wprzytula: I've made the test use the dry mode to avoid dependency on a running cluster.

wprzytula:
v1.1: resolved comments,
- made the unit test not require a running cluster.

roydahan:
There is no point in merging this into 0.15. branch, we're not going to release it another 0.15 release.

scylla/src/cluster/metadata.rs
```diff
@@ -601,7 +601,8 @@ impl MetadataReader {
     async fn fetch_metadata(&self, initial: bool) -> Result<Metadata, MetadataError> {
         // TODO: Timeouts?
         self.control_connection.wait_until_initialized().await;
-        let conn = ControlConnection::new(self.control_connection.random_connection()?);
+        let conn = ControlConnection::new(self.control_connection.random_connection()?)
+            .override_timeout(None);
```
Lorak-mmk: Commit "metadata: customise ControlConnection's request timeout"

Why call `.override_timeout(None)` here? It is none even without this call.

scylla/src/cluster/metadata.rs
```diff
@@ -773,30 +777,62 @@ mod control_connection {
     #[derive(Clone)]
     pub(super) struct ControlConnection {
         conn: Arc<Connection>,
+        overriden_timeout: Option<Duration>,
     }
```
Lorak-mmk: Could you change the name to indicate that it is about server-side timeout?

scylla/src/client/session.rs
```diff
@@ -278,6 +281,7 @@ impl SessionConfig {
             timestamp_generator: None,
             keyspaces_to_fetch: Vec::new(),
             fetch_schema_metadata: true,
+            metadata_request_serverside_timeout: Some(Duration::from_secs(2)),
             keepalive_interval: Some(Duration::from_secs(30)),
             keepalive_timeout: Some(Duration::from_secs(30)),
```
Lorak-mmk, line 286: I wonder if this is the right default value. As far as I understand, the goal is to prevent wasting work (which is wasted, if the cluster is overloaded and we hit the timeout).
Maybe 20 seconds would be better?

cc @mykaul

Lorak-mmk (changes requested):
It's awesome that you made the in-crate tests not require real db!

scylla/src/cluster/metadata.rs
```diff
@@ -813,277 +814,284 @@ mod control_connection {
...
-        .map(|result| result.map(|stream| stream.map_err(MetadataFetchErrorKind::NextRowError)))
-        .try_flatten()
-        // Add table context to the error.
-        .map_err(|error| MetadataFetchError {
-            error,
-            table: "system.peers",
-        })
-        .and_then(|row_result| future::ok((NodeInfoSource::Peer, row_result)));
+        let local_ip: IpAddr = self.get_connect_address().ip();
+        let local_address = SocketAddr::new(local_ip, connect_port);
```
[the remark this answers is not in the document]
wprzytula: `self` is consumed below the new location of `local_ip`. If I were to leave it where it had been before, I would have to clone the `Arc` once more, which I preferred to avoid.

scylla/src/cluster/metadata.rs
wprzytula: Very good idea.

scylla/src/cluster/metadata.rs
wprzytula: It's there just for explicitness. The `None` serves as a placeholder, which is replaced by an actual value in the next commit.

scylla/src/cluster/metadata.rs
```diff
@@ -710,28 +710,30 @@ impl MetadataReader {
...
-    keyspace_to_fetch: &[String],
-    fetch_schema: bool,
-) -> Result<Metadata, MetadataError> {
-    let peers_query = query_peers(conn, connect_port);
-    let keyspaces_query = query_keyspaces(conn, keyspace_to_fetch, fetch_schema);
+impl ControlConnection {
+    async fn query_metadata(
+        self,
+        connect_port: u16,
+        keyspace_to_fetch: &[String],
```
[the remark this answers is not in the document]
wprzytula: > ❓ Why does it take `self`? Previously it took `&Arc<Connection>` and I see no mention of this change in commit message.

It takes `self`, because it's supposed to be a method on `ControlConnection`, so that `ControlConnection` can hide its implementation details (at this time: just applying the custom server-side timeout) when executing metadata queries.

> Imo ControlConnection should not be possible to clone at all, hinting that we can't have multiple of themm

Although I understand your concern, I still believe the approach is justified. 
1. `ControlConnection` is clonable by sharing ownership only (because it has an `Arc` inside), so it's still keeping the single control connection.
2. There are more types that have such shared ownership in the driver, most notably `ExecutionProfile` and `ExecutionProfileHandle`. `ExecutionProfile` is shared when cloned, and immutable, which makes it very similar to current `ControlConnection`.
3. What could be a better alternative? Would you prefer to

scylla/src/cluster/metadata.rs
wprzytula: Done in a new commit.

scylla/src/cluster/metadata.rs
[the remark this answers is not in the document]
Lorak-mmk: > It takes self, because it's supposed to be a method on ControlConnection, so that ControlConnection can hide its implementation details (at this time: just applying the custom server-side timeout) when executing metadata queries.

What I meant is why it takes `self` instead of `&self`. Before the change it took a reference. I don't see why you decided to change this.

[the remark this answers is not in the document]
wprzytula: Then the answer is simple: it takes `self` because its callees take `self`. It's the same as what we would do about `String`: if a callee of `foo` takes `String`, then `foo` would take `String` too, not `&str` or `&String`, because this way we can avoid a clone.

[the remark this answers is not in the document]
muzarski: > Then the answer is simple: it takes `self` because its callees take `self`. It's the same as what we would do about `String`: if a callee of `foo` takes `String`, then `foo` would take `String` too, not `&str` or `&String`, because this way we can avoid a clone.

But do the callees need to take `self`? Why can't all `query_*` methods accept `&self`?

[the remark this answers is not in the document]
muzarski: Ahh, I see that `Connection` methods accept `self: Arc<Self>` (before this PR). I assumed that they accept `&self`. In such case, I think it's fine that `ControlConnectionMethods` accept `self` instead of `&self`.

muzarski (approved):

scylla/src/cluster/metadata.rs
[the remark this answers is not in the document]
wprzytula: > Ahh, I see that `Connection` methods accept `self: Arc<Self>` (before this PR). I assumed that they accept `&self`. In such case, I think it's fine that `ControlConnection` methods accept `self` instead of `&self`.

Yep.

> If they accepted `&self`, we would need to `Arc::clone()` the underlying connection anyway. Maybe it's better? Since the `Arc::clone()` is explicit, and not hidden behind `derive(Clone)` of `ControlConnection`?

I believe the current behaviour is in line with the general rule of passing ownership to the function if only it needs to consume the object, which is the case here. That's why I'm still convinced to leave it in the current state.

wprzytula:
Removed `#[derive(Clone)]` from `ControlConnection` according to @Lorak-mmk's suggestion.
Rewrote commits so that there are no longer fixes in later commits.

muzarski (approved):
I see that in the end, we decided to accept `&self` in `ControlConnection` methods and `Arc::clone()` the underlying `Connection` internally. Makes sense.

PR #1173, opened 2025-01-18
Get rid of custom scylla images

dkropachev:
Stop using custom images.

We used to use `scylladb/scylla-passauth` and `scylladb/scylla-tls` in our pipeline.
They were based on old scylla and has minimal config changes on them.
This PR removes them, to eleviate burdain to keep them updated.
Instead it makes them build on fly or just patch config via docker image cli.

## Pre-review checklist

### DIFF
--- a/test/tls/scylla.yaml
+++ b/test/tls/scylla.yaml
@@ -20,9 +20,6 @@
 # The more tokens, relative to other nodes, the larger the proportion of data
 # that this node will store. You probably want all nodes to have the same number
 # of tokens assuming they have equal hardware capability.
-#
-# If you already have a cluster with 1 token per node, and wish to migrate to
-# multiple tokens per node, see http://cassandra.apache.org/doc/latest/operating
 num_tokens: 256
 
 # Directory where Scylla should store all its files, which are commitlog,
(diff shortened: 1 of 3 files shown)


dkropachev:
@wprzytula and probably you will need to backport it to [scylladb:branch-0.15.x](https://github.com/scylladb/scylla-rust-driver/tree/branch-0.15.x)

test/tls/scylla.yaml
roydahan, line 1: Why do we need to store the entire scylla.yaml and maintain it here?
It's an error prone approach that we miss some updates on scylla side.

My suggestion is to have here only the params we add/override and have a function to "append/override" scylla.yaml.
(It can be a follow up PR because this one already make the situation better than now).

dkropachev: There is no good way to fix it on image,  it would be nice to have this feature there.

dkropachev: https://github.com/scylladb/scylladb/pull/22392

Lorak-mmk (changes requested):
Regarding the checklist: a convention we use is to:
- Check the items that apply and were done
- Cross-out items that are not applicable (`~~`)
- Leave unmarked the items that were not done for some reason.

So if the PR doesn't change the documentation (because it has not reason to), then please make it unmarked and crossed out.

This lets reviewers to know if something is not done on purpose, or because it is not applicable.

The changes themselves look fine (apart from some grammar issues in the commit messages).

dkropachev:
> Regarding the checklist: a convention we use is to:
> 
> * Check the items that apply and were done
> * Cross-out items that are not applicable (`~~`)
> * Leave unmarked the items that were not done for some reason.
> 
> So if the PR doesn't change the documentation (because it has not reason to), then please make it unmarked and crossed out.
> 
> This lets reviewers to know if something is not done on purpose, or because it is not applicable.
> 
> The changes themselves look fine (apart from some grammar issues in the commit messages).

Done, please take a look.

Lorak-mmk (approved):

PR #1180, opened 2025-01-27
errors: refactor errors on USE KEYSPACE execution path and clean up NewSessionError

muzarski:

## Motivation
`NewSessionError` has a lot of variants that can't even be constructed. This is all because of `From<QueryError> for NewSessionError` implementation. The only remaining execution path where we make use of this conversion is USE KEYSPACE request execution. (Previously, there was also `MetadataReader::read_metadata` that returned `QueryError`, but this was addressed in previous PR).

This is why we narrow the return error types in USE KEYSPACE execution path - from `QueryError` to new `UseKeyspaceError`.

### DIFF
--- a/scylla/src/network/connection_pool.rs
+++ b/scylla/src/network/connection_pool.rs
@@ -7,7 +7,9 @@ use super::connection::{
     ErrorReceiver, VerifiedKeyspaceName,
 };
 
-use crate::errors::{BrokenConnectionErrorKind, ConnectionError, ConnectionPoolError, QueryError};
+use crate::errors::{
+    BrokenConnectionErrorKind, ConnectionError, ConnectionPoolError, UseKeyspaceError,
+};
 use crate::routing::{Shard, ShardCount, Sharder};
 
 use crate::cluster::metadata::{PeerEndpoint, UntranslatedEndpoint};
@@ -317,7 +319,7 @@ impl NodeConnectionPool {
     pub(crate) async fn use_keyspace(
         &self,
         keyspace_name: VerifiedKeyspaceName,
-    ) -> Result<(), QueryError> {
+    ) -> Result<(), UseKeyspaceError> {
         let (response_sender, response_receiver) = tokio::sync::oneshot::channel();
 
         self.use_keyspace_request_sender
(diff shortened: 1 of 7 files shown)


muzarski:
group 0 test failure...

scylla/src/network/connection_pool.rs
```diff
@@ -1097,12 +1099,13 @@ impl PoolRefiller {
...
 
-            let use_keyspace_results: Vec<Result<(), QueryError>> = tokio::time::timeout(
+            let use_keyspace_results: Vec<Result<(), UseKeyspaceError>> = tokio::time::timeout(
                 connect_timeout,
                 futures::future::join_all(use_keyspace_futures),
             )
             .await
-            .map_err(|_| QueryError::TimeoutError)?;
+            // FIXME? Why are we using connect_timeout for use keyspace request???
+            .map_err(|_| UseKeyspaceError::RequestTimeout(connect_timeout))?;
```
wprzytula, line 1108: :thought_balloon: For me, it's quite intuitive, in the context that it's probably the most popular amongst the drivers to specify the keyspace to be used upon Session construction. Therefore, the `use_keyspace` is executed as part of `Session::connect`, essentially becoming part of the _connecting_. Thus, `connect_timeout` makes sense.

:question: Could you take a look at cpp-driver in this regard?

wprzytula (approved):
:partying_face: How wonderful to see `NewSessionError` finally free of those redundant variants!

muzarski:
Rebased on main

wprzytula (dismissed):

scylla/src/network/connection_pool.rs
muzarski: cpp-driver does not expose a `use_keyspace` method on session object. User has two ways to set current keyspace for the session:
1. choose the keyspace when connecting to the cluster (`cass_session_connect_keyspace[_n]`)
2. send "USE KEYSPACE <>" query manually

In the first case, the timeout would obviously be the connect timeout. In the second case, it would be the client request timeout. In rust-driver's case, we always choose the connect timeout. We could probably make the timeout for `USE KEYSPACE` requests configurable. This can be easily addressed after 1.0.0, though (if someone ever requests this). I wouldn't worry about it right now. Should I delete/adjust the FIXME comment in the code?

wprzytula: Let's adjust it to your above considerations.

muzarski:
v1.1: Adjusted the `FIXME` related to USE KEYSPACE request timeout

PR #1182, opened 2025-01-27
Rustls support

### LINKED ISSUE #293: rustls support

nrxus:
Fixes #293

This is based on the work and discussion on https://github.com/scylladb/scylla-rust-driver/pull/911

Movement on that PR seems to have stalled but it seemed like you all wanted to do this as part of the other breaking changes for v1.0 so I wanted to get this PR into your eyes prior to that. 

Main breaking changes are:
* Renamed `ssl` items that were meant to be generic over the TLS drivers o be `tls` instead as that is more accurate (although I know they are sometimes used interchangeably) (e.g., `ssl_context` to `tls_context`)
* Renamed `ssl` feature flag to be `openssl`
* The `cloud` feature flag no longer auto-enables `openssl`. This is done so users can choose if they want to enable `openssl` and/or `rustls`. We *could* add a check in the code that says that if `cloud` is enabled and there is no tls-related feature flag then hard compile error. It currently will already not compile but the error won't be as nice I think. 

Other caveats:

* For the `cloud` feature, if both `openssl` and `rustls` are enabled, the current code defaults to using `openssl` to keep it slightly more consistent with the old code. I *think* defaulting to `rustls` may provide a slightly better experience since it doesn't require any linking (dynamic nor static) but I wasn't sure what you all would prefer, it's easy enough to change. 

* For non-cloud, if both `openssl` and `rustls` are enabled, the user has a choice at runtime of what `TlsContext` to pass. 
* I do not have a cloud account for scylla so I have not manually tested the cloud changes. 
* The certificates in the test directory use a version that `rustls` does not support so they are not useful for the `rustls` example. I was unable to create new certs from the existing self-signed CA in the test directory because I do not know the passphrase. I have, however, tested that it works using my own certificates in a self-hosted 3-node Scylla instance in AWS that has TLS enabled for client-node communication. 
* The existing tests only test when using `openssl`, we could (and probably should) add tests for the `rustls` feature but that requires making the right certs that `rustls` can use. 

## Pre-review checklist

- [X] I have adjusted the documentation in `./docs/source/`.

### DIFF
--- a/scylla/src/cloud/config.rs
+++ b/scylla/src/cloud/config.rs
@@ -86,6 +86,32 @@ pub(crate) enum TlsInfo {
         key: openssl::pkey::PKey<openssl::pkey::Private>,
         cert: openssl::x509::X509,
     },
+    #[cfg(feature = "rustls")]
+    #[allow(dead_code)]
+    Rustls {
+        cert_chain: Vec<rustls::pki_types::CertificateDer<'static>>,
+        key: rustls::pki_types::PrivateKeyDer<'static>,
+    },
+}
+
+impl TlsInfo {
+    fn from_pem(cert: &[u8], key: &[u8]) -> Result<Self, TlsError> {
+        #[cfg(feature = "openssl")]
+        {
+            let cert = openssl::x509::X509::from_pem(cert)?;
+            let key = openssl::pkey::PKey::private_key_from_pem(key)?;
+            Ok(TlsInfo::OpenSsl { key, cert })
+        }
+
+        #[cfg(all(not(feature = "openssl"), feature = "rustls"))]
+        {
+            use rustls::pki_types::pem::PemObject;
+            let key = rustls::pki_types::PrivateKeyDer::from_pem_slice(key)?;
+            let cert_chain: Vec<_> = rustls::pki_types::CertificateDer::pem_slice_iter(cert)
+                .collect::<Result<_, _>>()?;
+            Ok(TlsInfo::Rustls { cert_chain, key })
+        }
+    }
 }
 
 impl AuthInfo {
@@ -109,6 +135,8 @@ impl AuthInfo {
 pub(crate) struct Datacenter {
     #[cfg(feature = "openssl")]
     openssl_ca: openssl::x509::X509,
+    #[cfg(feature = "rustls")]
+    rustls_ca: rustls::pki_types::CertificateDer<'static>,
     server: String,
     #[allow(unused)]
     tls_server_name: Option<String>,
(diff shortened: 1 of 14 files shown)


wprzytula:
@nrxus, please let us know the next time when you are planning to create a PR with nontrivial changes, so that we discuss it first. We've already started working on this issue; being in-sync could have brought benefit to both of us and avoid work duplication.

Apart from that, we're happy to see another contribution from you! I'm going to review it soon and compare it to our intended solution.

nrxus:
> @nrxus, please let us know the next time when you are planning to create a PR with nontrivial changes, so that we discuss it first. We've already started working on this issue; being in-sync could have brought benefit to both of us and avoid work duplication.
> 
> Apart from that, we're happy to see another contribution from you! I'm going to review it soon and compare it to our intended solution.

That's my bad, I totally meant to message on the old PR first but I was first waiting to see if I could even do this PR on my own or not (I am no TLS expert) and once it worked I got triggered happy on making the PR. I'll definitely contact you all first when I see a change becoming non-trivial.

Feel free to close this PR if taking it in is more work than just finishing what you all already had in mind (:

scylla/src/cloud/config.rs
```diff
@@ -74,21 +72,51 @@ impl CloudConfig {
...
+        #[cfg(all(not(feature = "openssl"), feature = "rustls"))]
+        {
+            use rustls::pki_types::pem::PemObject;
+            let key = rustls::pki_types::PrivateKeyDer::from_pem_slice(key)?;
+            let cert_chain: Vec<_> = rustls::pki_types::CertificateDer::pem_slice_iter(cert)
+                .collect::<Result<_, _>>()?;
+            Ok(TlsInfo::Rustls { cert_chain, key })
+        }
     }
+}
```
wprzytula, line 115: :thought_balloon:  This seems to work. In case of both `openssl` and `rustls` features being enabled, openssl is chosen. Fortunately, as this is the creation of `TlsInfo`, it's the only place where we have to do such logic based on features; in other places we're just matching on `TlsInfo`.

wprzytula:
@nrxus Will it be OK for you if I just cherry-pick your commits and rework them more or less for our implementation?

nrxus:
> @nrxus Will it be OK for you if I just cherry-pick your commits and rework them more or less for our implementation? This is a valuable work for us!

Yeah that's absolutely okay! I am glad it's useful, I look forward to all the nice changes in the next release

wprzytula:
@nrxus Could you please elaborate on the following:
```
don't ignore openssl connection error

otherwise the error shows up too late in a non-terminal way whereas
the error is terminal
```
?

What did you observe under what circumstances?

PR #1188, opened 2025-02-01
CI: cassandra workflow: bigger prepared cache

wprzytula:
Some tests (notably: `test_coalescing` - #864) exhibit flakiness when run against Cassandra, returning `DbError::Unprepared`. As an attempt to combat the flakiness, the prepared statements cache size is increased in the docker compose script.

The default is `MAX(10MiB, heap_size / 256)`. 100MB is the first guess; let's see if it's enough.

Fixes: #864
_(hopefully)_

## Pre-review checklist

### DIFF
--- a/test/cluster/cassandra/docker-compose.yml
+++ b/test/cluster/cassandra/docker-compose.yml
@@ -23,6 +23,11 @@ services:
       - CASSANDRA_BROADCAST_ADDRESS=172.42.0.2
       - HEAP_NEWSIZE=512M
       - MAX_HEAP_SIZE=2048M
+    # Only needed to set the custom prepared cache size in /etc/cassandra/cassandra.yaml.
+    entrypoint:
+      - /bin/bash
+      - -c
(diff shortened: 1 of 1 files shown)


muzarski:
Maybe better solution is to prepare the statement once instead of doing it multiple times in a loop in `test_coalescing` (based on our discussion on Slack)? It may be some bug in Cassandra, that occurs when user tries to prepare the same statement multiple times. Notice that there are no other cache issues apart from this test. The issue reproduces when running this test alone.

wprzytula:
> Maybe better solution is to prepare the statement once instead of doing it multiple times in a loop in `test_coalescing` (based on our discussion on Slack)? It may be some bug in Cassandra, that occurs when user tries to prepare the same statement multiple times. Notice that there are no other cache issues apart from this test. The issue reproduces when running this test alone.

I encourage you to try doing that. I'm leaving this PR for now as a backup if your proposed solution does not work.

Lorak-mmk:
I think we can close this since the test stopped failing. If the test starts to fail again, then we can deduplicate prepare calls, or increase the cache.

PR #1189, opened 2025-02-01
integration: direct node connectivity test

wprzytula:
## Problem

I noticed that we lacked a test that requires direct connectivity to all nodes in the cluster in order to pass. Other tests would often succeed even if only one node (the initial contact point) was directly reachable.

The issue has manifested in testing serverless cloud why working on rustls support: tests would pass even with address translation disabled...

### Side note 
An example of a test which has exercised direct connectivity to all nodes is `tablets.rs`, yet it could be enabled only for non-cloud case, as it uses the proxy.

## Solution
I wrote a test that iterates though all targets (pairs `(node, shard)`) and sends a request directly to them (using a load balancing policy that produces singleton query plans).

### Bonus
I extracted and refactored some utils from `tablets.rs` to `utils.rs`, so that they can be used by other tests.

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -19,6 +19,10 @@ use uuid::Uuid;
 use super::metadata::{Keyspace, Metadata, Strategy};
 use super::node::{Node, NodeRef};
 
+/// Represents the state of the cluster, including known nodes, keyspaces, and replica locator.
+///
+/// It is immutable after creation, and is replaced atomically upon a metadata refresh.
+/// Can be accessed through [Session::get_cluster_state()](crate::client::session::Session::get_cluster_state).
 #[derive(Clone)]
 pub struct ClusterState {
     /// All nodes known to be part of the cluster, accessible by their host ID.
--- a/scylla/tests/integration/load_balancing/tablets.rs
+++ b/scylla/tests/integration/load_balancing/tablets.rs
@@ -1,23 +1,18 @@
 use std::sync::Arc;
 
 use crate::utils::{
-    scylla_supports_tablets, setup_tracing, test_with_3_node_cluster, unique_keyspace_name,
-    PerformDDL,
+    scylla_supports_tablets, send_statement_everywhere, send_unprepared_query_everywhere,
+    setup_tracing, test_with_3_node_cluster, unique_keyspace_name, PerformDDL,
 };
 
 use futures::future::try_join_all;
 use futures::TryStreamExt;
 use itertools::Itertools;
 use scylla::client::session::Session;
-use scylla::cluster::ClusterState;
 use scylla::cluster::Node;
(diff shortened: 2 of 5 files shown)


scylla/tests/integration/tablets.rs
```diff
@@ -365,7 +366,7 @@ async fn test_tablet_feedback_not_sent_for_unprepared_queries() {
 
             // I expect Scylla to not send feedback for unprepared queries,
             // as such queries cannot be token-aware anyway
-            send_unprepared_query_everywhere(
+            execute_unprepared_statement_everywhere(
                 &session,
                 session.get_cluster_data().as_ref(),
                 &Query::new(format!("INSERT INTO {ks}.t (a, b, c) VALUES (1, 1, 'abc')")),
```
Lorak-mmk, line 372: 🤔 I don't really like "execute" used with unprepared statements, because EXECUTE is a CQL command to execute prepared statements. Would "send_unprepared_statement_everywhere" be an acceptable name?

wprzytula: I've been thinking that once we do the execution API refactor for 2.0, there will be only one method for all kinds of statements: `execute`. Do you agree? If so, then why not use it here?

Lorak-mmk: I've been thinking about it, and I think that even in new execution API we should keep separate methods for unprepared statements, prepared statements, and batches.
Why? The choice of the query type should be more concious, a single method makes this less obvious.
Also the interface would be simpler to learn: user would still have methods of the struct that accept simple types, and not traits which then user needs to research and learn what actually implements them.

What I think request execution refactor should be mostly about is enabling configuration of the request, meaning we can do `session.execute(something).with_timestamp(....).paging_iter().await` instead of having an exponential amount of methods.

wprzytula: OK, makes sense.

wprzytula: OTOH, the fact that the CQL protocol has some specific names for execution of prepared and unprepared statements **does not** imply that those names are a good fit for the names of high-level user-facing API functions. It's not intuitive at all for anyone not well-versed in the CQL protocol that "query" is related to unprepared statements and "execute" to prepared statements.

If in the future we were to have distinct names for execution of different types of statements, then I'd go for:
- `execute_unprepared()`,
- `execute_prepared()`,
- `execute_batch()`.

The names then make it perfectly clear that the action is _execution_ and the object is a particular kind of a statement.

Lorak-mmk: I agree that the names do not need to be the same (but in that case the documentation should clearly state which CQL command it corresponds to).
The proposed names are a bit too long for my taste.

wprzytula:
Addressed the comments.

PR #1190, opened 2025-02-02
Rustls support

wprzytula:
# ~~No cover letter, because GitHub ate it.~~

# The Big Picture

## Before

Before, `openssl` was the only supported TLS backend. (therefore, all TLS abstractions in the driver use the `Ssl` prefix). Written in C, it's a bit cumbersome to enable in the driver: one needs to install `openssl` library using the system package manager.

There were two distinct cases:
1. Global SSL context, set by a user for all connections to add encryption layer.
2. Serverless Cloud, where TLS is used mainly to set up SNI. Then, SSL context is different for every node (endpoint) node.

### SSL abstractions before
```
┌─←─ openssl::SslContext
│
│ gets wrapped in
│
↳ SslConfig (same for all connections)
    │               │
    │ gets cloned   ├─ either of these, depending on
    │               │  cloud or non-cloud scenario
    │  sets up SNI  │    
    ├─←─────────── CloudConfig (powered by openssl)
    │
    ↳ SslConfig (specific for the particular connection)
        │
        │ produces
        │
        ↳ openssl::Ssl (wrapper over TCP stream which adds encryption)
```

## After

The main goal of this PR is to add support for another TLS backend: `rustls`, which is written in Rust and hence easier to include in the project as a usual Rust dependency. As a bonus, its authors claims it's faster than `openssl`.

As we're going to support more than one TLS backend, we need to create abstractions that hide the implementation details, including which backend is used.
- **_-> done in this PR:_** For that, naming is adjusted from `Ssl*` abstractions specific for `openssl` (e.g., `SslContext`), to `Tls*` (e.g., `TlsContext`), which are made "generic" over the backend (actually, they are enums with `#[cfg(feature)]`-activated backend-specific variants).

Now, the two cases (global context and cloud) can both be done with `openssl` or `rustls` as a backend, so there are 4 combinations in total. As all combinations should be supported, new abstractions were designed to encapsulate the exact case and backend used. Thanks to that, the code is not only _not more_ complicated after this PR, but it's actually _less_ complicated! Comparison of the following schema with the previous one should convince you.

### TLS abstractions after
```
┌─←─ TlsContext (openssl::SslContext / rustls::ClientConfig)
│       │
│       ├─ either of these, depending on cloud or non-cloud scenario
│       │  
├─←─ CloudConfig (powered by either TLS backend)
│
│ gets wrapped in
│
↳ TlsProvider (same for all connections)
    │
    │ produces
    │
    ↳ TlsConfig (specific for the particular connection)
        │
        │ produces
        │
        ↳ Tls (wrapper over TCP stream which adds encryption)
```

# The Details

## Features
- introduced `"__tls"`, which gates the common `Tls*` abstractions.<details> <summary>Rationale:</summary> As those `Tls*` are enums that are empty (and this is problematic) when no TLS backend is enabled, an explicit compile error is raised in such scenario. Users should not enable this feature on their own; it's a dependency of all TLS-dependant features.
</details>

- renamed `"openssl"` -> `"openssl-010"` (#771)
- added `"rustls-023"` (#771)
- renamed `"cloud"` -> `"unstable-cloud"` <details><summary>Rationale:</summary> The Serverless Cloud is by no means a stable feature and it's likely that we'll have to break its API in the future. </details>

### Global TLS context case

- `SessionBuilder` now accepts `Option<impl Into<TlsContext>>`, allowing passing both `openssl::SslContext` and `rustls::ClientConfig`.
- Both backends can be supported by a single build of the driver, so users can even have distinct `Session`s that differ by the TLS backend use, which may be handy for comparative benchmarks between the backends and choosing the best fit.

### Serverless Cloud

Which backend is employed is now chosen based on the set of enabled features:
- if `"openssl-010"` feature is enabled, `openssl` is used for cloud;
- else if `"rustls-023"` feature is enabled, `rustls` is used for cloud;
- else a compile error is raised.

### Newly-introduced `Host*Config`

As @lorak-mmk correctly observed, we could leverage the type system to distinguish global pool/connection config from one that is already made specific for a particular endpoint. This way we avoid non-obvious mutation that happened in ConnectionConfig before this PR in the cloud case.

`Host{Pool,Connection}Config` is introduced to bring type safety and hide the necessary adjustments done in the cloud case (setting up SNI for a particular node) in the `ConnectionConfig` -> `HostConnectionConfig` conversion. In the future, there might be more differences between `*Config` and `Host*Config`, which will further justify this duplication.

### Refactored Serverless address translation

<details>
<summary> Use AddressTranslator instead of hand-crafted logic for serverless. </summary>

It seems that when I originally wrote serverless, for some reason I hand-crafted address translation logic specifically for serverless, making it fully separate from the standard `AddressTranslator` trait.

In this PR I successfully implement cloud address translation inside the standard `AddressTranslator` framework, decluttering `PoolRefiller` logic from cloud-related code. `CloudConfig` becomes an `AddressTranslator` itself, so it needn't be passed separately anymore. This further reduces code complication arising from cloud-related codepaths.
</details>

### Changes to `UntranslatedEndpoint`

`UntranslatedEndpoint` now:
- has getters and private fields (compared to `#[non_exhaustive]` public fields before) - for more flexibility in the future;
- is parametrised by a lifetime and stores datacenter and rack as `&str` instead of `String`, in order to elide allocations upon translation. `UntranslatedEndpoint` is constructed right before translation from `PeerEndpoint`, so now we avoid `String` allocation.

### Bumped `openssl` version
Older releases miss important security patches.

### Other fixes
Minor ones, you'll see them yourselves in the commits.

# Credits

Special thanks to @nrxus, who created a PR that I based mine on.
Thanks to @lorak-mmk and @nemosupremo for ideas and code inspiration. 

Fixes: #293
Closes: #1182
Closes: #911

## Pre-review checklist


scylla/src/network/connection.rs
```diff
@@ -369,7 +460,7 @@ pub(crate) struct ConnectionConfig {
     pub(crate) tcp_keepalive_interval: Option<Duration>,
     pub(crate) timestamp_generator: Option<Arc<dyn TimestampGenerator>>,
     #[cfg(feature = "__tls")]
-    pub(crate) tls_config: Option<TlsConfig>,
+    pub(crate) tls_provider: Option<TlsProvider>,
```
muzarski, line 257: Ok, now I understand the difference between `HostConnectionConfig` and `ConnectionConfig`.

scylla/src/network/connection.rs
```diff
@@ -388,14 +479,23 @@ pub(crate) struct ConnectionConfig {
 }
 
 impl ConnectionConfig {
-    pub(crate) fn to_host_connection_config(&self) -> HostConnectionConfig {
+    pub(crate) fn to_host_connection_config(
+        &self,
+        #[allow(unused)] endpoint: &UntranslatedEndpoint,
```
muzarski, line 276: Please add a comment why we need this allow. AFAIU, the endpoint is only used for cloud.

README.md
```diff
@@ -51,7 +51,7 @@ The driver supports the following:
 * Batch statements
 * Configurable load balancing policies
 * Driver-side metrics
-* TLS support - install openssl if you want to use it https://docs.rs/openssl/0.10.32/openssl/#automatic
+* TLS support. Supports either [OpenSSL](https://docs.rs/openssl/0.10.32/openssl/#automatic) or rustls
```
muzarski, line 54: We bumped the version in the driver, so probably link could be adjusted as well.

muzarski (commented):
I noticed that we do not do cargo checks when `unstable-cloud` and `rustls-x` features are enabled. I think we should do that. Probably should check the following combinations:
- `openssl-x`, `rustls-x`, `unstable-cloud`
- `openssl-x`, `unstable-cloud`,
- `rustls-x`, `unstable-cloud`
- `openssl-x`, `rustls-x`
- `openssl-x` (done in `tls.yml`, but it currently does not work)
- `rustls-x` (makes sense to put it in `tls.yml` once it's fixed )

Also, previously we would run a tls example in `tls.yml`. We should add a step to run tls-rustls example as well (once workflow is fixed).

scylla/src/network/connection_pool.rs
```diff
@@ -220,15 +229,15 @@ impl NodeConnectionPool {
                     ..
                 }) => (Some(host_id), address.into_inner(), datacenter.as_deref()),
             };
-            set_tls_config_for_scylla_cloud_host(host_id, dc, address, &mut pool_config.connection_config)
-                .unwrap_or_else(|err| warn!(
+
+            pool_config.connection_config.tls_config = make_tls_config_for_scylla_cloud_host(host_id, dc, address, cloud_config)
+                .inspect_err(|err| warn!(
```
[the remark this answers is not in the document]
Lorak-mmk: I guess that cargo check only verifies code that is actually enabled, and we don't do msrv cargo check with the feature that this line is guarded by?

scylla/Cargo.toml
```diff
@@ -39,6 +39,7 @@ full-serialization = [
     "num-bigint-04",
     "bigdecimal-04",
 ]
+__tls = []
```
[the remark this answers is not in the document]
wprzytula: It's the opposite: the driver throws an explicit compile-time error if user enables this feature without enabling any TLS backend.

scylla/src/cloud/config.rs
```diff
@@ -74,21 +72,36 @@ impl CloudConfig {
...
-    pub(crate) fn get_key(&self) -> &PKey<Private> {
-        &self.key
+#[derive(Debug)]
+pub(crate) enum TlsInfo {
+    #[cfg(feature = "openssl-010")]
+    OpenSsl010 {
+        key: openssl::pkey::PKey<openssl::pkey::Private>,
+        cert: openssl::x509::X509,
+    },
+}
```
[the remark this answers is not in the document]
wprzytula: Because the whole `cloud` module is hidden behind `unstable-cloud` feature, which depends on `__tls`.

scylla/src/network/connection.rs
```diff
@@ -212,35 +212,72 @@ struct TaskResponse {
...
+            }
+        }
+    }
+
+    #[cfg(feature = "openssl-010")]
+    impl From<openssl::error::ErrorStack> for TlsError {
+        fn from(error: openssl::error::ErrorStack) -> Self {
+            TlsError::OpenSsl010(error)
+        }
+    }
```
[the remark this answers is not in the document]
wprzytula: I took this code from @nrxus; will fix this.

scylla/src/network/connection.rs
```diff
@@ -387,6 +387,87 @@ pub(crate) struct ConnectionConfig {
...
+    pub(crate) authenticator: Option<Arc<dyn AuthenticatorProvider>>,
+    pub(crate) address_translator: Option<Arc<dyn AddressTranslator>>,
+    pub(crate) enable_write_coalescing: bool,
+
+    pub(crate) keepalive_interval: Option<Duration>,
+    pub(crate) keepalive_timeout: Option<Duration>,
+    pub(crate) tablet_sender: Option<mpsc::Sender<(TableSpec<'static>, RawTablet)>>,
+
+    pub(crate) identity: SelfIdentity<'static>,
+}
```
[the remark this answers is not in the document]
wprzytula: I will. It was all perfectly described in the cover letter (**which GitHub maliciously ate!**).

scylla/src/cloud/mod.rs
```diff
@@ -53,11 +40,15 @@ pub(crate) fn set_tls_config_for_scylla_cloud_host(
...
+        Ok(Some(TlsConfig::new_for_sni(
+            tls_context,
+            domain_name,
+            host_id,
+        )))
     } else {
         warn!("Datacenter {:?} of node {:?} with addr {} not described in cloud config. Proceeding without setting SNI for the node, which will most probably result in nonworking connections,.",
                dc, host_id, proxy_address);
+        // FIXME: Consider returning error here.
+        Ok(None)
```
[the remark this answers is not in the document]
wprzytula: You mean that it makes sense to return `Ok` or `Err`?

scylla/src/network/connection.rs
wprzytula: That's why we need this allow - because it's only used for cloud. Alternatively, I could sprinkle more `#[cfg(feature = "unstable_cloud")]` in various places, including at the calling spot. As it's just a reference being passed, it's such small overhead that I'd just let it be even without `unstable-cloud` feature being turned on. Especially that it makes sense from the abstract point of view that `HostConnectionConfig` needs to know its endpoint in order to be created from `ConnectionConfig`.

scylla/src/cloud/mod.rs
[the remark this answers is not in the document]
muzarski: To return `Err`. AFAIU, the only caller is `TlsProvider::make_tls_config`, which does not return a `Result`. It inspects the result of this function - if error is present, it's being logged (`warn!`), while the method returns `None`.

scylla/src/network/connection_pool.rs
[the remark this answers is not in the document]
muzarski: We fortunately do the check with the features: https://github.com/scylladb/scylla-rust-driver/blob/main/.github/workflows/rust.yml#L93-L96. It would be worrying if we didn't. This `inspect_err` is just replaced later in this PR, so we don't see the failure during the checks for the last commit.

scylla/src/cloud/mod.rs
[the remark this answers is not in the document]
wprzytula: The problem is, we have no suitable `TlsError` kind to return in this situation. Do you believe it's worthy introducing dedicated variants for this case, i.e., `TlsError::MissingNodeDatacenter` (when a DC is not given in node's entry in system tables) and `TlsError::MissingCloudConfigDatacenter` (when the node's DC is not described in the `CloudConfig`)?

wprzytula:
v1.1:
- addresses @muzarski's comments

wprzytula:
v1.2: rebased on main.

wprzytula:
1.2.1: (hopefully) fixed cloud code.

wprzytula:
The CI's SSL workflow is failing due to old ScyllaDB certificates used for testing. It has nothing to do with the code, hence the code is fully ready for review.

docs/source/connecting/tls.md
```diff
@@ -59,7 +59,7 @@ context_builder.set_verify(SslVerifyMode::PEER);
 
 let session: Session = SessionBuilder::new()
     .known_node("127.0.0.1:9142") // The the port is now 9142
-    .ssl_context(Some(context_builder.build()))
+    .tls_context(Some(context_builder.build()))
     .build()
     .await?;
```
Lorak-mmk, line 73: Commit: "codewide: wrap openssl usage with generic TLS wrappers "

❓ This may become clear later but a few lines later I see:
`.tls_context(Some(TlsContext::OpenSsl010(context_builder.build())))`

Why the difference?

scylla/src/client/session.rs
```diff
@@ -96,6 +94,21 @@ impl std::fmt::Debug for Session {
     }
 }
 
+#[cfg(feature = "__tls")]
+#[derive(Clone)] // Cheaply clonable - reference counted.
+#[non_exhaustive]
+pub enum TlsContext {
+    #[cfg(feature = "openssl-010")]
+    OpenSsl010(openssl::ssl::SslContext),
+}
```
Lorak-mmk: As I said during in-person conversation, I'd really want to remove this `__tls` feature so that `TlsContext` is not conditional - only its variants would be.

scylla/src/client/session.rs
```diff
@@ -806,6 +806,26 @@ impl Session {
...
+                            to CloudConfig being provided. This is certainly an API misuse - Cloud \
+                            may not be combined with user's own AddressTranslator."
+                    )
+                }
+
+                break 'translator Some(translator as Arc<dyn AddressTranslator>);
+            }
+
+            config.address_translator
+        };
```
Lorak-mmk: Please use `if else` instead of break with label and value.

scylla/src/client/session.rs
```diff
@@ -804,14 +806,36 @@ impl Session {
...
+
+                let provider = TlsProvider::new_cloud(cloud_config);
+                break 'provider Some(provider);
+            }
+            if let Some(tls_context) = config.tls_context {
+                let provider = TlsProvider::new_with_global_context(tls_context);
+                break 'provider Some(provider);
+            }
+            None
+        };
```
Lorak-mmk: Ditto

scylla/src/cluster/metadata.rs
```diff
@@ -171,14 +171,14 @@ impl Peer {
...
     pub(crate) host_id: Uuid,
     pub(crate) untranslated_address: SocketAddr,
-    pub(crate) datacenter: Option<String>,
-    pub(crate) rack: Option<String>,
+    pub(crate) datacenter: Option<&'a str>,
+    pub(crate) rack: Option<&'a str>,
 }
 
-impl UntranslatedPeer {
+impl UntranslatedPeer<'_> {
```
Lorak-mmk: It looks like address translation is the only use of UntranslatedPeer. Can you move it to `address_translator.rs`? I see no reason for it to be in metadata.

Cargo.lock.msrv
Lorak-mmk, line 1: Commit: "cloud: support rustls "

Commit message explains the logic based on features that chooses cloud backend, and invites to discussion, but does not provide a justification for such choice.

To me much more intuitive choice would  be to have this choice at runtime, just as we have with normal TLS provides after this PR. Is there a good reason to not do it this way? I want to avoid complexity in conditional compilation as much as possible, it complicates testing a lot.

.github/workflows/rust.yml
```diff
@@ -66,6 +66,22 @@ jobs:
...
+    # (openssl-x, rustls-x) is checked in tls.yml.
+
+    # Cloud-related feature sets.
+    - name: Cargo check with unstable-cloud and openssl-x features
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "unstable-cloud" --features "openssl-010"
+    - name: Cargo check with unstable-cloud and rustls-x features
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "unstable-cloud" --features "rustls-023" 
+    - name: Cargo check with unstable-cloud, openssl-x and rustls-x features
+      run: cargo check --all-targets --manifest-path "scylla/Cargo.toml" --features "unstable-cloud" --features "openssl-010" --features "rustls-023" 
+
```
Lorak-mmk, line 84: This is terrible, the static checks are already long :(

Lorak-mmk (changes requested):

.github/workflows/rust.yml
wprzytula: Any better ideas?

Lorak-mmk: Make the features independent (I think this is only problem with backend for cloud, right?), so that they are just adding new enum variants and implementations.
Then just test with no TLS features, and all of them.

docs/source/connecting/tls.md
wprzytula: I've made both lines the shorter version (without mentioning `TlsContext`).

scylla/src/client/session.rs
wprzytula: I wrote it this way because it appeared to me to be the most readable way. If you have a better way on your mind, please write it here.

scylla/src/client/session.rs
wprzytula: ditto

PR #1191, opened 2025-02-03
errors: Narrow return error type of Session::[query/execute]_iter

muzarski:

## Motivation

`NextPageError` and `NextRowError` were introduced specifically for `QueryPager` API. Currently (before this PR), the top-level Session methods `[execute/query]_iter()` return a `ExecutionError`. This is why `NextRowError` variant is present in `ExecutionError`. To remove that, we can narrow the return error type of these methods.

There is one subtle thing to pay attention to (and potentially discuss). `query_iter` method potentially prepares the statement (if values are not empty). `Session::prepare` currently returns a `ExecutionError`. This is why newly introduced `PagerExecutionError` now depends on `ExecutionError.` The question is whether we should narrow the return error type of `Session::prepare` to some new error type (`PrepareError`), or leave it as is.

## Pre-review checklist

### DIFF
--- a/scylla/src/errors.rs
+++ b/scylla/src/errors.rs
@@ -111,14 +111,6 @@ pub enum ExecutionError {
     /// 'USE KEYSPACE <>' request failed.
     #[error("'USE KEYSPACE <>' request failed: {0}")]
     UseKeyspaceError(#[from] UseKeyspaceError),
-
-    // TODO: This should not belong here, but it requires changes to error types
-    // returned in async iterator API. This should be handled in separate PR.
-    // The reason this needs to be included is that topology.rs makes use of iter API and returns ExecutionError.
-    // Once iter API is adjusted, we can then adjust errors returned by topology module (e.g. refactor MetadataError and not include it in ExecutionError).
-    /// An error occurred during async iteration over rows of result.
-    #[error("Failed to fetch next page of the result: {0}")]
-    NextPageError(#[from] NextPageError),
 }
 
 impl From<SerializationError> for ExecutionError {
--- a/scylla/src/errors.rs
+++ b/scylla/src/errors.rs
@@ -152,6 +152,23 @@ pub enum PrepareError {
     PreparedStatementIdsMismatch,
 }
 
+/// An error that occurred during construction of [`QueryPager`][crate::client::pager::QueryPager].
+#[derive(Error, Debug, Clone)]
+#[non_exhaustive]
+pub enum PagerExecutionError {
+    /// Failed to prepare the statement.
+    #[error("Failed to prepare the statement to be used by the pager: {0}")]
+    PrepareError(#[from] PrepareError),
+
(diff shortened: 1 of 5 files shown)


muzarski:
Flaky `test_coalescing` fails yet again (https://github.com/scylladb/scylla-rust-driver/issues/864).

wprzytula:
> There is one subtle thing to pay attention to (and potentially discuss). `query_iter` method potentially prepares the statement (if values are not empty). `Session::prepare` currently returns a `QueryError`. This is why newly introduced `PagerExecutionError` now depends on `QueryError.` The question is whether we should narrow the return error type of `Session::prepare` to some new error type (`PrepareError`), or leave it as is.

Now that we renamed `QueryError` to `ExecutionError`, it's even clearer that `PrepareError` should be introduced - because _preparing a statement_ is different from _executing a statement_.

wprzytula (dismissed):
LGTM, but see my comment about introducing `PrepareError`.

muzarski:
> Now that we renamed `QueryError` to `ExecutionError`, it's even clearer that `PrepareError` should be introduced - because _preparing a statement_ is different from _executing a statement_.

I 100% agree. New name clearly suggests that we should distinguish these two kinds of operations.

>     * `NextRowError` (no rows result comes from preparation).
BTW, `NextRowError` is not present in `ExecutionError` after this PR.

muzarski:
> LGTM, but see my comment about introducing `PrepareError`.

Opened a PR https://github.com/scylladb/scylla-rust-driver/pull/1193

scylla/src/errors.rs
```diff
@@ -151,6 +151,23 @@ impl From<response::Error> for ExecutionError {
...
 }
 
+/// An error that occurred during construction of [`QueryPager`][crate::client::pager::QueryPager].
+#[derive(Error, Debug, Clone)]
+#[non_exhaustive]
+pub enum PagerExecutionError {
+    /// Failed to prepare the statement.
+    #[error("Failed to prepare the statement to be used by the pager: {0}")]
+    PrepareError(#[from] ExecutionError),
+
```
Lorak-mmk, line 153: 🔧 I'm not a fan of this `#[from]`, it can be a footgun. `ExecutionError` is a general type, not restricted to statement preparation, so it is possible to accidently convert to this variant in other (non-prepare) paths, which will be difficult to spot.

Lorak-mmk (changes requested):

scylla/src/errors.rs
muzarski: Once rebased on #1193, it will be `#[from] PrepareError`

muzarski:
Rebased on main again (ssl CI fix, and removed legacy ser/deser api)

wprzytula (approved):

Lorak-mmk (approved):

Lorak-mmk (approved):

PR #1196, opened 2025-02-03
Replace `Node::is_down` with `Node::is_connected`

Lorak-mmk:
As found by @danielhe4rt `Node::is_down` method is unintuive. It is actually not really possible for the user to utulize this method
because status tracking with events is inherently unreliable. Instead we should treat node as up if we have some connections open to it.
This PR introduces `Node::is_connected` method that does just that, and un-pubs the old method.

## Pre-review checklist


scylla/src/network/connection_pool.rs
```diff
@@ -226,6 +226,17 @@ impl NodeConnectionPool {
...
     }
 
+    pub(crate) fn is_connected(&self) -> bool {
+        let maybe_conns = self.conns.load();
+        match maybe_conns.as_ref() {
+            MaybePoolConnections::Initializing => false,
+            MaybePoolConnections::Broken(_) => false,
+            // TODO: Is is possible for the pool to be "Ready", but for "_pool_connections"
+            // to be empty?
+            MaybePoolConnections::Ready(_pool_connections) => true,
```
wprzytula, line 236: :question: Is it?

wprzytula:
```rust
enum MaybePoolConnections {
    // The pool is being filled for the first time
    Initializing,

    // The pool is empty because either initial filling failed or all connections
    // became broken; will be asynchronously refilled. Contains an error
    // from the last connection attempt.
    Broken(ConnectionError),

    // The pool has some connections which are usable (or will be removed soon)
    Ready(PoolConnections),
}
```
From this enum I understand that if it's not buggy, then `Ready` implies some connections in the pool.

scylla/src/cluster/worker.rs
```diff
@@ -291,6 +291,16 @@ impl ClusterWorker {
...
+                                // What I mean by that?
+                                // - Don't store the status at all.
+                                // - When receiving down event, and the driver still sees the node
+                                //   as connected, then try to send a keepalive query to its connections.
+                                // - When receiving up event, and we have no connections to the node,
+                                //   then try to open new connections.
+
                                 match status {
                                     StatusChangeEvent::Down(addr) => self.change_node_down_marker(addr, true),
                                     StatusChangeEvent::Up(addr) => self.change_node_down_marker(addr, false),
```
wprzytula, line 306: :+1:

wprzytula (dismissed):

scylla/src/network/connection_pool.rs
Lorak-mmk: Edited the comment.

Lorak-mmk:
Rebased on main

muzarski (approved):

wprzytula (approved):

Issue #1167, opened 2025-01-09
Remove legacy serialization and deserialization frameworks


Lorak-mmk:
We must do this before 1.0 because this is a breaking change.

PR #1198, opened 2025-02-04
scylla_cql cleanup after legacy APIs removal

Lorak-mmk:
After removing the legacy serialization / deserialization APIs it became clear that some things in scylla_cql are not placed where they should be.
This PR tries to address it.

### Ser / deser tests

They are placed in several places, making it hard to follow what tests we actually have.
We should strive to put them together.
Where are the tests and how does this PR change it?
- In `deserialize` module in files separate from implementation (`row_tests.rs`, `value_tests.rs`). This is good and so left unchanged.
- In `serialize` module inside the files with implementation. I extracted those tests to separate files (`row_tests.rs`, `value_tests.rs`) to reduce file size and be more similar to `deserialize` module.
- In `frame/value_tests.rs` file. Those were some of the serialization tests for both values, rows, and batches. I moved them to relevant files in `serialize` module.
- In `frame/result/response.rs` file. Those were some of the deserialization tests for `CqlValue`. I moved them to `deserialize/value_tests.rs` file.

### `value` module

This module in `scylla_cql` is placed in `frame` module and contained multiple things:
1. Old serialization APIs (`Value` and `ValueList` traits and their implementations)
2. Our ser/deser related types like `CqlValue`, `CqlTimeuuid` etc

After removing (1), and removing `cql_to_rust` module which was also in `frame`, I find it a bit weird
for this module to be placed in `frame`. Why? Previously types like `CqlValue` and `CqlTimeuuid`
were strongly tied to frame serialization / deserialization, because they were used in CqlValue,
which was always used when deserializing result values.
With the new API this is no longer the case, so there is no more reason to keep those types in `frame` module.

I moved this file into the root folder, so `value` is now a top-level module.

### CqlValue and Row

Those were placed in `scylla-cql/frame/response/result.rs`. As described in the previous point, this location made
sense with the old deserialization API because `CqlValue` and `Row` were strongly tied to frame deserialization.

With the new API they are not anymore - they are normal types than can be serialized / deserialized, and are kept
mostly to address dynamic use cases. With this change, `result.rs` no longer seems like a good place for them.

I moved them to the newly created `value.rs` top-level module, so it is placed close to  `CqlTimeuuid` and other such types.
This may be a bit of a weird place for a `Row` struct, but I don't have any better ideas.

Fixes https://github.com/scylladb/scylla-rust-driver/issues/1167

## Pre-review checklist


muzarski (approved):
An idea: maybe extend `test_types_imports` test? Currently we only test for `_scylla::value::CqlValue`, but now that we explicitly re-export types from this module it makes sense to test imports for them as well.

Lorak-mmk:
Makes sense, I'll do it soon

Lorak-mmk:
Added the rest of the types to the test

Lorak-mmk:
Rebased on main

scylla-cql/src/frame/response/result.rs
```diff
@@ -301,13 +301,13 @@ impl CqlValue {
...
+    pub(crate) fn as_naive_date_04(&self) -> Option<chrono_04::NaiveDate> {
         self.as_cql_date().and_then(|date| date.try_into().ok())
     }
 
     #[cfg(test)]
     #[cfg(feature = "time-03")]
-    fn as_date_03(&self) -> Option<time_03::Date> {
+    pub(crate) fn as_date_03(&self) -> Option<time_03::Date> {
         self.as_cql_date().and_then(|date| date.try_into().ok())
     }
```
wprzytula: :thought_balloon: Wouldn't we like to have these functions available as `pub` out of test context, too? Such API might be useful e.g. in JS Driver.

scylla-cql/src/value.rs
```diff
@@ -779,3 +791,461 @@ pub struct CqlDuration {
     pub days: i32,
     pub nanoseconds: i64,
 }
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum CqlValue {
```
wprzytula, line 796: :camping: Missing docstring.

scylla-cql/src/value.rs
```diff
@@ -779,3 +791,461 @@ pub struct CqlDuration {
...
+    Ascii(String),
+    Boolean(bool),
+    Blob(Vec<u8>),
+    Counter(Counter),
+    Decimal(CqlDecimal),
```
wprzytula, line 801: :memo: Variants are missing docstrings, too, but these should be easily searchable in the ScyllaDB documentation.

scylla-cql/src/value.rs
```diff
@@ -779,3 +791,461 @@ pub struct CqlDuration {
...
+    #[cfg(feature = "chrono-04")]
+    pub(crate) fn as_datetime_04(&self) -> Option<chrono_04::DateTime<chrono_04::Utc>> {
+        self.as_cql_timestamp().and_then(|ts| ts.try_into().ok())
+    }
+
+    #[cfg(test)]
+    #[cfg(feature = "time-03")]
+    pub(crate) fn as_offset_date_time_03(&self) -> Option<time_03::OffsetDateTime> {
+        self.as_cql_timestamp().and_then(|ts| ts.try_into().ok())
+    }
```
wprzytula, line 880: :camping: I believe all these should be made `pub`.

scylla-cql/src/value.rs
```diff
@@ -779,3 +791,461 @@ pub struct CqlDuration {
...
+            _ => None,
+        }
+    }
+
+    pub fn into_pair_vec(self) -> Option<Vec<(CqlValue, CqlValue)>> {
+        match self {
+            Self::Map(s) => Some(s),
+            _ => None,
+        }
+    }
```
wprzytula, line 1055: :thought_balloon: :camping: Wouldn't we prefer to return `Result<CastT, Self>`, not to lose `self` in case of conversion failure?

scylla-cql/src/value.rs
```diff
@@ -779,3 +791,461 @@ pub struct CqlDuration {
...
+                .collect::<StdResult<_, _>>()?;
+            CqlValue::Tuple(t)
+        }
+    })
+}
+
+#[derive(Debug, Default, PartialEq)]
+pub struct Row {
+    pub columns: Vec<Option<CqlValue>>,
+}
```
wprzytula, line 1251: :camping: Missing docstring.

wprzytula (approved):

scylla-cql/src/value.rs
Lorak-mmk: I think this is the only comment that we can't really address after 1.0 (which doesn't mean we have to address it now - we could to that for 2.0). Would you like to open an issue?

PR #1203, opened 2025-02-06
tests: implement ccm integration

dkropachev:
CCM is cli tool to ease scylla/cassandra/dse cluster management. 
Using it in integration suit will help to expand test coverage.

## Pre-review checklist


scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+                )))
+            }
+        }
+    }
+
+    async fn stream_reader<T>(
+        stream: T,
+        writer: Arc<Mutex<File>>,
+        prefix: String,
+        mut buffer: Option<&mut Vec<u8>>,
```
wprzytula, line 199:
```suggestion
        buffer: Option<&mut Vec<u8>>,
```

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+    ) where
+        T: tokio::io::AsyncRead + Unpin + Send + 'static,
+    {
+        let reader = BufReader::new(stream);
+        let mut lines = reader.lines();
+        while let Some(line) = tokio::select! {
+            line = lines.next_line() => line.unwrap_or(None),
+        } {
```
wprzytula, line 207: `tokio::select!` is redundant here, as there is only one future to select from.

wprzytula:
```suggestion
        while let Some(line) = lines.next_line().ok().flatten() {
```

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+        self.file = Some(Arc::new(Mutex::new(file)));
+        Ok(())
+    }
+
+    pub(crate) async fn run_command(
+        &self,
+        command: &str,
+        args: &[&str],
+        opts: Option<RunOptions>,
+    ) -> Result<ExitStatus, Error> {
```
wprzytula, line 61: :wrench: This function is too long. Please extract helper functions from it, so that it's easier to follow its logic.

wprzytula (changes requested):
This is the first round of the review.
There are for sure more issues, but I'm unable to grasp them until I gain the high-level of the approach, which I failed to achieve so far due to no comments nor commit messages nor the cover letter.

scylla/Cargo.toml
```diff
@@ -56,7 +55,6 @@ tokio = { version = "1.34", features = [
     "rt",
     "macros",
 ] }
-snap = "1.0"
 uuid = { version = "1.0", features = ["v4"] }
```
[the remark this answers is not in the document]
dkropachev: same, not used.

scylla/Cargo.toml
```diff
@@ -67,7 +65,6 @@ openssl = { version = "0.10.32", optional = true }
 tokio-openssl = { version = "0.6.1", optional = true }
 arc-swap = "1.3.0"
 dashmap = "5.2"
-lz4_flex = { version = "0.11.1" }
 smallvec = "1.8.0"
```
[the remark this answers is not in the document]
dkropachev: again, don't see it is used

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+            ..Default::default()
+        })
+    };
+}
+
+#[derive(Default, Debug)]
+pub(crate) struct RunOptions {
+    pub env: HashMap<String, String>,
+    pub allow_failure: Option<bool>,
+}
```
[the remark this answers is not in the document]
dkropachev: why not, it will be limited by visilibility of the struct

scylla/Cargo.toml
```diff
@@ -43,7 +43,6 @@ full-serialization = [
 [dependencies]
 scylla-macros = { version = "0.7.0", path = "../scylla-macros" }
 scylla-cql = { version = "0.4.0", path = "../scylla-cql" }
-byteorder = "1.3.4"
 bytes = "1.0.1"
```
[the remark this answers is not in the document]
wprzytula: In such case, deletion of it should absolutely happen in a separate commit with a message that justifies this deletion. Such things mustn't sneak into commits that add new functionality!

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
+use tokio::process::Command;
+use tokio::sync::{Mutex};
+use tokio::task;
+
+pub(crate) struct LoggedCmd {
+    log_file: String,
+    file: Option<Arc<Mutex<File>>>,
+    run_id: AtomicI32,
+}
```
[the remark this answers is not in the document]
dkropachev: Not always, but rather give that option

scylla/Cargo.toml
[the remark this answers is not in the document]
wprzytula: This is a leftover from when I extracted `scylla_cql`.

scylla/Cargo.toml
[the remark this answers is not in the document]
wprzytula: This is a leftover from when I extracted `scylla_cql`.

scylla/Cargo.toml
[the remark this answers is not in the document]
wprzytula: This is a leftover from when I extracted `scylla_cql`.

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+        self.log_file = file_name.clone();
+        let file = OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(self.log_file.as_str())
+            .await
+            .context(format!("failed to open ccm log file {}", self.log_file))?;
+        self.file = Some(Arc::new(Mutex::new(file)));
+        Ok(())
+    }
```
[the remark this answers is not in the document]
dkropachev: As of now logging into file is optional, another option is to write same thing into stderr/stdout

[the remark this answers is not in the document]
dkropachev: Problem is this, file needs to be tokio::File, which means that method that opens it up should be `async`.
If you are ok with having constructor async, ok. 
Otherwise we can have method `async fn start() -> Result<(), Error>` that does that.
WDYT ?

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    pub(crate) async fn set_log_file(&mut self, file_name: String) -> Result<(), Error> {
+        self.log_file = file_name.clone();
+        let file = OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(self.log_file.as_str())
+            .await
+            .context(format!("failed to open ccm log file {}", self.log_file))?;
```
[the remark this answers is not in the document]
dkropachev: Could you please clarify that ?
Idea here is to preserve original error and add additional information to it.

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,344 @@
...
+        args: &[&str],
+        opts: Option<RunOptions>,
+    ) -> Result<ExitStatus, Error> {
+        let run_id = self
+            .run_id
+            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        let mut cmd = Command::new(command);
+        cmd.args(args).stdout(Stdio::piped()).stderr(Stdio::piped());
+
+        let writer = self.file.as_ref().unwrap();
```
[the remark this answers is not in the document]
dkropachev: True, I left it for later.

scylla/tests/ccm-integration/logged_cli.rs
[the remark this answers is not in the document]
wprzytula: Yes, let's have an async constructor. This is normal in Rust.

scylla/tests/ccm-integration/logged_cli.rs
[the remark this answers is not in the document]
wprzytula: The problem is, `context` evaluates its argument eagerly, so it allocates the error string even if no error appeared. In contrast, `with_context` only calls the function passed as an argument if an error appeared.

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,676 @@
...
+                    )));
+                }
+            }
+            Err(err) => match err.kind() {
+                std::io::ErrorKind::NotFound => {
+                    tokio::fs::create_dir_all(opts.config_dir().as_str())
+                        .await
+                        .context(
+                            format! {"failed to create root directory {}", opts.config_dir()},
+                        )?;
```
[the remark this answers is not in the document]
wprzytula:
```suggestion
                        .context(
                            format_args! {"failed to create root directory {}", opts.config_dir()},
                        )?;
```

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,676 @@
...
+                    )));
+                }
+            },
+        }
+
+        let mut lcmd = LoggedCmd::new();
+        let ccm_log_file = opts.ccm_log_file();
+        lcmd.set_log_file(ccm_log_file.clone())
+            .await
+            .context(format!("failed to set log file {}", ccm_log_file))?;
```
wprzytula, line 497:
```suggestion
            .context(format_args!("failed to set log file {}", ccm_log_file))?;
```

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,676 @@
...
+    pub(crate) async fn init(&mut self) -> Result<(), Error> {
+        let cluster_dir = PathBuf::from(self.opts.cluster_dir());
+        let config_dir = self.opts.config_dir();
+        if cluster_dir.exists() {
+            tokio::fs::remove_dir_all(&cluster_dir)
+                .await
+                .context(format!(
+                    "failed to remove config dir {}",
+                    cluster_dir.to_str().unwrap()
+                ))?;
```
wprzytula, line 524:
```suggestion
                .context(format_args!(
                    "failed to remove config dir {}",
                    cluster_dir.to_str().unwrap()
                ))?;
```

wprzytula (changes requested):

scylla/tests/ccm-integration/test_example.rs
```diff
@@ -0,0 +1,116 @@
...
+        let rows: Vec<String> = session
+            .query_unpaged("select data_center from system.local", &[])
+            .await
+            .context("failed to execute query")?
+            .into_rows_result()
+            .context("failed to get rows")?
+            .rows::<(String,)>()
+            .context("failed to deserialize rows")?
+            .map(|row| row.unwrap().0)
+            .collect();
```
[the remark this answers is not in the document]
muzarski: This `map` maps a single value of the iterator. In this case, the value type we iterate over is `Result<T, DeserializationError>` where T is some tyle we deserialize into. This unwrap inside map can panic if value is Err (i.e. deserialization failed for some reason).

I believe that in our case, if the deserialization of a single value fails, we want to fail the whole operation. For this, you can use https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect. This will collect the iterator of `Result<T, E>` values into `Result<Vec<T>, E>`.

scylla/tests/ccm-integration/logged_cli.rs
wprzytula: Try experimenting with matching on `buffer.as_ref()` or `buffer.as_mut()` instead of matching on plain `buffer`. Might solve the issue.

scylla/tests/ccm-integration/cluster.rs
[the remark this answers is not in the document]
wprzytula: In such case, let's try:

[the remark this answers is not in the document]
dkropachev: done

scylla/tests/ccm-integration/cluster.rs
dkropachev: fixed.

scylla/tests/ccm-integration/cluster.rs
dkropachev: fixed with `with_context`

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,676 @@
...
+                        .await
+                        .context(
+                            format! {"failed to create root directory {}", opts.config_dir()},
+                        )?;
+                }
+                _ => {
+                    return Err(Error::from(err).context(format!(
+                        "failed to create root directory {}",
+                        opts.config_dir()
+                    )));
```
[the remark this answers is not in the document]
dkropachev: this guy does not have `.with_context`

scylla/tests/ccm-integration/logged_cli.rs
dkropachev: `as_mut` requires `buffer` to be mut, `as_ref` goes wrong way, but `match buffer` worked well, take a look

scylla/tests/ccm-integration/logged_cli.rs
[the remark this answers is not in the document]
dkropachev: fixed

scylla/tests/ccm-integration/test_example.rs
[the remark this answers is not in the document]
dkropachev: addressed, please take a look.

scylla/tests/ccm-integration/test_example.rs
```diff
@@ -0,0 +1,116 @@
...
+        println!("{:?}", rows);
+        Ok(())
+    })
```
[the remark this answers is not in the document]
Lorak-mmk: `            .map(|row| row.unwrap().0)` - unwrap panics on error.

scylla/tests/ccm-integration/cluster.rs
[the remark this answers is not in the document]
wprzytula: My bad: in this particular case it's OK to use `context`, because we are returning error unconditionally, so there's no way to avoid allocations anyway.

scylla/tests/ccm-integration/test_example.rs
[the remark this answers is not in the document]
dkropachev: Thanks, it is fixed

scylla/tests/ccm-integration/test_example.rs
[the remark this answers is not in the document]
muzarski: Now this indeed works as expected. However, using `try_fold` is an overkill. This can still be simplified. I see that the issue arises from the fact that `String` does not implement `DeserializeRow` (while `(String,)` does). We want to flatten the row type from a 1-value tuple to just this value (to have `Vec<String>`, instead of `Vec<(String,)>`). Here is simplified code (I also put some comments to explain what's going on - do not include them in the fix):
```rust
        let rows = session
            .query_unpaged("select data_center from system.local", &[])
            .await
            .context("failed to execute query")?
            .into_rows_result()
            .context("failed to get rows")?
            .rows::<(String,)>()
            .context("failed to deserialize rows")?
            // This is Iterator::map ( res is Result<(String,), DeserializationError> )
            .map(|res| {
                // This is Result::map ( row is (String,) )
                // It maps Result<(String,), _> to Result<String, _>.
                // If res is Ok(row), then we map it to Ok(row.0).
                // Err is preserved.
                res.map(|row| row.0)
            })
            // This is Iterator::collect with type annotation (turbofish).
            // The type needs to implement FromIterator<Result<String, DeserializationError>>.
            // Result<Vec<String>, DeserializationError> indeed implements it.
            // The value will be Ok(vec) if all row deserializations suceeded, otherwise
            // it will be Err(e), where e is the error of the first failed deserialization.
            .collect::<Result<Vec<_>, _>>()?;
```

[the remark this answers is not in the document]
muzarski: And here you can read more about how `Result<_, _>` implements `FromIterator<_>`: https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromIterator%3CResult%3CA,+E%3E%3E-for-Result%3CV,+E%3E

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,832 @@
...
+
+pub(crate) enum NodeStartOption {
+    // Don't wait node to start
+    NoWait,
+    // Wait till other nodes recognize started node
+    WaitOtherNotice,
+    // Wait till started node report that client port is opened and operational
+    WaitForBinaryProto,
+}
+
```
Lorak-mmk, line 145: My first thought was that it is nice that you translated CCM options to enum, because Rust Driver contributor shouldn't have to know anything about CCM, but now I see that it is not that obvious. Each variant corresponds to a command line option, and `Node::start` accepts `Option<&[NodeStartOption]>`.
So this requires the user of those utilities to understand how CCM handles repeated / conflicting command line arguments. What happends if I pass `Some(&[NodeStartOption::NoWait, NodeStartOption::WaitOtherNotice])`? It is not obvious to me after reading ccm's help and briefly even source code.
I think this should be a struct, with fields `wait_other_notice: bool` and `wait_for_binary_proto: bool`. This way the user specifices what they want to wait for, and only the ccm wrapper needs to know how to translate this to command line options.

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,832 @@
...
+pub(crate) enum NodeStopOption {
+    // Don't wait node to properly stop
+    NoWait,
+    // Force-terminate node with `kill -9`
+    Kill,
+}
```
Lorak-mmk, line 151: As above, this would be better expressed as a struct with boolean fields.

scylla/tests/ccm-integration/test_example.rs
```diff
@@ -0,0 +1,128 @@
...
+
+lazy_static! {
+    static ref CLUSTER_VERSION: String =
+        std::env::var("SCYLLA_TEST_CLUSTER").unwrap_or("release:6.2.2".to_string());
+    static ref TEST_KEEP_CLUSTER_ON_FAILURE: bool = !std::env::var("TEST_KEEP_CLUSTER_ON_FAILURE")
+        .unwrap_or("".to_string())
+        .parse::<bool>()
+        .unwrap_or(false);
+}
+
```
Lorak-mmk, line 19: The files with tests should not interact with environment at all. Cluster version selection (and thus reading the env var) should be done by the ccm utils.

Preferably, all interaction with env variables should be done in one place in the code.

scylla/tests/ccm-integration/test_example.rs
```diff
@@ -0,0 +1,128 @@
...
+    let cluster_arc = cb().await;
+    {
+        let res = test_body(cluster_arc.clone()).await;
+        if res.is_err() && *TEST_KEEP_CLUSTER_ON_FAILURE {
+            println!("Test failed, keep cluster alive, TEST_KEEP_CLUSTER_ON_FAILURE=true");
+            cluster_arc.write().await.set_keep_on_drop(true);
+        }
+    }
+}
+
```
Lorak-mmk, line 60: This is also something that should not be present in the test code, but rather in a centralized location.

scylla/tests/ccm-integration/test_example.rs
```diff
@@ -0,0 +1,128 @@
...
+async fn get_session(cluster: &Cluster) -> Session {
+    let endpoints = cluster.nodes.get_contact_endpoints().await;
+    SessionBuilder::new()
+        .known_nodes(endpoints)
+        .build()
+        .await
+        .unwrap()
+}
```
Lorak-mmk, line 68: Two issues with this:
- This should not be in test file, but rather a method on `Cluster`
- This should return a `SessionBuilder` so that a test can customize the session further. See https://github.com/scylladb/scylla-rust-driver/blob/bceac6aaa852d51699b73820fc5317d75e3e21d4/scylla/tests/integration/utils.rs#L137-L167

scylla/tests/ccm-integration/test_example.rs
```diff
@@ -0,0 +1,128 @@
...
+    T: FnOnce(Arc<RwLock<Cluster>>) -> TFut,
+    TFut: std::future::Future<Output = Result<(), Error>>,
+{
+    let cluster_arc = cb().await;
+    {
+        let res = test_body(cluster_arc.clone()).await;
+        if res.is_err() && *TEST_KEEP_CLUSTER_ON_FAILURE {
+            println!("Test failed, keep cluster alive, TEST_KEEP_CLUSTER_ON_FAILURE=true");
+            cluster_arc.write().await.set_keep_on_drop(true);
+        }
```
Lorak-mmk, line 57: Tests usually fail by panicking, not by returning an error. This should catch the panic, do this, and rethrow it.

scylla/tests/ccm-integration/node_config.rs
```diff
@@ -0,0 +1,365 @@
...
+pub(crate) enum NodeConfig {
+    Null,
+    Bool(bool),
+    Int(i64),
+    Float(f64),
+    String(String),
+    List(Vec<NodeConfig>),
+    Map(HashMap<String, NodeConfig>),
+}
+
```
Lorak-mmk, line 16: This is basically the same as `serde_yaml::Value`. Why did you need to do this, instead of having a newtype `NodeConfig(Value)`?

scylla/tests/ccm-integration/node_config.rs
```diff
@@ -0,0 +1,365 @@
...
+        let mut val = HashMap::new();
+        val.insert(
+            "endpoint_snitch".to_string(),
+            NodeConfig::String(
+                "org.apache.cassandra.locator.GossipingPropertyFileSnitch".to_owned(),
+            ),
+        );
+        NodeConfig::Map(val)
+    }
+}
```
Lorak-mmk, line 28: Its not obvious to me why this specific value is the only one that must be in the default config. Please justify it in the comment.

Lorak-mmk: Btw is there a reason for making this a dynamic type instead of a struct with fields corresponding to config values, besides the fact that it would probably be more work?

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,832 @@
...
+impl Default for ClusterOptions {
+    fn default() -> Self {
+        ClusterOptions {
+            name: "".to_string(),
+            db_type: DBType::SCYLLA,
+            version: "".to_string(),
+            ip_prefix: NetPrefix::empty(),
+            root_dir: "/tmp/ccm".to_string(),
+            nodes: Vec::new(),
+            smp: DEFAULT_SMP,
```
Lorak-mmk, line 81: Putting it in `/tmp/ccm` will prevent running multiple instances of tests simultaneously.
Either put in in some folder in the repo (probably the better option), or (my less preferred option) use some lib like https://docs.rs/mktemp/latest/mktemp/ to make a unique folder in tmp.

scylla/tests/ccm-integration/logged_cmd.rs
```diff
@@ -0,0 +1,349 @@
...
+use tokio::sync::Mutex;
+use tokio::task;
+
+// A simple abstraction to run commands, log command, exit code its stderr, stdout to a file
+//  and optionally to own stderr/stdout
+// It should allow to run multiple commands in parallel
+pub(crate) struct LoggedCmd {
+    file: Arc<Mutex<File>>,
+    run_id: AtomicI32,
+}
```
Lorak-mmk, line 25: The comment explains what does this struct do, but not why.
Why would we need to log ccm output to a file? In other words why is it not enough to log it to normal output, like other stuff in tests do? Note that if you write something to stdoud/stderr it will be properly caught, and printed in the test fails.
To demonstrate, I wrote a tiny integration test:
```rust
#[tokio::test]
async fn test_output() {
    setup_tracing();
    debug!("Some debug output");
    warn!("Some warn output");
    println!("Some stdout output");
    eprintln!("Some stderr output");
    panic!("To see if the test fails");
}
```

If I run it with `cargo test test_output` I get:
image

And if I enable tracing with `RUST_LOG=trace cargo test test_output` I get the tracing output too:
image


Is it for some reason not enough to just normally log ccm commands?

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,832 @@
...
+                    continue;
+                }
+                let ip_prefix: IpAddr = Ipv4Addr::new(127, a, b, 0).into();
+                if !used_ips.contains(&ip_prefix) {
+                    return Ok(ip_prefix.into());
+                }
+            }
+        }
+        Err(Error::msg("All ip ranges are busy"))
+    }
```
Lorak-mmk, line 418: Its absurd that we have to do this, it should be ccm's job to find a free ips for the nodes.

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,832 @@
...
+    pub(crate) version: String,
+    // CCM allocates node ip addresses based on this prefix:
+    // if ip_prefix = `127.0.1.`, then `node1` address is `127.0.1.1`, `node2` address is `127.0.1.2`
+    pub(crate) ip_prefix: NetPrefix,
+    // CCM does not allow to have one active cluster within one config directory
+    // To have more than two active CCM cluster at the same time we isolate each cluster into separate
+    // config director, each config directory is created in `root_dir`.
+    // Example: root_dir = `/tmp/ccm`, config directory for cluster `test_cluster_1` is going be `/tmp/ccm/test_cluster_1`
+    //  and cluster directory (that is created and managed  by CCM) for this cluster is going to be `/tmp/ccm/test_cluster_1/test_cluster_1`
+    pub(crate) root_dir: String,
```
Lorak-mmk, line 39: Again, ccm should be changed so that the commands can specify which cluster to operate on.

scylla/tests/ccm-integration/cluster.rs
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    fn get_ccm_env(&self) -> HashMap<String, String> {
+        let mut env: HashMap<String, String> = HashMap::new();
+        env.insert(
+            "SCYLLA_EXT_OPTS".to_string(),
+            format!("--smp={} --memory={}M", self.opts.smp, self.opts.memory),
+        );
+        env
+    }
```
Lorak-mmk, line 594: I am unable to imagine why would someone think it is a good idea to pass some options throught env instead normal arguments.....

Lorak-mmk (changes requested):
By no means a thorough review, just posting a few things that I noticed during a brief read.

Few other things:
- The utils for tests should be put in some folder inside `ccm-integration`, instead of being on the same level as the test files.
- Before merging, we will need to update `Makefile` and `CONTRIBUTING.md`.

muzarski (commented):
I think that we should firstly focus on having the CI pass. Currently, the code does not even compile. Once we fix this, it will be much easier to test the code locally (which will be helpful in understanding the logic). I checked out on this PR locally, to play around with new API/run some tests but I was not able to do so, because it does not compile.

I suggest fixing the issues listed by: `cargo clippy --verbose --all-features --all-targets -- -D warnings`. Or if you want to not only do the checks, but run the tests as well, I suggest using `make ci`.

scylla/tests/ccm-integration/logged_cli.rs
```diff
@@ -0,0 +1,349 @@
...
+            RunOptions::AllowFailure => {
+                allow_failure = true;
+            }
+            RunOptions::AllowFailureWithEnv(new_env) => {
+                env.extend(new_env);
+                allow_failure = true;
+            }
+        }
+        (env, allow_failure)
+    }
```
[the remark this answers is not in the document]
dkropachev: fixed

scylla/tests/ccm-integration/test_example.rs
```diff
@@ -0,0 +1,128 @@
...
+    SessionBuilder::new()
+        .known_nodes(endpoints)
+        .build()
+        .await
+        .unwrap()
+}
+
+#[tokio::test]
+async fn test_cluster_lifecycle() {
+    run_ccm_test(cluster_1_node, async move |cluster| -> Result<(), Error> {
```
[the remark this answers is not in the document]
dkropachev: Thanks, it works.

dkropachev:
> I think that we should firstly focus on having the CI pass. Currently, the code does not even compile. Once we fix this, it will be much easier to test the code locally (which will be helpful in understanding the logic). I checked out on this PR locally, to play around with new API/run some tests but I was not able to do so, because it does not compile.
> 
> I suggest fixing the issues listed by: `cargo clippy --verbose --all-features --all-targets -- -D warnings`. Or if you want to not only do the checks, but run the tests as well, I suggest using `make ci`.

Thanks, done

scylla/tests/ccm-integration/cluster.rs
dkropachev: )))

scylla/tests/ccm-integration/cluster.rs
dkropachev: well, it does not do that and we have to dial with it.

scylla/tests/ccm-integration/cluster.rs
dkropachev: true, but for now we have to dial with current behavior.

scylla/tests/ccm-integration/logged_cmd.rs
dkropachev: When test is complex it is very handy to be able to see a flow of ccm commands it executed.
Even if we go with dumping everything into stderr, stdout, we will still need this wrapper.

Lorak-mmk: When the complex test fails there are many things that are handy to see, I don't see a reason to single-out CCM to be dumped to the file.

The wrapper can still exists to dump CCM commands to the output. 
I'd suggest using trace/debug for the actual output, and info for the command, env, and return code.

Lorak-mmk:
I'll be opening new PR with ccm integration.

PR #1204, opened 2025-02-06
errors: final cleanups (restructure ExecutionError and introduce SchemaAgreementError)

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/519

This PR originated from our offline discussion with @Lorak-mmk .

# Changes

## ProtocolError
We remove this type altogether. We noticed that this error type was a catch-all for sub-errors/variants we did not know where to put otherwise.

## ExecutionError
Before this PR, it would contain a lot of top-level variants without providing some meaningful context for them. Most of the variants came from the `RequestAttemptError::into_execution_error` method.

### Inlining RequestAttemptError
We decided  that it's best to inline `RequestAttemptError`, so users have additional context about how request execution looks like when they match the error. It allows us to remove a lot of variants from ExecutionError and ProtocolError (which is later removed).

## SchemaAgreementError
New error type - returned by public API related to awaiting schema agreement (Session::await/check_schema_agreement).

## BadQuery
Me and @Lorak-mmk agreed that final version (after this PR) of `BadQuery` is just fine. I did not do any changes to this error type, apart from moving one of the ProtocolError variants to it. I remember that @wprzytula did not like the name of this error - any suggestions?

# Potential improvements for the future
## SerializationError appears in two places
1. `BadQuery::SerializationError`
This is an error that appears on standard execution path for `Session::execute` and `Session::batch`, when we serialize the values (before doing the retries etc.)

2. `ExecutionError::LastAttemptError(RequestAttemptError::SerializationError)`
This error is constructed ONLY for `Session::query` with non-empty values. Why do we currently need a SerializationError in RequestAttemptError - i.e. why do we serialize on each attempt? This is because we need to firstly prepare the statement during each attempt - the statement needs to be firstly prepared on the node during current attempt. Only after preparation, have we access to the metadata and are able to serialize the values.

After discussion, we decided to currently leave it as is. This is something to be potentially fixed in the future. The fix requires non-trivial changes to execution logic (which is quite complex). The idea is to prepare the statement only once before the retries - then we can serialize the values and pass the serialized values to the context of each attempt. I think it deserves a separate issue in the repository, since it not only will fix the hierarchy of errors, but will improve the overall performance as well - we won't need to prepare the statement on each attempt (we can send QUERY cql request to the node).

## Pre-review checklist

### DIFF
--- a/scylla/src/errors.rs
+++ b/scylla/src/errors.rs
@@ -69,10 +69,6 @@ pub enum ExecutionError {
     #[error("No connections in the pool: {0}")]
     ConnectionPoolError(#[from] ConnectionPoolError),
 
-    /// Protocol error.
-    #[error("Protocol error: {0}")]
-    ProtocolError(#[from] ProtocolError),
-
     /// Failed to run a request within a provided client timeout.
     #[error(
         "Request execution exceeded a client timeout of {}ms",
@@ -152,16 +148,6 @@ pub enum NewSessionError {
     #[error("'USE KEYSPACE <>' request failed: {0}")]
     UseKeyspaceError(#[from] UseKeyspaceError),
 }
-
-/// A protocol error.
-///
-/// It indicates an inconsistency between CQL protocol
-/// and server's behavior.
-/// In some cases, it could also represent a misuse
-/// of internal driver API - a driver bug.
-#[derive(Error, Debug, Clone)]
-#[non_exhaustive]
-pub enum ProtocolError {}
 
 /// An error that occurred during `USE KEYSPACE <>` request.
 #[derive(Error, Debug, Clone)]
--- a/scylla/src/errors.rs
+++ b/scylla/src/errors.rs
@@ -80,13 +80,13 @@ pub enum ExecutionError {
     )]
     RequestTimeout(std::time::Duration),
 
-    /// Schema agreement timed out.
-    #[error("Schema agreement exceeded {}ms", std::time::Duration::as_millis(.0))]
-    SchemaAgreementTimeout(std::time::Duration),
-
     /// 'USE KEYSPACE <>' request failed.
     #[error("'USE KEYSPACE <>' request failed: {0}")]
     UseKeyspaceError(#[from] UseKeyspaceError),
+
+    /// Failed to await automatic schema agreement.
+    #[error("Failed to await schema agreement: {0}")]
+    SchemaAgreementError(#[from] SchemaAgreementError),
 }
 
 impl From<SerializationError> for ExecutionError {
@@ -162,10 +162,6 @@ pub enum NewSessionError {
 #[derive(Error, Debug, Clone)]
 #[non_exhaustive]
 pub enum ProtocolError {
-    /// A protocol error appeared during schema version fetch.
-    #[error("Schema version fetch protocol error: {0}")]
-    SchemaVersionFetch(#[from] SchemaVersionFetchError),
-
     /// Unable extract a partition key based on prepared statement's metadata.
     #[error("Unable extract a partition key based on prepared statement's metadata")]
     PartitionKeyExtraction,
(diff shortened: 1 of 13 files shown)


scylla/src/errors.rs
```diff
@@ -1,8 +1,29 @@
...
+pub use scylla_cql::frame::frame_errors::{
+    CqlAuthChallengeParseError, CqlAuthSuccessParseError, CqlAuthenticateParseError,
+    CqlErrorParseError, CqlEventParseError, CqlRequestSerializationError, CqlResponseParseError,
+    CqlResultParseError, CqlSupportedParseError, FrameBodyExtensionsParseError,
+    FrameHeaderParseError,
+};
+pub use scylla_cql::frame::request::CqlRequestKind;
 pub use scylla_cql::frame::response::error::{DbError, OperationType, WriteType};
+pub use scylla_cql::frame::response::CqlResponseKind;
+pub use scylla_cql::serialize::SerializationError;
```
Lorak-mmk: 🤔 Re-exporting errors from scylla_cql of course makes sense. I'm wondering about errors from `scylla`. Should we re-export them here, or maybe move them here?
And if we decide to use re-exports, which errors should be defined in the errors module, and which whould only be re-exported here?
 
@wprzytula @muzarski

muzarski: > 🔧 In any case, please move re-exports after imports.

Done

muzarski: > 🤔 Re-exporting errors from scylla_cql of course makes sense. I'm wondering about errors from `scylla`. Should we re-export them here, or maybe move them here? 
I'm fine with either approach. 

> And if we decide to use re-exports, which errors should be defined in the errors module, and which whould only be re-exported here?
I believe errors such as `NextRowError`/`IntoRowsResultError` were introduced in different module, since they are strictly tied to one object (in these cases it's `QueryPager` and `QueryResult`). Errors in `errors` module are rather related to methods on `Session`. But I'm pretty sure this does not apply to all errors in this module (e.g. `RequestError` used only in `RetrySession` API). So, actually, I think we could move ALL error types to `scylla::errors` module. Current division is a bit misleading.

Lorak-mmk (commented):
I noticed some pub types without `non_exhaustive`. Please tell me if they should remain that way, have `non_exhaustive` added, or be unpubed.
- `FirstRowError`
- `IntoRowsResultError`
- `MaybeFirstRowError`
- `ResultNotRowsError`
- `RowsError`
- `SingleRowError`
cc @wprzytula because they come from deserialization refactor.

Also:
- `DbError` (!)
- `OperationType`
- `WriteType`


We also have
```rust
/// Type to represent an authentication error message.
pub type AuthError = String;
```
is it indentional?

muzarski:
> I noticed some pub types without `non_exhaustive`. Please tell me if they should remain that way, have `non_exhaustive` added, or be unpubed.
> 
>     * `FirstRowError`
> 
>     * `IntoRowsResultError`
> 
>     * `MaybeFirstRowError`
> 
>     * `ResultNotRowsError`
> 
>     * `RowsError`
> 
>     * `SingleRowError`
>       cc @wprzytula because they come from deserialization refactor.

I remember asking about it during deserialization PRs. I don't remember why we decided not to mark them as `non_exhaustive`.

> Also:
> 
>     * `DbError` (!)
>
>     * `OperationType`
> 
>     * `WriteType`

These should definitely be `non_exhaustive`. Especially the `DbError` (if Scylla ever introduces new error types).

> We also have
> 
> ```rust
> /// Type to represent an authentication error message.
> pub type AuthError = String;
> ```
> 
> is it indentional?

Why wouldn't it be? We could always convert it to `Arc<dyn Error`> or something like that. I.e. do something like with `Serialization/DeserializationError`s. It has to be a dynamic error type (in this case just a `String`), because users need to be able to define their custom types - the error is returned from a public trait.

Lorak-mmk:
> Why wouldn't it be?

I'm not saying it is wrong, I'm just asking if it is intentionally this way or was just overlooked.

muzarski:
> > Why wouldn't it be?
> 
> I'm not saying it is wrong, I'm just asking if it is intentionally this way or was just overlooked.

Ok. So would you leave it as a `String`, or do something like:

```rust
pub AuthError(Arc<dyn Error>);

// Errors returned by driver's  default implementations of the trait (if there are any - I'm not sure)
pub enum BuiltinAuthErrorKind {
...
}
```

This would look like the `[De]SerializationError`.

Lorak-mmk:
Let's leave it as string.

muzarski:
v1.1:
- marked DbError as non_exhaustive
- narrow the return error type of public `ClusterState` methods related to token calculation

We still need to discuss and agree on whether we want to pack all the error types used in the driver to `scylla::errors` or not.

Lorak-mmk (approved):

PR #1212, opened 2025-02-10
feature(structre): extract utils into seperate common module

CodeLieutenant:
Since there will be more folders in scylla/tests, not just integration tests, then we'll need to extract common logic need by all tests folders into seperate module (`common`) before moving forward.

## Pre-review checklist


scylla/tests/common/retry_policy.rs
```diff
@@ -0,0 +1,26 @@
...
+
+impl RetrySession for NoRetrySession {
+    fn decide_should_retry(&mut self, _request_info: RequestInfo) -> RetryDecision {
+        RetryDecision::DontRetry
+    }
+
+    fn reset(&mut self) {
+        *self = Default::default()
+    }
+}
```
Lorak-mmk, line 26: We already have `scylla::policies::retry::fallthrough` which is exactly the same policy as this one. Please remove this.

Lorak-mmk (changes requested):

wprzytula (dismissed):

Lorak-mmk (approved):

CodeLieutenant:
@Lorak-mmk What do you think to set in `common/mod.rs`

```rust
pub(crate) use utils::*;
```

so that every import in other modules can be
```rust
#[path = "../common/mod.rs"]
mod utils;
```

Lorak-mmk:
> ```rust
> #[path = "../common/mod.rs"]
> mod utils;
> ```

Are you sure you meant to paste this snipper? This is not an import statement.

CodeLieutenant:
> > ```rust
> > #[path = "../common/mod.rs"]
> > mod utils;
> > ```
> 
> Are you sure you meant to paste this snippet? This is not an import statement.

Yes, not an import statement, but needed for the tests, each integration test file in rust is compiled as separate crate and each directory is separate crate, to include common in depth-2 without creating module and `integration_test.rs` in tests/ root, this is what's need to make it compile and run.

Lorak-mmk:
Sorry, I'm not following :( What change would you like to make? Right now (because this PR was merged) integration tests can `use utils::something` like before, and ccm tests can `use common::utils::something`. Would you like to make the ccm tests imports the same as in integration tests? That makes sense.

CodeLieutenant:
> Sorry, I'm not following :( What change would you like to make? Right now (because this PR was merged) integration tests can `use utils::something` like before, and ccm tests can `use common::utils::something`. Would you like to make the ccm tests imports the same as in integration tests? That makes sense.

yes exactly, to unify the imports (mod), right now, integration-tests need just the utils, but as more tests are added, there will be more utils 100%, and if we change it to code I suggested, would unify the `mod` statements.

```rust
#[path = "../common/mod.rs"]
mod utils;
```

exposing everything from utils, so that we dont have to change all the imports in integration-tests (backward compatible and at least for now, minize the number of conflicts - basically if i change `crate::utils::function` to `crate::common::utils::function`, there would be ton of conflicts as many people are working the the code right now). Like the temporary fix for the hackathon and later can be made to properly fix it with `crate::common::utils::my_needed_function`.

Lorak-mmk:
Can you please do the following:
- In both integration and ccm_integration do
```
#[path = "../common/mod.rs"]
mod common;
```

- In both do 
```
pub(crate) use common::utils;
```

This way they will have access to utils module without too many `path = ` hacks.

PR #1216, opened 2025-02-10
Add tests for {query,execute}_iter API

sylwiaszunejko:
## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/query_result.rs
+++ b/scylla/tests/integration/query_result.rs
@@ -1,4 +1,6 @@
+use assert_matches::assert_matches;
 use futures::TryStreamExt;
+use scylla::errors::PagerExecutionError;
 use scylla::{
     batch::{Batch, BatchType},
     client::session::Session,
@@ -137,6 +139,44 @@ async fn query_iter_should_iterate_over_all_pages_asynchronously_cross_partition
 }
 
 #[tokio::test]
+async fn query_iter_no_results() {
+    let (session, ks) = initialize_cluster_two_partitions().await;
+
+    let query = Query::new(format!("SELECT * FROM {}.t where k0 = ?", ks));
+
+    let query_result = session.query_iter(query, ("part3",)).await.unwrap();
+    let mut iter = query_result.rows_stream::<(String, i32, i32)>().unwrap();
+
+    assert_eq!(iter.try_next().await.unwrap(), None);
+}
+
+#[tokio::test]
+async fn query_iter_prepare_error() {
+    let (session, ks) = initialize_cluster_two_partitions().await;
+
+    // Wrong table name
+    let query = Query::new(format!("SELECT * FROM {}.test where k0 = ?", ks));
+
+    assert_matches!(
+        session.query_iter(query, (PARTITION_KEY1,)).await,
+        Err(PagerExecutionError::PrepareError(_))
+    );
+}
+
+#[tokio::test]
+async fn query_iter_serialization_error() {
+    let (session, ks) = initialize_cluster_two_partitions().await;
+
+    let query = Query::new(format!("SELECT * FROM {}.t where k0 = ?", ks));
+
+    // Wrong value type
+    assert_matches!(
+        session.query_iter(query, (1,)).await,
+        Err(PagerExecutionError::SerializationError(_))
+    );
+}
+
+#[tokio::test]
 async fn execute_single_page_should_only_iterate_over_rows_in_current_page() {
     let (session, ks) = initialize_cluster_two_partitions().await;
 
@@ -220,4 +260,38 @@ async fn execute_iter_should_iterate_over_all_pages_asynchronously_cross_partiti
         i += 1;
     }
     assert_eq!(i, 2 * ROWS_PER_PARTITION);
+}
+
+#[tokio::test]
+async fn execute_iter_no_results() {
+    let (session, ks) = initialize_cluster_two_partitions().await;
+
+    let prepared_query = session
+        .prepare(format!("SELECT * FROM {}.t where k0 = ?", ks))
+        .await
+        .unwrap();
+
+    let query_result = session
+        .execute_iter(prepared_query, ("part3",))
+        .await
+        .unwrap();
+    let mut iter = query_result.rows_stream::<(String, i32, i32)>().unwrap();
+
+    assert_eq!(iter.try_next().await.unwrap(), None);
+}
+
+#[tokio::test]
+async fn execute_iter_serialization_error() {
+    let (session, ks) = initialize_cluster_two_partitions().await;
+
+    let prepared_query = session
+        .prepare(format!("SELECT * FROM {}.t where k0 = ?", ks))
+        .await
+        .unwrap();
+
+    // Wrong value type
+    assert_matches!(
+        session.execute_iter(prepared_query, (1,)).await,
+        Err(PagerExecutionError::SerializationError(_))
+    )
 }
--- a/scylla/tests/integration/query_result.rs
+++ b/scylla/tests/integration/query_result.rs
@@ -1,0 +1,223 @@
+use futures::TryStreamExt;
+use scylla::{
+    batch::{Batch, BatchType},
+    client::session::Session,
+    query::Query,
+};
+use scylla_cql::frame::request::query::{PagingState, PagingStateResponse};
+
+use crate::utils::{create_new_session_builder, setup_tracing, unique_keyspace_name, PerformDDL};
+
+const PAGE_SIZE: i32 = 100;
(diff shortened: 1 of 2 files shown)


scylla/tests/ccm_integration/query_result.rs
```diff
@@ -0,0 +1,218 @@
...
+        .unwrap();
+    let page_results = rs_manual
+        .into_rows_result()
+        .unwrap()
+        .rows::<(String, i32, i32)>()
+        .unwrap()
+        .collect::<Result<Vec<_>, _>>()
+        .unwrap();
+
+    assert_eq!(page_results.len(), PAGE_SIZE as usize);
```
muzarski, line 81: I know that this test is based on the Java, but I think we should not assert equality here. See the short discussion: https://github.com/scylladb/scylla-rust-driver/pull/1068#issuecomment-2315226528. TLDR, we can never expect a specific number of rows when using a paged query (even if we expect only a single row!). This is because Scylla can send us some empty pages. Only when we disable paging, can we assume that Scylla sent whole result set.

This is why, I suggest to convert this equality check into `page_results.len() <= PAGE_SIZE`.
cc: @wprzytula  @Lorak-mmk

scylla/tests/ccm_integration/query_result.rs
```diff
@@ -0,0 +1,218 @@
...
+use crate::common::utils::{
+    create_new_session_builder, setup_tracing, unique_keyspace_name, PerformDDL,
+};
+
+const PAGE_SIZE: i32 = 100;
+const ROWS_PER_PARTITION: i32 = 1000;
+const PARTITION_KEY1: &str = "part";
+const PARTITION_KEY2: &str = "part2";
+
+async fn initialize_cluster() -> (Session, String) {
```
muzarski, line 19: nit: let's use some more descriptive name (e.g. `initialize_cluster_two_partitions` or sth like that)

muzarski: And please, add a docstring describing what the function does

scylla/tests/ccm_integration/query_result.rs
```diff
@@ -0,0 +1,218 @@
...
+        .unwrap();
+    let page_results = rs_manual
+        .into_rows_result()
+        .unwrap()
+        .rows::<(String, i32, i32)>()
+        .unwrap()
+        .collect::<Result<Vec<_>, _>>()
+        .unwrap();
+
+    assert_eq!(page_results.len(), PAGE_SIZE as usize);
```
muzarski, line 198: ditto: change `==` to `<=`

scylla/tests/ccm_integration/query_result.rs
```diff
@@ -131,6 +132,46 @@ async fn query_should_iterate_over_all_pages_asynchronously_cross_partition() {
...
+
+    let query = Query::new(format!("SELECT * FROM {}.test where k0 = ?", ks));
+
+    let query_result = session.query_iter(query, (PARTITION_KEY1,)).await;
+
+    assert!(matches!(
+        query_result,
+        Err(PagerExecutionError::PrepareError(_))
+    ));
+}
```
muzarski, line 159: We get `PrepareError`, because the table name is incorrect, right? If so, please add a small comment about it. It took me a while to spot the difference between this and non-failing scenario.

muzarski: Oh, and BTW there is the `assert_matches!` macro from the crate of the same name: https://docs.rs/assert_matches/latest/assert_matches/. We use it in the tests.

sylwiaszunejko:
@muzarski I addressed your comments and moved the file to tests/integration as it doesn't need ccm

scylla/tests/integration/query_result.rs
```diff
@@ -0,0 +1,223 @@
...
+const PARTITION_KEY2: &str = "part2";
+
+/// Initialize a cluster with a table and insert data into two partitions.
+/// Returns a session and the keyspace name.
+///
+/// #Example
+/// ```
+/// let (session, ks) = initialize_cluster().await;
+/// ```
+async fn initialize_cluster_two_partitions() -> (Session, String) {
```
wprzytula, line 25: :wrench: naming mismatch.

wprzytula:
```suggestion
/// # Example
/// ```rust
/// let (session, ks) = initialize_cluster_two_partitions().await;
/// ```
async fn initialize_cluster_two_partitions() -> (Session, String) {
```

wprzytula (changes requested):

scylla/tests/integration/query_result.rs
```diff
@@ -0,0 +1,297 @@
...
+    let prepared_query = session
+        .prepare(format!("SELECT * FROM {}.t where k0 = ?", ks))
+        .await
+        .unwrap();
+
+    // Wrong value type
+    match session.execute_iter(prepared_query, (1,)).await {
+        Ok(_) => panic!("the query should have failed due to a serialization error"),
+        Err(e) => assert_matches!(e, PagerExecutionError::SerializationError(_)),
+    }
```
wprzytula, line 296: :pick: 
```suggestion
    assert_matches!(session.execute_iter(prepared_query, (1,)).await, Err(PagerExecutionError::SerializationError(_)));
```

scylla/tests/integration/query_result.rs
```diff
@@ -0,0 +1,297 @@
...
+    while let Ok(Some((a, b, c))) = iter.try_next().await {
+        if i < ROWS_PER_PARTITION {
+            assert_eq!(a, PARTITION_KEY1);
+        } else {
+            assert_eq!(a, PARTITION_KEY2);
+        }
+        assert_eq!(b, i);
+        assert_eq!(c, i);
+        i += 1;
+    }
```
wprzytula, line 261: :wrench: This does not ensure that all rows are deserialized correctly.
```suggestion
    while let Some((a, b, c)) = iter.try_next().await.unwrap() {
        if i < ROWS_PER_PARTITION {
            assert_eq!(a, PARTITION_KEY1);
        } else {
            assert_eq!(a, PARTITION_KEY2);
        }
        assert_eq!(b, i);
        assert_eq!(c, i);
        i += 1;
    }
```

scylla/tests/integration/query_result.rs
```diff
@@ -0,0 +1,297 @@
...
+    let mut iter = query_result.rows_stream::<(String, i32, i32)>().unwrap();
+
+    let mut i = 0;
+
+    while let Ok(Some((a, b, c))) = iter.try_next().await {
+        assert_eq!(a, PARTITION_KEY1);
+        assert_eq!(b, i);
+        assert_eq!(c, i);
+        i += 1;
+    }
```
wprzytula, line 234: :wrench: ditto below

scylla/tests/integration/query_result.rs
```diff
@@ -0,0 +1,297 @@
...
+async fn query_iter_serialization_error() {
+    let (session, ks) = initialize_cluster_two_partitions().await;
+
+    let query = Query::new(format!("SELECT * FROM {}.t where k0 = ?", ks));
+
+    // Wrong value type
+    match session.query_iter(query, (1,)).await {
+        Ok(_) => panic!("the query should have failed due to a serialization error"),
+        Err(e) => assert_matches!(e, PagerExecutionError::SerializationError(_)),
+    }
```
wprzytula, line 176: :pick: ditto below

scylla/tests/integration/query_result.rs
```diff
@@ -0,0 +1,297 @@
...
+        if i < ROWS_PER_PARTITION {
+            assert_eq!(a, PARTITION_KEY1);
+        } else {
+            assert_eq!(a, PARTITION_KEY2);
+        }
+        assert_eq!(b, i);
+        assert_eq!(c, i);
+        i += 1;
+    }
+    assert_eq!(i, 2 * ROWS_PER_PARTITION);
```
wprzytula, line 138: :wrench: ditto below

scylla/tests/integration/query_result.rs
sylwiaszunejko: this gives me an error: "`QueryPager` doesn't implement `Debug`
the trait `Debug` is not implemented for `QueryPager`
the trait `Debug` is implemented for `std::result::Result<T, E>`"

wprzytula: It should implement `Debug`. I'll fix it in a PR to main, coming soon.

wprzytula: The fix is already on `branch-hackathon`.

wprzytula (approved):

PR #1218, opened 2025-02-10
tests: introduce new test for compression

fruch:
test are trying both algorithms (snappy, lz4), while counting the size of the frames coming into scylla

for reference we also have one test doing the same without compression

test main structure was borrowed from java-driver, the usage of proxy to count the packets is unique to this change.

Closes: https://github.com/scylladb/scylla-rust-driver/issues/1424

## Pre-review checklist

- [] I added relevant tests for new features and bug fixes.


fruch:
and now, with the file :)

scylla/tests/integration/compression.rs
```diff
@@ -0,0 +1,114 @@
...
+
+async fn test_compression(compression: Option<Compression>, text_size: usize, expected_frame_total_size_range: std::ops::Range<usize>) {
+    let res = test_with_3_node_cluster(ShardAwareness::QueryNode, |proxy_uris, translation_map, mut running_proxy| async move {
+    
+        let request_rule = |tx| {
+            RequestRule(
+                Condition::or(Condition::RequestOpcode(RequestOpcode::Execute), Condition::RequestOpcode(RequestOpcode::Query)),
+                RequestReaction::noop().with_feedback_when_performed(tx),
+            )
+        };
```
wprzytula, line 20: :question: Isn't it erroneous that you don't filter out the control connection? Depending on when the driver issues metadata queries on it and what responses it gets, the sum of frames' lengths can vary.

scylla/tests/integration/compression.rs
```diff
@@ -0,0 +1,114 @@
...
+use scylla::client::session_builder::SessionBuilder;
+use scylla::statement::query::Query;
+use scylla::frame::Compression;
+
+use scylla_proxy::{ProxyError, WorkerError, ShardAwareness, RequestRule, Reaction, RequestReaction, Condition, RequestOpcode};
+
+use tokio::sync::mpsc;
+use std::sync::Arc;
+
+async fn test_compression(compression: Option<Compression>, text_size: usize, expected_frame_total_size_range: std::ops::Range<usize>) {
```
wprzytula, line 12: :wrench: Could you please add a docstring / comment explaining the semantics of all parameters and the logic inside?

scylla/tests/integration/compression.rs
```diff
@@ -0,0 +1,114 @@
...
+        session.ddl(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 3}}", ks)).await.unwrap();
+        session.use_keyspace(ks, false).await.unwrap();
+        session
+            .ddl("CREATE TABLE test (k text PRIMARY KEY, t text, i int, f float)")
+            .await
+            .unwrap();
+
+        let q = Query::from("INSERT INTO test (k, t, i, f) VALUES (?, ?, ?, ?)");
+        let large_string = "a".repeat(text_size);
+        session.query_unpaged(q.clone(), ("key", large_string.clone(), 42_i32, 24.03_f32)).await.unwrap();
```
wprzytula, line 45: :pick: It should be able to pass `&str` as a bind value instead of cloning a `String`.

wprzytula:
```suggestion
        session.query_unpaged(q.clone(), ("key", large_string.as_str(), 42_i32, 24.03_f32)).await.unwrap();
```

scylla/tests/integration/compression.rs
fruch: I think that no all of the control parts are getting compressed, so they are less relevant

I can include everything back again, but then we'll need to re-calibrate the possible ranges of data length

scylla/tests/integration/compression.rs
fruch: 👍🏼

scylla/tests/integration/compression.rs
fruch: add a docstring (I think it's in the right convention)

wprzytula (changes requested):

scylla/tests/integration/compression.rs
```diff
@@ -0,0 +1,132 @@
...
+    expected_frame_total_size_range: std::ops::Range<usize>,
+) {
+    let res = test_with_3_node_cluster(ShardAwareness::QueryNode, |proxy_uris, translation_map, mut running_proxy| async move {
+
+        let request_rule = |tx| {
+            RequestRule(
+                Condition::True,
+                RequestReaction::noop().with_feedback_when_performed(tx),
+            )
+        };
```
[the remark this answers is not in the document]
fruch: It's not gonna be accurate anyhow, regardless of what filter we'll be doing

and it's the only thing we need to assess here, if it's compress or not, and we are using the estimated size (with or without control connection activity)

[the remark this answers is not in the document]
fruch: o.k. I've took now the logic you suggested
calibrated again the ranges, and let's see if CI is fine with that.

I think the outcome is the same either way (i.e. it's just a estimated values, and won't be completely  accurate)

scylla/tests/integration/compression.rs
```diff
@@ -0,0 +1,134 @@
...
+#[cfg(not(scylla_cloud_tests))]
+async fn should_execute_queries_without_compression() {
+    test_compression(None, 1024, 200..9000).await;
+}
+
+#[tokio::test]
+#[cfg(not(scylla_cloud_tests))]
+async fn should_execute_queries_without_compression_10mb() {
+    test_compression(None, 1024 * 10, 10000..19000).await;
+}
```
wprzytula, line 110: Just to be sure: length is given in bytes, so `1024 * 10` is 10 kiB, not 10 MiB, correct? Because the name of the test is `...mb`, and this confuses me.

wprzytula:
The test [has failed for Cassandra](https://github.com/scylladb/scylla-rust-driver/actions/runs/13282334636/job/37083208141?pr=1218#step:6:1405):
`assertion failed: expected_frame_total_size_range.contains(&total_frame_size)`

scylla/tests/integration/compression.rs
fruch: yes I was confusing both thing
it's doesn't make a big difference in this tests if it's MB or MiB 

I hope it's a bit clearer now

wprzytula:
Closing in favour of https://github.com/scylladb/scylla-rust-driver/pull/1896.

PR #1221, opened 2025-02-11
Implement 'From<Option>' for 'MaybeUnset'

stanislav-tkach:
I think it is convenient to simplify conversion from `Option<T>` to `MaybeUnset<T>`, but maybe I'm missing something.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/value.rs
+++ b/scylla-cql/src/value.rs
@@ -33,6 +33,12 @@ pub enum MaybeUnset<V> {
     Set(V),
 }
 
+impl<V> From<Option<V>> for MaybeUnset<V> {
+    fn from(value: Option<V>) -> Self {
+        value.map_or(Self::Unset, MaybeUnset::Set)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


wprzytula:
:question: Did you consider that `Option::None` already represents presence of the CQL null value, which is different from absence of value that `MaybeUnset` represent?

stanislav-tkach:
I see, but `From` conversion is explicit, so I still consider it useful. But feel free to close the pull request if you think it will create confusion.

wprzytula:
> @wprzytula I see, but `From` conversion is explicit, so I still consider it useful. But feel free to close the pull request if you think it will create confusion.

I'm in two minds. @Lorak-mmk @muzarski WDYT?

Lorak-mmk:
I'm against such conversions. It doesn't save a lot of typing, but it may introduce unwanted behavior.

stanislav-tkach:
Ok, no problem.

PR #1226, opened 2025-02-11
tests: add integration tests for batch statements

dimakr:
Test cases/ideas are inspired by Java driver batch statements tests.

## Pre-review checklist


wprzytula:
I intended to open a PR instead of pushing to your branch directly, sorry @dimakr...

dimakr:
@wprzytula I've made the changes as per the comments above

scylla/tests/integration/batch.rs
```diff
@@ -76,3 +139,310 @@ async fn batch_statements_and_values_mismatch_detected() {
...
+        let row = query_result.single_row::<(i64,)>().unwrap();
+        let (c,) = row;
+        assert_eq!(c, i as i64);
+    }
+}
+
+/// TODO: Remove #[ignore] once counters are supported with tablets.
+#[tokio::test]
+#[ignore]
+async fn test_fail_logged_batch_with_counter_increment() {
```
wprzytula, line 395: :thought_balloon: Perhaps we could run these tablets-incompatible tests with tablets explicitly turned off for the keyspace?

dimakr: Disabled tablets explicitly for these tests.

dimakr:
Turned out that we have a few batch tests in scylla/tests/integration/session.rs test suite.
Moved them to the dedicated `batch.rs` test suite, for which we are creating new test cases in this PR.

wprzytula (approved):

scylla/tests/integration/batch_statements.rs
```diff
@@ -0,0 +1,403 @@
...
+        batch.append_statement(simple_statement);
+    }
+    session.batch(&batch, vec![(); BATCH_COUNT]).await.unwrap();
+    verify_batch_insert(&session, &ks, &test_name, BATCH_COUNT).await;
+}
+
+#[tokio::test]
+async fn test_batch_of_bound_statements() {
+    setup_tracing();
+    let test_name = String::from("test_batch_bound_statements");
```
Lorak-mmk: Could you name this differently? Bound statements and bound batches are something we plan to introduce in the future, this naming will become confusing then.

scylla/tests/integration/batch_statements.rs
```diff
@@ -0,0 +1,403 @@
...
+    let mut batch = Batch::new(BatchType::Unlogged);
+    let mut batch_values: Vec<Vec<CqlValue>> = Vec::with_capacity(BATCH_COUNT);
+    for i in 0..BATCH_COUNT as i32 {
+        batch.append_statement(prepared.clone());
+        batch_values.push(vec![
+            CqlValue::Text(test_name.clone()),
+            CqlValue::Int(i),
+            CqlValue::Int(i + 1),
+        ]);
+    }
```
Lorak-mmk: You can use `Vec<(String, i32, i32)>` for `batch_values` - no need to go through another vector, and CqlValue.

scylla/tests/integration/batch_statements.rs
```diff
@@ -0,0 +1,403 @@
...
+    session.batch(&batch, batch_values).await.unwrap();
+    verify_batch_insert(&session, &ks, &test_name, BATCH_COUNT).await;
+}
+
+/// TODO: Remove #[ignore] once LWTs are supported with tablets.
+#[tokio::test]
+#[ignore]
+async fn test_cas_batch() {
+    setup_tracing();
+    let test_name = String::from("test_cas_batch");
```
Lorak-mmk: I guess you got the inspiration from `test_lwt_optimization_works_with_tablets`. The difference is that this test does not depend on tablets, while `test_lwt_optimization_works_with_tablets` does. 
Please disable tablets explicitly during keyspace creation. This  will allow the test to actually run.

scylla/tests/integration/batch_statements.rs
```diff
@@ -0,0 +1,403 @@
...
+        .unwrap();
+    assert!(applied2, "Second CAS batch should not be applied");
+}
+
+/// TODO: Remove #[ignore] once counters are supported with tablets.
+#[tokio::test]
+#[ignore]
+async fn test_counter_batch() {
+    setup_tracing();
+    let test_name = String::from("test_counter_batch");
```
Lorak-mmk: ditto

scylla/tests/integration/batch_statements.rs
```diff
@@ -0,0 +1,403 @@
...
+        let (c,) = row;
+        assert_eq!(c, i as i64);
+    }
+}
+
+/// TODO: Remove #[ignore] once counters are supported with tablets.
+#[tokio::test]
+#[ignore]
+async fn test_fail_logged_batch_with_counter_increment() {
+    setup_tracing();
```
Lorak-mmk: ditto for this and other tests you disabled

scylla/tests/integration/batch_statements.rs
Lorak-mmk, line 1: Commit: "don't use CqlValue unnecessarily " 

We try to not have one commit do something in a wrong way and then have another that fixes it. This commit should be squashed into the previous one.

scylla/tests/common/utils.rs
```diff
@@ -42,7 +42,7 @@ pub(crate) fn unique_keyspace_name() -> String {
         SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .unwrap()
-            .as_secs(),
+            .as_micros(),
```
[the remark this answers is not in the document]
Lorak-mmk: I'd also really like to know why did you make this change. I don't see anything about it on PR cover lether nor in commit message.

Lorak-mmk (changes requested):

scylla/tests/common/utils.rs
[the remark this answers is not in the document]
dimakr: I had to commit this separately, agree.. Did it now.

This change was done as ks name uniqueness could not be guarantied in some test suite setups. E.g. in batch test suite there is a dozen of tests which share the same setup helper. Most of the cases require the same KS (with the same parameters), so its OK for them to reuse KS if any exists already.
But few cases require tablets disabled, so the KS params differ. This change allows to ensure unique value for KS name for parallelized tests.

scylla/tests/integration/batch_statements.rs
dimakr: Changed the name

scylla/tests/integration/batch_statements.rs
dimakr: Squashed the commits

scylla/tests/integration/batch.rs
```diff
@@ -76,3 +144,549 @@ async fn batch_statements_and_values_mismatch_detected() {
...
+        batch.append_statement(Query::new(query_str.clone()));
+        batch_values.push((&test_name, i, i + 1));
+    }
+    let prepared_batch = session.prepare_batch(&batch).await.unwrap();
+    session.batch(&prepared_batch, batch_values).await.unwrap();
+    verify_batch_insert(&session, &test_name, BATCH_COUNT).await;
+}
+
+#[tokio::test]
+async fn test_batch_of_bound_statements_with_unset_values() {
```
wprzytula, line 344: :wrench: As noted by @Lorak-mmk, we don't yet have bound statements in the driver, so the name may be misleading.

scylla/tests/integration/batch.rs
```diff
@@ -76,3 +144,549 @@ async fn batch_statements_and_values_mismatch_detected() {
...
+        rows.len()
+    );
+    for (k0, k1, v) in rows {
+        assert_eq!(k0, test_name);
+        assert_eq!(v, if k1 % 20 == 0 { k1 + 1 } else { k1 + 2 });
+    }
+}
+
+#[tokio::test]
+async fn test_batch_of_bound_statements_named_variables() {
```
wprzytula, line 406: :wrench: ditto: naming

scylla/tests/integration/batch.rs
```diff
@@ -76,3 +144,549 @@ async fn batch_statements_and_values_mismatch_detected() {
...
+        values.insert("k1", CqlValue::Int(i));
+        values.insert("v", CqlValue::Int(i + 1));
+        batch_values.push(values);
+    }
+    session.batch(&batch, batch_values).await.unwrap();
+    verify_batch_insert(&session, &test_name, BATCH_COUNT).await;
+}
+
+#[tokio::test]
+async fn test_batch_of_mixed_bound_and_simple_statements() {
```
wprzytula, line 432: :wrench: ditto: naming

scylla/tests/integration/batch.rs
```diff
@@ -76,3 +144,549 @@ async fn batch_statements_and_values_mismatch_detected() {
...
+    let batch_deserializer = batch_res.into_rows_result().unwrap();
+
+    // Scylla returns 4 columns, but Cassandra returns only 1
+    let is_scylla: bool = batch_deserializer.column_specs().len() == 4;
+
+    if is_scylla {
+        test_batch_lwts_for_scylla(&session, &batch, &batch_deserializer, &test_name).await;
+    } else {
+        test_batch_lwts_for_cassandra(&session, &batch, &batch_deserializer, &test_name).await;
+    }
```
wprzytula, line 516: :wrench: Please provide links to documentation explaining this difference.

wprzytula (changes requested):

Lorak-mmk:
> @wprzytula @Lorak-mmk addressed the last comments.

In commit "tests: add integration tests for batch statements" some tests are still marked with `#[ignore]` instead of disabling tablets. Maybe you forgot to fix that?

scylla/tests/common/utils.rs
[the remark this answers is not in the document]
dimakr: This is how the batch tests fail. if `as_sec()` is used to generate suffix for KS name:
```
──── STDERR:             scylla::integration batch::test_batch_lwts
thread 'batch::test_batch_lwts' panicked at scylla/tests/integration/batch.rs:506:76:
called `Result::unwrap()` on an `Err` value: LastAttemptError(DbError(Invalid, "Cannot use LightWeight Transactions for table test_rust_1739527888_0.test_counter_batch: LWT is not yet supported with tablets"))
...
──── STDERR:             scylla::integration batch::test_fail_counter_batch_with_non_counter_increment
thread 'batch::test_fail_counter_batch_with_non_counter_increment' panicked at scylla/tests/integration/batch.rs:54:14:
called `Result::unwrap()` on an `Err` value: LastAttemptError(DbError(Invalid, "Cannot use the 'counter' type for table test_rust_1739527888_0.counter1: Counters are not yet supported with tablets"))
...
```

The `batch` test suite uses the common helper to create a KS and table. Then when each test is executed and calls the helper it would try to get its own, unique, KS name. But tests in this test suite are executed within 1 second so the `unique_keyspace_name` cannot guarantee unique value. And in this particular test suite part of tests require KS with tablets disabled, so the existing KS cannot be reused.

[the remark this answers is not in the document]
dimakr: Probably for the affected tests here some other approach could be used, i.e. do not use`unique_keyspace_name`, but some dedicated KS name generation logic. But I thought that the symptom might pop up in other tests as well, and this change would help.

dimakr:
> In commit "tests: add integration tests for batch statements" some tests are still marked with `#[ignore]` instead of disabling tablets. Maybe you forgot to fix that?

@Lorak-mmk the tablets disabling was done as part of "[tests: move batch statements tests to batch.rs module](https://github.com/scylladb/scylla-rust-driver/pull/1226/commits/24cdc2550c3b9a20b477c2558af89a4b57b61523)" commit. I've adjusted that commit message to clearly mention this.

scylla/tests/common/utils.rs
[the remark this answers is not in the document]
Lorak-mmk: > But tests in this test suite are executed within 1 second so the unique_keyspace_name cannot guarantee unique value.

Please the the source code of unique_keyspace_name. It uses an atomic counter to guarantee uniqueness.

Lorak-mmk:
> @Lorak-mmk the tablets disabling was done as part of "[tests: move batch statements tests to batch.rs module](https://github.com/scylladb/scylla-rust-driver/pull/1226/commits/24cdc2550c3b9a20b477c2558af89a4b57b61523)" commit. I've adjusted that commit message to clearly mention this.

Is there a reason for doing it in the commit that moves tests, instead of doing it that way from the beggining?

scylla/tests/common/utils.rs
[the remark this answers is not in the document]
dimakr: It fails when specific module is executed, e.g.:
```
> cargo nextest r -E 'test(/^batch::/)'
```
But it is ok when all tests are executed at once (make test, etc.). So I'll drop that change.

dimakr:
> Is there a reason for doing it in the commit that moves tests, instead of doing it that way from the beginning?
There was a bit of mess in the series of commits in this PR from the beginning after a commit from another person was accidentally pushed in (that one is squashed eventually). But to avoid further issues I've made 'disable tablets'-change in the subsequent commit.

scylla/tests/common/utils.rs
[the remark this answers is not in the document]
Lorak-mmk: `cargo nextest` is not something we ever tested, or recommend in `CONTRIBUTING.md`.
It uses process-per-test architecture, making a counter that guarantees uniqueness useless - because each process, and thus test, will have its own copy of it, thus always using 0.

Lorak-mmk:
Two questions:
- I don't think you fixed all tests in regards to tablets. In the "Files changed view" I see that `test_cas_batch`, `test_counter_batch`, `test_fail_logged_batch_with_counter_increment`, `test_fail_counter_batch_with_non_counter_increment` in `batch_statement.rs` are still ignored
- I see that after this PR we have 2 files: batch.rs and batch_statement.rs. Why? What future tests should go into the latter, and what into the former?

dimakr:
> Two things:

It was OK, but due to this number of nit related changes one of recent rebases went wrong. Will correct it

dimakr:
> > Two things:
> 
> It was OK, but due to this number of nit related changes one of recent rebases went wrong. Will correct it

@Lorak-mmk  This time should be ok - got rid of `batch_statement.rs` as per comment https://github.com/scylladb/scylla-rust-driver/pull/1226#discussion_r1951070923

wprzytula:
Closing in favour of #1897.

PR #1228, opened 2025-02-11
Add integration tests for `query_unpaged` and `execute_unpaged` APIs

vponomaryov:
## Pre-review checklist


scylla/tests/integration/query_result.rs
```diff
@@ -295,3 +296,171 @@ async fn execute_iter_serialization_error() {
...
+    let cql_create_ks = format!(
+        "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION =\
+         {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}",
+        ks
+    );
+    session.ddl(cql_create_ks).await.unwrap();
+    session.use_keyspace(ks, false).await.unwrap();
+
+    let cql_drop_table = format!("DROP TABLE IF EXISTS {}", table_name);
+    session.ddl(cql_drop_table).await.unwrap();
```
Lorak-mmk, line 313: No need to drop the table because we use random keyspace names.

scylla/tests/integration/query_result.rs
```diff
@@ -295,3 +296,171 @@ async fn execute_iter_serialization_error() {
...
+}
+
+async fn unpaged_error(use_prepared_statements: bool) {
+    let table_name = if use_prepared_statements {
+        "execute_unpaged"
+    } else {
+        "query_unpaged"
+    };
+    let session: Session = create_session(table_name).await;
+    session.await_schema_agreement().await.unwrap();
```
Lorak-mmk, line 384: No need to await schema agreement after DDL statements - it is awaited automatically. unless explicitly disabled in SessionBuilder.

scylla/tests/integration/query_result.rs
```diff
@@ -295,3 +296,171 @@ async fn execute_iter_serialization_error() {
...
+
+#[tokio::test]
+async fn test_query_unpaged_error() {
+    unpaged_error(false).await
+}
+
+#[tokio::test]
+async fn test_execute_unpaged_error() {
+    unpaged_error(true).await
+}
```
Lorak-mmk, line 416: I'd like to not make tests unnecessarily slow, which means reducing the amount of schema changes.
Those 2 tests could be merged into one which:
- Creates KS and table
- Makes the unprepared check (because it does not remove the table)
- Makes the prepared check.

scylla/tests/integration/query_result.rs
```diff
@@ -295,3 +296,171 @@ async fn execute_iter_serialization_error() {
...
+
+async fn check_query_unpaged(insert_num: u64, use_prepared_statements: bool) {
+    let table_name = if use_prepared_statements {
+        "execute_unpaged"
+    } else {
+        "query_unpaged"
+    };
+    let session: Session = create_session(table_name).await;
+    session.await_schema_agreement().await.unwrap();
+
```
Lorak-mmk, line 332: Ditto: no need to await schema agreement

scylla/tests/integration/query_result.rs
```diff
@@ -295,3 +296,171 @@ async fn execute_iter_serialization_error() {
...
+#[tokio::test]
+async fn test_execute_unpaged_hundred_rows() {
+    check_query_unpaged(100, true).await;
+}
+
+#[tokio::test]
+async fn test_execute_unpaged_thousand_rows() {
+    check_query_unpaged(1000, true).await;
+}
```
Lorak-mmk, line 466: Those tests could be merged into one test (or two - one for prepared, one for unprepared), in order to re-use the created session and table.

vponomaryov: All the 23 tests in this module take about 5 seconds not considering the compilation time.

Current approach is the test isolation, reusing session and tables we create side-effects.
Having 5 seconds for 23 tests I don't think that the gained test run time worth the side-effects.

Then, existing 11 tests follow the same approach - isolated logic. Which, I think, is proper approach for "tests".

scylla/tests/integration/query_result.rs
vponomaryov: Answered below.

scylla/tests/integration/query_result.rs
vponomaryov: Done.

scylla/tests/integration/query_result.rs
Lorak-mmk: 5 seconds of execution time is not short.
I know that writing them as one tests is a bit more involved (e.g. you need to `DELETE` stuff from the table between test cases) but I think its still worth it.

vponomaryov: It is not just some delete's or truncate's, it is losing test isolation.
Also, delete and truncate calls take some time too.
Adding new tests we should not bother ourselves in which order tests run and how they may fail in the middle and cause other tests to fail due to unexpected DB table state.
I still value the test isolation more than the really small possible time gain.

Also, note that 5 seconds is taken by all the tests in this module including existing ones.

vponomaryov: Proper test isolation reduces the investigation time in case of failures.

So, looks like we need other opinions here to get a quorum.

Lorak-mmk: Investigating failing tests has never been a problem in Rust Driver. Iteration time is something that constantly affects development speed.

dkropachev: I think we need to merge them, another reason is that there is chance for ddl statement to fail when ran in parralell due to Scylla bug, to make suit more stable we need reduce them as much as possible.

PR #1229, opened 2025-02-11
test(query_result): added metadata tests

soyacz:
Added tests around query response metadata:
1. execution info upon exception
2. metadata basic verification (state)
3. basic tracing verification

based on:
com.datastax.oss.driver.core.session.ExceptionIT
com.datastax.oss.driver.core.metadata.MetadataIT
com.datastax.oss.driver.core.cql.QueryTraceIT

## Pre-review checklist


scylla/tests/integration/query_result.rs
```diff
@@ -295,3 +307,131 @@ async fn execute_iter_serialization_error() {
...
+    assert!(tracing_info.parameters.unwrap().contains_key("query"));
+
+    // Check events
+    for event in tracing_info.events {
+        assert!(!event.activity.clone().unwrap().is_empty());
+        assert!(event.source.is_some());
+        assert!(event.source_elapsed.unwrap() >= 0);
+        assert!(!event.activity.unwrap().is_empty());
+    }
+}
```
Lorak-mmk, line 437: Could you look into `test_tracing_` tests in `integration/session.rs`? It looks to me like those tests mostly duplicate the ones already existing - maybe instead of adding new ones you could improve existing ones with the checks that you wrote?

scylla/tests/integration/query_result.rs
```diff
@@ -295,3 +307,131 @@ async fn execute_iter_serialization_error() {
...
+            running_proxy
+        },
+    )
+    .await;
+
+    match res {
+        Ok(()) => (),
+        Err(err) => panic!("{}", err),
+    }
+}
```
Lorak-mmk, line 357: Isn't that mostly a `history` tests? What does it verify that tests in `integration/history.rs` don't?

Lorak-mmk (changes requested):

scylla/tests/integration/query_result.rs
soyacz: basically I wanted to test more details in errors, like Java driver does provide - but Rust driver don't expose them.
Somehow I have problems with finding similar tests - like also tracing tests I didn't notice already exist. Still I'm adapting to new IDE and ctrl+f mislead me.

Lorak-mmk: The structure of our tests (which test resides in which file) is unfortunately not the best :( We'd like to improve it - which is why I made my comments.
Feel free to ask if you can't find something that you suspect may already exist.
Nice way is to check for references to the function / method / field etc that you want to use in the test.
For `Session::get_tracing_info()` it does correctly find a reference in `integration/session.rs`:
image
specifically, in `test_get_tracing_info`. This function is in turn called by `test_tracing` in the same file, and there you can see our existing test cases for tracing.

soyacz: mostly my confusion is taken (again) out of python: test files usually start with `test_` so when I search for something and I see it in `test_xxx` file then I know it's in testing. Here when searching it is shown in `session.rs` which mislead me (I missed looking at the directories).

soyacz:
@Lorak-mmk I adjusted to your's comments. Other state tests will be moved in different PR.

scylla/tests/integration/session.rs
```diff
@@ -831,6 +831,39 @@ async fn test_get_tracing_info(session: &Session, ks: String) {
...
+    if session
+        .get_cluster_state()
+        .get_nodes_info()
+        .first()
+        .unwrap()
+        .sharder()
+        .is_some()
+    {
+        assert!(tracing_info.duration.unwrap() > 0);
+    }
```
Lorak-mmk, line 848: Is there some Scylla issue about the incompatibility? Or maybe a documentation if the incompatibility is intentional? It should be linked here.

scylla/tests/integration/cluster_state_tests.rs
```diff
@@ -0,0 +1,30 @@
...
+        .get_nodes_info()
+        .iter()
+        .map(|node| node.address.to_string())
+        .collect();
+
+    assert_eq!(
+        got_addresses, expected_addresses,
+        "Cluster node addresses do not match environment variables"
+    );
+}
```
wprzytula, line 30: :thinking: Maybe: `test_session_should_have_metadata` -> `test_session_should_have_topology_metadata`? Schema metadata is not checked here at all.

scylla/tests/integration/session.rs
```diff
@@ -831,6 +831,39 @@ async fn test_get_tracing_info(session: &Session, ks: String) {
...
+    // Verify started_at timestamp is present
+    assert!(tracing_info.started_at.unwrap().0 > 0);
+
+    // Check parameters
+    assert!(tracing_info
+        .parameters
+        .clone()
+        .unwrap()
+        .contains_key("consistency_level"));
+    assert!(tracing_info.parameters.unwrap().contains_key("query"));
```
wprzytula, line 858: :wrench: Instead of `clone()`ing an `Option<T>` in order to `unwrap()` it, it's better to use `as_ref()` to `unwrap()` on a borrowed version of the `Option`. See:

```suggestion
    // Check parameters
    assert!(tracing_info
        .parameters
        .as_ref()
        .unwrap()
        .contains_key("consistency_level"));
    assert!(tracing_info.parameters.as_ref().unwrap().contains_key("query"));
```

scylla/tests/integration/session.rs
```diff
@@ -831,6 +831,39 @@ async fn test_get_tracing_info(session: &Session, ks: String) {
...
+
+    // Check events
+    for event in tracing_info.events {
+        assert!(!event.activity.clone().unwrap().is_empty());
+        assert!(event.source.is_some());
+        assert!(event.source_elapsed.unwrap() >= 0);
+        assert!(!event.activity.unwrap().is_empty());
```
wprzytula, line 865:
```suggestion
        assert!(!event.activity.as_ref().unwrap().is_empty());
        assert!(event.source.is_some());
        assert!(event.source_elapsed.unwrap() >= 0);
        assert!(!event.activity.as_ref().unwrap().is_empty());
```

wprzytula (changes requested):

wprzytula (approved):

scylla/tests/integration/session.rs
soyacz: adjusted

scylla/tests/integration/session.rs
soyacz: adjusted

scylla/tests/integration/cluster_state_tests.rs
soyacz: adjusted

scylla/tests/integration/session.rs
soyacz: I adjusted the test to verify behavior when cassandra is used (waiting for duration to appear - as cassandra has some delay)

soyacz:
@Lorak-mmk `cargo clippy --all-targets` started to fail after update rust from 1.84 to 1.85. I believe this is fixed on master - but this branch bases on branch-hackathon

PR #1231, opened 2025-02-12
Fix improper ip address formatting in CCM integration.

Lorak-mmk:
When passing `NetPrefix` as an argument to `ccm populate` it is
formatted using `.to_string()` method. Result of this is then concatenated
with the node id, and the result is an "ip" that ccm tries to find in
Scylla logs to determine that the node is UP.

`.to_string()` implementation was displaying the whole ip, instead of
it's first 3 octets, resulting in CCM looking for ips like `127.0.1.02`.
Instead the `to_str()` method should be used, that produces ccm-compatible
form of the ip range.

## Pre-review checklist


vponomaryov (approved):
LGTM

muzarski (approved):

PR #1233, opened 2025-02-12
ccm: Add simple support for `updateconf` and setting up TLS

muzarski:
This PR adds support for `ccm updateconf` command (for both cluster and node) in our new API. It also adds a possibility to start a cluster with TLS enabled (by using `updateconf`). I added a simple test case for tls to check that everything works.

Apart from that, I removed the `tls.yml` workflow. There is no need for it, since we can write integration tests using ccm API. Originally, the workflow would only run the tls example.

## Pre-review checklist


muzarski:
v1.1: Added some docstrings with example usage of new API

Lorak-mmk (approved):
Please remove scylla.yaml as discussed. Apart from that looks good.

Lorak-mmk (approved):

Lorak-mmk:
I'll merge once the CI passes. I'm really happy we can move some stuff from custom pipelines to CCM!

PR #1234, opened 2025-02-12
ccm: Make 'opts' parameter of the 'stop' method be optional

vponomaryov:
It will allow to use the 'stop' method just providing "None" parameter:
```
  ... .stop(None).await;
```

## Pre-review checklist


scylla/tests/ccm_integration/ccm/cluster.rs
```diff
@@ -174,6 +174,7 @@ impl NodeStartOptions {
 /// Options to stop the node with.
 /// It allows to control the value of `--no-wait` and `--not-gently` ccm options.
 #[allow(dead_code)]
+#[derive(Default)]
```
muzarski, line 177: Both booleans will be set to `false`. This is expected, right?

muzarski (approved):

Lorak-mmk (approved):

PR #1235, opened 2025-02-12
Add integration tests for query parameters

sylwiaszunejko:
Moved some tests from different files and added new ones inspired by java-driver tests.

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/statement.rs
+++ b/scylla/tests/integration/statement.rs
@@ -1,0 +1,273 @@
+use crate::utils::{create_new_session_builder, setup_tracing, unique_keyspace_name, PerformDDL};
+use scylla::{
+    batch::{Batch, BatchType},
+    query::Query,
+    response::PagingState,
+};
+use std::collections::HashSet;
+use std::sync::{Arc, Mutex};
+
+#[tokio::test]
+async fn test_timestamp() {
+    setup_tracing();
+    let session = create_new_session_builder().build().await.unwrap();
+    let ks = unique_keyspace_name();
+
+    session.ddl(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks)).await.unwrap();
+    session.use_keyspace(ks, false).await.unwrap();
+    session
+        .ddl("CREATE TABLE IF NOT EXISTS t_timestamp (a text, b text, primary key (a))")
+        .await
+        .unwrap();
+
+    let query_str = "INSERT INTO t_timestamp (a, b) VALUES (?, ?)";
+
+    // test regular query timestamps
+
+    let mut regular_query = Query::new(query_str.to_string());
+
+    regular_query.set_timestamp(Some(420));
+    session
+        .query_unpaged(regular_query.clone(), ("regular query", "higher timestamp"))
+        .await
+        .unwrap();
+
+    regular_query.set_timestamp(Some(42));
+    session
+        .query_unpaged(regular_query.clone(), ("regular query", "lower timestamp"))
+        .await
+        .unwrap();
+
+    // test regular query iter timestamps
+
+    let mut regular_query = Query::new(query_str.to_string());
+
+    regular_query.set_timestamp(Some(420));
+    session
+        .query_iter(
+            regular_query.clone(),
+            ("regular query iter", "higher timestamp"),
+        )
+        .await
+        .unwrap();
+
+    regular_query.set_timestamp(Some(42));
+    session
+        .query_iter(
+            regular_query.clone(),
+            ("regular query iter", "lower timestamp"),
+        )
+        .await
+        .unwrap();
+
+    // test regular query single page timestamps
+
+    let mut regular_query = Query::new(query_str.to_string());
+
+    regular_query.set_timestamp(Some(420));
+    session
+        .query_single_page(
+            regular_query.clone(),
+            ("regular query single page", "higher timestamp"),
+            PagingState::start(),
+        )
+        .await
+        .unwrap();
+
+    regular_query.set_timestamp(Some(42));
+    session
+        .query_single_page(
+            regular_query.clone(),
+            ("regular query single page", "lower timestamp"),
+            PagingState::start(),
+        )
+        .await
+        .unwrap();
+
+    // test prepared statement timestamps
+
+    let mut prepared_statement = session.prepare(query_str).await.unwrap();
+
+    prepared_statement.set_timestamp(Some(420));
+    session
+        .execute_unpaged(&prepared_statement, ("prepared query", "higher timestamp"))
+        .await
+        .unwrap();
+
+    prepared_statement.set_timestamp(Some(42));
+    session
+        .execute_unpaged(&prepared_statement, ("prepared query", "lower timestamp"))
+        .await
+        .unwrap();
+
+    // test prepared statement iter timestamps
+
+    let mut prepared_statement = session.prepare(query_str).await.unwrap();
+
+    prepared_statement.set_timestamp(Some(420));
+    session
+        .execute_iter(
+            prepared_statement.clone(),
+            ("prepared query iter", "higher timestamp"),
+        )
+        .await
+        .unwrap();
+
+    prepared_statement.set_timestamp(Some(42));
+    session
+        .execute_iter(
+            prepared_statement,
+            ("prepared query iter", "lower timestamp"),
+        )
+        .await
+        .unwrap();
+
+    // test prepared statement single page timestamps
+
+    let mut prepared_statement = session.prepare(query_str).await.unwrap();
+
+    prepared_statement.set_timestamp(Some(420));
+    session
+        .execute_single_page(
+            &prepared_statement,
+            ("prepared query single page", "higher timestamp"),
+            PagingState::start(),
+        )
+        .await
+        .unwrap();
+
+    prepared_statement.set_timestamp(Some(42));
+    session
+        .execute_single_page(
+            &prepared_statement,
+            ("prepared query single page", "lower timestamp"),
+            PagingState::start(),
+        )
+        .await
+        .unwrap();
+
+    // test batch statement timestamps
+
+    let mut batch: Batch = Default::default();
+    batch.append_statement(regular_query);
+    batch.append_statement(prepared_statement);
+
+    batch.set_timestamp(Some(420));
+    session
+        .batch(
+            &batch,
+            (
+                ("first query in batch", "higher timestamp"),
+                ("second query in batch", "higher timestamp"),
+            ),
+        )
+        .await
+        .unwrap();
+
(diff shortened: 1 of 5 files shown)


scylla/tests/integration/query.rs
```diff
@@ -0,0 +1,300 @@
...
+    let mut regular_query = Query::new(query_str.to_string());
+
+    regular_query.set_timestamp(Some(420));
+    session
+        .query_iter(
+            regular_query.clone(),
+            ("regular query iter", "higher timestamp"),
+        )
+        .await
+        .unwrap();
```
muzarski, line 80: I'm actually shocked that it works. The usual use case for `[query/execute]_iter` is to use them with SELECT statements (here we use INSERT). I was almost sure that we throw an error in case response is not `RESULT:Rows`. But then I saw this in `pager.rs`:
```rust
            Ok(NonErrorQueryResponse {
                response: NonErrorResponse::Result(_),
                tracing_id,
                ..
            }) => {
                // We have most probably sent a modification statement (e.g. INSERT or UPDATE),
                // so let's return an empty iterator as suggested in #631.

                // We must attempt to send something because the iterator expects it.
                let (proof, _) = self.sender.send_empty_page(tracing_id).await;
                Ok(ControlFlow::Break(proof))
            }
```

In such case, it's nice you extended the test cases by `query/execute_iter` scenarios.

scylla/tests/integration/query.rs
```diff
@@ -309,3 +318,698 @@ async fn test_timestamp_generator() {
...
+}
+
+/// ScyllaDB does not distinguish empty collections from nulls. That is, INSERTing an empty collection
+/// is equivalent to nullifying the corresponding column.
+/// As pointed out in [#1001](https://github.com/scylladb/scylla-rust-driver/issues/1001), it's a nice
+/// QOL feature to be able to deserialize empty CQL collections to empty Rust collections instead of
+/// `None::<RustCollection>`. This test checks that.
+#[tokio::test]
+async fn test_deserialize_empty_collections() {
+    // Setup session.
```
muzarski, line 445: I think this test does not belong here. It should be placed in deserialization test suite (the closest one is `integration/cql_types.rs` I believe) cc: @Lorak-mmk

scylla/tests/integration/query.rs
```diff
@@ -309,3 +318,698 @@ async fn test_timestamp_generator() {
...
+    let result = session.execute_iter(prepared_query.clone(), (KEY, 1)).await;
+    assert_matches!(result, Err(PagerExecutionError::SerializationError(_)));
+
+    let result = session
+        .execute_single_page(&prepared_query, (KEY, 1), PagingState::start())
+        .await;
+    assert_matches!(
+        result,
+        Err(ExecutionError::BadQuery(BadQuery::SerializationError(_)))
+    );
```
muzarski, line 590: I'm glad we have some test cases which kinda show the difference between `ExecutionError::BadQuery::SerializationError` and `ExecutionError::LastAttemptError::SerializationError`. I know that it must be confusing that `SerializationError` appears twice in errors hierarchy. This is something we will be trying to fix in the future. But the fix for this is non-trivial, and we won't be able to do that before 1.0.0,

scylla/tests/integration/query.rs
```diff
@@ -309,3 +318,698 @@ async fn test_timestamp_generator() {
...
+            RequestAttemptError::SerializationError(_)
+        ))
+    );
+
+    let prepared_query = session.prepare(query).await.unwrap();
+
+    let result = session.execute_unpaged(&prepared_query, (KEY, 1)).await;
+    assert_matches!(
+        result,
+        Err(ExecutionError::BadQuery(BadQuery::SerializationError(_)))
```
muzarski, line 578: I think that we can do even more specific match here.

```rust
    match result {
        Err(ExecutionError::LastAttemptError(RequestAttemptError::SerializationError(e))) => {
            match e.downcast_ref::<row::BuiltinTypeCheckError>().unwrap().kind {
                row::BuiltinTypeCheckErrorKind::WrongColumnCount {
                    rust_cols: 2,
                    cql_cols: 1,
                } => {}
                _ => panic!("Expected WrongColumnCount error"),
            }
        }
        _ => panic!("Expected WrongColumnCount error"),
    }
```

muzarski: And the same applies to other `SerializationError(_)` matches (when we know what kind of serialization error appears)

muzarski (commented):
Looks good. I'm thinking if `query.rs` is a good name for the module (we are trying to rename the `Query` struct before the 1.0.0 - most probably it will be renamed to `Statement`). I think it should be maybe called `statement.rs`? These tests indeed verify correctness binding values to the statement, and some statement config parameters (e.g. timestamps)

scylla/tests/integration/query.rs
Lorak-mmk: I didn't yet read this PR nor https://github.com/scylladb/scylla-rust-driver/pull/1232 , but please make sure that you are not repeating the same work

sylwiaszunejko:
I changed the file name as suggested, moved `test_deserialize_empty_collections` and made error matches more specific.

scylla/tests/integration/statement.rs
```diff
@@ -309,3 +317,754 @@ async fn test_timestamp_generator() {
...
+
+    let wrongmaps: Vec<HashMap<&str, i32>> = vec![
+        HashMap::from([("pk", 7), ("fefe", 42), ("ck", 13)]),
+        HashMap::from([("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx", 7)]),
+        HashMap::new(),
+        HashMap::from([("ck", 9)]),
+    ];
+    for wrongmap in wrongmaps {
+        assert!(session.execute_unpaged(&prepared, &wrongmap).await.is_err());
+    }
```
Lorak-mmk, line 374: Instead of only checking that it results in an error, could you also verify what this error is?
In other tests we did have situations that unrelated bug (e.g. wrong query string) throws an error, resulting in a test being useless.

Lorak-mmk (changes requested):

scylla/tests/integration/statement.rs
```diff
@@ -309,3 +317,754 @@ async fn test_timestamp_generator() {
...
+            match e
+                .downcast_ref::<serialize::row::BuiltinTypeCheckError>()
+                .unwrap()
+                .kind
+            {
+                serialize::row::BuiltinTypeCheckErrorKind::WrongColumnCount {
+                    rust_cols: 2,
+                    cql_cols: 1,
+                } => {}
+                _ => panic!("Expected WrongColumnCount error"),
```
wprzytula, line 499: :wrench: `assert_matches!` prints more information upon no match, so let's use it.

wprzytula:
```suggestion
            let kind = &e
                .downcast_ref::<serialize::row::BuiltinTypeCheckError>()
                .unwrap();
                .kind;
            assert_matches!(kind, serialize::row::BuiltinTypeCheckErrorKind::WrongColumnCount {
                rust_cols: 2,
                cql_cols: 1,
            });
```

scylla/tests/integration/statement.rs
```diff
@@ -309,3 +317,754 @@ async fn test_timestamp_generator() {
...
+            }
+        }
+        _ => panic!("Expected WrongColumnCount error"),
+    }
```
wprzytula, line 503: `let else` syntax is perfect here.

wprzytula:
```suggestion
    let Err(ExecutionError::LastAttemptError(RequestAttemptError::SerializationError(e))) = result else {
        panic!("Expected WrongColumnCount error");
    };
    match e
        .downcast_ref::<serialize::row::BuiltinTypeCheckError>()
        .unwrap()
        .kind
    {
        serialize::row::BuiltinTypeCheckErrorKind::WrongColumnCount {
            rust_cols: 2,
            cql_cols: 1,
        } => {}
        _ => panic!("Expected WrongColumnCount error"),
    }
```

wprzytula: :wrench: Similarly, the below cases can be made more readable, too.

scylla/tests/integration/statement.rs
```diff
@@ -309,3 +317,754 @@ async fn test_timestamp_generator() {
...
+
+    let result = session
+        .query_single_page(query.clone(), (), PagingState::start())
+        .await;
+    assert_matches!(
+        result,
+        Err(ExecutionError::LastAttemptError(
+            RequestAttemptError::DbError(_, _)
+        ))
+    );
```
wprzytula, line 634: :wrench: Let's verify that the `DbError` variant is what we expect.

wprzytula (changes requested):

scylla/tests/integration/statement.rs
```diff
@@ -0,0 +1,296 @@
...
+        .unwrap()
+        .into_rows_result()
+        .unwrap();
+
+    let timestamps_locked = timestamps.lock().unwrap();
+    assert!(query_rows_result
+        .rows::<(i32, i32, i64)>()
+        .unwrap()
+        .map(|row_result| row_result.unwrap())
+        .all(|(_a, _b, writetime)| timestamps_locked.contains(&writetime)));
```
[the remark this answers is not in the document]
sylwiaszunejko: I didn't write this test, only moved it from different file, I am not sure TBH

scylla/tests/integration/query.rs
```diff
@@ -7,6 +7,21 @@ use scylla::{
...
+    let mut query = Query::new("SELECT * FROM system_schema.tables");
+    query.set_is_idempotent(true);
+    query.set_page_size(42);
+
+    let prepared_statement = session.prepare(query).await.unwrap();
+
+    assert!(prepared_statement.get_is_idempotent());
+    assert_eq!(prepared_statement.get_page_size(), 42);
+}
+
```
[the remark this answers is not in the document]
sylwiaszunejko: I added additional checks, is there something more I should add?

scylla/tests/integration/statement.rs
```diff
@@ -0,0 +1,296 @@
...
+        .await
+        .unwrap();
+
+    session.await_schema_agreement().await.unwrap();
+
+    let query_str = format!("INSERT INTO {}.t_timestamp (a, b) VALUES (?, ?)", ks);
+
+    // test regular query timestamps
+
+    let mut regular_query = Query::new(query_str.to_string());
```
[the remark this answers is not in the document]
sylwiaszunejko: `to_string()` is still there, but there is no `format!` after I started using `USE ks`

Lorak-mmk (approved):

muzarski (approved):
Some commit titles still mention `query.rs` (which we decided to rename to `statement.rs`)

sylwiaszunejko:
I changed the commit message to mention correct file

wprzytula:
Closing in favour of #1900.

PR #1236, opened 2025-02-12
tests: port AUTH tests to ccm

muzarski:
This PR moves existing auth tests from `integration` to `ccm_integration`. The corresponding custom workflow in CI is now deleted.

## Pre-review checklist


Lorak-mmk (approved):

scylla/tests/ccm_integration/ccm/cluster.rs
```diff
@@ -705,6 +705,13 @@ impl Cluster {
         self.updateconf(args).await
     }
 
+    /// Enables the `PasswordAuthenticator` for the cluster.
+    pub(crate) async fn enable_authentication(&self) -> Result<(), Error> {
```
dkropachev, line 709:
```suggestion
    pub(crate) async fn enable_password_authentication(&self) -> Result<(), Error> {
```

muzarski: Makes sense, I'll change it tomorrow. BTW, are there any other kind of authenticators? I've seen this https://opensource.docs.scylladb.com/stable/operating-scylla/security/authentication.html and assumed that only two possible values are `AllowAllAuthenticator` and `PasswordAuthenticator`.

dkropachev: https://github.com/scylladb/scylladb/blob/529ff3efa57553eef6b0239b03b81581b70fb9ed/db/config.cc#L1045-L1051

muzarski: I went with the first suggestion. For now, I don't think we plan on testing other auth methods. If we ever need support for other auth methods, I'll gladly introduce such enum. Currently it would introduce unnecessary noise (and compiler complaining about never constructed variants).

wprzytula (approved):

PR #1237, opened 2025-02-12
Integration test for `CachingSession`

Bouncheck:
Adds the integration test `caching_session::ensure_cache_is_used` for `CachingSession`. The test sends through the `scylla_proxy` several queries and verifies that the number of prepare requests is exactly as expected. If the cache is working correctly no extra prepare requests should appear. If it's not, the assertions will bring attention to that.

## Pre-review checklist

*No public items introduced*
*No docs to change for the test. Docs for CachingSession are separate issue.*
*No open issues regarding integration tests for CachingSession*

### DIFF
--- a/scylla/tests/integration/caching_session.rs
+++ b/scylla/tests/integration/caching_session.rs
@@ -1,0 +1,193 @@
+use std::sync::Arc;
+
+use crate::utils::test_with_3_node_cluster;
+use crate::utils::{setup_tracing, unique_keyspace_name, PerformDDL};
+use scylla::batch::Batch;
+use scylla::batch::BatchType;
+use scylla::client::caching_session::CachingSession;
+use scylla_proxy::RequestOpcode;
+use scylla_proxy::RequestReaction;
+use scylla_proxy::RequestRule;
+use scylla_proxy::ShardAwareness;
+use scylla_proxy::{Condition, ProxyError, Reaction, RequestFrame, TargetShard, WorkerError};
+use tokio::sync::mpsc;
+
+fn consume_current_feedbacks(
+    rx: &mut mpsc::UnboundedReceiver<(RequestFrame, Option<TargetShard>)>,
+) -> usize {
+    std::iter::from_fn(|| rx.try_recv().ok()).count()
+}
+
+#[tokio::test]
+#[cfg(not(scylla_cloud_tests))]
+async fn ensure_cache_is_used() {
+    use scylla::client::execution_profile::ExecutionProfile;
+
+    use crate::utils::SingleTargetLBP;
+
+    setup_tracing();
+    let res = test_with_3_node_cluster(
+        ShardAwareness::QueryNode,
+        |proxy_uris, translation_map, mut running_proxy| async move {
+            let session = scylla::client::session_builder::SessionBuilder::new()
+                .known_node(proxy_uris[0].as_str())
+                .address_translator(Arc::new(translation_map))
+                .build()
+                .await
+                .unwrap();
+
+            let cluster_size: usize = 3;
+            let (feedback_txs, mut feedback_rxs): (Vec<_>, Vec<_>) = (0..cluster_size)
+                .map(|_| mpsc::unbounded_channel::<(RequestFrame, Option<TargetShard>)>())
+                .unzip();
+            for (i, tx) in feedback_txs.iter().cloned().enumerate() {
+                running_proxy.running_nodes[i].change_request_rules(Some(vec![RequestRule(
+                    Condition::and(
+                        Condition::RequestOpcode(RequestOpcode::Prepare),
+                        Condition::not(Condition::ConnectionRegisteredAnyEvent),
+                    ),
+                    RequestReaction::noop().with_feedback_when_performed(tx),
+                )]));
+            }
+
+            let ks = unique_keyspace_name();
+            let rs = "{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}";
+            session
+                .ddl(format!(
+                    "CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {}",
+                    ks, rs
+                ))
+                .await
+                .unwrap();
+            session.use_keyspace(ks, false).await.unwrap();
+            session
+                .ddl("CREATE TABLE IF NOT EXISTS tab (a int, b int, c int, primary key (a, b, c))")
+                .await
+                .unwrap();
+            // Assumption: all nodes have the same number of shards
+            let nr_shards = session
+                .get_cluster_state()
+                .get_nodes_info()
+                .first()
+                .expect("No nodes information available")
+                .sharder()
+                .map(|sharder| sharder.nr_shards.get() as usize)
+                .unwrap_or_else(|| 1); // If there is no sharder, assume 1 shard.
+
+            // Consume all feedbacks so far to ensure we will not count something unrelated.
+            let _feedbacks = feedback_rxs
+                .iter_mut()
+                .map(consume_current_feedbacks)
+                .sum::<usize>();
+
+            let caching_session: CachingSession = CachingSession::from(session, 100);
+
+            let batch_size: usize = 4;
+            let mut batch = Batch::new(BatchType::Logged);
+            for i in 1..=batch_size {
+                let insert_b_c = format!("INSERT INTO tab (a, b, c) VALUES ({}, ?, ?)", i);
+                batch.append_statement(insert_b_c.as_str());
+            }
+            let batch_values: Vec<(i32, i32)> = (1..=batch_size as i32).map(|i| (i, i)).collect();
+
+            // First batch that should generate prepares for each shard.
+            caching_session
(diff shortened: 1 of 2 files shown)


scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+            let ks = unique_keyspace_name();
+
+            let (feedback_txs, mut feedback_rxs): (Vec<_>, Vec<_>) = (0..3)
+                .map(|_| mpsc::unbounded_channel::<(RequestFrame, Option<TargetShard>)>())
+                .unzip();
+            for (i, tx) in feedback_txs.iter().cloned().enumerate() {
+                running_proxy.running_nodes[i].change_request_rules(Some(vec![RequestRule(
+                    Condition::and(Condition::RequestOpcode(RequestOpcode::Prepare),Condition::not(Condition::ConnectionRegisteredAnyEvent)),
+                    RequestReaction::noop().with_feedback_when_performed(tx),
+                )]));
```
wprzytula, line 45: :wrench: This doesn't seem like formatted correctly, even though the format check has passed...
Please format this manually, as it's a pain to read.

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+
+            session.ddl(format!("CREATE KEYSPACE IF NOT EXISTS {} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}", ks)).await.unwrap();
```
wprzytula, line 49: :wrench: ditto: formatting here.

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+            session.use_keyspace(ks, false).await.unwrap();
+            session
```
Lorak-mmk, line 51: Could you split this string into two lines? Here you are running into a nasty edge case in rustfmt, where if a line length exceeds its limit, and its not possible to make it shorted (e.g. a too long string literal) it silently gives up formatting the whole function. This causes your whole test to not be formatted properly.

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+                .ddl("CREATE TABLE IF NOT EXISTS tab (a int, b int, c int, primary key (a, b, c))")
+                .await
+                .unwrap();
+            // Assumption: all nodes have the same number of shards
+            let nr_shards = match session.get_cluster_state().get_nodes_info().first() {
+                Some(node_info) => match node_info.sharder() {
+                    Some(sharder) => sharder.nr_shards,
+                    None => NonZero::new(1).unwrap(), // If there is no sharder, assume 1 shard
+                },
+                None => panic!("No nodes information available"),
```
wprzytula, line 61: :pick:
```suggestion
            let nr_shards = session.get_cluster_state()
                .get_nodes_info()
                .first()
                .expect("No nodes information available")
                .map(|sharder| sharder.nr_shards)
                .unwrap_or(NonZero::new(1).unwrap()); // If there is no sharder, assume 1 shard.
```

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+            prepared_batch_res_rows.sort();
+            let expected_rows = vec![(1, 1, 1), (2, 2, 2), (3, 3, 3), (4, 4, 4)];
+            assert_eq!(prepared_batch_res_rows, expected_rows);
+
+
+            let feedbacks: usize = feedback_rxs.iter_mut().map(consume_current_feedbacks).sum();
+
+            let expected_feedbacks = (4 + 1) * nr_shards.get() as usize * 3; // 4 inserts + 1 select, 3 nodes
+            assert_eq!(feedbacks, expected_feedbacks);
+
```
Lorak-mmk, line 109: Can you verify the feedbacks amount after first execution of the batch, and then again after other executions to verify there are no new prepares?

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+            batch.append_statement(insert_3_b_c);
+            batch.append_statement(insert_4_b_c);
+            // First batch that should generate prepares for each shard.
+            let batch_values = vec![(1, 1), (2, 2), (3, 3), (4, 4)];
+            caching_session.batch(&batch, batch_values.clone()).await.unwrap();
+            // Few extra runs. Those batches should not result in any prepares being sent.
+            caching_session.batch(&batch, batch_values.clone()).await.unwrap();
+            caching_session.batch(&batch, batch_values.clone()).await.unwrap();
+            caching_session.batch(&batch, batch_values.clone()).await.unwrap();
+            caching_session.batch(&batch, batch_values.clone()).await.unwrap();
```
wprzytula, line 84: :pick: Let's use `for` loop instead of repetition here.

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+            let prepared_batch_res_rows: Vec<(i32, i32, i32)> = caching_session
+                .execute_unpaged("SELECT * FROM tab", &[])
+                .await
+                .unwrap()
+                .into_rows_result()
+                .unwrap()
+                .rows()
+                .unwrap()
+                .map(|r| r.unwrap())
+                .collect();
```
wprzytula, line 95: :pick: 
```suggestion
                .collect::<Result<_, _>>()
                .unwrap();
```

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,140 @@
...
+            // Consume those frames.
+            feedback_rxs.iter_mut().map(consume_current_feedbacks).sum::<usize>();
+
+            // Run batches many times per node to ensure we hit every one of them.
+            // Server side cache was invalidated, but the reprepare will be required only if
+            // we hit the node that had its cache invalidated. Only the corresponding shards
+            // will be targeted, so this time expected number of prepares is not multiplied by nr_shards
+            for _ in 0..(64*3) {
+                caching_session.batch(&batch, batch_values.clone()).await.unwrap();
+            }
```
Lorak-mmk, line 126: Instead of making a lot of queries (which can actually fail to reach all shards if you are unlucky) you can use a custom load balancing policy to send a query to each shard exactly once. See tablets.rs integration tests and `SingleTargetLBP` inside it (it is used in `send_statement_everywhere` and `send_unprepared_query_everywhere` helper functions).

Lorak-mmk (changes requested):

wprzytula (approved):

scylla/tests/integration/caching_session.rs
Lorak-mmk: As I mentioned in my comment, this is caused by rustfmt giving up on formatting due to a too long string literal. Breaking this literal up to two lines will fix the issue.

scylla/tests/integration/caching_session.rs
Bouncheck: I've put replication settings in another variable. Looks like `cargo fmt --all` went further this time.
Changes will be visible once pushed.

scylla/tests/integration/caching_session.rs
Bouncheck: I've added more frequent checks. Now the number of prepares is verified after each query. I've also put those numbers into descriptive variables like `batch_size`, `cluster_size`.

scylla/tests/integration/caching_session.rs
Bouncheck: Applied same logic as in `send_statement_everywhere` but in a synchronous manner. Moved some code from `tablets.rs` to `common/utils.rs` in a separate commit to use it.

scylla/tests/integration/caching_session.rs
Bouncheck: Applied, will be visible after push.

scylla/tests/integration/caching_session.rs
Bouncheck: Switched to for loops from 1 to `=batch_size`

Bouncheck:
Pushed a new version. Let me know if should I mark conversations as resolved or if it's on the reviewer if he agrees with the new version.

scylla/tests/common/utils.rs
```diff
@@ -300,3 +300,31 @@ impl PerformDDL for CachingSession {
         self.execute_unpaged(query, &[]).await.map(|_| ())
     }
 }
+
+#[derive(Debug)]
+#[allow(dead_code)]
```
Bouncheck, line 305: I don't understand why I had to add this line here even though I use this struct in my test. Should it not be not a dead_code because of that?

Lorak-mmk:
> Pushed a new version. Let me know if should I mark conversations as resolved or if it's on the reviewer if he agrees with the new version.

In this repo usually the author of the PR marks them as resolved. Is that the best way? I have no idea, maybe the reviewer should do this - we can discuss it in the future.

muzarski (approved):

PR #1246, opened 2025-02-14
Proxy: compresssion support

wprzytula:
# Motivation

During hackathon, there emerged a need to test compressed communication through proxy. However, proxy did not yet support compression.

# What's done

### Exposed some `scylla-cql` compression utilities
These are now used by the proxy.

### New impls
- `impl FromStr for Compression`
- `impl DeserializableRequest for Startup`

The impls are used by the proxy to read the negotiated compression from STARTUP frame.

### `impl (Partial)Eq for {Request,Response}Frame`
It's convenient to be able to compare the whole frame at once.

### Fixed some debug prints
Before, they printed the shard number next to the driver, not the node, which was confusing.

### Stopped abusing `COMPRESSED` frame flag in tests randomly
    
Some tests just set random CQL flags and assert that they are simply passed through the proxy. It used to work, but as now we make the proxy clever enough to interpret at least one of them, namely the compression flag, we needed to stop abusing it blindly.

### Combat timing-based flakiness
    
The test `proxy_processes_requests_concurrently` showed to be flaky on my machine. As its timeout there was arbitrarily tight, I increased it.

## Compression abstractions

`CompressionReader` and `CompressionWriter` work in tandem to propagate the compression that the driver negociated with the node (in STARTUP frame) across the proxy connection workers.

- `CompressionWriter` is used by the `request_processor` to set the compression globally for the whole bunch of proxy workers for that connection once a STARTUP message is intercepted.
- `CompressionReader` is used by all workers that perform ser/de (`{sender_to,receiver_from}_{driver,cluster}`) to (de)compress frames as they come. **Proxy always stores the frames and feeds them into feedback channels in the decompressed form, so that it is for example possible to use `Condition`s based on uncompressed body contents.**

##  Unit test for compression mechanisms in the proxy
    
Outline of the test:
1. "driver" sends an, uncompressed, e.g., QUERY frame, feedback returns its uncompressed body, and "node" receives the uncompressed frame.
2. "node" responds with an uncompressed RESULT frame, feedback returns its uncompressed body, and "driver" receives the uncompressed frame.
3. "driver" sends an uncompressed STARTUP frame, feedback returns its uncompressed body, and "node" receives the uncompressed frame.
4. "driver" sends a compressed, e.g., QUERY frame, feedback returns its uncompressed body, and "node" receives the compressed frame.
5. "node" responds with a compressed RESULT frame, feedback returns its uncompressed body, and "driver" receives the compressed frame.

## Pre-review checklist

### DIFF
--- a/scylla-proxy/src/frame.rs
+++ b/scylla-proxy/src/frame.rs
@@ -321,7 +321,9 @@ pub(crate) async fn read_frame(
         }
     }
 
-    Ok((frame_params, opcode, body.into_inner().into()))
+    let body = compression.maybe_decompress_body(flags, body.into_inner().into())?;
+
+    Ok((frame_params, opcode, body))
 }
 
 pub(crate) async fn read_request_frame(
--- a/scylla-proxy/src/frame.rs
+++ b/scylla-proxy/src/frame.rs
@@ -62,12 +62,14 @@ impl RequestFrame {
     pub(crate) async fn write(
         &self,
         writer: &mut (impl AsyncWrite + Unpin),
+        compression: &CompressionReader,
     ) -> Result<(), tokio::io::Error> {
         write_frame(
             self.params,
             FrameOpcode::Request(self.opcode),
             &self.body,
             writer,
+            compression,
         )
         .await
     }
@@ -136,12 +138,14 @@ impl ResponseFrame {
     pub(crate) async fn write(
         &self,
         writer: &mut (impl AsyncWrite + Unpin),
+        compression: &CompressionReader,
     ) -> Result<(), tokio::io::Error> {
         write_frame(
             self.params,
             FrameOpcode::Response(self.opcode),
             &self.body,
             writer,
+            compression,
         )
         .await
     }
@@ -235,6 +239,7 @@ pub(crate) async fn write_frame(
     opcode: FrameOpcode,
     body: &[u8],
     writer: &mut (impl AsyncWrite + Unpin),
+    compression: &CompressionReader,
 ) -> Result<(), tokio::io::Error> {
     let mut header = [0; HEADER_SIZE];
 
--- a/scylla-proxy/src/frame.rs
+++ b/scylla-proxy/src/frame.rs
@@ -11,6 +11,8 @@ use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 
 use tracing::warn;
 
+use crate::errors::ReadFrameError;
+
 const HEADER_SIZE: usize = 9;
 
 // Parts of the frame header which are not determined by the request/response type.
@@ -253,7 +255,7 @@ pub(crate) async fn write_frame(
 pub(crate) async fn read_frame(
     reader: &mut (impl AsyncRead + Unpin),
     frame_type: FrameType,
-) -> Result<(FrameParams, FrameOpcode, Bytes), FrameHeaderParseError> {
+) -> Result<(FrameParams, FrameOpcode, Bytes), ReadFrameError> {
     let mut raw_header = [0u8; HEADER_SIZE];
     reader
         .read_exact(&mut raw_header[..])
@@ -269,7 +271,7 @@ pub(crate) async fn read_frame(
             FrameType::Response => (FrameHeaderParseError::FrameFromClient, 0x80, "response"),
         };
         if version & 0x80 != valid_direction {
-            return Err(err);
+            return Err(err.into());
         }
         let protocol_version = version & 0x7F;
         if protocol_version != 0x04 {
@@ -311,10 +313,9 @@ pub(crate) async fn read_frame(
             .map_err(|err| FrameHeaderParseError::BodyChunkIoError(body.remaining_mut(), err))?;
         if n == 0 {
             // EOF, too early
-            return Err(FrameHeaderParseError::ConnectionClosed(
-                body.remaining_mut(),
-                length,
-            ));
+            return Err(
+                FrameHeaderParseError::ConnectionClosed(body.remaining_mut(), length).into(),
+            );
         }
     }
 
@@ -323,7 +324,7 @@ pub(crate) async fn read_frame(
 
 pub(crate) async fn read_request_frame(
     reader: &mut (impl AsyncRead + Unpin),
-) -> Result<RequestFrame, FrameHeaderParseError> {
+) -> Result<RequestFrame, ReadFrameError> {
     read_frame(reader, FrameType::Request)
         .await
         .map(|(params, opcode, body)| RequestFrame {
@@ -338,7 +339,7 @@ pub(crate) async fn read_request_frame(
 
 pub(crate) async fn read_response_frame(
     reader: &mut (impl AsyncRead + Unpin),
-) -> Result<ResponseFrame, FrameHeaderParseError> {
+) -> Result<ResponseFrame, ReadFrameError> {
     read_frame(reader, FrameType::Response)
         .await
         .map(|(params, opcode, body)| ResponseFrame {
--- a/scylla-proxy/src/frame.rs
+++ b/scylla-proxy/src/frame.rs
@@ -22,13 +22,13 @@ pub struct FrameParams {
 }
 
 impl FrameParams {
-    pub fn for_request(&self) -> FrameParams {
+    pub const fn for_request(&self) -> FrameParams {
         Self {
             version: self.version & 0x7F,
             ..*self
         }
     }
-    pub fn for_response(&self) -> FrameParams {
+    pub const fn for_response(&self) -> FrameParams {
         Self {
             version: 0x80 | (self.version & 0x7F),
             ..*self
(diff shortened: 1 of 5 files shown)


Lorak-mmk:
Is that useful for @fruch 's compression tests?

fruch:
> Is that useful for @fruch 's compression tests?

In my tests I wasn't reading the content, just using the total sizes of  frames

I can help as extra validation that it's compressed, and which algorithm was used

wprzytula:
> > Is that useful for @fruch 's compression tests?
> 
> In my tests I wasn't reading the content, just using the total sizes of frames
> 
> I can help as extra validation that it's compressed, and which algorithm was used

Yes: once proxy detects COMPRESSION key in STARTUP frame, it starts to interpret further frames having compression flag set as compressed, and decompresses them automatically using the negotiated compression. So it proxy fails to decompress a frame with an error, then we now that the frame was not compressed even though it should, or it was ill-compressed (perhaps with a wrong compression method).

scylla-proxy/src/proxy.rs
```diff
@@ -2715,4 +2725,193 @@ mod tests {
...
+                .unwrap();
+
+            let (captured_frame, _) = request_feedback_rx.recv().await.unwrap();
+            assert_eq!(captured_frame, sent_frame);
+
+            let received_frame = read_request_frame(&mut node_conn, &no_compression())
+                .await
+                .unwrap();
+            assert_eq!(received_frame, sent_frame);
+        }
```
muzarski, line 2865: This step also triggers the `CompressionWriter::set()` in the proxy, so the associated `CompressionReader`s are notified about it (and can use negotiated compression algorithm to (de)compress the frames sent in steps 4. and 5.), correct? If so, I think this deserves to be put in the comment.

muzarski (changes requested):

scylla-proxy/src/frame.rs
```diff
@@ -56,7 +56,7 @@ pub struct RequestFrame {
 }
 
 impl RequestFrame {
-    pub async fn write(
+    pub(crate) async fn write(
```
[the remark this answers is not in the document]
wprzytula: I'm not sure that `scylla-proxy` used to comply with semver. Anyway, releasing a new major makes perfect sense.

wprzytula:
Rebased on main.

scylla-proxy/src/proxy.rs
```diff
@@ -932,6 +945,26 @@ mod compression {
...
+                    compress_append(body, compression, &mut buf).map_err(|err| {
+                        let CqlRequestSerializationError::SnapCompressError(err) = err else {unreachable!("BUG: compress_append returned variant different than SnapCompressError")};
+                        CompressionError::SnapCompressError(err)
+                    })?;
+                    Ok(Some(Bytes::from(buf)))
+                }
+                (true, None) => Err(CompressionError::NoCompressionNegotiated),
+                (false, _) => Ok(None),
+            }
+        }
```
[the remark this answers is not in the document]
wprzytula: The reason why it's done this way in the compress case is that I want to return a narrow enough error type, and `CqlRequestSerializationError` is quite the opposite. Therefore, I explicitly expect `SnapCompressError` (which is not very robust, but I guess it's acceptable in a testing tool, which proxy is) variant and extract its inner error into proxy-defined `CompressionError`.

OTOH, `decompress()` already returns a narrow error type: `FrameBodyExtensionsParseError`, which is perfect for my case. So I don't have to perform any extraction in the decompress case.

wprzytula:
Addressed @muzarski comments. Thanks!

scylla-proxy/src/frame.rs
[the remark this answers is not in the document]
Lorak-mmk: But why would we need `0.1.0`? `0.0.5` is also a new major.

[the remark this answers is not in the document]
wprzytula: semver says:
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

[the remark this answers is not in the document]
wprzytula: So it seems that it was just us that had the convention that in `0.y.z`, `y` was the major and `z` was the minor version number.

muzarski (approved):
LGTM!

scylla-proxy/src/proxy.rs
```diff
@@ -2587,9 +2602,10 @@ mod tests {
             .await
             .unwrap();
 
-            let received_response = read_frame(client_socket_ref, FrameType::Response, &no_compression())
-                .await
-                .unwrap();
+            let received_response =
+                read_frame(client_socket_ref, FrameType::Response, &no_compression())
+                    .await
+                    .unwrap();
```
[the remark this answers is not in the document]
muzarski: Yes, the commits that introduce (de)compression logic are much cleaner now,

scylla-proxy/src/frame.rs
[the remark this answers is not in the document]
Lorak-mmk: > So it seems that it was just us that had the convention that in 0.y.z, y was the major and z was the minor version number.

This is also incorrect. Official semver spec differs from the semver used in Cargo. See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements

> Versions are considered compatible if their left-most non-zero major/minor/patch component is the same. This is different from [SemVer](https://semver.org/) which considers all pre-1.0.0 packages to be incompatible.

scylla-proxy/src/proxy.rs
```diff
@@ -633,12 +633,25 @@ impl Doorkeeper {
...
+            compression_reader_receiver_from_cluster,
+            compression_reader_sender_to_driver,
+            compression_reader_sender_to_cluster,
+        ) = compression::make_compression_infra();
+
+        tokio::task::spawn(new_worker().receiver_from_driver(
+            driver_read,
+            tx_request,
+            compression_reader_receiver_from_driver,
+        ));
```
Lorak-mmk, line 649: 🤔 This is not the first time we need to share some state between the tasks - we already share event_register_flag. I think it is probable that we will need more stuff in the future, for example if we decide to add more advanced Conditions (like those matching the query string of prepared statement).

Maybe we should have some `ConnectionState` and/or  `ProxyState` for the shared state? They would be shared and could contain `OnceLock<Option<Compression>>`, atomic bool for events etc. It would make it easier to add new such items in the future.

scylla-proxy/src/proxy.rs
```diff
@@ -848,6 +867,61 @@ impl Doorkeeper {
...
+            CompressionWriter(info.clone()),
+            CompressionReader(info.clone()),
+            CompressionReader(info.clone()),
+            CompressionReader(info.clone()),
+            CompressionReader(info),
+        )
+    }
+}
+pub(crate) use compression::{CompressionReader, CompressionWriter};
+
```
Lorak-mmk: Why so much code? Why do you need separate types?
This is not a library abstraction that will be reused, the compression is used in just a few places. It is also only a test-related code, not something used in "production", so the ease of reading / editing is more important than full bug-proofing.

Can we just have `get` and `set` methods on `CompressionInfo`? 
It would also remove the need for `make_compression_infra`.

scylla-proxy/src/proxy.rs
```diff
@@ -942,6 +942,19 @@ mod compression {
...
+            let once = OnceLock::new();
+            once.set(compression).unwrap();
+            once
+        }))
+    }
+
+    // Compression explicitly turned off.
+    pub(crate) fn no_compression() -> CompressionReader {
+        mock_compression_reader(None)
+    }
```
Lorak-mmk, line 1009: The only uses of this seem to be in tests. Why not move this to tests, or at least mark as cfg(test)?

Lorak-mmk (changes requested):

scylla-proxy/src/proxy.rs
wprzytula: `no_compression()` is used in multiple places out of tests, including `Doorkeeper`.

scylla-proxy/src/proxy.rs
wprzytula: > Why so much code? Why do you need separate types?

Rust, in general, promotes encoding information in types. This is not only for hardening and bug-proofing; for me, this makes code comprehension smoother. Once I see a `CompressionReader` being parameter of a function, I instantly know that the function only reads compression. Same for `CompressionWriter`. This makes analysis of compression-related logic much simpler. What's the cost? Some boilerplate code, which is all limited to one extracted module. I believe this is a right way to write non-prototyping code in Rust, which the proxy is.

If you read the code top-down, then existence of `CompressionReader/Writer` does not, at least for me, decrease readability; it's the opposite. What _might_ decrease readability could be diving into details of the `compression` module, but that's no longer top-down approach to understanding code.

scylla-proxy/src/proxy.rs
wprzytula: Makes sense. This is however out of scope of this PR. I can try to refactor this according to your suggestion in a follow-up.

Lorak-mmk (approved):

scylla-proxy/src/proxy.rs
Lorak-mmk: Please do so, or at least open an issue about it.

Issue #1152, opened 2024-12-16
Remove uses of SerializedValues from public API


Lorak-mmk:
Those uses are a leftover from previous API and should be removed - most likely replaced with `impl SerializeRow`

Usages to be removed:
https://docs.rs/scylla/latest/scylla/transport/partitioner/fn.calculate_token_for_partition_key.html
https://docs.rs/scylla/latest/scylla/transport/struct.ClusterData.html#method.compute_token
https://docs.rs/scylla/latest/scylla/transport/struct.ClusterData.html#method.get_endpoints

Lorak-mmk:
I think it has dependency on https://github.com/scylladb/scylla-rust-driver/issues/691
The reason is that we need to create `RowSerializationContext`, which operates on `ColumnSpec` (which contains `ColumnType`), using `Table`, containing `pub columns: HashMap<String, Column>,` - and `Column` contains `CqlType`.

PR #1252, opened 2025-02-18
Remove SerializedValues from public API of `scylla` crate.

Lorak-mmk:
Some APIs on ClusterState still accept SerializedValues instead of `impl SerializeRow` / `&dyn SerializeRow`.
This is a leftover from the old serialization API, and should be changed. Why? Because there is no easy way for the user to create the SerializedValues.
That means that if a user wants to calculate the token / replicas, for some partition key of some table they need to:
- Get the `Table` struct from `ClusterState`
- Create new SerializedValues
- Iterate over `partition_key` field and retrieve relevant `ColumnSpec`s from `columns` field.
- For each column, call `add_value` on the SerializedValues and handle the error
- Pass the result to `compute_token` / `get_endpoints`

This is a lot of unnecessary work.
This PR changes those methods to accepts `&dyn SerializeRow`.

As a small side change I removed `RowSerializationContext::column_by_name`.
Columns need to be serialized in order, so there is no reason to get the column out of order.
This is confirmed by this method being unused.

## Creating RowSerializationContext from Table

See the commit "Table: Add Vec of primary key column specs" for an explanation.
I'm not sure that the approach I've taken is the best one, so I'm open t suggestions in this matter.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1152

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/metadata.rs
+++ b/scylla/src/cluster/metadata.rs
@@ -77,6 +77,10 @@ type PerKsTableResult<T, E> = PerKsTable<Result<T, E>>;
 pub(crate) enum SingleKeyspaceMetadataError {
     #[error(transparent)]
     MissingUDT(MissingUserDefinedType),
+    #[error("Partition key column with position {0} is missing from metadata")]
+    IncompletePartitionKey(i32),
+    #[error("Clustering key column with position {0} is missing from metadata")]
+    IncompleteClusteringKey(i32),
 }
 
 /// Allows to read current metadata from the cluster
@@ -1606,7 +1610,7 @@ async fn query_tables_schema(
     let mut all_partitioners = query_table_partitioners(conn).await?;
     let mut result = HashMap::new();
 
-    for ((keyspace_name, table_name), table_result) in tables_schema {
+    'tables_loop: for ((keyspace_name, table_name), table_result) in tables_schema {
         let keyspace_and_table_name = (keyspace_name, table_name);
 
         #[allow(clippy::type_complexity)]
@@ -1621,15 +1625,51 @@ async fn query_tables_schema(
                 continue;
             }
         };
-        let mut partition_key = vec!["".to_string(); partition_key_columns.len()];
-        for (position, column_name) in partition_key_columns {
-            partition_key[position as usize] = column_name;
+
+        fn validate_key_columns(mut key_columns: Vec<(i32, String)>) -> Result<Vec<String>, i32> {
+            key_columns.sort_unstable_by_key(|(position, _)| *position);
+
+            key_columns
+                .into_iter()
+                .enumerate()
+                .map(|(idx, (position, column_name))| {
+                    // unwrap: I don't see the point of handling the scenario of fetching over
+                    // 2 * 10^9 columns.
+                    let idx: i32 = idx.try_into().unwrap();
+                    if idx == position {
+                        Ok(column_name)
+                    } else {
+                        Err(idx)
+                    }
+                })
+                .collect::<Result<Vec<_>, _>>()
         }
 
-        let mut clustering_key = vec!["".to_string(); clustering_key_columns.len()];
-        for (position, column_name) in clustering_key_columns {
-            clustering_key[position as usize] = column_name;
-        }
+        let partition_key = match validate_key_columns(partition_key_columns) {
+            Ok(partition_key_columns) => partition_key_columns,
+            Err(position) => {
+                result.insert(
+                    keyspace_and_table_name,
+                    Err(SingleKeyspaceMetadataError::IncompletePartitionKey(
+                        position,
+                    )),
+                );
+                continue 'tables_loop;
+            }
+        };
+
+        let clustering_key = match validate_key_columns(clustering_key_columns) {
+            Ok(clustering_key_columns) => clustering_key_columns,
+            Err(position) => {
+                result.insert(
+                    keyspace_and_table_name,
+                    Err(SingleKeyspaceMetadataError::IncompleteClusteringKey(
+                        position,
+                    )),
+                );
+                continue 'tables_loop;
+            }
+        };
 
         let partitioner = all_partitioners
             .remove(&keyspace_and_table_name)
--- a/scylla/src/cluster/metadata.rs
+++ b/scylla/src/cluster/metadata.rs
@@ -35,6 +35,7 @@ use futures::Stream;
 use itertools::Itertools;
 use rand::seq::{IndexedRandom, SliceRandom};
 use rand::{rng, Rng};
+use scylla_cql::frame::response::result::{ColumnSpec, TableSpec};
 use scylla_macros::DeserializeRow;
 use std::borrow::BorrowMut;
 use std::cell::Cell;
@@ -197,13 +198,14 @@ pub struct Keyspace {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Table {
     pub columns: HashMap<String, Column>,
-    /// Names of the column of partition key. 
+    /// Names of the column of partition key.
     /// All of the names are guaranteed to be present in `columns` field.
     pub partition_key: Vec<String>,
-    /// Names of the column of clustering key. 
+    /// Names of the column of clustering key.
     /// All of the names are guaranteed to be present in `columns` field.
     pub clustering_key: Vec<String>,
     pub partitioner: Option<String>,
+    pub(crate) pk_column_specs: Vec<ColumnSpec<'static>>,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -1410,6 +1412,7 @@ async fn query_tables(
             partition_key: vec![],
             clustering_key: vec![],
             partitioner: None,
+            pk_column_specs: vec![],
         }));
 
         let mut entry = result
@@ -1461,6 +1464,7 @@ async fn query_views(
                 partition_key: vec![],
                 clustering_key: vec![],
                 partitioner: None,
+                pk_column_specs: vec![],
             }))
             .map(|table| MaterializedView {
                 view_metadata: table,
@@ -1679,6 +1683,20 @@ async fn query_tables_schema(
             .remove(&keyspace_and_table_name)
             .unwrap_or_default();
 
+        // unwrap of get() result: all column names in `partition_key` are at this
+        // point guaranteed to be present in `columns`. See the construction of `partition_key`
+        let pk_column_specs = partition_key
+            .iter()
+            .map(|column_name| (column_name, columns.get(column_name).unwrap().clone().typ))
+            .map(|(name, typ)| {
+                let table_spec = TableSpec::owned(
+                    keyspace_and_table_name.0.clone(),
+                    keyspace_and_table_name.1.clone(),
+                );
+                ColumnSpec::owned(name.to_owned(), typ, table_spec)
+            })
+            .collect();
+
         result.insert(
             keyspace_and_table_name,
             Ok(Table {
@@ -1686,6 +1704,7 @@ async fn query_tables_schema(
                 partition_key,
                 clustering_key,
                 partitioner,
+                pk_column_specs,
             }),
         );
     }
--- a/scylla/src/cluster/metadata.rs
+++ b/scylla/src/cluster/metadata.rs
@@ -67,6 +67,18 @@ type PerTable<T> = HashMap<String, T>;
 type PerKsTable<T> = HashMap<(String, String), T>;
 type PerKsTableResult<T, E> = PerKsTable<Result<T, E>>;
 
+/// Indicates that reading metadata failed, but in a way
+/// that we can handle, by throwing out data for a keyspace.
+/// It is possible that some of the errors could be handled in even
+/// more granular way (e.g. throwing out a single table), but keyspace
+/// granularity seems like a good choice given how independent keyspaces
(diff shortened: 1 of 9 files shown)


scylla/src/cluster/state.rs
```diff
@@ -203,17 +203,22 @@ impl ClusterState {
...
-        let partitioner = self
+        partition_key: &dyn SerializeRow,
+    ) -> Result<Token, PartitionKeyError> {
+        let table = self
             .keyspaces
             .get(keyspace)
-            .and_then(|k| k.tables.get(table))
+            .and_then(|k| k.tables.get(table));
+        let values = SerializedValues::from_serializable(
+            &RowSerializationContext::from_specs(table.unwrap().pk_column_specs.as_slice()),
```
muzarski, line 213: This unwrap is not ok. We should probably return an error if we don't have metadata for given `keyspace.table`. Now I wonder if extending `PartitionKeyError` by additional enum is fine, or we should rather introduce new error type. New variant would never be constructed in `PreparedStatement` API - e.g. `PreparedStatement::compute_token()`.

BTW, previously user could provide some garbage keyspace and table names, and we would still compute the token (with the default partitioner - murmur3). I wonder if that was OK.

scylla/src/cluster/state.rs
```diff
@@ -203,17 +203,22 @@ impl ClusterState {
         &self,
         keyspace: &str,
         table: &str,
-        partition_key: &SerializedValues,
-    ) -> Result<Token, TokenCalculationError> {
-        let partitioner = self
+        partition_key: &dyn SerializeRow,
```
muzarski, line 206: We accept `&dyn SerializeRow` here, but `impl SerializeRow` in `Session::[query/execute]_*` API. We already discussed it shortly on the meeting, but I want to bring attention to this in case it requires further discussion.

muzarski (commented):
I haven't yet thought of alternative way to solve the `RowSerializationContext` - `Table` issue. I'll do it later.

scylla/src/cluster/state.rs
Lorak-mmk: Ooops, forgot about this unwrap, sorry.
I'll let you judge if it is better to introduce new error type or extend PartitionKeyError. I lean slightly towards extending, but it's not a strong opinion.

> BTW, previously user could provide some garbage keyspace and table names, and we would still compute the token (with the default partitioner - murmur3). I wonder if that was OK.

Imo it was not ok. It would default to murmur partitioner, so it would silently break if trying to calculate a token for non-existent CDC table.

muzarski: > Ooops, forgot about this unwrap, sorry. I'll let you judge if it is better to introduce new error type or extend PartitionKeyError. I lean slightly towards extending, but it's not a strong opinion.

Let's extend the `PartitionKeyError`. I don't think there's much sense in introducing the distinction and having two separate types. Both sets of methods are related to computation of token based on PK anyway.

Lorak-mmk: Two questions:
- There is `into_execution_error` method on `PartitionKeyError`. I don't really see how to convert this new variant. I don't see how to solve that. Maybe we do need new type after all?
- Should I include the name of the table and keyspace that are missing in the error? Since this is user-facing API this doesn't sound very useful (because user knows what names they passed), and it would require an allocation to store those Strings.

muzarski: > Two questions:
> 
>     * There is `into_execution_error` method on `PartitionKeyError`. I don't really see how to convert this new variant. I don't see how to solve that. Maybe we do need new type after all?

Oh, true. I totally forgot about it. Yeah, I think that this strongly suggests us to introduce a new type (`ClusterStateError`??? - maybe you will come up with some better name). 

>     * Should I include the name of the table and keyspace that are missing in the error? Since this is user-facing API this doesn't sound very useful (because user knows what names they passed), and it would require an allocation to store those Strings.

I'd include them nonetheless. There might be some complex use cases where ks and table names are not passed directly as literals, but are dynamically computed in some manner.

scylla/src/cluster/metadata.rs
```diff
@@ -196,6 +197,7 @@ pub struct Table {
     pub partition_key: Vec<String>,
     pub clustering_key: Vec<String>,
     pub partitioner: Option<String>,
+    pub(crate) pk_column_specs: Vec<ColumnSpec<'static>>,
```
wprzytula, line 200: :ok: I believe the taken approach is perfectly acceptable.

:thought_balloon: What I'm not convinced about is having a HashMap of columns. HashMaps are generally heavyweight structures, whereas tables never contain that many columns that HashMap's performance would be better than Vec's.

:wrench: My another concern are metadata-related structs in `metadata.rs` which are `pub` and have `pub` fields. For now, at least `Column` and `MaterializedView` are such structs and are not `#[non_exhaustive]`, which might be a potential issue in the future.
I strongly suggest `pub(crate)`'ing all those fields and exposing only getters for them. This leaves maximum freedom for us in the future.

scylla/src/cluster/state.rs
Lorak-mmk: @wprzytula wdyt, should we change session and caching session that way?
It makes sense especially for `do_query_iter`, where we can't just serialize in the beginning (like we do for execute_*) because we don't have context yet. This may result in a lot of code duplication.
Serializing will be just one virtual call per execution, it should not affect performance in any way.

scylla/src/cluster/metadata.rs
Lorak-mmk: > This would also allow us to implement [`cass_table_meta_column`](https://github.com/scylladb/cpp-rust-driver/blob/master/include/cassandra.h#L3688-L3699)

It is possible (but problematic) to implement it now.
Afaik, the order of the columns is always:
- partition key columns (in order defined in the key)
- clustering key columns (in order defined in the key)
- the rest of the column, in alphabetical order

cass_table_meta_column could sort the columns according to this, and retrieve the nth one. This is of course really ineffective.

> 💭 What I'm not convinced about is having a HashMap of columns. HashMaps are generally heavyweight structures, whereas tables never contain that many columns that HashMap's performance would be better than Vec's.

Are you talking about the performance of retrieving columns by name? Do you know at which size hashmaps start outperforming hashmaps (key length needs to be taken into consideration)?
Afaik it is not very unusual to have dozens or even hundreds columns in a table - Scylla / Cassandra design promotes using denormalized data.

Lorak-mmk: > 🔧 My another concern are metadata-related structs in metadata.rs which are pub and have pub fields. For now, at least Column and MaterializedView are such structs and are not #[non_exhaustive], which might be a potential issue in the future.
I strongly suggest pub(crate)'ing all those fields and exposing only getters for them. This leaves maximum freedom for us in the future.

Makes sense.

Lorak-mmk: Regarding HashMaps vs Vec, the argument that convinces me more than performance is that we don't expose all the information about the table (we don't expose column order used by Scylla).
Do you have an idea what should the `Table` struct look like?

scylla/src/cluster/state.rs
wprzytula: I believe virtual calls are cheap enough not to influence the request execution time if done once or a couple of times on the execution path. At the same time, monomorphised code bloat clutters cache lines on the request path.
My conclusion: where possible, let's move to `&dyn`.

scylla/src/cluster/metadata.rs
```diff
@@ -1606,13 +1610,26 @@ async fn query_tables_schema(
...
+            .iter()
+            .map(|column_name| (column_name, columns.get(column_name).unwrap().clone().typ))
+            .map(|(name, typ)| {
+                let table_spec = TableSpec::owned(
+                    keyspace_and_table_name.0.clone(),
+                    keyspace_and_table_name.1.clone(),
+                );
+                ColumnSpec::owned(name.to_owned(), typ, table_spec)
+            })
+            .collect();
```
wprzytula, line 1623: :wrench: Convince me that this `unwrap()` won't panic.

scylla/src/lib.rs
```diff
@@ -163,10 +163,6 @@ pub mod serialize {
             BuiltinSerializationError, BuiltinSerializationErrorKind, BuiltinTypeCheckError,
             BuiltinTypeCheckErrorKind,
         };
-
-        // Not part of the old framework, but something that we should
-        // still aim to remove from public API.
-        pub use scylla_cql::serialize::row::{SerializedValues, SerializedValuesIterator};
```
wprzytula, line 169: :question: Aren't these used by benchmarks?

Lorak-mmk: Benchmarks import them using `scylla_cql::` path instead of those re-exports.

scylla/src/cluster/metadata.rs
Lorak-mmk: Interesting, I can't - because there is actually no guarantee that it won't be tripped!

How the code works:
- First it fetches the columns table. Each row of this table describes a column in some table.
- Each such row has a column type, and a "position" field, which tells the order of the column among the column of the same type.
- We collect those columns in a hash map per type - so a hashmap <int, name> for pk, similar hashmap for ck, and a hashmap from name to `Column` for all column (this one does not use "position").
- After processing all of the rows, we move to next stage.

In the next stage, we transform those hashmaps into the vectors that you can see in the `Table` struct.
We do this by creating vector of strings, with the same size as respective hashmap, initialized to empty strings.
Then we assign data from hashmap to the vec:
```rust
        let mut partition_key = vec!["".to_string(); partition_key_columns.len()];
        for (position, column_name) in partition_key_columns {
            partition_key[position as usize] = column_name;
        }
```

The problem here is that there is no guarantee that all indexes from `0` to `partition_key_columns.len() - 1` will be present in the hashmap!
What if Scylla returns columns with pk positions [0, 1, 3]? Then my unwrap would fail.

Note that this is a pre-existing problem. As a user of our Table struct ( https://docs.rs/scylla/latest/scylla/transport/topology/struct.Table.html ),
I would expect that I could safely do this:
```
for name in table.partition_key.iter() {
    let column = table.columns.get(name).unwrap();
    <some code that uses column>
}
```

but this is not the case - theoretically some of the partition key column names could be empty (and thus not present in `columns` field).



How to solve this?

A few PRs ago, I introduced the possibility of erroring out a single keyspace during metadata fetch, instead of whole fetch.
It was then only used for `MissingUserDefinedType`, and so some methods in `metadata.rs` now return types that contain this error.

I think we should extend this. Instead of using `MissingUserDefinedType`, I would introduce `SingleKeyspaceMetadataError` (non-pub), which for now would have variants:
- MissingUserDefinedType
- some error for non-complete column positions

Then we would verify that we received all the columns, and if not - error out the problematic keyspace.

WDYT (cc @muzarski )?

wprzytula: > Interesting, I can't - because there is actually no guarantee that it won't be tripped!

This was my guess.

> I think we should extend this. Instead of using MissingUserDefinedType, I would introduce SingleKeyspaceMetadataError (non-pub), which for now would have variants:
>
>  - MissingUserDefinedType
>  - some error for non-complete column positions
>
> Then we would verify that we received all the columns, and if not - error out the problematic keyspace.

I believe this is a good solution.

muzarski: > WDYT (cc @muzarski )?

Sounds good, let's do that

scylla/src/cluster/state.rs
```diff
@@ -203,17 +202,29 @@ impl ClusterState {
...
```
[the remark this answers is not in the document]
Lorak-mmk: Done

Lorak-mmk:
Rebased on main

Lorak-mmk:
Incorporated changes proposed by @wprzytula in https://github.com/Lorak-mmk/scylla-rust-driver/pull/13

scylla/src/cluster/state.rs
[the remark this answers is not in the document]
wprzytula: For me it's enough, GJ.

scylla/src/lib.rs
```diff
@@ -257,3 +257,22 @@ pub(crate) mod utils;
...
+    pub fn calculate_token_for_partition_key(
+        serialized_partition_key_values: &SerializedValues,
+        partitioner: &PartitionerName,
+    ) -> Result<Token, TokenCalculationError> {
+        crate::routing::partitioner::calculate_token_for_partition_key(
+            serialized_partition_key_values,
+            partitioner,
+        )
+    }
+}
```
wprzytula, line 274: :question: Once we have this, can we deduplicate some test utilities that used to be present for both `scylla` unit tests and integration tests?

wprzytula (approved):

scylla/src/lib.rs
Lorak-mmk: Theoretically we could deduplicate them by only having them in the lib (instead of having them in integration tests) under this feature, like we have `calculate_token_for_partition_key` now.
This is a tradeoff: running integration tests would require passing a flag to `cargo test` to enable this feature, otherwise the integration tests target will not exist at all.
It's not a problem for CI, nor when using Makefile, but it is making it less convenient to call `cargo test` directly.

For benchmark doing this seemed ok to me (because of how rarely we run them). For integration tests I'm not sure.

PR #1255, opened 2025-02-21
Update `rand` crate 0.8.5 -> 0.9.0

gsson:
Also updates the following to align versions and avoid duplicates in transitive dependencies
* `rand_pcg` 0.3.1 -> 0.9.0
* `rand_chacha` 0.3.1 -> 0.9.0

MSRV for all changes is 1.63, so should hopefully be good

I'm currently trying to reduce the number of duplicate dependencies to reduce some bloat in my projects, and `rand` is one of them.

## Pre-review checklist

### DIFF
--- a/Cargo.lock.msrv
+++ b/Cargo.lock.msrv
@@ -26,7 +26,7 @@ dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.34",
 ]
 
 [[package]]
@@ -651,7 +651,19 @@ checksum = "c4567c8db10ae91089c99af84c68c38da3ec2f087c3f82960bcdbf3656b6f4d7"
 dependencies = [
  "cfg-if",
  "libc",
- "wasi",
+ "wasi 0.11.0+wasi-snapshot-preview1",
+]
+
+[[package]]
+name = "getrandom"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43a49c392881ce6d5c3b8cb70f98717b7c07aabbdff06687b9030dbfbe2725f8"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "wasi 0.13.3+wasi-0.2.2",
+ "windows-targets 0.52.5",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


gsson:
Assume I need to update Cargo.lock.msrv 🤔 

I'll get on that shortly!

muzarski:
> Assume I need to update Cargo.lock.msrv 🤔
> 
> I'll get on that shortly!

Update procedure is documented here: https://github.com/scylladb/scylla-rust-driver/blob/main/CONTRIBUTING.md#min_rust-workflow-and-cargolockmsrv

Lorak-mmk:
Clippy failed, but that's most likely not related to your changes, but to the recent release of Rust 1.85 - we will have to fix that.
min_rust failed because you updated dependencies without updating the lock file for MSRV. In contributing.md there are instructions for fixing that: https://github.com/scylladb/scylla-rust-driver/blob/main/CONTRIBUTING.md#min_rust-workflow-and-cargolockmsrv . Could you do that?

One other minor comment: in this case I'd prefer all the changes in the single commit. There are not many of them, so it won't make the commit difficult to read, but it will prevent having commits with deprecation warnings.

gsson:
> One other minor comment: in this case I'd prefer all the changes in the single commit. There are not many of them, so it won't make the commit difficult to read, but it will prevent having commits with deprecation warnings.

And here I went through all the trouble splitting my original commit 😁 

I'll sort that as well

gsson:
@Lorak-mmk : I'd probably "fix" the clippy lints like this
```rust
impl CqlTimeuuid {
    /// Read 8 most significant bytes of timeuuid from serialized bytes
    fn msb(&self) -> u64 {
        // Scylla and Cassandra use a standard UUID memory layout for MSB:
        // 4 bytes    2 bytes    2 bytes
        // time_low - time_mid - time_hi_and_version
        let bytes = self.0.as_bytes();
        u64::from_be_bytes([
            bytes[6], bytes[7], bytes[4], bytes[5], bytes[0], bytes[1], bytes[2], bytes[3],
        ])
    }

    fn lsb(&self) -> u64 {
        let bytes = self.0.as_bytes();
        u64::from_be_bytes([
            bytes[8], bytes[9], bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15],
        ])
    }

    fn lsb_signed(&self) -> u64 {
        self.lsb() ^ 0x8080808080808080
    }
}
```

Gets rid of most of the exposed bit twiddling. As far as I've been able to tell where I've done it, the generated code isn't really any worse.

Lorak-mmk:
> #1256 should hopefully fix the clippy lints

I've merged it. Please rebase this PR on main.

wprzytula (approved):

Lorak-mmk (approved):

Lorak-mmk:
Thanks!

PR #1260, opened 2025-02-28
Address unnecessary doc(hidden)

Lorak-mmk:
There are some places were we use `doc(hidden)` and we probably shouldn't.
In some cases the function can be just removed, in others it can be made pub(crate), or the doc(hidden) can be just removed.
First four commits should be quite obvious, I don't expect them to be controversial.

The last commit is the one I'm not sure about.

It removes `doc(hidden)` from items that are to be used only by our proc macros.
Why? There are a few arguments for that - but also arguments against that.

A more theoretical / ideological issue is that the macros should be just that - macros, a way to automatically generate some code instead of writing it by hand.
If the macros use some private API that the user should not use, then this is no longer true - because writing the same code by hand is no longer possible without opting out of semver.

One exception that could be doc(hidden) without breaking this rule is items that themselves only use the public API, so they could be considered macro-internal items (e.g. macro could generate their code directly).
Macros of course can't export non-macro functions, so the function would need to be generated as a nested function, which would increase compile time. `doc(hidden)` seems like a good solution here.
Example of such item is `mk_typck_err`.

The more practical argument is that the items used by scylla-macros have to (currently, but not inherently) partially adhere to semver even if they are `doc(hidden)`.
If they are doc(hidden) then this is harder, because `cargo-semver-checks` no longer checks them (because they are not part of public API).

Why do those items need to (partially) adhere to semver in our case?
`scylla` crate currently depends on `scylla-macros` in a way we would on most other dependencies: by specifying minimum version, and declaring that versions that are semver-compatible with this one are also ok.
`scylla-macros = { version = "0.7.0", path = "../scylla-macros" }` means that using `0.7.0` is fine, but so is using `0.7.39`.
Let's say we have `scylla` crate 1.1, with dependency on scylla-macros version = "1.1".
Now let's say in 1.2 we want to change the API used by macros, so we remove some stuff from `macro_internal` mod, and adjust the macros.

This would break `scylla` 1.1 because it would happily use new version of the `scylla-macros` crate, that uses incompatible items.
We can fix it by treating our proc macro API as a public one. This would address both the theoretical and practical issue.

The solution that seem to be easier in practice is to only depend on a specific version of `scylla-macros`. 
That is what `clap` and `serde` do - they depend on a specific version of their macro crates.
In our case this solution is not totally obvious, because `scylla-cql` is the one that depends on `scylla-macros`, and defines the items in `macro_internal` module.
If we wanted to go this way, we should probably:
- make `scylla-cql` depend on fixed version on `scylla-macros`.
- keep `scylla` dependencies loose as they are now, because `scylla` crates doesn't really influence macros.

The second solution allows us to modify the interface between macros and the code.
It's worth mentioning that macros are only a separate crate out of necessity (Rust doesn't allow proc macros to be defined inside lib crate),
so depending on a specific version of macros makes sense - it's basically treating both crates as one.
Unfortunately this solution doesn't address the first problem at all - the interface between macros and the crate still remains inaccessible magic :(

## Pre-review checklist


scylla-cql/src/deserialize/value.rs
```diff
@@ -675,12 +675,14 @@ where
...
+    pub(crate),
     typck_error_replace_rust_name,
     TypeCheckError,
     BuiltinTypeCheckError
 );
 
 make_error_replace_rust_name!(
+    pub,
     deser_error_replace_rust_name,
     DeserializationError,
```
wprzytula, line 687: :question: Why is once of these functions used by external code and the other is not? They seem to be very similar, so I see no immediate reason.
If they are so similar, I reason that they should have the same visibility. Same goes for `row.rs` replace functions.

scylla-cql/src/lib.rs
```diff
@@ -16,7 +16,6 @@ pub mod value;
 
 pub use crate::frame::types::Consistency;
 
-#[doc(hidden)]
 pub mod _macro_internal {
```
wprzytula, line 19: :thinking: Now the module does appear in the crate docs and clutters it. We did want to hide it from the docs, that's why the attribute is here.
IIUC, the reason why you remove the attribute is to trigger automatic semver checks for the module, to prevent silent breakage of the API. Can't we somehow have a cake and eat it, that is, have the module hidden from the docs but still checked by semver-checks? This would be ideal for this particular case IMO.

scylla-cql/src/deserialize/value.rs
Lorak-mmk: `typck_error_replace_rust_name` is only used in `DeserializeValue` impls written in the `value.rs` file.
`deser_error_replace_rust_name` is also used in those impls, but also in `scylla-macros` in generated deserialize method impls.
Why? I'm not sure, and its not easy to read the macro code - since you've written it I invite you to check that.

scylla-cql/src/lib.rs
Lorak-mmk: Afaik there is not such way.

Please see PR cover letter for more considerations. I see 2 options. Either we make this a public API (and thus not doc(hidden)) or we tie scylla-cql to a fixed version of scylla-macros.

The more I think about it the more I like the second solution - even though it makes macros more "magical". But this is something we can be careful about manually. I think the only doc(hidden) functions that macros are currently using fall into the exceptional category that I described in the PR, so it seems that for now we are doing good on that front.

wprzytula: If we make sure that `macro_internal` only contains re-exports of items that are public for ordinary users, too, then macros are no longer "magical" even if we tie `scylla-cql` to fixed version of `scylla-macros` - because even though they use some special non-public doc-hidden functionalities present in scylla-cql, one could implement those functionalities oneself. As you noticed in the cover letter:
> Macros of course can't export non-macro functions, so the function would need to be generated as a nested function, which would increase compile time. doc(hidden) seems like a good solution here.
> Example of such item is mk_typck_err.

This is merely an optimisation, not "magical" functionality that is not available to ordinary (non-macro) users.

So to sum up, I'm in favour of tying `scylla-cql` to fixed version of `scylla-macros` and leaving `_macro_internal` a `doc_hidden` module.

Lorak-mmk: Ok, I'll do that. I'll also put a note about this in contributing.md or maintenance.md

Lorak-mmk:
- Dropped the commit that removed doc(hidden) from macro-related items
- Made scylla-cql depend on exact version of scylla-macros
- Moved macro docs to scylla-macros crate
- Cleaned up macro-related paths / modules / re-exports

scylla-cql/Cargo.toml
```diff
@@ -11,7 +11,7 @@ categories = ["database"]
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-scylla-macros = { version = "0.7.0", path = "../scylla-macros" }
+scylla-macros = { version = "=0.7.0", path = "../scylla-macros" }
 byteorder = "1.3.4"
```
wprzytula, line 15: :wrench: Commit: `scylla-cql: Depend on exact version of scylla-macros`: future blame's readers will appreciate putting justification of this change in the commit message.

examples/custom_deserialization.rs
```diff
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use scylla::client::session::Session;
 use scylla::client::session_builder::SessionBuilder;
-use scylla::deserialize::DeserializeValue;
```
wprzytula, line 4: :pick: typo (`Remve`) in commit `Remve DeserializeValue/Row re-exports in deserialize/mod.rs `.

scylla-macros/Cargo.toml
```diff
@@ -21,3 +21,7 @@ proc-macro2 = "1.0"
 [lints.rust]
 unnameable_types = "warn"
 unreachable_pub = "warn"
+
+[dev-dependencies]
+scylla = { version = "0.15.0", path = "../scylla"}
+scylla-cql = { version = "0.4.0", path = "../scylla-cql"}
```
wprzytula, line 27: :thinking: This looks like an unexpected place for such dependency. What I'm afraid of is notorious forgetting about bumping versions here. How can we make this more discoverable and easier to remember about?

scylla-macros/src/lib.rs
```diff
@@ -23,9 +23,117 @@ impl FromMeta for Flavor {
...
+/// its Rust name.
+///
+/// `#[scylla(skip)]`
+///
+/// Don't use the field during serialization.
+///
+/// ---
+///plain here.
 #[proc_macro_derive(SerializeValue, attributes(scylla))]
 pub fn serialize_value_derive(tokens_input: TokenStream) -> TokenStream {
```
wprzytula, line 157: :question: :wrench: What is the `--- plain here.` part?

scylla-macros/src/lib.rs
```diff
@@ -34,9 +142,104 @@ pub fn serialize_value_derive(tokens_input: TokenStream) -> TokenStream {
...
+/// its Rust name.
+///
+/// `#[scylla(skip)]`
+///
+/// Don't use the field during serialization.
+///
+/// ---
+///
 #[proc_macro_derive(SerializeRow, attributes(scylla))]
 pub fn serialize_row_derive(tokens_input: TokenStream) -> TokenStream {
```
wprzytula, line 263: :wrench: :question: ditto: `---`

scylla-macros/src/lib.rs
```diff
@@ -47,6 +250,119 @@ pub fn serialize_row_derive(tokens_input: TokenStream) -> TokenStream {
...
+/// be initialized with `Default::default()`.
+///
+/// `#[scylla(rename = "field_name")]`
+///
+/// By default, the generated implementation will try to match the Rust field
+/// to a column with the same name. This attribute allows to match to a column
+/// with provided name.
 #[proc_macro_derive(DeserializeRow, attributes(scylla))]
 pub fn deserialize_row_derive(tokens_input: TokenStream) -> TokenStream {
     match deserialize::row::deserialize_row_derive(tokens_input) {
```
wprzytula, line 388: :thinking: And here `---` is not present.

scylla-macros/Cargo.toml
Lorak-mmk: If you don't update those lines when bumping major version of scylla-cql / scylla, then Cargo will yell at you.
If you don't update those lines when bumping minor version, then nothing bad will happen. New minor version will be compatible with the one specified here, so it will just be used.

Lorak-mmk:
Rebased on main.

scylla-macros/src/lib.rs
Lorak-mmk: Ooops, I don't know how it appeared here.

scylla-macros/src/lib.rs
Lorak-mmk: Removed

Lorak-mmk:
Addressed @wprzytula comments

scylla-macros/Cargo.toml
Lorak-mmk: Ok

Lorak-mmk:
Rebased on main

Lorak-mmk:
Fixed book building, added a comment requested by @wprzytula

wprzytula (approved):
:100: Awesome job!

This is true quality stabilisation that we need for 1.0.

PR #1261, opened 2025-03-03
mark metadata public structs as `non_exhaustive`

muzarski:
I decided to introduce getters for (and un-pub the fields of) `Keyspace` and `Table`. I think these are the structs whose current fields may be modified in the future.

Remaining types (`ColumnKind`, `Column`, `MaterializedView` and `Strategy`) are simple enough for me to believe we won't need to modify their current fields/variants in the future. OTOH, we want to be able to introduce new fields/variants in the future - thus I marked them as `non_exhaustive`.

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/metadata.rs
+++ b/scylla/src/cluster/metadata.rs
@@ -183,7 +183,9 @@ impl Peer {
     }
 }
 
+/// Describes a keyspace in the cluster.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
 pub struct Keyspace {
     pub strategy: Strategy,
     /// Empty HashMap may as well mean that the client disabled schema fetching in SessionConfig
@@ -194,7 +196,9 @@ pub struct Keyspace {
     pub user_defined_types: HashMap<String, Arc<UserDefinedType<'static>>>,
 }
 
+/// Describes a table in the cluster.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
 pub struct Table {
     pub columns: HashMap<String, Column>,
     /// Names of the column of partition key.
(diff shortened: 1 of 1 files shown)


wprzytula (approved):
LGTM

Lorak-mmk (changes requested):
First and last commit are fine. I still don't see the point of second and third commit.
What changes we may want to perform on those specific structs that would warrant moving to such getters?

- We discussed moving to Vecs instead of HashMaps. Those getters won't allow us to do this after 1.0
- We could use different hashmap (e.g. hashbrown). Those getters won't allow us to do this.
- Do we have any potential plans to touch `strategy` field? I don't think so.

I'm strongly in favor of just marking those struct non_exhaustive for now instead of introducing such getters. We can change them in 2.0 if the need arises.

wprzytula:
> What changes we may want to perform on those specific structs that would warrant moving to such getters?

We can, for instance, introduce more nesting.
E.g., we can define
```rust
struct A {
  tables: HashMap<String, Table>,
  some_else_field: AnotherType,
}
```
and then store `A` in `Keyspace` instead of plain `tables`. Then, getter allows us to do such change without breaking changes.

Lorak-mmk:
I don't see what would be the purpose of such change here.

Lorak-mmk:
> And I don't see why it would be good to expose fields of those structs. After all, those fields are only meant to be readable (and immutable). Getters are a usual way to achieve that.

In general we should have a reason to introduce more code (getters are additional code, fields are the simplest).
If we accepted this argument, then we should not have any public fields in any struct, which I don't think is a reasonable position.
By the way, why did you approve this PR? It only marked `MaterializedView` and `Column` as non_exhaustive. Consistency in your argumentation would require you to demand un-pubbing the fields there too.

We should judge each case individually and think about what changes we may want to introduce. I don't see us needing to change Keyspace / Table in the specific way that you described.
If you disagree, please tell me a scenario where that would happen.

muzarski:
I'm leaning towards public fields of `Keyspace` and `Table` (non_exhaustive). Is it fine with you @wprzytula ?

muzarski:
Rebased on main

muzarski:
v2: Removed the getters from `Keyspace` and `Table` API. The fields remain as `pub`. Added `non_exhaustive` to both types.

Lorak-mmk (approved):

wprzytula (approved):

Issue #330, opened 2021-11-05
Make metrics collection optional/faster


havaker:
Benchmarks done some time ago showed that the driver can spend a significant amount of time logging query latency (`scylla::transport::metrics::Metrics::log_query_latency` 11.68% on [this flamegraph](https://raw.githubusercontent.com/cvybhu/rust-driver-benchmarks/main/images/flame100mil8192.svg)). It is probably due to the fact that logging latency to the metrics histogram can lead to congestion of a [mutex guarding this histogram](https://github.com/scylladb/scylla-rust-driver/blob/af828e0c081a16229d0081698506565f04fa029e/scylla/src/transport/metrics.rs#L38) (logging takes place after each query completion).

Mutex congestion problem could be fixed by using lock-free histogram, but unfortunately I can't find any crate providing one on crates.io. Another way to reduce this problem could be sharding histograms - making one per thread, in a similar way to [this one](https://github.com/piodul/scylla-chair/blob/b1974643e7e8accfb26bd19a5bfd676b27c7629f/src/sharded_histogram.rs#L21). The last option sacrifices read performance, so it would be good to add some caching mechanism to `Metrics` to avoid slowdowns caused by e.g. speculative execution policies asking for latency percentile for each query.

I suppose we could also make metrics collection optional. It would be an obvious speedup for people who do not want to use metrics.

wprzytula:
When tackling with this, let's consider cpp-rust-driver requirements wrt metrics.

roydahan:
The current PR require resolving conflicts and addressing comments.

PR #1263, opened 2025-03-03
Make metrics collection optional/faster

wprzytula:
This is a touch-up of #1147. Pasting the original cover letter below.
-------

This patch contains:
- Implementation of lock-free histogram with hot and cold bucket pools
- Introduction of crate feature "metrics" to make using them optional
- Functionality to gather latency statistics via histogram snapshots
- Implementation of [rates of queries per second](https://github.com/scylladb/cpp-driver/blob/9d6b05c9d4ebd0a6d7006af4df3e33fcdf956eeb/src/metrics.hpp#L39C1-L252C5) based on cpp-driver implementation.
- Initial implementation of connection metrics (changes required after review)

Fixes: #330

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -882,6 +882,8 @@ impl Session {
             can_use_shard_aware_port: !config.disallow_shard_aware_port,
         };
 
+        let metrics = Arc::new(Metrics::new());
+
         let cluster = Cluster::new(
             known_nodes,
             pool_config,
@@ -890,6 +892,7 @@ impl Session {
             config.host_filter,
             config.cluster_metadata_refresh_interval,
             tablet_receiver,
+            Arc::clone(&metrics),
         )
         .await?;
 
@@ -899,7 +902,7 @@ impl Session {
             cluster,
             default_execution_profile_handle,
             schema_agreement_interval: config.schema_agreement_interval,
-            metrics: Arc::new(Metrics::new()),
+            metrics,
             schema_agreement_timeout: config.schema_agreement_timeout,
             schema_agreement_automatic_waiting: config.schema_agreement_automatic_waiting,
             refresh_metadata_on_auto_schema_agreement: config
@@ -1108,7 +1111,7 @@ impl Session {
                 statement,
                 execution_profile,
                 self.cluster.get_state(),
-                self.metrics.clone(),
+                Arc::clone(&self.metrics),
             )
             .await
             .map_err(PagerExecutionError::NextPageError)
@@ -1123,7 +1126,7 @@ impl Session {
                 values,
                 execution_profile,
                 cluster_state: self.cluster.get_state(),
-                metrics: self.metrics.clone(),
+                metrics: Arc::clone(&self.metrics),
             })
             .await
             .map_err(PagerExecutionError::NextPageError)
@@ -1382,7 +1385,7 @@ impl Session {
             values: serialized_values,
             execution_profile,
             cluster_state: self.cluster.get_state(),
-            metrics: self.metrics.clone(),
+            metrics: Arc::clone(&self.metrics),
         })
         .await
         .map_err(PagerExecutionError::NextPageError)
@@ -1597,7 +1600,7 @@ impl Session {
     /// Driver collects various metrics like number of queries or query latencies.
     /// They can be read using this method
     pub fn get_metrics(&self) -> Arc<Metrics> {
-        self.metrics.clone()
+        Arc::clone(&self.metrics)
     }
 
     /// Access cluster state visible by the driver.
@@ -1814,7 +1817,7 @@ impl Session {
                     };
 
                     let context = speculative_execution::Context {
-                        metrics: self.metrics.clone(),
+                        metrics: Arc::clone(&self.metrics),
                     };
 
                     speculative_execution::execute(
@@ -1856,9 +1859,12 @@ impl Session {
             .request_timeout
             .or(execution_profile.request_timeout);
         let result = match effective_timeout {
-            Some(timeout) => tokio::time::timeout(timeout, runner)
-                .await
-                .unwrap_or_else(|_| Err(RequestError::RequestTimeout(timeout))),
+            Some(timeout) => tokio::time::timeout(timeout, runner).await.unwrap_or_else(
+                |_: tokio::time::error::Elapsed| {
+                    self.metrics.inc_request_timeouts();
+                    Err(RequestError::RequestTimeout(timeout))
+                },
+            ),
             None => runner.await,
         };
 
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -20,6 +20,7 @@ use crate::network::tls::TlsProvider;
 use crate::network::{Connection, ConnectionConfig, PoolConfig, VerifiedKeyspaceName};
 use crate::observability::driver_tracing::RequestSpan;
 use crate::observability::history::{self, HistoryListener};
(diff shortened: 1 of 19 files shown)


scylla/src/observability/metrics.rs
```diff
@@ -134,6 +178,58 @@ impl Metrics {
...
+            min = u64::min(min, lower_bound);
+            max = u64::max(max, upper_bound);
+        }
+
+        if min > max {
+            Err(MetricsError::Empty)
+        } else {
+            Ok((min, max))
+        }
+    }
```
Lorak-mmk, line 485: In this comment: https://github.com/scylladb/scylla-rust-driver/pull/1147#discussion_r1963774054 it was suggested by @brayniac that it would be better to use `percentiles` method from histogram to implement this - we already call it , so we can add more arguments to it. It was marked as resolved without response and without changes, which I don't understand.

I do understand that it may not be feasible to implement it that way. For example, from what I can tell from histogram crate source you can't get `min` using `0` percentile - because histogram will return the first bucket, no matter if it has non-0 count. I don't see any problem with using `100` percentile for max though.

If there is a reason not to do that, it should be stated as a response to the review comment, instead of marking it resolved with no action.

scylla/src/observability/metrics.rs
```diff
@@ -120,4 +114,63 @@ impl Metrics {
...
+            count += bucket.count() as u128;
+        }
+
+        if count != 0 {
+            Ok((weighted_sum / count) as u64)
+        } else {
+            Err(MetricsError::Empty)
+        }
+    }
+}
```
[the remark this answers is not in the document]
Lorak-mmk: I'd like to note that in my previous review I wrote "Commit message needs to explain why this method was deleted". It was not done, but the comment was marked as resolved.

scylla/src/observability/metrics.rs
```diff
@@ -33,13 +33,181 @@ pub struct Snapshot {
...
+            let rate = f64::from_bits(self.rate.load(Ordering::Acquire));
+            self.rate.store(
+                f64::to_bits(rate + self.alpha * (instant_rate - rate)),
+                Ordering::Release,
+            );
+        } else {
+            self.rate
+                .store(f64::to_bits(instant_rate), Ordering::Release);
+            *is_initialized = true;
+        }
```
Lorak-mmk, line 98: In my comment I suggested first using `try_lock()`, and if it fails, printing a warning.
Do you think that is a bad idea?

Lorak-mmk, line 98: I think this code is still not fully correct. Previously I suggested locking the mutex before the `count` is read. Why?
Consider the scenario, where driver has a constant rate of queries, and was just started, and we got 2 concurrent calls to this method.
Following execution is possible:
- Thread 1 reads and zeros count
- Thread 2 reads 0
- Thread 2 gets mutex, `is_initialized` is false, so goes to the second branch
- It stores `0` rate
- Thread 1 gets mutex, stores new rate

Now the rate will slowly increase, instead of being correct from the beginning.
Please move the lock before count read, I see no downsides to it.

scylla/src/cluster/metadata.rs
```diff
@@ -459,6 +463,7 @@ impl MetadataReader {
             control_connection_endpoint.clone(),
             &control_connection_pool_config,
             control_connection_repair_requester.clone(),
+            metrics.clone(),
         );
```
Lorak-mmk: Please use `Arc::clone`

scylla/src/client/session.rs
```diff
@@ -890,6 +892,7 @@ impl Session {
             config.host_filter,
             config.cluster_metadata_refresh_interval,
             tablet_receiver,
+            metrics.clone(),
```
Lorak-mmk: ditto

scylla/src/cluster/metadata.rs
```diff
@@ -575,6 +581,7 @@ impl MetadataReader {
                 self.control_connection_endpoint.clone(),
                 &self.control_connection_pool_config,
                 self.control_connection_repair_requester.clone(),
+                self.metrics.clone(),
```
Lorak-mmk: ditto

scylla/src/cluster/metadata.rs
```diff
@@ -673,6 +680,7 @@ impl MetadataReader {
                         self.control_connection_endpoint.clone(),
                         &self.control_connection_pool_config,
                         self.control_connection_repair_requester.clone(),
+                        self.metrics.clone(),
                     );
```
Lorak-mmk: ditto

scylla/src/cluster/state.rs
```diff
@@ -98,6 +101,7 @@ impl ClusterState {
                         pool_config,
                         used_keyspace.clone(),
                         is_enabled,
+                        metrics.clone(),
                     ))
```
Lorak-mmk: ok I'll stop now, there are probably more instances of that.

Lorak-mmk (changes requested):

scylla/src/observability/metrics.rs
brayniac: You appear to be right. I’ll call that a bug in the histogram crate. I’ll submit and release a fix tomorrow. 

Let me know if any other changes would be helpful. 

Linking the issue here: https://github.com/pelikan-io/rustcommon/issues/133

scylla/src/observability/metrics.rs
```diff
@@ -134,6 +178,58 @@ impl Metrics {
...
+
+    fn percentiles(h: &Histogram, percentiles: &[f64]) -> Result<Vec<u64>, MetricsError> {
+        let res = h.percentiles(percentiles);
+
+        match res {
+            Err(err) => Err(MetricsError::HistogramError(Arc::new(err))),
+
+            Ok(None) => Err(MetricsError::Empty),
+
+            Ok(Some(ps)) => Ok(ps.into_iter().map(|(_, bucket)| bucket.count()).collect()),
```
muzarski, line 443: Since this function is used only internally (it's private and used only in `get_snapshot`), what's the point of mapping the `(f64, histogram::Bucket)` pairs to `u64` (`Bucket::count()`)? Notice that currently `Metrics::percentiles` potentially allocates two vectors - first one in `Histogram::percentiles` (which is immediately converted to iterator), then the second one during `.collect()` call.

I suggest returning either `Vec<(f64, Bucket)>` (return type from Histogram::percentiles) or `impl Iterator<Item = (f64, Bucket)>`. The caller can then map the elements of resulting vector/iterator appropriately.

muzarski: Now that I think of it, returning `impl Iterator<Item = u64>` (after mapping) makes sense as well.

scylla/src/observability/metrics.rs
wprzytula: The warning here is not very important IMO. Let's not complicate the logic with it.

scylla/src/observability/metrics.rs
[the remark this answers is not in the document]
wprzytula: I put the comment quoting @brayniac's answer:
https://github.com/scylladb/scylla-rust-driver/pull/1147#discussion_r1920620551

wprzytula:
Addressed comments.

wprzytula:
Rebased on main.

Lorak-mmk (dismissed):

muzarski (approved):

Lorak-mmk (approved):

PR #1266, opened 2025-03-04
state: Adjust ClusterState keyspace info getters

muzarski:
We don't want to expose the underlying structure of how we store keyspace info. This is because, we may want to change this structure in the future.

One possible idea for the future is to wrap `Keyspace` in an `Arc`. This would allow us to do following:
- share keyspace info in multiple places
- cheap cloning of keyspace info (e.g. currently we do the clone of whole `Keyspace` structure here: https://github.com/scylladb/scylla-rust-driver/blob/main/scylla/src/cluster/state.rs#L118-L121)

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -110,24 +110,30 @@ impl ClusterState {
         }
 
         let keyspaces: HashMap<String, Keyspace> = metadata
-        .keyspaces
-        .into_iter()
-        .filter_map(|(ks_name, ks)| match ks {
-            Ok(ks) => Some((ks_name, ks)),
-            Err(e) => {
-                if let Some(old_ks) = old_keyspaces.get(&ks_name) {
-                    warn!("Encountered an error while processing metadata of keyspace \"{ks_name}\": {e}.\
-                           Re-using older version of this keyspace metadata");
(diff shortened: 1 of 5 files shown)


scylla/src/cluster/state.rs
```diff
@@ -185,11 +191,14 @@ impl ClusterState {
...
-        &self.keyspaces
+    /// Access keyspace details collected by the driver.
+    pub fn get_keyspace(&self, keyspace: impl AsRef<str>) -> Option<&Keyspace> {
+        self.keyspaces.get(keyspace.as_ref())
+    }
+
+    /// Returns an iterator over keyspaces.
+    pub fn get_keyspaces_iter(&self) -> impl Iterator<Item = (&String, &Keyspace)> {
+        self.keyspaces.iter()
     }
```
Lorak-mmk, line 202: Nit: This name sounds like there is already some iterator in ClusterState, and we are retrieving it.
I'd change this to `keyspaces_iter()`.

Lorak-mmk (changes requested):

Lorak-mmk (approved):

wprzytula (approved):

Lorak-mmk (approved):

PR #901, opened 2024-01-03
CI: Use Rust cache

oeb25:
This adds [Swatinem/rust-cache@v2](https://github.com/Swatinem/rust-cache) to all CI jobs, with the intention of speeding them up. Hopefully we can compare some runs to get empirical data of the improvements it gives.

## Pre-review checklist

- [X] I have split my patch into logically separate commits.
- [X] All commit messages clearly explain what they change and why.
- [ ] I added relevant tests for new features and bug fixes.
- [ ] All commits compile, pass static checks and pass test.
- [X] PR description sums up the changes and reasons why they should be introduced.
- [ ] I have provided docstrings for the public items that I want to introduce.
- [ ] I have adjusted the documentation in `./docs/source/`.
- [ ] I added appropriate `Fixes:` annotations to PR description.

PR #1274, opened 2025-03-10
Use cache in CI

Lorak-mmk:
This PR performs some CI improvements. The most important one is utilizing caching, with the help of https://github.com/actions-rust-lang/setup-rust-toolchain action.
This should speed up CI, but its unclear how much exactly. I'll try to check that and post some results.

Apart from caching, this PR:
- fixes a bug, where clippy warnings were not rejected during cpp-rust-driver cfg check.
- makes some minor adjustments, like replacing "--manifest-path" with "--package".
- removes redundant build step.
- replaces scylla-cql build with clippy check.
- removes old workaround for a rust bug

Replaces https://github.com/scylladb/scylla-rust-driver/pull/901

## Pre-review checklist

### DIFF
--- a/.github/workflows/authenticate_test.yml
+++ b/.github/workflows/authenticate_test.yml
@@ -22,8 +22,8 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v3
 
-    - name: Update rust toolchain
-      run: rustup update
+    - name: Setup rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@v1
 
     - name: Start the cluster
       run: docker compose -f test/cluster/docker-compose-passauth.yml up -d
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -32,7 +32,9 @@ jobs:
         # We can't use cache here, or mdbook will complain about multiple candidates for dependency.
         cache: false
     - name: Install mdbook
-      run: cargo install mdbook --no-default-features
+      uses: taiki-e/install-action@v2
+      with:
+        tool: mdbook
     - name: Build the project
       run: cargo build --examples
     - name: Build the book
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -26,10 +26,8 @@ jobs:
         options: --health-cmd "cqlsh --debug scylladb" --health-interval 5s --health-retries 10
     steps:
     - uses: actions/checkout@v3
-    # Set to 1.81 because of regression in 1.82: https://github.com/rust-lang/rust/issues/131893
-    # TODO: change back to latest stable after this bug is fixed.
     - name: Update rust toolchain
-      run: rustup default 1.81
+      run: rustup update
     - name: Install mdbook
       run: cargo install mdbook --no-default-features
     - name: Build the project
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -26,8 +26,11 @@ jobs:
         options: --health-cmd "cqlsh --debug scylladb" --health-interval 5s --health-retries 10
     steps:
     - uses: actions/checkout@v3
-    - name: Update rust toolchain
-      run: rustup update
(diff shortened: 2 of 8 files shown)


Lorak-mmk:
Improvements of the first run (so without cache), in comparison to #1171:
authenticate: 1:52 vs 2:11
book: 3:53 vs 4:49
cassandra: 8:47 vs 9:33
cargo docs: 0:43 vs 0:35
min_rust: 1:50 vs 1:49
static checks: 4:23 vs 7:50
scylla tests: 4:22 vs 5:32
semver checks: 6:38 vs 5:39
serverless: 5:52 vs 6:47
ssl: 1:43 vs 2:05

Even without cache we have really nice gains, especially on static checks. I'll push some commit now, so the CI will run again, with the cache working correctly (hopefully)

Lorak-mmk:
Ok, CI failed because book got the well-known  error:
```
multiple candidates for `rlib` dependency `X` found
```

it looks like we won't be able to utilize caching there. 
Not a big problem, it is not a long workflow.
Other workflows (caching run vs no caching run vs #1171):
authenticate: 1:39 vs 1:52 vs 2:11
book: fail vs 3:53 vs 4:49
cassandra: 7:58 vs 8:47 vs 9:33
cargo docs: 0:24 vs 0:43 vs 0:35
min_rust: 0:39 vs 1:50 vs 1:49
static checks: 2:38 vs 4:23 vs 7:50 (!!!)
scylla tests: 3:18 vs 4:22 vs 5:32
semver checks: 6:40 vs 6:38 vs 5:39
serverless: 3:25 vs 5:52 vs 6:47
ssl: 1:04 vs 1:43 vs 2:05

The speedup is imo great. 

With the cache the longest workflows are:
Cassandra 7:58
Semver checks 6:40
Book 3:53
Serverless 3:25

We should be able to speed up semver checks and book significantly bu using https://github.com/cargo-bins/cargo-binstall . I can try to do this.

I don't really know what to do about Cassandra. Timings there are:
- Cluster setup 3:36
- Tests 3:47
- Cluster teardown 0:25
- Other steps are insignificant

Do we need to test with 3-node Cassandra cluster? Maybe a single node is enough? @wprzytula @muzarski

Lorak-mmk:
With binary install, book job took 2:33, and semver checks took 1:47. There is now quite significant gap between cassandra, and second-slowest workflow.

wprzytula:
> Do we need to test with 3-node Cassandra cluster? Maybe a single node is enough? @wprzytula @muzarski

Let's keep it. Apart from being a regular Cassandra that we want to ensure compatibility with, it's still routed using the token ring, which is no longer tested in our Scylla cluster (which uses tablets).

wprzytula (approved):
:rocket: AWESOME job! :tada: :sparkles:

Lorak-mmk:
> > Do we need to test with 3-node Cassandra cluster? Maybe a single node is enough? @wprzytula @muzarski
> 
> Let's keep it. Apart from being a regular Cassandra that we want to ensure compatibility with, it's still routed using the token ring, which is no longer tested in our Scylla cluster (which uses tablets).
> 
> What I could consider is selecting a smaller subset of tests to be run against Cassandra.

Ok, we can look at how to speed up C* tests in another PR.

muzarski (approved):
`setup-rust-toolchain` mentions that it can be also used to set RUSTFLAGS: https://github.com/actions-rust-lang/setup-rust-toolchain?tab=readme-ov-file#rustflags. I'm not sure if we want to ditch the runner env variable (`-Dwarnings`) in favor of this.

Lorak-mmk:
> `setup-rust-toolchain` mentions that it can be also used to set RUSTFLAGS: https://github.com/actions-rust-lang/setup-rust-toolchain?tab=readme-ov-file#rustflags. I'm not sure if we want to ditch the runner env variable (`-Dwarnings`) in favor of this.

I saw that, but I personally see no reason to. Setting this flag ourselves is more explicit and obvious.

PR #1276, opened 2025-03-11
CI: Improve Cassandra startup time

Lorak-mmk:
Similarly to Scylla before Raft, Cassandra needs a lot of time to start the cluster. We can reduce this time for non-production clusters with some options, similar to those that were used for testing older versions of Scylla in test.py etc.

The options I used reduce cluster startup time from 3.5 minutes to 45 seconds on my machine.
IIRC it usually also took 3.5 minutes in the CI, so I'm expecting a similar speedup. If it turns out that those options cause some flakiness, we can increase the values or revert the change.

## Pre-review checklist


Lorak-mmk:
Longest workflows in this run:
- Cassandra tests 5:11
- Serverless 4:00
- Scylla tests 3:52
- Static checks 3:26

So Cassandra is still the slowest one after this change, but the difference is not so drastic now.

Issue #1030, opened 2024-07-09
Expose request coordinator id (host_id, address) in `QueryResult`


wprzytula:
It may be a useful information what node served as coordinator of the request.
cpp-driver has API for checking that, so we need to implement this at least for cpp-rust-driver.

Lorak-mmk:
This could potentially be implemented in cpp-rust-driver using history or similar mechanism

muzarski:
Notice that extending `QueryResult` may not be enough to be fully compatible with cpp-driver. The `cass_future_coordinator` method is defined on a `ResponseFuture`, which is used for `PREPARE` requests as well.

In cpp-driver, the "coordinator" of `PREPARE` request, is the first node the driver prepares the statement on.

However, I don't think it's that important to support `PREPARE` requests in this case. For now, I'm fine with implementing it only for requests that execute some CQL statement on the server (i.e. QUERY, EXECUTE and BATCH) and return `QueryResult`. If somebody ever asks for this feature for PREPARE requests, we can always implement it later.

wprzytula:
> For now, I'm fine with implementing it only for requests that execute some CQL statement on the server (i.e. QUERY, EXECUTE and BATCH) and return `QueryResult`.

I'd also add it to the `TypedRowStream`. It shouldn't be hard to store the last coordinator's ID as part of the stream state. I.e., once the `TypedRowStream` has returned a row (i.e., `next()` returned `Some`), the last coordinator will be available by the (non-`async`) call to `get_last_coordinator()`. WDYT?

Lorak-mmk:
> > For now, I'm fine with implementing it only for requests that execute some CQL statement on the server (i.e. QUERY, EXECUTE and BATCH) and return `QueryResult`.
> 
> I'd also add it to the `TypedRowStream`. It shouldn't be hard to store the last coordinator's ID as part of the stream state. I.e., once the `TypedRowStream` has returned a row (i.e., `next()` returned `Some`), the last coordinator will be available by the (non-`async`) call to `get_last_coordinator()`. WDYT?

Why last? It's not consistent with tracing IDs and warnings where we return a vec

wprzytula:
> > > For now, I'm fine with implementing it only for requests that execute some CQL statement on the server (i.e. QUERY, EXECUTE and BATCH) and return `QueryResult`.
> > 
> > 
[...]

PR #1287, opened 2025-03-20
Expose request coordinator

wprzytula:
This is the second approach to #1030.

### `Coordinator`

`Coordinator` is defined as follows:
```rust
/// The coordinator of a CQL request, i.e., the node+shard that receives
/// and processes the request, and hopefully eventually sends a response.
#[derive(Debug, Clone)]
pub struct Coordinator {
    /// Translated address, i.e., one that the connection is opened against.
    connection_address: SocketAddr,
    /// The node that served as coordinator.
    node: Arc<Node>,
    /// Number of the shard, if applicable (present for ScyllaDB nodes, absent for Cassandra).
    shard: Option<Shard>,
}
```

It has `pub` getters for all of its fields.

`Coordinator` is built from `NodeRef` (`untranslated_address` and `host_id`), `Shard` (`shard`), and `Connection` (`connection_address`).

### Use

`Coordinator` is stored in `QueryResult` (as well as `QueryRowsResult`) and `QueryPager` (together with `TypedRowStream`). `QueryPager` exposes an iterator of references to `Coordinator`s, each of which served one page.

### Testing
No tests yet.

Fixes: #1030

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -417,6 +417,8 @@ pub(crate) enum RunRequestResult<ResT> {
     Completed(ResT),
 }
 
+/// Represents a CQL session, which can be used to communicate
+/// with the database
 impl Session {
     /// Sends a request to the database and receives a response.\
     /// Executes an unprepared CQL statement without paging, i.e. all results are received in a single response.
@@ -803,11 +805,7 @@ impl Session {
     ) -> Result<QueryResult, ExecutionError> {
         self.do_batch(batch, values).await
     }
-}
 
-/// Represents a CQL session, which can be used to communicate
-/// with the database
-impl Session {
     /// Estabilishes a CQL session with the database
     ///
     /// Usually it's easier to use [SessionBuilder](crate::client::session_builder::SessionBuilder)
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -30,7 +30,9 @@ use crate::policies::retry::{RequestInfo, RetryDecision, RetrySession};
 use crate::policies::speculative_execution;
 use crate::policies::timestamp_generator::TimestampGenerator;
 use crate::response::query_result::{MaybeFirstRowError, QueryResult, RowsError};
-use crate::response::{NonErrorQueryResponse, PagingState, PagingStateResponse, QueryResponse};
+use crate::response::{
+    Coordinator, NonErrorQueryResponse, PagingState, PagingStateResponse, QueryResponse,
+};
 use crate::routing::partitioner::PartitionerName;
 use crate::routing::{Shard, ShardAwarePortRange};
 use crate::statement::batch::batch_values;
@@ -1049,7 +1051,10 @@ impl Session {
 
         let span = RequestSpan::new_query(&statement.contents);
         let span_ref = &span;
-        let run_request_result = self
+        let (run_request_result, coordinator): (
+            RunRequestResult<NonErrorQueryResponse>,
+            Coordinator,
+        ) = self
             .run_request(
                 statement_info,
                 &statement.config,
@@ -1112,7 +1117,8 @@ impl Session {
         self.handle_set_keyspace_response(&response).await?;
         self.handle_auto_await_schema_agreement(&response).await?;
 
-        let (result, paging_state_response) = response.into_query_result_and_paging_state()?;
+        let (result, paging_state_response) =
+            response.into_query_result_and_paging_state(coordinator)?;
         span.record_result_fields(&result);
 
         Ok((result, paging_state_response))
@@ -1438,7 +1444,10 @@ impl Session {
             }
         }
 
-        let run_request_result: RunRequestResult<NonErrorQueryResponse> = self
+        let (run_request_result, coordinator): (
+            RunRequestResult<NonErrorQueryResponse>,
+            Coordinator,
+        ) = self
             .run_request(
                 statement_info,
                 &prepared.config,
@@ -1481,7 +1490,8 @@ impl Session {
         self.handle_set_keyspace_response(&response).await?;
         self.handle_auto_await_schema_agreement(&response).await?;
 
-        let (result, paging_state_response) = response.into_query_result_and_paging_state()?;
+        let (result, paging_state_response) =
+            response.into_query_result_and_paging_state(coordinator)?;
         span.record_result_fields(&result);
 
         Ok((result, paging_state_response))
@@ -1563,7 +1573,10 @@ impl Session {
 
         let span = RequestSpan::new_batch();
 
-        let run_request_result: RunRequestResult<NonErrorQueryResponse> = self
+        let (run_request_result, coordinator): (
+            RunRequestResult<NonErrorQueryResponse>,
+            Coordinator,
+        ) = self
             .run_request(
                 statement_info,
                 &batch.config,
@@ -1593,9 +1606,9 @@ impl Session {
             .await?;
 
         let result = match run_request_result {
-            RunRequestResult::IgnoredWriteError => QueryResult::mock_empty(),
+            RunRequestResult::IgnoredWriteError => QueryResult::mock_empty(coordinator),
             RunRequestResult::Completed(non_error_query_response) => {
-                let result = non_error_query_response.into_query_result()?;
+                let result = non_error_query_response.into_query_result(coordinator)?;
                 span.record_result_fields(&result);
                 result
             }
@@ -1847,7 +1860,7 @@ impl Session {
         execution_profile: Arc<ExecutionProfileInner>,
         run_request_once: impl Fn(Arc<Connection>, Consistency, &ExecutionProfileInner) -> QueryFut,
         request_span: &'a RequestSpan,
-    ) -> Result<RunRequestResult<ResT>, ExecutionError>
+    ) -> Result<(RunRequestResult<ResT>, Coordinator), ExecutionError>
     where
         QueryFut: Future<Output = Result<ResT, RequestAttemptError>>,
         ResT: AllowedRunRequestResTType,
@@ -2017,7 +2030,7 @@ impl Session {
         run_request_once: impl Fn(Arc<Connection>, Consistency, &ExecutionProfileInner) -> QueryFut,
         execution_profile: &ExecutionProfileInner,
         mut context: ExecuteRequestContext<'a>,
-    ) -> Option<Result<RunRequestResult<ResT>, RequestError>>
+    ) -> Option<Result<(RunRequestResult<ResT>, Coordinator), RequestError>>
     where
         QueryFut: Future<Output = Result<ResT, RequestAttemptError>>,
         ResT: AllowedRunRequestResTType,
@@ -2056,6 +2069,9 @@ impl Session {
                     connection = %connect_address,
                     "Sending"
                 );
+                let coordinator =
+                    Coordinator::new(node, node.sharder().is_some().then_some(shard), &connection);
+
                 let attempt_id: Option<history::AttemptId> =
                     context.log_attempt_start(connect_address);
                 let request_result: Result<ResT, RequestAttemptError> =
@@ -2075,7 +2091,7 @@ impl Session {
                             elapsed,
                             node,
                         );
-                        return Some(Ok(RunRequestResult::Completed(response)));
+                        return Some(Ok((RunRequestResult::Completed(response), coordinator)));
                     }
                     Err(e) => {
                         trace!(
@@ -2130,7 +2146,7 @@ impl Session {
                     RetryDecision::DontRetry => break 'nodes_in_plan,
 
                     RetryDecision::IgnoreWriteError => {
-                        return Some(Ok(RunRequestResult::IgnoredWriteError))
+                        return Some(Ok((RunRequestResult::IgnoredWriteError, coordinator)))
                     }
                 };
             }
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -2052,13 +2052,14 @@ impl Session {
                 self.metrics.inc_total_nonpaged_queries();
                 let request_start = std::time::Instant::now();
 
+                let connect_address = connection.get_connect_address();
                 trace!(
                     parent: &span,
-                    connection = %connection.get_connect_address(),
+                    connection = %connect_address,
                     "Sending"
                 );
                 let attempt_id: Option<history::AttemptId> =
-                    context.log_attempt_start(connection.get_connect_address());
+                    context.log_attempt_start(connect_address);
                 let request_result: Result<ResT, RequestAttemptError> =
                     run_request_once(connection, current_consistency, execution_profile)
                         .instrument(span.clone())
--- a/scylla/src/network/connection.rs
+++ b/scylla/src/network/connection.rs
@@ -23,8 +23,7 @@ use crate::policies::address_translator::{AddressTranslator, UntranslatedPeer};
 use crate::policies::timestamp_generator::TimestampGenerator;
 use crate::response::query_result::QueryResult;
 use crate::response::{
-    NonErrorAuthResponse, NonErrorStartupResponse, PagingState, PagingStateResponse, QueryResponse,
-    RawPreparedStatement,
+    NonErrorAuthResponse, NonErrorStartupResponse, PagingState, QueryResponse, RawPreparedStatement,
 };
 use crate::routing::locator::tablets::{RawTablet, TabletParsingError};
 use crate::routing::{Shard, ShardAwarePortRange, ShardInfo, Sharder, ShardingError};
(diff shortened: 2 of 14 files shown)


Lorak-mmk:
Why did you introduce new struct (`Coordinator`) instead of using `Node`?

wprzytula:
> Why did you introduce new struct (`Coordinator`) instead of using `Node`?

1. I did not realise that we do have access to `Node` in all places where we need to create `Coordinator` -- this fortunately seems to be true.
2. `Coordinator` is not a struct. It's just a type alias that is used for convenience. I imagined it could be a pair `(Uuid, SocketAddr)` in the final version of this PR.

muzarski:
> Why did you introduce new struct (`Coordinator`) instead of using `Node`?

Just `Node` is not enough for cpp-rust-driver purposes. In cpp-driver, the coordinator is the target (i.e. node + shard) identified by `SocketAddr`.

muzarski:
> > > Why did you introduce new struct (`Coordinator`) instead of using `Node`?
> > 
> > 
> > Just `Node` is not enough for cpp-rust-driver purposes. In cpp-driver, the coordinator is the target (i.e. node + shard) identified by `SocketAddr`.
> 
> Is there really an API in cpp-driver to retrieve a shard that the request was sent to?

No, I might have phrased it wrong. Now that I think of it, I did, because there are no shards in Cassandra. What I meant is just that the request coordinator is identified via ip + port (`SocketAddr`).

Lorak-mmk:
> 
> What I meant is just that the request coordinator is identified via ip + port (`SocketAddr`).

Than I don't understand why `Node` is not enough. It has [addr field](https://docs.rs/scylla/latest/scylla/cluster/struct.Node.html#structfield.address) which contains `SocketAddr` inside: https://docs.rs/scylla/latest/scylla/cluster/enum.NodeAddr.html

Lorak-mmk:
We can of course consider wrapping `Node` (or whatever else we need) in a new `Coordinator` struct if we may want to add new stuff there (maybe a Option<shard number>?).

muzarski:
> > What I meant is just that the request coordinator is identified via ip + port (`SocketAddr`).
> 
> Than I don't understand why `Node` is not enough. It has [addr field](https://docs.rs/scylla/latest/scylla/cluster/struct.Node.html#structfield.address) which contains `SocketAddr` inside: https://docs.rs/scylla/latest/scylla/cluster/enum.NodeAddr.html

But the `port` in this `SocketAddr` is not always the port of the connection we routed the request to. AFAIU, this port is set to `self.control_connection_endpoint.address().port()` obtained here: https://github.com/scylladb/scylla-rust-driver/blob/v1.0.0/scylla/src/cluster/metadata.rs#L608. This port is later propagated for other peers as well (is it even correct???). Notice that the queried `rpc_address` from `system.local/peers` is an ip address without any port information.

wprzytula:
> : https://github.com/scylladb/scylla-rust-driver/blob/v1.0.0/scylla/src/cluster/metadata.rs#L608. This port is later propagated for other peers as well (is it even correct???).

Thanks for spotting the bug!

wprzytula:
> This port is later propagated for other peers as well (is it even correct???).

I took a blame walk and this is what I found:
image

wprzytula:
> > > : https://github.com/scylladb/scylla-rust-driver/blob/v1.0.0/scylla/src/cluster/metadata.rs#L608. This port is later propagated for other peers as well (is it even correct???).
> > 
> > 
> > Thanks for spotting the bug!
> 
> Is it a bug?

Is it not? Is the assumption that all nodes in the cluster expose the same port correct?

Lorak-mmk:
> 
> Is it not? Is the assumption that all nodes in the cluster expose the same port correct?

I don't know. If it is now, how can we discover other ports?

wprzytula:
> > Is it not? Is the assumption that all nodes in the cluster expose the same port correct?
> 
> I don't know. If it is not, how can we discover other ports?

OK, so in an internal discussion we established that all nodes shall follow the same configuration, including the _native_transport_port_. Thus, this is not a bug.

wprzytula:
> Why did you introduce new struct (`Coordinator`) instead of using `Node`?

`Node` is not clonable. `Arc<Node>` is clonable quite cheaply, but not as cheaply as custom structs filled with a bunch of `Copy` types are. Therefore, I still believe a tuple (be it a named tuple or a struct) is a better fit here.

wprzytula:
> We can of course consider wrapping `Node` (or whatever else we need) in a new `Coordinator` struct if we may want to add new stuff there (maybe a Option?).

Ah, I know why I didn't use `Node`. Then it would necessarily by passed by cloning `Arc`, which is an overhead I didn't want to introduce. Conversely, copying `(SocketAddr, Shard)` is very cheap. WDYT?

scylla/src/client/session.rs
```diff
@@ -412,6 +412,55 @@ impl Default for SessionConfig {
...
+pub struct Coordinator {
+    /// Translated address, i.e., one that the connection is opened against.
+    connection_address: SocketAddr,
+    /// Untranslated address, i.e., one that the node broadcast.
+    node_address: SocketAddr,
+    /// Unique ID of the node in the cluster.
+    host_id: Uuid,
+    /// Number of the shard, if applicable (present for ScyllaDB nodes, absent for Cassandra).
+    shard: Option<Shard>,
+}
```
muzarski, line 427: commit: `session: introduce Coordinator`

> 2. @muzarski - please make sure that `Coordinator`'s API meets the needs
   of cpp-rust-driver.

I think this is more than enough. The main use case of `cass_future_coordinator` is something like
```c++
// CassNode would be a wrapper over `Coordinator`
CassNode *n = cass_future_coordinator(cass_session_execute(...));
// This would enforce `SingleTargetLoadBalancingPolicy` using the host_id/node_address 
// (and most probably the shard).
cass_statement_set_node(some_statement, n);
```


The API you proposed, along with `SingleTargetLoadBalancingPolicy` is more than enough to implement this on cpp-rust-driver side.

scylla/src/network/connection.rs
```diff
@@ -22,9 +22,7 @@ use crate::frame::{
 use crate::policies::address_translator::{AddressTranslator, UntranslatedPeer};
```
muzarski, line 23: commit: `connection: remove dead code`

> There's no use in keeping them, especially that internal refactors often require adjusting them, too

I agree. This was really annoying during errors refactor.

scylla/src/client/session.rs
```diff
@@ -461,6 +461,18 @@ impl Coordinator {
...
+/// including non-[Session] APIs (`MetadataReader::query_metadata()`, `Connection::query_iter()`, etc.).
+///
+/// Care should be taken not to leak this [Coordinator] to any user-facing API.
+pub(crate) const MOCK_COORDINATOR: Coordinator = Coordinator {
+    host_id: Uuid::from_u128(0x_feed_dead_deaf_deed_cafe),
+    connection_address: SocketAddr::new(std::net::IpAddr::V6(std::net::Ipv6Addr::LOCALHOST), 42),
+    node_address: SocketAddr::new(std::net::IpAddr::V4(std::net::Ipv4Addr::LOCALHOST), 2137),
+    shard: Some(2137),
+};
+
```
muzarski, line 475: commit: `codewide: introduce Coordinator to QueryResult`

Ok, so I believe it's either this or holding `Coordinator` under `Option<>` in `QueryResult`. TBH, I don't have a strong opinion on this. Both have its pros and cons. Maybe I'm leaning slightly towards `Option<Coordinator>`, as it is not as bug prone as `MOCK_COORDINATOR` solution.

muzarski: If we decide to stay by `MOCK_COORDINATOR` solution, I'd at least 0-initialize the struct.

muzarski (commented):
Looks good

scylla/src/client/session.rs
wprzytula: Hmm. I though that using some crazy values will bring up our (or someone else's) attention quicker should anything go wrong.

wprzytula: But you're right in that zero IP address and zero host ID is obviously not a valid state.

wprzytula: > Ok, so I believe it's either this or holding `Coordinator` under `Option<>` in `QueryResult`. TBH, I don't have a strong opinion on this. Both have its pros and cons. Maybe I'm leaning slightly towards `Option<Coordinator>`, as it is not as bug prone as `MOCK_COORDINATOR` solution.

If we put `Option<Coordinator>` into `QueryResult`, shall we return `Option<Coordinator>` from `QueryResult::request_coordinator()` or rather panic on `None`?

What my point was, was that user should never be able to observe a `None` `Coordinator`. This is because the only requests whose `Coordinator` is unknown are driver-private, and their results are not exposed to the user. Thus the user should not be told to even consider the `None` case. Out of the two options:
1) to assert the correct behaviour in runtime by panicking at invalid use point (`request_coordinator()` is called on a `QueryResult` without a known `Coordinator`, or
2) to hide the type-level possibility of incorrect behaviour by introducing a mock. This let us pretend on the `QueryResult` level that there's no hazard - because I thought there would be no benefit from such visual threat.

wprzytula:
v1.1: changed contents of `MOCK_COORDINATOR` to be more blunt about the value being invalid.

Lorak-mmk (changes requested):
I'm not convinced about the approach taken here - I still think that `Coordinator` struct should contain `Arc<Node>` and `Option<Shard>`.
- The stated reason to do otherwise (performance) is not convincing at all. A single arc clone is extremely unlikely to have any impact on performance. If you believe otherwise and want this to affect the code, there should be a proof of the impact.
- You are removing one arc clone, but in exchange you increased the size of `QueryResult` from 104 bytes to 192 bytes (so nearly twice!). It is not at all obvious to me that it is a good performance tradeoff.
- Apart from performance what other reasons are there? I see that Coordinator stores both `connection_address` and `node_address`, but I see no explanation on why is that necessary.
- There are ergonomic issues with the current approach: in order to retrieve `Node` user needs to consult `ClusterState` which means error handling and additional code. And needing `Arc<Node>` is not a hard to imagine scenario, because it has multiple use cases that current coordinator does not fullfill: using it in LBP, calling methods like `is_connected`, `is_enabled` or `sharder`, retrieving its datacenter and rack (!!!)

wprzytula:
> I'm not convinced about the approach taken here - I still think that `Coordinator` struct should contain `Arc<Node>` and `Option<Shard>`.

What solution to the problem described [here](https://github.com/scylladb/scylla-rust-driver/pull/1287#discussion_r2053499175) does your approach suggest?

wprzytula:
> * I see that Coordinator stores both `connection_address` and `node_address`, but I see no explanation on why is that necessary.

I believe the user may be interested in both the untranslated and the translated address of the node. Depending on what is the exact scenario that address translation is used in, either of them may be more desirable.

wprzytula:
> * There are ergonomic issues with the current approach: in order to retrieve `Node` user needs to consult `ClusterState` which means error handling and additional code. And needing `Arc<Node>` is not a hard to imagine scenario, because it has multiple use cases that current coordinator does not fullfill: using it in LBP, calling methods like `is_connected`, `is_enabled` or `sharder`, retrieving its datacenter and rack (!!!)

This is true, thanks for bringing this up.

Lorak-mmk:
> What solution to the problem described https://github.com/scylladb/scylla-rust-driver/pull/1287#discussion_r2053499175 does your approach suggest?

Using Option (with `unwrap()` hidden from the user) is fine with me, because if I understand correctly, trying to access it when it is `None` means that there is a bug in the driver. Signaling bugs in the driver is a valid use case for panics.

Lorak-mmk:
> I believe the user may be interested in both the untranslated and the translated address of the node. Depending on what is the exact scenario that address translation is used in, either of them may be more desirable.

I was mostly asking for the use case of translated address, given that the translation is a process that the user controls.

wprzytula:
> > I believe the user may be interested in both the untranslated and the translated address of the node. Depending on what is the exact scenario that address translation is used in, either of them may be more desirable.
> 
> I was mostly asking for the use case of translated address, given that the translation is a process that the user controls.

Consider the Serverless Cloud case. Then the address translation is not that much the user's duty.

wprzytula:
Rebased on main.

wprzytula:
v1.2:
- `Coordinator` now stores `Arc<Node>`,
- `QueryResult` now holds `Option<Coordinator>`,
- `MOCK_COORDINATOR` is gone.

Lorak-mmk:
Regarding the semver issue with UnwindSafe:
After reading https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html I don't think it is critically important for impls of this trait to be precise.
"Note, however, that this is not an unsafe trait, so there is not a succinct contract that this trait is providing. Instead it is intended as more of a “speed bump” to alert users of catch_unwind that broken invariants may be witnessed and may need to be accounted for."
"the lack of unsafe means it is mostly an advisory. The [AssertUnwindSafe](https://doc.rust-lang.org/std/panic/struct.AssertUnwindSafe.html) wrapper struct can be used to force this trait to be implemented for any closed over variables passed to catch_unwind."

We can get around the error for now by:
- implementing UnwindSafe / RefUnwindSafe for some error types that contain `Arc<dyn Error>`. It should be perfectly fine to do in general. I see that  `FrameBodyExtensionsParseError`, `CqlRequestSerializationError`, `BrokenConnectionError` and `ConnectionError` require it.
- bumping Tokio to 1.40 (because the mpsc Sender starts to implement those traits in this version).
- Implementing the traits for `NodeConnectionPool`. It does not implement them automatically because of RemoteHandle. After reading the docs ( https://docs.rs/futures-util/latest/futures_util/future/struct.RemoteHandle.html ) I think the way we use it means we can implement it but I'm not sure.

It may be a good idea to actually analyze panic safety of our code. There are multiple places where we use user-supplied code (ser / deser traits, policies etc). Are we sure that our logical invariants are not broken if such user code panics?
I am definitely not sure. Maybe checking this is a good idea for 2.0? If you agree, then please open an issue about it and for now just implement my proposed workaround.

wprzytula:
Rebased on main.

wprzytula:
Solved the problem around lack of `UnwindSafe` autoimplementation.

scylla/src/client/session.rs
```diff
@@ -414,6 +414,46 @@ impl Default for SessionConfig {
...
+        &self.node
+    }
+
+    /// Number of the shard, if applicable (present for ScyllaDB nodes, absent for Cassandra).
+    #[inline]
+    pub fn shard(&self) -> Option<Shard> {
+        self.shard
+    }
+}
+
```
Lorak-mmk, line 456: Regarding a place for this struct, I also don't know which is best. Alternatives I think are worth considering:
1. `client/session.rs` (current)
2. `client/coordinator.rs` (new file)
3. `cluster/node.rs`
4. `cluster/coordinator.rs` (new file)
5. `response/coordinator.rs` (new file)

Let's discuss where it should go.If you see any arguments for any of the places please share them.

One small disadvantage I see with 1 is that this struct will also be used by CachingSession, so it is a bit weird for it to be in the session.

scylla/src/client/session.rs
Lorak-mmk, line 1: Commit: `codewide: introduce Coordinator to QueryResult`

One possible solution (which we should definitely not pursue now, maybe in distant future it this becomes a problem) is to not use QueryResult internally, and instead have some InternalQueryResult that does not contains Coordinator.

scylla/src/response/request_response.rs
```diff
@@ -55,29 +56,54 @@ impl NonErrorQueryResponse {
         }
     }
 
-    pub(crate) fn into_query_result_and_paging_state(
+    fn into_query_result_and_paging_state_with_maybe_unknown_coordinator(
         self,
+        request_coordinator: Option<Coordinator>,
     ) -> Result<(QueryResult, PagingStateResponse), RequestAttemptError> {
-        let (raw_rows, paging_state_response) = match self.response {
+        let Self {
```
Lorak-mmk, line 63: I hate names that are this long, but I don't see a better solution.

scylla/src/client/session.rs
wprzytula: > Regarding a place for this struct, I also don't know which is best. Alternatives I think are worth considering:
> 
>   1. `client/session.rs` (current)
> 
>   2. `client/coordinator.rs` (new file)
> 
>   3. `cluster/node.rs`
> 
>   4. `cluster/coordinator.rs` (new file)
> 
>   5. `response/coordinator.rs` (new file)
> 
> 
> Let's discuss where it should go.If you see any arguments for any of the places please share them.

1.
    - I'd prefer not to make `session.rs` any larger (-);
    - `Coordinator` is only exposed on response results (`QueryResult`, `QueryPager`) returned from `Session` API (+);
2.
    - Makes sense in general. (+)
    - OTOH, `Coordinator` is such a small struct that it seems overkill to introduce a separate module for it. (-)
3.
4.
    - I'm strongly against. `cluster` represents entities that are independent from requests and stable across requests. At the same time, `Coordinator` is strictly about a specific request, that's why it does not feel good here (--)
5.
    - It's actually the response that the users can obtain `Coordinator` from, so it makes sense (+)
    - It's more the _request's_ thing which node is a `Coordinator`, not the _response's_ thing. But this is a minor nit. (-)

muzarski (approved):

scylla/src/client/session.rs
muzarski: I like response/coordinator.rs most

Lorak-mmk: `response/coordinator.rs` is fine with me too

wprzytula:
Rebased on main.

wprzytula:
v1.3:
- Moved `Coordinator` to `response/coordinator.rs`.

muzarski (approved):

Lorak-mmk (approved):

PR #1294, opened 2025-03-25
Fix example type annotations

muzarski:
## Pre-review checklist

### DIFF
--- a/examples/compare-tokens.rs
+++ b/examples/compare-tokens.rs
@@ -53,7 +53,7 @@ async fn main() -> Result<()> {
             )
             .await?
             .into_rows_result()?
-            .single_row()?;
+            .single_row::<(i64,)>()?;
         assert_eq!(t, qt);
         println!("token for {}: {}", pk, t);
     }


muzarski:
@Lorak-mmk I did no changes yet - the CI indeed fails

examples/compare-tokens.rs
```diff
@@ -53,7 +53,7 @@ async fn main() -> Result<()> {
             )
             .await?
             .into_rows_result()?
-            .single_row()?;
+            .single_row::<(i64,)>()?;
```
wprzytula, line 56: :question: Should we file a bug report in rustc?

muzarski: I don't think it's a rustc bug. The only place the type can be inferred from is `assert_eq!(t, qt);`, where `t: i64`. I think that error message sounds reasonable:
```
 57 |         assert_eq!(t, qt);
   |         ----------------- type must be known at this point
   |
   = note: multiple `impl`s satisfying `i64: std::cmp::PartialEq<_>` found in the following crates: `core`, `deranged`:
           - impl std::cmp::PartialEq for i64;
           - impl<MIN, MAX> std::cmp::PartialEq<deranged::RangedI64<MIN, MAX>> for i64
             where the constant `MIN` has type `i64`, the constant `MAX` has type `i64`;
```

This might be due to the `deranged` dependency update. See: https://docs.rs/deranged/0.4.1/deranged/struct.RangedI64.html#impl-PartialEq%3CRangedI64%3CMIN,+MAX%3E%3E-for-i64

wprzytula (approved):

Lorak-mmk (approved):

Issue #1273, opened 2025-03-06
Murmur3Partitioner and hash_one() method should be public


muzarski:
There is a use case for that in cql-stress. See: https://github.com/scylladb/cql-stress/blob/8951aaff854b51bbba653ab82134a33597392c05/src/bin/cql-stress-cassandra-stress/java_generate/values/mod.rs#L152-L168.

`hash_one()` method is actually a `Partitioner` trait method. There are multiple solutions to that:
- pubify the `Partitioner` trait and implementors again (revert https://github.com/scylladb/scylla-rust-driver/pull/1222/commits/f805725d8868f77cda10026b77f3fdbb89dbcb6e)
- pubify `Murmur3Partitioner` and expose some equivalent method on the struct
- introduce some standalone public method that hashes a blob using murmur3

muzarski:
cc: @soyacz

PR #1222, opened 2025-02-11
`partitioner`: Change visibility of  `Partitioner`-related items

wprzytula:
# Motivation

We want to extract `session_test.rs` to integration tests directory - out of `scylla` crate.

Tests in `session_test.rs` call `get_partitioner_name()` to assert some things. We must therefore expose `get_partitioner_name().

# What's done
1. `PartitionerName` and `PreparedStatement::get_partitioner_name()` are made `pub`.
2. `PartitionerName` gets missing `Eq` impl and `#[non_exhaustive]` attribute.
3. `calculate_token_for_partition_key()` now accepts `PartitionerName` instead of `impl Partitioner`.
4. `Partitioner(Hasher)` traits and their impls are un`pub`bed.

## Bonus
Implemented `Debug` for `QueryPager` and `TypedRowStream`.

## Pre-review checklist

- [x] I have split my patch into logically separate commits.
- [x] All commit messages clearly explain what they change and why.
- ~~[ ] I added relevant tests for new features and bug fixes.~~
- [ ] All commits compile, pass static checks and pass test.
- [x] PR description sums up the changes and reasons why they should be introduced.
- ~~[ ] I have provided docstrings for the public items that I want to introduce.~~
- ~~[ ] I have adjusted the documentation in `./docs/source/`.~~
- ~~[ ] I added appropriate `Fixes:` annotations to PR description.~~

PR #1298, opened 2025-03-26
partitioner: expose public method to compute murmur3 hash

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1273

I pubified `Murmur3Partitioner` and introduced a public method that computes the hash. This is the second solution from the solutions listed in the issue.

The `partitioner` module is already pub, so no need to re-export the struct.

## Pre-review checklist


Lorak-mmk:
I'm not a fan of exposing something such specific.

If we want to allow users to compute hash of some bytes, is there a reason to not reverse un-pubbing Partitioner trait and implementations?
I was under the impression that this interface is quite good and stable and we don't expect changes to it.

Another reason is that the Partitioner interface is better for performance, which was afaik the main reason to introduce it (and there were benchmarks for that). 

This is the commit that un-pubbed Partitioner and impls: https://github.com/scylladb/scylla-rust-driver/pull/1222/commits/f805725d8868f77cda10026b77f3fdbb89dbcb6e
The rationale there is:
- `We do not expect users to implement their own Partitioners` - this is still true, so let's make it sealed
- `nor use their methods` - turns out to not be true, we have important use case for that. So let's make it public again.

muzarski:
> This is the commit that un-pubbed Partitioner and impls: [f805725](https://github.com/scylladb/scylla-rust-driver/commit/f805725d8868f77cda10026b77f3fdbb89dbcb6e) The rationale there is:
> 
>     * `We do not expect users to implement their own Partitioners` - this is still true, so let's make it sealed
> 
>     * `nor use their methods` - turns out to not be true, we have important use case for that. So let's make it public again.

Makes sense, so I'll seal the traits and pubify the implementors.

Lorak-mmk (approved):

scylla/src/routing/partitioner.rs
```diff
@@ -75,15 +76,19 @@ impl PartitionerHasher for PartitionerHasherAny {
 }
 
 mod sealed {
-    pub(crate) trait Sealed {}
+    // This is a sealed trait - its whole purpose is to be unnameable.
+    // This means we need to disable the check.
+    #[allow(unknown_lints)] // Rust 1.70 doesn't know this lint
```
wprzytula, line 81: :question: Is our MSRV still 1.70?

wprzytula: Oh, it actually still is. That's because we haven't ported the MSRV bump from hackathon branch yet, right?

scylla/src/routing/partitioner.rs
```diff
@@ -37,6 +37,7 @@ impl PartitionerName {
     }
 }
 
+impl sealed::Sealed for PartitionerName {}
 impl Partitioner for PartitionerName {
```
wprzytula, line 41: :wrench: commit message mistake: `We will un-pub them in the next commit.` -> should be without `un-`.

wprzytula (approved):

scylla/src/routing/partitioner.rs
Lorak-mmk: Yes, it is part of my PR with CCM backport.

muzarski:
Fixed commit message typo

wprzytula (approved):

Lorak-mmk (approved):

wprzytula:
The PR's name and cover letter are outdated. @muzarski Please make it reflect its actual content.

Issue #1027, opened 2024-07-04
Enable configuring local address and range of ports for client sockets to be bound to


wprzytula:
## Motivation

Cpp driver exposes the following functions. Their docstrings contain motivation for both.
```c
/**
 * Sets the local address to bind when connecting to the cluster,
 * if desired.
 *
 * @public @memberof CassCluster
 *
 * @param[in] cluster
 * @param[in] name IP address to bind, or empty string for no binding.
 * Only numeric addresses are supported; no resolution is done.
 * @return CASS_OK if successful, otherwise an error occurred.
 */
CASS_EXPORT CassError
cass_cluster_set_local_address(CassCluster* cluster,
                               const char* name);
/**
 * Sets the range of outgoing port numbers (ephemeral ports) to be
 * used when establishing the shard-aware connections. This is
 * applicable when the routing of connection to shard is based on
 * the client-side port number.
 * 
 * When application connects to multiple CassCluster-s it is advised
 * to assign mutually non-overlapping port intervals to each. It is assumed
 * that the supplied range is allowed by the OS (e.g. it fits inside
 * /proc/sys/net/ipv4/ip_local_port_range on *nix systems)
 *
 * @public @memberof CassCluster
 *
 * @param[in] cluster
 * @param[in] lo The lower bound (inclusive)
 * @param[in] hi The upper bound (exclusive)
 * @return CASS_OK if successful, otherwise the range is invalid.
 */
CassError cass_cluster_set_local_port_range(CassCluster* cluster,
                                            int lo,
                                            int hi);
```

Implementing them both make sense for Rust driver itself, as well as to support them in cpp-rust-driver.

## Solution

Add the following methods to `SessionBuilder` (and corresponding config to `SessionConfig`, and respect it codewide):

```rust
fn driver_bind_address(self, addr: SocketAddr) -> Self;

fn driver_bind_port_range(self, port_range: Range<u16>) -> Self;
```

PR #1300, opened 2025-03-27
Implement local bind address configuration option

muzarski:
Ref: https://github.com/scylladb/scylla-rust-driver/issues/1027

This PR solves the first part of the issue: implementing local ip address binding logic.

I implemented the tests for this feature as well. 

~Actually, I have a question regarding these:
I defined a test `valid_local_ip_address` which binds the socket to `172.42.0.1`, which should be the address associated with the interface assigned to docker subnet defined in the docker-compose file (for both Scylla and Cassandra). This means that the test will obviously fail if the cluster is accessible via some other interface. For this reason, I introduced `docker_cluster_tests` cfg option and enable it for scylla (non-cloud) and cassandra tests. I wonder if this is a good solution. Maybe introducing the `cfg` option is an overkill, and we can assume that developers run the tests locally against the docker-compose cluster (e.g. by running `make up`)?~ <- Solved

## Pre-review checklist

- ~[] I have adjusted the documentation in `./docs/source/`.~

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -905,6 +905,7 @@ impl Session {
         };
 
         let connection_config = ConnectionConfig {
+            local_ip_address: config.local_ip_address,
             compression: config.compression,
             tcp_nodelay: config.tcp_nodelay,
             tcp_keepalive_interval: config.tcp_keepalive_interval,
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -47,7 +47,7 @@ use scylla_cql::serialize::batch::BatchValues;
 use scylla_cql::serialize::row::{SerializeRow, SerializedValues};
 use std::borrow::Borrow;
 use std::future::Future;
-use std::net::SocketAddr;
+use std::net::{IpAddr, SocketAddr};
 use std::num::NonZeroU32;
 use std::sync::Arc;
 use std::time::Duration;
@@ -154,6 +154,13 @@ pub struct SessionConfig {
     /// Session will connect to these nodes to retrieve information about other nodes in the cluster.
     /// Each node can be represented as a hostname or an IP address.
     pub known_nodes: Vec<KnownNode>,
+
+    /// A local ip address to bind all driver's TCP sockets to.
+    ///
+    /// By default set to None, which is equivalent to:
+    /// - `INADDR_ANY` for IPv4 ([`Ipv4Addr::UNSPECIFIED`][std::net::Ipv4Addr::UNSPECIFIED])
+    /// - `in6addr_any` for IPv6 ([`Ipv6Addr::UNSPECIFIED`][std::net::Ipv6Addr::UNSPECIFIED])
+    pub local_ip_address: Option<IpAddr>,
 
     /// Preferred compression algorithm to use on connections.
     /// If it's not supported by database server Session will fall back to no compression.
@@ -293,6 +300,7 @@ impl SessionConfig {
     pub fn new() -> Self {
         SessionConfig {
             known_nodes: Vec::new(),
+            local_ip_address: None,
             compression: None,
             tcp_nodelay: true,
             tcp_keepalive_interval: None,
--- a/scylla/src/client/session_builder.rs
+++ b/scylla/src/client/session_builder.rs
@@ -16,7 +16,7 @@ use crate::policies::timestamp_generator::TimestampGenerator;
 use crate::statement::Consistency;
 use std::borrow::Borrow;
 use std::marker::PhantomData;
-use std::net::SocketAddr;
+use std::net::{IpAddr, SocketAddr};
 use std::num::NonZeroU32;
 #[cfg(feature = "unstable-cloud")]
 use std::path::Path;
@@ -395,6 +395,31 @@ impl CloudSessionBuilder {
 // This block contains configuration options that make sense both for Cloud and non-Cloud
 // `Session`s. If an option fit only one of them, it should be put in a specialised block.
 impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
+    /// Sets the local ip address all TCP sockets are bound to.
+    ///
+    /// By default, this option is set to `None`, which is equivalent to:
+    /// - `INADDR_ANY` for IPv4 ([`Ipv4Addr::UNSPECIFIED`][std::net::Ipv4Addr::UNSPECIFIED])
+    /// - `in6addr_any` for IPv6 ([`Ipv6Addr::UNSPECIFIED`][std::net::Ipv6Addr::UNSPECIFIED])
+    ///
+    /// # Example
+    /// ```
+    /// # use scylla::client::session::Session;
+    /// # use scylla::client::session_builder::SessionBuilder;
+    /// # use std::net::Ipv4Addr;
+    /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let session: Session = SessionBuilder::new()
+    ///     .known_node("127.0.0.1:9042")
+    ///     .local_ip_address(Some(Ipv4Addr::new(192, 168, 0, 1)))
+    ///     .build()
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn local_ip_address(mut self, local_ip_address: Option<impl Into<IpAddr>>) -> Self {
+        self.config.local_ip_address = local_ip_address.map(Into::into);
+        self
+    }
+
     /// Set preferred Compression algorithm.
     /// The default is no compression.
     /// If it is not supported by database server Session will fall back to no encryption.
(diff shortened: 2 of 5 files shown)


Lorak-mmk (changes requested):

scylla/src/client/session.rs
```diff
@@ -134,6 +134,13 @@ pub struct SessionConfig {
     /// Each node can be represented as a hostname or an IP address.
     pub known_nodes: Vec<KnownNode>,
 
+    /// A local ip address to bind the TCP socket to.
+    ///
+    /// By default set to None, which is equivalent to:
+    /// - `INADDR_ANY` for IPv4 ([`Ipv4Addr::UNSPECIFIED`][std::net::Ipv4Addr::UNSPECIFIED])
+    /// - `in6addr_any` for IPv6 ([`Ipv6Addr::UNSPECIFIED`][std::net::Ipv6Addr::UNSPECIFIED])
+    pub local_ip_address: Option<IpAddr>,
+
```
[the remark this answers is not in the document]
muzarski: Good question. I need to check

scylla/tests/integration/new_session.rs
```diff
@@ -36,3 +38,60 @@ async fn all_hostnames_invalid() {
...
+    let session_builder = create_new_session_builder()
+        // Bind the address to local host - it is used by serverless cluster.
+        .local_ip_address(Some(Ipv4Addr::new(127, 0, 0, 1)));
+    let session = session_builder.build().await.unwrap();
+
+    session
+        .query_unpaged("SELECT host_id FROM system.local WHERE key='local'", &[])
+        .await
+        .unwrap();
+}
```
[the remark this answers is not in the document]
muzarski: > One way for that could be to:
> 
>     * Get the IP used by some node

Do you mean just reading (and resolving) the `SCYLLA_URIx` env variables?

>     * Read the network of that IP. We could parse `ip addr` output, or parse `/proc` or `/sys` (I'm sure it is there somewhere) or use 3rd party library.

Makes sense.

>     * Bind to some random address in that network (even better - find a free address like you did in ip allocator in ccm).

Will it work? Doesn't the socket need to be bound to the interface address? I quickly checked it locally, and modified the `valid_ip_port` test so we bind to `172.42.0.5` (instead of `172.42.0.1` - the output from `ip a`). The `bind` operation fails with `EADDRNOTAVAIL`:
```
       EADDRNOTAVAIL
              A nonexistent interface was requested or the requested address was not local.
```

scylla/src/client/session.rs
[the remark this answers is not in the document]
muzarski: It seems that Scylla supports it: https://opensource.docs.scylladb.com/stable/operating-scylla/admin.html#ipv6-addresses.

scylla/src/client/session.rs
```diff
@@ -134,6 +134,13 @@ pub struct SessionConfig {
     /// Each node can be represented as a hostname or an IP address.
     pub known_nodes: Vec<KnownNode>,
 
+    /// A local ip address to bind the TCP socket to.
+    ///
+    /// By default set to None, which is equivalent to:
+    /// - `INADDR_ANY` for IPv4 ([`Ipv4Addr::UNSPECIFIED`][std::net::Ipv4Addr::UNSPECIFIED])
+    /// - `in6addr_any` for IPv6 ([`Ipv6Addr::UNSPECIFIED`][std::net::Ipv6Addr::UNSPECIFIED])
+    pub local_ip_address: Option<IpAddr>,
```
wprzytula, line 142: :wrench:

scylla/src/client/session_builder.rs
```diff
@@ -395,6 +395,31 @@ impl CloudSessionBuilder {
 // This block contains configuration options that make sense both for Cloud and non-Cloud
 // `Session`s. If an option fit only one of them, it should be put in a specialised block.
 impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
+    /// Sets the local ip address bound to the TCP socket.
```
wprzytula, line 398: :wrench:

wprzytula:
```suggestion
    /// Sets the local ip address all TCP sockets are bound to.
```

scylla/src/network/connection.rs
```diff
@@ -2034,26 +2033,28 @@ pub(super) async fn open_connection_to_shard_aware_port(
...
+    source_port: Option<u16>,
 ) -> Result<TcpStream, std::io::Error> {
-    match addr {
+    // Binding to port 0 is equivalent to choosing random ephemeral port.
+    let source_port = source_port.unwrap_or(0);
+
+    match connect_address {
         SocketAddr::V4(_) => {
+            // If source_ip not provided, bind to INADDR_ANY.
+            let source_ipv4 = source_ip.unwrap_or(Ipv4Addr::new(0, 0, 0, 0).into());
```
wprzytula, line 2051: :pick:

wprzytula:
```suggestion
            // If source_ip not provided, bind to INADDR_ANY.
            let source_ipv4 = source_ip.unwrap_or(Ipv4Addr::UNSPECIFIED.into());
```

scylla/src/network/connection.rs
```diff
@@ -2034,26 +2033,28 @@ pub(super) async fn open_connection_to_shard_aware_port(
...
-                Ipv4Addr::new(0, 0, 0, 0).into(),
-                source_port,
-            ))?;
-            Ok(socket.connect(addr).await?)
+            socket.bind(SocketAddr::new(source_ipv4, source_port))?;
+            Ok(socket.connect(connect_address).await?)
         }
         SocketAddr::V6(_) => {
+            // If source_ip not provided, bind to in6addr_any.
+            let source_ipv6 = source_ip.unwrap_or(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 0).into());
```
wprzytula, line 2058: :pick:

wprzytula:
```suggestion
            // If source_ip not provided, bind to in6addr_any.
            let source_ipv6 = source_ip.unwrap_or(Ipv6Addr::UNSPECIFIED.into());
```

scylla/tests/integration/new_session.rs
```diff
@@ -36,3 +38,60 @@ async fn all_hostnames_invalid() {
...
+
+#[cfg_attr(not(scylla_cloud_tests), ignore)]
+#[tokio::test]
+async fn cloud_valid_local_ip_address() {
+    setup_tracing();
+
+    let session_builder = create_new_session_builder()
+        // Bind the address to local host - it is used by serverless cluster.
+        .local_ip_address(Some(Ipv4Addr::new(127, 0, 0, 1)));
```
wprzytula, line 90: :pick:

wprzytula:
```suggestion
        // Bind the address to local host - it is used by serverless cluster.
        .local_ip_address(Some(Ipv4Addr::LOCALHOST));
```

wprzytula (changes requested):

Lorak-mmk:
> I wanted to do similarly for analogous cloud test, but I don't see a way to obtain the node ip address from driver's public API (and SCYLLA_URI env variable is not passed in cloud case).

What about https://docs.rs/scylla/latest/scylla/cluster/struct.Node.html#structfield.address ?
You could first create a normal session, retrieve IP using this field, and then create another session with local-address binding. No need for separate cloud test then.

muzarski:
> You could first create a normal session, retrieve IP using this field, and then create another session with local-address binding.

Sounds a bit like a hack, but I have nothing against that (since this is the test).

muzarski:
v1.2: Applied @Lorak-mmk suggestion - merged non-cloud and cloud test cases into one. Thanks!

Lorak-mmk (dismissed):

scylla/tests/integration/utils.rs
```diff
@@ -30,6 +31,31 @@ pub(crate) fn setup_tracing() {
...
+/// This function uses the `ip route get` command to get the routing information for the destination IP.
+pub(crate) fn find_local_ip_for_destination(dest: IpAddr) -> Option<IpAddr> {
+    let output = Command::new("ip")
+        .arg("route")
+        .arg("get")
+        .arg(dest.to_string())
+        .output()
+        .ok()?;
+
+    let output_str = String::from_utf8_lossy(&output.stdout);
```
wprzytula, line 45: :question: Why not https://doc.rust-lang.org/std/str/fn.from_utf8.html?
```suggestion
    let output_str = std::str::from_utf8(&output.stdout).ok()?;
    
```

wprzytula (dismissed):
If only you verified manually that the driver's sockets actually get bound to the given address, LGTM.

wprzytula (approved):

muzarski:
Rebased on main. Once CI passes, I think we can merge

PR #1303, opened 2025-04-01
Shard aware port range

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1027

This PR introduces configuration option for shard-aware local port range.

Since not all ranges are valid (e.g. empty one, or one including reserved ports i.e. ports < 1024), I introduced a simple wrapper called `ShardAwarePortRange` which validates provided range in constructor.

Currently, `Sharder` is `pub`, so are its fields, and its methods responsible for computing local ports. This is why, I couldn't modify the sharder in any of the following ways:
- adding `ShardAwarePortRange` field to the `Sharder`
- providing `ShardAwarePortRange` parameter to `Sharder` methods

This is why, I introduced separate set of methods which additionally accept the port range. Current public methods are using new methods by providing a default range (ephemeral port range `[49152, 65535]`).

## Pre-review checklist

### DIFF
--- a/scylla/src/routing/sharding.rs
+++ b/scylla/src/routing/sharding.rs
@@ -1,10 +1,50 @@
 use std::collections::HashMap;
 use std::num::NonZeroU16;
+use std::ops::RangeInclusive;
 
 use rand::Rng as _;
 use thiserror::Error;
 
 use super::Token;
+
+/// A range of ports that can be used for shard-aware connections.
+///
+/// The range is inclusive and has to be a sub-range of [1024, 65535].
+#[derive(Debug, Clone)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+#[allow(dead_code)] // TODO: remove later in this PR.
+pub struct ShardAwarePortRange(RangeInclusive<u16>);
+
+impl ShardAwarePortRange {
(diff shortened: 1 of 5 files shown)


scylla/src/routing/sharding.rs
```diff
@@ -1,11 +1,50 @@
...
+    /// 1. Provided range is empty (`end` < `start`).
+    /// 2. Provided range starts at a port lower than 1024. Ports 0-1023 are reserved and
+    ///    should not be used by application.
+    pub fn new(range: impl Into<RangeInclusive<u16>>) -> Result<Self, InvalidShardAwarePortRange> {
+        let range = range.into();
+        if range.is_empty() || range.start() < &1024 {
+            return Err(InvalidShardAwarePortRange);
+        }
+        Ok(Self(range))
+    }
```
wprzytula, line 33: :pick: Let's add `#[inline]` to allow cross-crate inlining of constructor calls.

wprzytula (dismissed):

Lorak-mmk (dismissed):

Lorak-mmk (approved):

wprzytula (approved):

PR #1307, opened 2025-04-02
Add information about server application version to metadata

### LINKED ISSUE #1286: Add server release version info to metadata
This would directly correspond to [`cass_schema_meta_version`](https://github.com/scylladb/cpp-rust-driver/blob/v0.3.0/include/cassandra.h#L3328-L3338).

What cpp-driver means by server version, is the output of `SELECT release_version FROM * system.local` query executed on control connection.

The same `release_version` column is also present in `system.peers`. As a bonus, we could also fetch version information for each peer and store it in `Peer` and `Node` structs.

While it may not be useful for Scylla users, because of https://github.com/scylladb/scylladb/issues/8740 (TLDR; release_version is hardcoded to 3.0.8), it definitely is a nice feature to have for C* users.

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1286

This PR adds server application version information to the metadata.

It is exposed in two places:
1. `Node` struct - a server version of specific node (`release_version` from `system.local/system.peers`)
2. `ClusterState` - a cluster version, which is the version of the node currently used for control connection (`release_version` from `system.local`)

I'm not entirely sure about the second place. Maybe it's not really needed. I added some context to the corresponding commit message, so we can discuss this.

I implemented the test which validates the version for Scylla. It's possible, because Scylla currently hardcodes the `release_version` column to `3.0.8` and it does not seem like it's going to be addressed in the near future (the reason is some java-driver compatibility issues).

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/session.rs
+++ b/scylla/tests/integration/session.rs
@@ -1013,6 +1013,45 @@ async fn test_raw_use_keyspace() {
         .is_ok());
 }
 
+/// This test will work only for Scylla, as currently `release_version`
+/// column always contains '3.0.8' value.
+/// See: https://github.com/scylladb/scylladb/issues/8740
+#[cfg_attr(cassandra_tests, ignore)]
+#[tokio::test]
+async fn test_metadata_scylla_release_version() {
+    setup_tracing();
+
+    let session = create_new_session_builder().build().await.unwrap();
+    let hardcoded_scylla_version = "3.0.8";
+
+    // Get release_version manually.
+    let release_version = session
+        .query_unpaged(
+            "SELECT release_version FROM system.local WHERE key='local'",
+            &[],
+        )
+        .await
+        .unwrap()
+        .into_rows_result()
+        .unwrap()
+        .single_row::<(String,)>()
+        .unwrap()
+        .0;
+    assert_eq!(&release_version, hardcoded_scylla_version);
+
+    let cluster_state = session.get_cluster_state();
+
+    let cluster_version = cluster_state.cluster_version().unwrap();
+    assert_eq!(cluster_version, hardcoded_scylla_version);
+
+    for node in cluster_state.get_nodes_info() {
+        assert_eq!(
+            node.server_version.as_deref(),
+            Some(hardcoded_scylla_version)
+        );
+    }
+}
+
 #[tokio::test]
 async fn test_fetch_system_keyspace() {
     setup_tracing();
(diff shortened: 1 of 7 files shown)


scylla/tests/integration/session.rs
```diff
@@ -1013,6 +1013,45 @@ async fn test_raw_use_keyspace() {
...
+    assert_eq!(cluster_version, hardcoded_scylla_version);
+
+    for node in cluster_state.get_nodes_info() {
+        assert_eq!(
+            node.server_version.as_deref(),
+            Some(hardcoded_scylla_version)
+        );
+    }
+}
+
```
Lorak-mmk, line 1054: You can make this test work with C* too by getting rid of hardcoded version and using the result of manuala select as the source of truth.

Lorak-mmk (commented):
I did not review the code carefully yet, this is just a general comment.

I think that having the version in ClusterState too (so not only in Node) makes sense, because it makes the field much better to use. One change I'd do in this regard is putting a comment saying that the semantic of this field may change in the future, for example we may:
- Use `None` if not all nodes have the same version
- Concat versions if not all are the same to represent cluster that is being upgraded.


This PR makes sense for Cassandra, and for cpp-rust-driver, but is not very useful for Scylla users, who are our primary target.

Maybe we could expose Scylla version in ClusterState too?
I see that it is available for local node in `system.versions` (table existing only on Scylla).
We would need to make sure that this table is stable enough to be used in the driver (cc @piodul ).

scylla/tests/integration/session.rs
muzarski: True! I'll do that

piodul:
> We would need to make sure that this table is stable enough to be used in the driver (cc @piodul ).

@Lorak-mmk could you elaborate on that?

Looks like the table was introduced back in https://github.com/scylladb/scylladb/pull/9517 and its schema hasn't changed since.

Lorak-mmk:
> @Lorak-mmk could you elaborate on that?

> Looks like the table was introduced back in https://github.com/scylladb/scylladb/pull/9517 and its schema hasn't changed since.

Do we know that this table / column won't be removed in the future?

piodul:
> > @Lorak-mmk could you elaborate on that?
> 
> > Looks like the table was introduced back in [scylladb/scylladb#9517](https://github.com/scylladb/scylladb/pull/9517) and its schema hasn't changed since.
> 
> Do we know that this table / column won't be removed in the future?

I doubt it, but I think it's best to ask the author. cc @denesb

denesb:
In general we maintain backwards compatibility for public interfaces of the DB and virtual tables are such a public interface. So I don't see this table or column going away, without some elaborate deprecation schedule at the very least.
To be really sure, add a test which is run as part of CI and will break when we want to release a ScyllaDB version which happens to remove this column.

Lorak-mmk:
Let's close this for now. This feature turns out to not be so easy to implement. When we decide on an approach, we can open a new PR.

PR #1309, opened 2025-04-03
appease 1.86 toolchain's clippy

wprzytula:
One thing: I disagree that the indentation in tablets.rs is bad. Therefore, I disabled the lint there.

## Pre-review checklist

### DIFF
--- a/scylla/src/policies/load_balancing/default.rs
+++ b/scylla/src/policies/load_balancing/default.rs
@@ -2862,7 +2862,7 @@ mod latency_awareness {
                 | RequestAttemptError::DbError(DbError::IsBootstrapping, _)
                 | RequestAttemptError::DbError(DbError::Unavailable { .. }, _)
                 | RequestAttemptError::DbError(DbError::Unprepared { .. }, _)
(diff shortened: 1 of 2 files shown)


Lorak-mmk:
Which part of the comment does clippy complain about?
I agree with the indentation of arguments (`/// * `removed_nodes`:` etc), but there is also this section at the bottom:
image

Looking at the example at clippy docs ( https://rust-lang.github.io/rust-clippy/master/index.html#doc_overindented_list_items ) it looks to me like it may be complaining about this section, and rightfully so - the indentation there is actually wrong.

wprzytula:
> Which part of the comment does clippy complain about? I agree with the indentation of arguments (`` /// * `removed_nodes`: `` etc), but there is also this section at the bottom: image
> 
> Looking at the example at clippy docs ( https://rust-lang.github.io/rust-clippy/master/index.html#doc_overindented_list_items ) it looks to me like it may be complaining about this section, and rightfully so - the indentation there is actually wrong.

You're right about the other issue, I missed it. Fixed now. Clippy still complains about the indentation of arguments, so I still need to disable the lint there.

Lorak-mmk (approved):

PR #1315, opened 2025-04-07
tests: udt deserialization with metadata present but value missing

muzarski:
This test checks that `DeserializeValue` macro behaves correctly in case of https://github.com/apache/cassandra/blob/4a80daf32eb4226d9870b914779a1fc007479da6/doc/native_protocol_v4.spec#L1003.

1. Serialize the UDT with a missing field value
2. Deserialize the serialized UDT, but include the missing field in metadata
3. Expect that the missing field value is treated as null

I don't know of any way to simulate this using the Scylla/C* cluster. CQL protocol mentions that such scenario is possible, but does not specify when it can appear.

Side note: cpp-driver throws an error in such case. After the discussion with @wprzytula, we concluded that such behaviour is not correct, since protocol specifications clearly states that this scenario (metadata present, serialized bytes missing) is possible - it should not be treated as error.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types_tests.rs
+++ b/scylla-cql/src/types_tests.rs
@@ -154,6 +154,47 @@ mod derive_macros_integration {
                 assert_eq!(deserialized_udt, expected_udt);
             }
         }
+
+        #[test]
+        fn derive_ser_de_with_metadata_present_but_value_missing_during_deser() {
+            #[derive(scylla_macros::SerializeValue, Debug)]
+            #[scylla(crate = "crate")]
+            struct SerUdt<'a> {
+                a: &'a str,
+            }
+
+            #[derive(scylla_macros::DeserializeValue, Debug, PartialEq, Eq)]
+            #[scylla(crate = "crate")]
+            struct DeserUdt<'a> {
+                a: &'a str,
+                b: Option<i32>,
+            }
+
+            let ser_udt = SerUdt {
+                a: "The quick brown fox",
+            };
+            // Do not serialize `b` field.
+            let serialization_type =
+                udt_def_with_fields([("a", ColumnType::Native(NativeType::Text))]);
+
+            let expected_deserialized_udt = DeserUdt {
+                a: "The quick brown fox",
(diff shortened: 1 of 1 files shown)


scylla-cql/src/types_tests.rs
```diff
@@ -154,6 +154,47 @@ mod derive_macros_integration {
                 assert_eq!(deserialized_udt, expected_udt);
             }
         }
+
+        #[test]
+        fn derive_serialize_and_deserialize_metadata_present_value_missing_during_deser() {
```
wprzytula, line 159: :pick:

wprzytula:
```suggestion
        fn derive_ser_de_with_metadata_present_but_value_missing_during_deser() {
```

wprzytula: The name is long, but at the same time a bit hard to parse.

wprzytula (approved):

wprzytula (dismissed):

scylla-cql/src/types_tests.rs
```diff
@@ -154,6 +154,47 @@ mod derive_macros_integration {
...
+            let deserialization_type = udt_def_with_fields([
+                ("a", ColumnType::Native(NativeType::Text)),
+                ("b", ColumnType::Native(NativeType::Int)),
+            ]);
+
+            let serialized_udt = Bytes::from(do_serialize(&ser_udt, &serialization_type));
+
+            let deserialized_udt =
+                deserialize::<DeserUdt<'_>>(&deserialization_type, &serialized_udt).unwrap();
+            assert_eq!(deserialized_udt, deser_udt);
```
Lorak-mmk, line 196: Nit: `expected_deserialized_udt` would be a better name than `deser_udt`.

Lorak-mmk (dismissed):

Lorak-mmk (approved):

wprzytula (approved):

Issue #1003, opened 2024-05-21
Add `MaybeUnset::from_option()` to reduce boilerplate


kakserpom:
Current: `my_option.map_or(MaybeUnset::Unset, |x| MaybeUnset::Set(x))`
Proposed: `MaybeUnset::from_option(my_option)`

([slack thread](https://scylladb-users.slack.com/archives/C01D7LCL44D/p1716268091641299))

PR #1317, opened 2025-04-11
feat: add MaybeUnset::from_option() to reduce boilerplate

vardbabayan:
This patch adds from_option() method for MaybeUnset enum
It enables easy conversion from Option<T> into MaybeUnset<T>

Fixes: #1003 

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/serialize/value_tests.rs
+++ b/scylla-cql/src/serialize/value_tests.rs
@@ -2349,6 +2349,24 @@ fn unset_value() {
         do_serialize(set_i32, &ColumnType::Native(NativeType::Int)),
         vec![0, 0, 0, 4, 0, 0, 0, 32]
     );
+
+    let unset_option_i32: Option<i32> = None;
+    assert_eq!(
+        do_serialize(
+            MaybeUnset::from_option(unset_option_i32),
+            &ColumnType::Native(NativeType::Int)
+        ),
+        &(-2_i32).to_be_bytes()[..]
+    );
+
+    let set_option_i32: Option<i32> = Some(44);
+    assert_eq!(
+        do_serialize(
+            MaybeUnset::from_option(set_option_i32),
+            &ColumnType::Native(NativeType::Int)
+        ),
+        vec![0, 0, 0, 4, 0, 0, 0, 44]
+    );
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


vardbabayan:
Hi and Cześć :)
I could not access the slack thread linked in the issue description since slack requires pro plan to read old messages
So I hope the implementation matches the original idea of the ticket

scylla-cql/src/value.rs
```diff
@@ -33,6 +33,21 @@ pub enum MaybeUnset<V> {
...
+        }
+    }
+}
+
+impl<V> From<Option<V>> for MaybeUnset<V> {
+    fn from(opt: Option<V>) -> Self {
+        Self::from_option(opt)
+    }
+}
+
```
wprzytula, line 50: :pick: Let's add `#[inline]` attribute to both functions to allow inlining across crates.

scylla-cql/src/serialize/value_tests.rs
```diff
@@ -2349,6 +2349,24 @@ fn unset_value() {
...
+            MaybeUnset::from_option(unset_option_i32),
+            &ColumnType::Native(NativeType::Int)
+        ),
+        &(-2_i32).to_be_bytes()[..]
+    );
+
+    let set_option_i32: Option<i32> = Some(44);
+    assert_eq!(
+        do_serialize(
+            MaybeUnset::from_option(set_option_i32),
```
wprzytula, line 2365: :pick: Let's use `from`, to check the `From` impl, too.

wprzytula (changes requested):

scylla-cql/src/value.rs
Lorak-mmk, line 50: We actually had a similar PR before: https://github.com/scylladb/scylla-rust-driver/pull/1221
My opinion has not changed since then, I'm still against using "From / Into" for such conversions.

The "from_option" method is fine.

Lorak-mmk (changes requested):

scylla-cql/src/value.rs
vardbabayan: Done

scylla-cql/src/serialize/value_tests.rs
vardbabayan: left same, since From trait conversion removed

scylla-cql/src/value.rs
vardbabayan: Agree that from_option is enough, done.

wprzytula (approved):

Lorak-mmk (approved):

Issue #949, opened 2024-03-06
Sprinkle `#[inline]` across `scylla-cql` crate to allow inlining into `scylla` crate


wprzytula:
## Motivation
Having read [the guidelines on inlining](https://matklad.github.io/2021/07/09/inline-in-rust.html) (_highly recommended for everyone!_), I've understood that adding `#[inline]` explicitly is required to allow inlining across crates.

## Problem
`scylla-cql` crate's functions are largely short ones and, at the same time, are heavily called by `scylla` crate's routines.

## Suggestion
Adding `#[inline]` to `scylla-cql` short functions (as already done in e.g. [recently added `RowSerializationContext`](https://github.com/scylladb/scylla-rust-driver/blob/main/scylla-cql/src/types/serialize/row.rs?plain=1#L28)) can bring considerable performance benefits on the *ser/de* code path.

Additionally, we can think about small functions in `scylla` that could make the driver's users benefit from their inlining.

### Note
Do not forget that `#[inline]` [requires transitivity](https://github.com/matklad/benchmarks/tree/91171269f0a6e260a27111d07661021a89d20085/rust-inline). In the following code:
```rust
fn foo(x: i32) -> i32 {
  x * 2
}

#[inline]
pub fn bar(x: i32) -> i32 {
  foo(x) + 1
}
```
if `bar` is called from another crate, it won't be inlined (however, IIUC, it's probable that the call to `foo()` will be inlined into `bar` body).
In order to enable inlining `bar()`, **it is mandatory that both `foo` and `bar` functions are marked with `#[inline]`**.

mykaul:
Do we compile / recommend to compile with LTO?

Lorak-mmk:
Worth noting: @piodul found out that rustc already inlines cross-crate small functions without asserts / other calls: https://doc.rust-lang.org/stable/releases.html#version-1750-2023-12-28

> Do we compile / recommend to compile with LTO?

We don't compile - we have no way to, profile settings for dependencies are ignored.
We don't recommend. Should we? I'm not sure. I think it highly depends on a specific application and it's not specific to the driver at all, so it's not our place to make such recommendations. Also, Rust already gets a lot of bad rep for compile times, so I get why people might be reluctant to enable LTO.

wprzytula:
> Worth noting: @piodul found out that rustc already cross-crate inlines small functions without asserts / other calls: https://doc.rust-lang.org/stable/releases.html#version-1750-2023-12-28

[...]

PR #1318, opened 2025-04-12
Sprinkle `#[inline]` across the library

wprzytula:
## What's done

I went through all the code in our library (well, that's not true: I forgot about the macros; will cover them in new commits soon) and added `#[inline]` attribute to `pub` functions that are short enough to make sense to be inlined.

## Results

These are the changes in the binary size (left = after, right = before):
```
❯ diff -y <(ls target/release/deps/integration-67e7ceb0bf5d0a39 --size) <(ls target/release-non-inline/deps/integration-67e7ceb0bf5d0a39 --size)
25072 target/release/deps/integration-67e7ceb0bf5d0a39	      |	25032 target/release-non-inline/deps/integration-67e7ceb0bf5d
```

```
❯ diff -y <(ls target/release/ --size) <(ls target/release-non-inline/ --size)
total 22016						      |	total 21916
   0 build							   0 build
   0 deps							   0 deps
   0 examples							   0 examples
   0 incremental						   0 incremental
   4 libscylla_cql.d						   4 libscylla_cql.d
7820 libscylla_cql.rlib					      |	7752 libscylla_cql.rlib
   8 libscylla.d						   8 libscylla.d
   4 libscylla_macros.d						   4 libscylla_macros.d
4120 libscylla_macros.so					4120 libscylla_macros.so
   4 libscylla_proxy.d						   4 libscylla_proxy.d
1256 libscylla_proxy.rlib					1256 libscylla_proxy.rlib
8800 libscylla.rlib					      |	8768 libscylla.rlib
```

### Conclusion

As you can see, there is a slight increase in the binary size of both libraries, as well as the integration test suite.
This is something that I'd expect if my changes do affect the compiler's inlining decisions.

Fixes: #949

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/deserialize/mod.rs
+++ b/scylla-cql/src/deserialize/mod.rs
@@ -252,6 +252,7 @@ impl TypeCheckError {
     }
 
     /// Retrieve an error reason by downcasting to specific type.
+    #[inline]
     pub fn downcast_ref<T: std::error::Error + 'static>(&self) -> Option<&T> {
         self.0.downcast_ref()
     }
@@ -284,6 +285,7 @@ impl DeserializationError {
     }
 
     /// Retrieve an error reason by downcasting to specific type.
+    #[inline]
     pub fn downcast_ref<T: Error + 'static>(&self) -> Option<&T> {
         self.0.downcast_ref()
     }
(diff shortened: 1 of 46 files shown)


Lorak-mmk:
Please read:
https://github.com/rust-lang/hashbrown/pull/119
https://nnethercote.github.io/perf-book/inlining.html
https://std-dev-guide.rust-lang.org/policy/inline.html

You did not provide any benchmarks suggesting that this improves performance. You did not provide comparisions of release compile times.
We, as far as I know, have absolutely no reason to believe that Rust Driver has performance issues because of inlining heuristics in LLVM (which are afaik quite good).
I see no point of adding some voodoo "performance improvements" without having a tangible reason to, and evaluating the downsides.

Issue #1290, opened 2025-03-20
Don't prepare the statement on all connections


muzarski:
Similarly to https://github.com/scylladb/scylla-rust-driver/issues/1289, driver unnecessarily tries to prepare the statement on all connections.

It looks like scylla prepares the statement on every shard when processing `PREPARE` request: https://github.com/scylladb/scylladb/blob/8f0d0daf53397aa68312571ab9d01d8b75cd1770/transport/server.cc#L1104-L1114.

We should prepare the statement on each node once.

Lorak-mmk:
See my comment in #1289 : https://github.com/scylladb/scylla-rust-driver/issues/1289#issuecomment-2740151748 , it applies here too - at least partially.

muzarski:
> See my comment in [#1289](https://github.com/scylladb/scylla-rust-driver/issues/1289) : [#1289 (comment)](https://github.com/scylladb/scylla-rust-driver/issues/1289#issuecomment-2740151748) , it applies here too - at least partially.

AFAIU, the preparation error is not as critical as schema version fetch error. It's because driver will silently "reprepare" the statement on the failed node later, correct?

Lorak-mmk:
It's definitely not as critical. The worst that could happen is unnecessarily returning an error from `prepare`: Let's say we have a 1-node cluster, and we have a broken connection, and we don't retry on another - we would retry error to the user despite possibly being able to prepare.

wprzytula:
> > It's definitely not as critical. The worst that could happen is unnecessarily returning an error from `prepare`: Let's say we have a 1-node cluster, and we have a broken connection, and we don't retry on another - we would retry error to the user despite possibly being able to prepare.
> 
> Currently we ignore errors, unless all of the requests failed - only then we return the error to the user (`PrepareError::AllAttemptsFailed`). I think we can follow the same approach after the fix - return error to the user only if `PREPARE` request failed for all nodes.

Just to make sure:
What follows such a situation when only one node responds correctly to `PREPARE` request is: the default load balancing policy first tries to execute the prepared statement on a replica. Then, depending on its availability:
1. If it got available, `DbError::Unprepared` is returned, so silent repreparation is done and the statement is executed on that replica;
2. If replicas are unavailable, the single node that had prepared the statement correctly before serves the request (hopefully it manages to contact the required number of replicas, otherwise a serverside error is returned - probably `DbError::{Read,Write}{Error,Timeout}`).

Can you confirm?

PR #1320, opened 2025-04-12
session: Prepare on one shard per node only

wprzytula:
## Overview

This PR comes with two main enhancements / optimisations:
1. **Prepare on every node, not on every shard.** This corresponds to #1290.
2. ~~**Wait only for one prepare attempt to succeed, not for all prepare attempts.** This is inspired by cpp-driver's logic.~~ (dropped)

Below, I describe both changes in detail.

## Prepare on every node, not on every shard

Just as other drivers do, now the Rust driver also prepares statements on a single connection to every node, not on a single connection to every shard. This brings performance benefits for both driver and cluster, as it lowers the overhead of handling duplicated prepare requests on both sides.

### Problem: unlucky random choice
As mentioned in https://github.com/scylladb/scylla-rust-driver/issues/1290#issuecomment-2740175939, we might be unlucky enough to end up with all randomly chosen connections being non-working. For example, the targeted shards might be overloaded.

### Solution
Fallback logic is introduced. Once we fail to prepare the statement on any node though the randomly chosen connections, the preparation is re-attempted **on all connections** (which is essentially how it used to work before this PR).

## ~~Wait only for one prepare attempt to succeed~~ (dropped)
~~Having taken a look at cpp-driver's handling of statement preparation, I noticed that the cpp-driver waits only for the first preparation to succeed. Preparation on all remaining nodes is done in the background, which decreases latency of the prepare operation from the driver user's PoV. To understand why, consider a node that is stuck/overloaded/temporarily unavailable. If we wait for prepare responses from all nodes, we are limited by the slowest response. This is not what ScyllaDB is designed for - it should be available and fast even if some nodes happen to be unavailable or overloaded.~~

~~I decided to implement this behaviour in the Rust driver. This is achieved by spawning a tokio worker task which prepares the statement on all nodes. It feeds all results, successes or failures, into a channel and this way signals the parent task. The parent task finishes early once it receives a successful response (a deserialized `PreparedStatement`). Meanwhile, the worker task keeps handling the remaining preparations in the background.~~

~~The change brings two main benefits:~~
    ~~1. reduces driver's latency upon statement preparation;~~
    ~~2. prevents the situation when one stuck node freezes the driver upon
       preparation - **there are no preparation timeouts!** (sic!).~~

## Tests

TODO. Waiting for #1246 to be merged. Then I'll add a new feature to the proxy to be aware of which shard was targeted by the intercepted frame. With that feature available, I'll write tests.

Fixes: #1290
## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1246,7 +1246,7 @@ impl Session {
         statement: &Statement,
     ) -> Result<PreparedStatement, PrepareError> {
         let cluster_state = self.get_cluster_state();
-        let connections_iter = cluster_state.iter_working_connections()?;
+        let connections_iter = cluster_state.iter_working_connections_to_shards()?;
 
         // Prepare statements on all connections concurrently
         let handles = connections_iter.map(|c| async move { c.prepare_raw(statement).await });
@@ -2108,7 +2108,7 @@ impl Session {
 
     pub async fn check_schema_agreement(&self) -> Result<Option<Uuid>, SchemaAgreementError> {
         let cluster_state = self.get_cluster_state();
-        let connections_iter = cluster_state.iter_working_connections()?;
+        let connections_iter = cluster_state.iter_working_connections_to_shards()?;
 
         let handles = connections_iter.map(|c| async move { c.fetch_schema_version().await });
         let versions = try_join_all(handles).await?;
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1181,7 +1181,7 @@ impl Session {
             // Making QueryPager::new_for_query work with values is too hard (if even possible)
             // so instead of sending one prepare to a specific connection on each iterator query,
             // we fully prepare a statement beforehand.
-            let prepared = self.prepare(statement).await?;
+            let prepared = self.prepare_nongeneric(statement).await?;
             let values = prepared.serialize_values(&values)?;
             QueryPager::new_for_prepared_statement(PreparedPagerConfig {
                 prepared,
@@ -1237,6 +1237,14 @@ impl Session {
         statement: impl Into<Statement>,
     ) -> Result<PreparedStatement, PrepareError> {
         let statement = statement.into();
+        self.prepare_nongeneric(statement).await
+    }
+
+    // Introduced to avoid monomorphisation of this large function.
+    async fn prepare_nongeneric(
+        &self,
+        statement: Statement,
+    ) -> Result<PreparedStatement, PrepareError> {
         let statement_ref = &statement;
 
         let cluster_state = self.get_cluster_state();
@@ -1583,7 +1591,7 @@ impl Session {
                 .iter_mut()
                 .map(|statement| async move {
                     if let BatchStatement::Query(query) = statement {
-                        let prepared = self.prepare(query.clone()).await?;
+                        let prepared = self.prepare_nongeneric(query.clone()).await?;
                         *statement = BatchStatement::PreparedStatement(prepared);
                     }
                     Ok::<(), PrepareError>(())
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1274,7 +1274,7 @@ impl Session {
         }
 
         prepared.set_partitioner_name(
-            self.extract_partitioner_name(&prepared, &self.cluster.get_state())
+            Self::extract_partitioner_name(&prepared, &self.cluster.get_state())
                 .and_then(PartitionerName::from_str)
                 .unwrap_or_default(),
         );
@@ -1283,7 +1283,6 @@ impl Session {
     }
 
     fn extract_partitioner_name<'a>(
-        &self,
         prepared: &PreparedStatement,
         cluster_state: &'a ClusterState,
     ) -> Option<&'a str> {
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1246,7 +1246,7 @@ impl Session {
         statement: &Statement,
     ) -> Result<PreparedStatement, PrepareError> {
         let cluster_state = self.get_cluster_state();
-        let connections_iter = cluster_state.iter_working_connections_to_shards()?;
+        let connections_iter = cluster_state.iter_working_connections_to_nodes()?;
 
         // Prepare statements on all connections concurrently
         let handles = connections_iter.map(|c| async move { c.prepare_raw(statement).await });
@@ -1255,7 +1255,7 @@ impl Session {
         // If at least one prepare was successful, `prepare()` returns Ok.
         // Find the first result that is Ok, or Err if all failed.
 
-        // Safety: there is at least one node in the cluster, and `Cluster::iter_working_connections()`
+        // Safety: there is at least one node in the cluster, and `Cluster::iter_working_connections_to_nodes()`
         // returns either an error or an iterator with at least one connection, so there will be at least one result.
         let first_ok: Result<RawPreparedStatement, RequestAttemptError> =
             results.by_ref().find_or_first(Result::is_ok).unwrap();
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1285,6 +1285,12 @@ impl Session {
         // Validate prepared ids equality.
         for another_raw_prepared in raw_prepared_statements_results_iter.flatten() {
             if prepared.get_id() != &another_raw_prepared.prepared_response.id {
+                tracing::error!(
+                    "Got differing ids upon statement preparation: statement \"{}\", id1: {:?}, id2: {:?}",
+                    prepared.get_statement(),
+                    prepared.get_id(),
+                    another_raw_prepared.prepared_response.id
+                );
                 return Err(PrepareError::PreparedStatementIdsMismatch);
             }
 
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1199,6 +1199,9 @@ impl Session {
     /// Prepares a statement on the server side and returns a prepared statement,
     /// which can later be used to perform more efficient requests.
     ///
+    /// The statement is prepared on all nodes. This function finishes once all nodes respond
+    /// with either success or an error.
+    ///
     /// Prepared statements are much faster than unprepared statements:
     /// * Database doesn't need to parse the statement string upon each execution (only once)
     /// * They are properly load balanced using token aware routing
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1181,7 +1181,7 @@ impl Session {
             // Making QueryPager::new_for_query work with values is too hard (if even possible)
             // so instead of sending one prepare to a specific connection on each iterator query,
             // we fully prepare a statement beforehand.
-            let prepared = self.prepare_nongeneric(statement).await?;
+            let prepared = self.prepare_nongeneric(&statement).await?;
             let values = prepared.serialize_values(&values)?;
             QueryPager::new_for_prepared_statement(PreparedPagerConfig {
                 prepared,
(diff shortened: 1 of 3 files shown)


scylla/src/client/session.rs
```diff
@@ -1252,44 +1248,107 @@ impl Session {
...
-        // returns either an error or an iterator with at least one connection, so there will be at least one result.
-        let first_ok: Result<PreparedStatement, RequestAttemptError> =
-            results.by_ref().find_or_first(Result::is_ok).unwrap();
-        let mut prepared: PreparedStatement =
-            first_ok.map_err(|first_attempt| PrepareError::AllAttemptsFailed { first_attempt })?;
-
-        // Validate prepared ids equality
-        for statement in results.flatten() {
-            if prepared.get_id() != statement.get_id() {
-                return Err(PrepareError::PreparedStatementIdsMismatch);
```
muzarski, line 1261: commit: `session: prepare() waits only for first success `

Notice that this optimization is a trade-off. We no longer check for statement id equality between the nodes - we cannot inform the user about it if it happens.

I'm not entirely sure if this is a good trade-off. While I like the optimization (because we are not blocked by one node that hangs), maybe a better idea would actually be to introduce (configurable) timeout and await all the preparations - then we could retain the statement verification logic.

IIUC, this would also allow us to ditch the complicated logic that spawns the separate tokio task (preparation_worker).

WDYT @Lorak-mmk ?

scylla/src/client/session.rs
```diff
@@ -1252,44 +1248,107 @@ impl Session {
...
+            let statement_ref = &statement;
+            let preparations = connections_iter.map(|c| async move {
+                let res = c.prepare(statement_ref).await;
+                let _ = tx_ref.send(res).await;
+            });
+            join_all(preparations).await;
+        }
+
+        tokio::task::spawn(preparation_worker(
+            cluster_state.clone(),
```
muzarski: commit: `session: prepare() waits only for first success `

nit: `Arc::clone()`

scylla/src/client/session.rs
```diff
@@ -1304,62 +1304,80 @@ impl Session {
...
-                            .and_then(PartitionerName::from_str)
-                            .unwrap_or_default(),
-                    );
-                    return Ok(prepared);
-                }
-                Err(attempt_error) => {
-                    if first_error.is_none() {
-                        first_error = Some(attempt_error);
+            tokio::task::spawn(preparation_worker(
+                cluster_state.clone(),
```
muzarski, line 1337: commit: `session: prepare: extract prepare_on_all()`

ditto: `Arc::clone()`

muzarski (changes requested):
Looks really good. I just wonder - maybe a better idea than spawning a separate task is to just introduce configurable timeout for each attempt (see my other comment).

scylla/src/cluster/state.rs
```diff
@@ -344,6 +344,28 @@ impl ClusterState {
...
+
+        // First we try to find the first working pool of connections.
+        // If none is found, return error.
+        let first_working_pool = peers_iter
+            .by_ref()
+            .map(|node| node.get_random_connection())
+            .find_or_first(Result::is_ok)
+            .expect("impossible: known_peers was asserted to be nonempty")?;
+
+        let remaining_pools_iter = peers_iter.flat_map(|node| node.get_random_connection());
```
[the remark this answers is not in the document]
wprzytula: > But... As I mentioned, it looks like the errors in `Node::get_random_connection()` are handled per-pool, and not per-connection. So this looks fine to me - just wanted to point it out.

Yep. `ConnectionPool::random_connection()` fails iff there's no working connection in the whole pool.

scylla/src/client/session.rs
wprzytula: The complex logic of the `preparation_worker` is not a problem to me. It's encapsulated, it's (I hope you agree) well documented, and it uses tokio abstractions to be more high-level.

What I'm not fully satisfied with is the overhead introduced by spawning a new tokio task upon each `prepare()`. However, I assume that `Session::prepare()` is called infrequently (perhaps only once per statement), so it should not hurt users.

Keep in mind that automatic/transparent preparation, which may occur more frequently than calls to `Session::prepare`, is not affected. By this I mean:
1. transparent preparation on a single connection upon `Session::query_*` with nonempty parameters,
2. automatic repreparation on a given connection upon `UNPREPARED` error received from the DB.

wprzytula: > Notice that this optimization is a trade-off. We no longer check for statement id equality between the nodes - we cannot inform the user about it if it happens.

Well, we can still do that. It's easy to add such verification logic to the `preparation_worker` so that it logs a warning / an error-level message when it detects inequality.

muzarski: > Should I add such checks?

I'll let you decide. If it's fairly easy to implement, and won't drastically increase the complexity of the `preparation_workter`, I'd say why not. I think error-level message is appropriate - the id mismatch indicates a serious bug on the server side. UUIC, the id is just a hash of the statement contents, so it definitely should be the same on different nodes.

wprzytula:
v.1.0.2: rebased on main.

wprzytula:
v1.1:
- replaced internal calls to `Session::prepare()` with `Session::prepare_nongeneric()`;
- added verification of prepared id equality in the worker task.

wprzytula:
v1.1: rebased on main. Incorporated changes from #1329.

scylla/src/client/session.rs
```diff
@@ -1253,41 +1251,91 @@ impl Session {
...
+                    let mut prepared = raw_prepared.into_prepared_statement();
+
+                    // This is the first preparation that succeeded.
+                    // Let's return the PreparedStatement.
+                    // Preparation on other nodes will continue in the background tokio task.
+                    prepared.set_partitioner_name(
+                        Session::extract_partitioner_name(&prepared, cluster_state)
+                            .and_then(PartitionerName::from_str)
+                            .unwrap_or_default(),
+                    );
```
muzarski, line 1308: nit: Outdated comment

scylla/src/client/session.rs
```diff
@@ -1253,41 +1251,91 @@ impl Session {
...
-        );
+        let cluster_state = self.get_cluster_state();
 
-        Ok(prepared)
+        // Start by attempting preparation on a single (random) connection to every node.
+        {
+            let on_all_nodes_result = prepare_on_all(statement, &cluster_state, false).await;
+            if let Ok(prepared) = on_all_nodes_result {
+                // We succeeded in preparing the statement on at least one node. We're done; at the same time,
+                // the background tokio task attempts preparation on remaining nodes.
```
muzarski, line 1324: ditto: outdated comment

scylla/src/client/session.rs
```diff
@@ -1259,15 +1259,9 @@ impl Session {
         async fn prepare_on_all(
             statement: &Statement,
             cluster_state: &ClusterState,
-            on_all_shards: bool,
+            working_connections: &mut (dyn Iterator<Item = Arc<Connection>> + Send),
```
muzarski, line 1262: I agree that `dyn Iterator` is cleaner.

muzarski (approved):

wprzytula:
v2.0.1: fixed outdated comments.

scylla/src/cluster/state.rs
```diff
@@ -344,6 +344,28 @@ impl ClusterState {
...
+
+        Ok(std::iter::once(first_working_pool).chain(remaining_pools_iter))
+        // The returned iterator is nonempty, because it returns at least `first_working_pool`.
```
Lorak-mmk, line 366: This iterator usage is too complicated for me to read.
I see that we return Result, with the Ok variant being an iterator of connections (no results of connections).
I don't see where errors from `node.get_random_connection` in `remaining_pools_iter` are being handled.

Please add either explicit types, or comments, so that this function is easier to understand.

scylla/src/client/session.rs
```diff
@@ -1199,6 +1199,14 @@ impl Session {
...
     /// which can later be used to perform more efficient requests.
     ///
+    /// The statement is prepared on all nodes. This function finishes once all nodes respond
+    /// with either success or an error.
+    // TODO: There are no timeouts here. So, just one stuck node freezes the driver here, potentially indefinitely long.
+    // Also, what the driver requires to get from the cluster is the prepared statement metadata.
+    // It suffices that it gets only one copy of it, just from one success response. Therefore, it's a possible
+    // optimisation that the function only waits for one preparation to finish successfully, and then it returns.
+    // For it to be done, other preparations must continue in the background, on a separate tokio task.
+    ///
```
Lorak-mmk, line 1207: The described idea is imo not a TODO, because it is not obvious it is something we should do. It should be an issue instead.

scylla/src/client/session.rs
```diff
@@ -1253,41 +1251,91 @@ impl Session {
...
+                    // Let's return the PreparedStatement.
+                    prepared.set_partitioner_name(
+                        Session::extract_partitioner_name(&prepared, cluster_state)
+                            .and_then(PartitionerName::from_str)
+                            .unwrap_or_default(),
+                    );
+                    Ok(prepared)
+                }
+                Err(err) => Err(PrepareError::AllAttemptsFailed { first_attempt: err }),
+            }
```
Lorak-mmk: I don't see why this is a nested function. Please make it just a private top-level function, like prepare_nongeneric.

scylla/src/client/session.rs
```diff
@@ -1253,41 +1251,91 @@ impl Session {
...
+            statement: &Statement,
+            cluster_state: &ClusterState,
+            on_all_shards: bool,
+        ) -> Result<PreparedStatement, PrepareError> {
+            // Find the first result that is Ok, or Err if all failed.
+            let working_connections = if on_all_shards {
+                itertools::Either::Left(cluster_state.iter_working_connections_to_shards()?)
+            } else {
+                itertools::Either::Right(cluster_state.iter_working_connections_to_nodes()?)
+            };
```
Lorak-mmk: Instead of having weird boolean parameter you could accept dyn Iterator

scylla/src/client/session.rs
```diff
@@ -1253,41 +1251,91 @@ impl Session {
...
+            let raw_prepared_statements_results = join_all(preparations).await;
+
+            let mut raw_prepared_statements_results_iter =
+                raw_prepared_statements_results.into_iter();
+
+            // Safety: We pass a nonempty iterator, so there will be at least one result.
+            let first_ok_or_error = raw_prepared_statements_results_iter
+                .by_ref()
+                .find_or_last(|res| res.is_ok())
+                .unwrap(); // Safety: there is at least one connection.
```
Lorak-mmk, line 1276: Commit: "session: prepare: introduce fallback to all shards "

Previously `find_or_first` was used, now you changed it to `find_or_last`. Why?

scylla/src/client/session.rs
Lorak-mmk, line 1262: Commit: session: prepare_on_all() uses dyn Iterator

Ah, I didn't know about existance of this commit when writing my comment about dyn Iterator.

Lorak-mmk (changes requested):

scylla/src/client/session.rs
wprzytula: Why shouldn't it be a TODO? Issues have the problem that they are not immediately visible when someone is reading the code. OTOH, such a TODO catches attention quickly and makes the reader aware of the described problem and considered solutions.

wprzytula:
v2.1:
- squashed the `&mut dyn Iterator` commit into the fallback commit,
- extracted more commits to make review smoother,
- improved readability of the `iter_working_connections*()` methods,
- brought back the, accidentally removed, logic that collects prepared tracing IDs from prepared statements on all connections.

muzarski (approved):

scylla/src/cluster/state.rs
```diff
@@ -324,22 +324,35 @@ impl ClusterState {
...
+        // 2. or found the first working pool of connections.
+        let first_working_pool: Vec<Arc<Connection>> = first_working_pool_or_error?;
+
+        // We retrieve connection pools for remaining nodes (those that are left in the iterator
+        // once the first working pool has been found).
+        let remaining_pools_iter = connection_pool_per_node_iter;
+        // Pools are flattened, so now we have `impl Iterator<Item = Result<Arc<Connection>, ConnectionPoolError>>`.
+        let remaining_connections_iter = remaining_pools_iter.flatten_ok();
+        // Errors (non-working pools) are filtered out.
+        let remaining_working_connections_iter = remaining_connections_iter.flatten();
```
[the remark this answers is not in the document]
wprzytula: > There I was confused again, and I had to check in documentation that Result implements IntoIterator, returning iterator that will have 1 element for Ok, and 0 for Err.

TBH, I thought this is a widely known and used property, similat to how `Option` implements `IntoIterator`.



> I understand that this is correct, but I also think this is a bit confusing. Maybe `.filter_map(Result::ok)` would be more explicit and obvious?

No problem, I can write it this way.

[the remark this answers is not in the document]
Lorak-mmk: > TBH, I thought this is a widely known and used property, similarly to how Option implements IntoIterator.

I knew about Option. I did not suspect Result does that too, because for Result such impl ignores errors in imo too silent way,

wprzytula:
v2.1.1: adressed @Lorak-mmk comments.

wprzytula:
Rebased on main.

Lorak-mmk (approved):

muzarski (approved):

PR #1325, opened 2025-04-18
Fix shard out of range panic

muzarski:
In this PR implements verification logic for the picked shard. The shard should not be out of range for the chosen node. I also added a regression test where we would previously panic in such scenario.

## Pre-review checklist


Lorak-mmk:
The way I understand it the root cause of the problem is the bracket indexing here: https://github.com/scylladb/scylla-rust-driver/blob/v1.1.0/scylla/src/network/connection_pool.rs#L302
This PR does not change this code, and instead introduces more code to verify the passed shard number. Could you explain why you chose this approach?

The approach that feels simpler to me is to change the linked bracket indexing to get, and print a warning if `None` is returned.

muzarski:
> The way I understand it the root cause of the problem is the bracket indexing here: https://github.com/scylladb/scylla-rust-driver/blob/v1.1.0/scylla/src/network/connection_pool.rs#L302 This PR does not change this code, and instead introduces more code to verify the passed shard number. Could you explain why you chose this approach?
> 
> The approach that feels simpler to me is to change the linked bracket indexing to get, and print a warning if `None` is returned.

Yeah, I wanted to mention that in the cover letter - I forgot about it. Sorry about that.

Firstly, we need to answer ourselves the question: if LBP picks shard that is out of range, should we just print a warning and try on some random shard? If so, the code becomes much simpler.

If not, I assume that we should move to the next target from the plan. So if we wanted to handle this check in `NodeConnectionPool`, methods that pick the connection should somehow "inform" the caller about the problem.
I see two possibilities:
1. Change return type from `Result<_, ConnectionPoolError>` to `Option<Result<_, ConnectionPoolError>>` (`None` indicating shard out of range).
2. Add the `ShardOutOfRange` variant to `ConnectionPoolError`.

I don't like `2.`, because picking shard out of range by LBP is not a connection pool error. We could do `1.`. If you don't like `1.`, then the only thing we can do is to check the `shard` before calling `Node::connection_for_shard` - this is currently done in this PR.

I listed possible alternative solutions, and I'm open to suggestions. The only thing I'm against is adding `ShardOutOfRange` variant to `ConnectionPoolError`.

Lorak-mmk:
The way I understand it the shard chosen by LBP is not a strict requirement, but a preference. In the linked connection pool code we see that if there are no connections to selected shard, then some other shard is chosen instead.
Shard out of range is another type of situation where we can't get a connection to preferred shard, so it should also not be a critical error warranting skipping entire plan element. Printing a warning and choosing another shard is imo perfectly reasonable thing to do here.

muzarski:
> The way I understand it the shard chosen by LBP is not a strict requirement, but a preference. In the linked connection pool code we see that if there are no connections to selected shard, then some other shard is chosen instead. Shard out of range is another type of situation where we can't get a connection to preferred shard, so it should also not be a critical error warranting skipping entire plan element. Printing a warning and choosing another shard is imo perfectly reasonable thing to do here.

Makes sense. I'll update this (this will significantly reduce the size of this PR).

muzarski:
v2: After the discussion with @Lorak-mmk , I adjusted the logic accordingly - now, when shard is out of range, we fallback to picking the connection to some other random shard.

Lorak-mmk (dismissed):
I'm okay with merging as-is, but I'd prefer the warning to be added.
As you mentioned, there is currently a log. The problem is that it uses `trace` level, so it won't get noticed.

While both situations (out of range shard, no connections to shard) allow us to continue with the request attempt on this target, the first one suggests a bug in LBP, so warning about it with higher level seems like a good idea.

muzarski:
> I'm okay with merging as-is, but I'd prefer the warning to be added. As you mentioned, there is currently a log. The problem is that it uses `trace` level, so it won't get noticed.
> 
> While both situations (out of range shard, no connections to shard) allow us to continue with the request attempt on this target, the first one suggests a bug in LBP, so warning about it with higher level seems like a good idea.

Done

Lorak-mmk (approved):

scylla/src/network/connection_pool.rs
```diff
@@ -299,7 +299,17 @@ impl NodeConnectionPool {
...
         // Try getting the desired connection
-        if let Some(conn) = Self::choose_random_connection_from_slice(&shard_conns[shard as usize])
+        if let Some(conn) = shard_conns
+            .get(shard as usize)
+            .or_else(|| {
+                warn!(
+                    shard = shard,
+                    "Requested shard is out of bounds.\
+                    This is most probably a bug in custom LoadBalancingPolicy implementation!"
+                );
```
wprzytula, line 309: :wrench:

wprzytula:
```suggestion
                warn!(
                    shard = shard,
                    "Requested shard is out of bounds.\
                    This is most probably a bug in custom LoadBalancingPolicy implementation!\
                    Targetting a random/arbitrary shard."
                );
```

wprzytula (changes requested):

wprzytula (dismissed):
LGTM.

wprzytula:
I took a look into `LoadBalancingPolicy` docstring and there's nothing about the shard being considered just a preference / hint (not a strict requirement). What I see there is:

```
That plan is a list of
targets (target is a node + an optional shard) to which
the driver will try to send the request. The first elements of the plan are the targets which are
the best to contact (e.g. they might have the lowest latency).
```

I believe we must clarify our understanding of the shard selection in the LBP docstring. Namely, let's inform that we ignore the shard number (and choose another randomly) if it's out of bounds for the given node.

muzarski:
v2.1: Addressed @wprzytula comments. Mentioned that picked shard serves as a hint in the LBP docstring.

scylla/src/network/connection_pool.rs
```diff
@@ -299,7 +299,18 @@ impl NodeConnectionPool {
...
         // Try getting the desired connection
-        if let Some(conn) = Self::choose_random_connection_from_slice(&shard_conns[shard as usize])
+        if let Some(conn) = shard_conns
+            .get(shard as usize)
+            .or_else(|| {
+                warn!(
+                    shard = shard,
+                    "Requested shard is out of bounds.\
+                    This is most probably a bug in custom LoadBalancingPolicy implementation!\
+                    Targetting a random/arbitrary shard."
```
wprzytula, line 309: Sorry, I made a typo myself.

wprzytula:
```suggestion
                    Targeting a random/arbitrary shard."
```

wprzytula (dismissed):

scylla/src/network/connection_pool.rs
muzarski: > Sorry, I made a typo myself.

And I didn't notice:)

wprzytula (approved):

Lorak-mmk (approved):

Issue #1031, opened 2024-07-09
Enable enforcing coordinator for request


wprzytula:
## Motivation

cpp-driver provides an option to explicitly set a node to be coordinator for a specific request. It argues for that option this way:
> In general, this should not be used, but it can be useful in the following situations:
>  - To query node-local tables such as system and virtual tables.
>  - To apply a sequence of schema changes where it makes sense for all the changes to be applied on a single node.

## Solution

Let's expose a setter for explicit coordinator - consider identifying by address or host_id.
If an explicit coordinator is set for a statement, then:
- load balancing policy shall be ignored,
- speculative execution shall be disabled,
- retries shall be only performed on the same node (consider if `RetryNextNode` decisions should be then changed into `RetrySameNode` or into `DontRetry`).

wprzytula:
This should be in line with #738 requirements about enforcing a particular coordinator.

Lorak-mmk:
This can already be done by setting a custom LBP on the query, which yields only 1 element.

wprzytula:
> This can already be done by setting a custom LBP on the query, which yields only 1 element.

Yes, it _can_ be done, but it's fairly inconvenient. The driver's duty is to provide convenient and shorthands to operate with Scylla, and by cpp-driver's documentation (see Motivation) the shorthand for setting an explicit coordinator for a statement is useful.

roydahan:
@muzarski is this one not required for cpp-rust-driver 1.0?

muzarski:
> [@muzarski](https://github.com/muzarski) is this one not required for cpp-rust-driver 1.0?

I just moved it out of the 1.1. I'm not sure if it's going to be prioritized in 1.2 or pushed back to 1.3. Thus, I moved it to 1.x.

roydahan:
If we need to complete cpp-rust-driver work by Q3, it must be either in 1.1 or in 1.2.
If there are other items that are more urgent, we need to understand why.

CC @dkropachev

Lorak-mmk:
> If we need to complete cpp-rust-driver work by Q3, it must be either in 1.1 or in 1.2. If there are other items that are more urgent, we need to understand why.

Why is this implication true? We want to release once per month, so there will be more versions before end of Q3.

PR #1326, opened 2025-04-18
Enforce request coordinator

muzarski:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1031

## SingleTargetLoadBalancingPolicy

This load balancing policy returns a single target in `pick`.  The fallback
always returns empty iterator.

The node is picked based on the `NodeIdentifier` that user can provide
when constructing the policy.
- NodeIdentifier::Node -> just `Arc<Node>` - it's returned unconditionally from pick()
- NodeIdentifier::Address -> lbp does the lookup in the cluster metadata
   and returns a node with the given address  (if found)
- NodeIdentifier::HostId -> same as for the address, but the lookup is
   done by the host id

## StatementConfig

In addition, we expose a utility method on all statements, that allow to overwrite current LBP, and enforce the specific target for the request.

## Pre-review checklist

- ~~[ ] I have adjusted the documentation in `./docs/source/`.~

### DIFF
--- a/docs/source/load-balancing/load-balancing.md
+++ b/docs/source/load-balancing/load-balancing.md
@@ -36,8 +36,12 @@ awareness enabled and latency-awareness disabled.
 
 ## Configuration
 
-Load balancing policies can be configured via execution profiles. In the code
-sample provided, a new execution profile is created using
+Load balancing policies can be configured in three different ways (sorted by descending precedence):
+1. directly on the `Statement`, `PreparedStatement` or `Batch` (`Statement::set_load_balancing_policy()`)
+2. execution profile set on the statement (`Statement::set_execution_profile_handle()`)
+3. default execution profile set on the session (`SessionBuilder::default_execution_profile_handle()`)
+
+ In the code sample provided, a new execution profile is created using
 `ExecutionProfile::builder()`, and the load balancing policy is set to the
 `DefaultPolicy` using `.load_balancing_policy(policy)`.
 
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -141,7 +141,6 @@ struct PagerWorker<'a, QueryFunc, SpanCreatorFunc> {
     query_is_idempotent: bool,
     query_consistency: Consistency,
     retry_session: Box<dyn RetrySession>,
-    execution_profile: Arc<ExecutionProfileInner>,
     #[cfg(feature = "metrics")]
     metrics: Arc<Metrics>,
 
@@ -744,7 +743,6 @@ impl QueryPager {
                 query_consistency: consistency,
                 load_balancing_policy,
                 retry_session,
-                execution_profile,
                 #[cfg(feature = "metrics")]
                 metrics,
                 paging_state: PagingState::start(),
@@ -868,7 +866,6 @@ impl QueryPager {
                 query_consistency: consistency,
                 load_balancing_policy,
                 retry_session,
-                execution_profile: config.execution_profile,
                 #[cfg(feature = "metrics")]
                 metrics: config.metrics,
                 paging_state: PagingState::start(),
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -34,7 +34,7 @@ use crate::observability::driver_tracing::RequestSpan;
 use crate::observability::history::{self, HistoryListener};
 #[cfg(feature = "metrics")]
 use crate::observability::metrics::Metrics;
-use crate::policies::load_balancing::{self, RoutingInfo};
+use crate::policies::load_balancing::{self, LoadBalancingPolicy, RoutingInfo};
 use crate::policies::retry::{RequestInfo, RetryDecision, RetrySession};
 use crate::response::query_result::ColumnSpecs;
 use crate::response::{NonErrorQueryResponse, QueryResponse};
@@ -136,6 +136,7 @@ struct PagerWorker<'a, QueryFunc, SpanCreatorFunc> {
     // AsyncFn(Arc<Connection>, Option<Arc<[u8]>>) -> Result<QueryResponse, RequestAttemptError>
     page_query: QueryFunc,
 
+    load_balancing_policy: Arc<dyn LoadBalancingPolicy>,
     statement_info: RoutingInfo<'a>,
     query_is_idempotent: bool,
     query_consistency: Consistency,
@@ -162,7 +163,7 @@ where
 {
     // Contract: this function MUST send at least one item through self.sender
     async fn work(mut self, cluster_state: Arc<ClusterState>) -> PageSendAttemptedProof {
-        let load_balancer = self.execution_profile.load_balancing_policy.clone();
+        let load_balancer = Arc::clone(&self.load_balancing_policy);
         let statement_info = self.statement_info.clone();
         let query_plan =
             load_balancing::Plan::new(load_balancer.as_ref(), &statement_info, &cluster_state);
@@ -336,8 +337,7 @@ where
                 let _ = self.metrics.log_query_latency(elapsed.as_millis() as u64);
                 self.log_attempt_success();
                 self.log_request_success();
-                self.execution_profile
-                    .load_balancing_policy
+                self.load_balancing_policy
                     .on_request_success(&self.statement_info, elapsed, node);
 
                 request_span.record_raw_rows_fields(&rows);
@@ -370,9 +370,12 @@ where
             Err(err) => {
                 #[cfg(feature = "metrics")]
                 self.metrics.inc_failed_paged_queries();
-                self.execution_profile
-                    .load_balancing_policy
-                    .on_request_failure(&self.statement_info, elapsed, node, &err);
+                self.load_balancing_policy.on_request_failure(
+                    &self.statement_info,
+                    elapsed,
+                    node,
+                    &err,
+                );
                 Err(err)
             }
             Ok(NonErrorQueryResponse {
@@ -392,9 +395,12 @@ where
                 self.metrics.inc_failed_paged_queries();
                 let err =
                     RequestAttemptError::UnexpectedResponse(response.response.to_response_kind());
-                self.execution_profile
-                    .load_balancing_policy
-                    .on_request_failure(&self.statement_info, elapsed, node, &err);
+                self.load_balancing_policy.on_request_failure(
+                    &self.statement_info,
+                    elapsed,
+                    node,
+                    &err,
+                );
                 Err(err)
             }
         }
@@ -694,6 +700,8 @@ impl QueryPager {
             ..Default::default()
         };
 
+        let load_balancing_policy = Arc::clone(&execution_profile.load_balancing_policy);
+
         let retry_session = statement
             .get_retry_policy()
             .map(|rp| &**rp)
@@ -734,6 +742,7 @@ impl QueryPager {
                 statement_info: routing_info,
                 query_is_idempotent: statement.config.is_idempotent,
                 query_consistency: consistency,
+                load_balancing_policy,
                 retry_session,
                 execution_profile,
                 #[cfg(feature = "metrics")]
@@ -769,6 +778,8 @@ impl QueryPager {
             .unwrap_or(config.execution_profile.serial_consistency);
 
         let page_size = config.prepared.get_validated_page_size();
+
+        let load_balancing_policy = Arc::clone(&config.execution_profile.load_balancing_policy);
 
         let retry_session = config
             .prepared
@@ -855,6 +866,7 @@ impl QueryPager {
                 statement_info,
                 query_is_idempotent: config.prepared.config.is_idempotent,
                 query_consistency: consistency,
+                load_balancing_policy,
                 retry_session,
                 execution_profile: config.execution_profile,
                 #[cfg(feature = "metrics")]
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -699,7 +699,11 @@ impl QueryPager {
             ..Default::default()
         };
 
-        let load_balancing_policy = Arc::clone(&execution_profile.load_balancing_policy);
+        let load_balancing_policy = Arc::clone(
+            statement
+                .get_load_balancing_policy()
+                .unwrap_or(&execution_profile.load_balancing_policy),
+        );
 
         let retry_session = statement
             .get_retry_policy()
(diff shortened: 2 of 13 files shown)


scylla/src/statement/batch.rs
Lorak-mmk, line 1: Commit: "statement: expose configuration to enforce request coordinator "

I was quite surprised when reading this commit, because the PR description and previous commits suggested that a different direction will be taken.

Why add `statement: expose configuration to enforce request coordinator `, instead of having a more general Arc<dyn LBP>? Previous commits suggested that user will be able to set LBP for a statement.

In that case we would have methods `set_load_balancing_policy`, `unset_load_balancing_policy`, and the coordinator enforcing would use them underneath.

This allows us more flexibility in the future as corrdinator enforcing is no longer special cased in any way.
We could for example introduce another policy that prefers, but not enforces, selected coordinator.

muzarski: Yeah, what you suggested makes much more sense. I'll fix it.

muzarski:
v2: Addressed @Lorak-mmk comment

scylla/src/policies/load_balancing/enforce_target.rs
```diff
@@ -0,0 +1,85 @@
...
+/// If `NodeIdentifier::Node` is used, it will be returned immediately without
+/// the lookup in the cluster metadata.
+///
+/// If `NodeIdentifier::NodeId` or `NodeIdentifier::NodeAddress` is used,
+/// the node will be looked up in the cluster metadata based on its host_id/address.
+#[derive(Debug, Clone)] // <- Cheaply clonable
+#[non_exhaustive]
+pub enum NodeIdentifier {
+    Node(Arc<Node>),
+    NodeId(Uuid),
```
wprzytula, line 22: :question: Everywhere in the codebase (driver's and core's) there is a notion of _host ID_. I haven't seen _node ID_ anywhere, have you?

scylla/src/policies/load_balancing/enforce_target.rs
```diff
@@ -0,0 +1,85 @@
...
+    NodeAddress(SocketAddr),
```
wprzytula, line 23: :thought_balloon: Note that the way it's used implies use of **untranslated** addresses. This must be clearly stated.

scylla/src/policies/load_balancing/enforce_target.rs
```diff
@@ -0,0 +1,85 @@
...
+                .all_nodes
+                .iter()
+                .find(|node| SocketAddr::new(node.address.ip(), node.address.port()) == *addr),
+        };
+
+        match node {
+            Some(node) => Some((node, self.shard)),
+            None => {
+                tracing::warn!(
+                    "EnforceTargetLoadBalancingPolicy failed to find requested node {:?} in cluster metadata.",
```
wprzytula, line 66: :wrench: The policy's name which appears in the log message does not match the policy's name in the code.

scylla/src/policies/load_balancing/enforce_target.rs
```diff
@@ -0,0 +1,85 @@
...
+        &'a self,
+        _request: &'a RoutingInfo,
+        _cluster: &'a ClusterState,
+    ) -> super::FallbackPlan<'a> {
+        Box::new(std::iter::empty())
+    }
+
+    fn name(&self) -> String {
+        "EnforceTargetLoadBalancingPolicy".to_string()
+    }
```
wprzytula, line 84: ditto

scylla/src/client/pager.rs
```diff
@@ -868,7 +866,6 @@ impl QueryPager {
                 query_consistency: consistency,
                 load_balancing_policy,
                 retry_session,
-                execution_profile: config.execution_profile,
                 #[cfg(feature = "metrics")]
```
wprzytula, line 876: :wrench: Now that we no longer need ownership over `execution_profile`, we can accept `&ExecutionProfileInner` instead of `Arc<ExecutionProfileInner>`, saving an `Arc` clone.

scylla/src/statement/mod.rs
```diff
@@ -38,6 +39,7 @@ pub(crate) struct StatementConfig {
     pub(crate) history_listener: Option<Arc<dyn HistoryListener>>,
 
     pub(crate) execution_profile_handle: Option<ExecutionProfileHandle>,
+    pub(crate) load_balancing_policy: Option<Arc<dyn LoadBalancingPolicy>>,
     pub(crate) retry_policy: Option<Arc<dyn RetryPolicy>>,
```
wprzytula, line 43: :warning: Please make sure that `PreparedStatement` inherits this setting upon preparation in `Session`, and also upon execution in `CachingSession`.

scylla/src/client/pager.rs
```diff
@@ -699,7 +699,10 @@ impl QueryPager {
             ..Default::default()
         };
 
-        let load_balancing_policy = Arc::clone(&execution_profile.load_balancing_policy);
+        let load_balancing_policy = statement
+            .get_load_balancing_policy()
+            .map(Arc::clone)
+            .unwrap_or_else(|| Arc::clone(&execution_profile.load_balancing_policy));
```
wprzytula, line 705: :question: Does this work?

wprzytula:
```suggestion
            .unwrap_or_else(|| &execution_profile.load_balancing_policy)
            .map(Arc::clone);
```

scylla/src/client/pager.rs
```diff
@@ -777,7 +780,11 @@ impl QueryPager {
 
         let page_size = config.prepared.get_validated_page_size();
 
-        let load_balancing_policy = Arc::clone(&config.execution_profile.load_balancing_policy);
+        let load_balancing_policy = config
+            .prepared
+            .get_load_balancing_policy()
+            .map(Arc::clone)
+            .unwrap_or_else(|| Arc::clone(&config.execution_profile.load_balancing_policy));
```
wprzytula, line 787: ditto

scylla/tests/integration/statement.rs
```diff
@@ -55,3 +66,97 @@ async fn test_prepared_statement_col_specs() {
...
+        )
+        .await;
+
+        query_system_local_and_verify(
+            &session,
+            NodeIdentifier::NodeAddress(SocketAddr::new(node_address.ip(), node_address.port())),
+            node_id,
+            node_address.ip(),
+        )
+        .await;
```
wprzytula, line 144: :memo: Note that this works because we operate on **untranslated** addresses both in `NodeIdentifier::NodeAddress` and in `Node.address`.

docs/source/load-balancing/load-balancing.md
```diff
@@ -36,8 +36,12 @@ awareness enabled and latency-awareness disabled.
 
 ## Configuration
 
-Load balancing policies can be configured via execution profiles. In the code
-sample provided, a new execution profile is created using
+Load balancing policies can be configured in three different ways (sorted by precedence):
```
wprzytula, line 39: :wrench:

wprzytula (changes requested):

scylla/src/policies/load_balancing/enforce_target.rs
muzarski: Good catch, I renamed it after implementing this

muzarski: I'll also rename the module from `enforce_target.rs` to `single_target.rs`.

scylla/src/client/pager.rs
muzarski: The `ExecutionProfileHandle::access()` returns an `Arc` anyway. We do not clone anything later.

```rust
        let execution_profile = statement
            .get_execution_profile_handle()
            .unwrap_or_else(|| self.get_default_execution_profile_handle())
            .access();
```

scylla/src/statement/mod.rs
muzarski: 1. preparation in `Session`: we clone entire `StatementConfig`, so the answer is yes.
2. `CachingSession` - again, we inherit entire `StatementConfig`.

scylla/src/client/pager.rs
wprzytula: > This would work:
> 
> ```rust
>         let load_balancing_policy = Arc::clone(
>             statement
>                 .get_load_balancing_policy()
>                 .unwrap_or_else(|| &execution_profile.load_balancing_policy),
>         );
> ```
> 
> Which one do you prefer?

I like the version you proposed the most.

muzarski:
v2.1: Addressed @wprzytula comments

scylla/src/policies/load_balancing/single_target.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// If `NodeIdentifier::Node` is used, it will be returned immediately without
+/// the lookup in the cluster metadata.
+///
+/// If `NodeIdentifier::HostId` or `NodeIdentifier::NodeAddress` is used,
+/// the node will be looked up in the cluster metadata based on its host_id/address.
+///
+/// Note: `NodeIdentifier::NodeAddress` assumes that provided address is **untranslated**.
+/// In other words, it should be the address of form `<rpc_address>:<cql_port>`, where
+/// `<rpc_address>` is the address of the node stored in the system tables, as well as the [`Node`] struct.
```
wprzytula, line 21:
```suggestion
/// Note: `NodeIdentifier::NodeAddress` assumes that provided address is **untranslated**.
/// In other words, it should be the address of form `<rpc_address>:<cql_port>`, where
/// `<rpc_address>` is the address of the node stored in the system tables, as well as
/// in the [`Node`] struct.
```

wprzytula: (I added `in` and split the line)

wprzytula (changes requested):

wprzytula (dismissed):

scylla/src/client/pager.rs
```diff
@@ -699,7 +699,11 @@ impl QueryPager {
             ..Default::default()
         };
 
-        let load_balancing_policy = Arc::clone(&execution_profile.load_balancing_policy);
+        let load_balancing_policy = Arc::clone(
+            statement
+                .get_load_balancing_policy()
+                .unwrap_or_else(|| &execution_profile.load_balancing_policy),
+        );
```
Lorak-mmk, line 706: Commit: `session/pager: overwrite LBP if set on statement `

? I'm curious, why do you use `unwrap_or_else` for a simple field access? It was my understanding that this method is useful when computing the default value is expensive, and that `unwrap_or` is better for simple cases like this.
And you actually do use `unwrap_or` in session.

scylla/src/policies/load_balancing/single_target.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl SingleTargetLoadBalancingPolicy {
+    pub fn new(node_identifier: NodeIdentifier, shard: Option<Shard>) -> Self {
+        Self {
+            node_identifier,
+            shard,
+        }
+    }
+}
```
Lorak-mmk, line 47: Wouldn't it be easier for the user if `new` returned `Arc`? All the use cases of this struct that can exist require wrapping it in Arc anyway, right?

If there ever emerges use case for creating this struct without Arc then we can add `new_raw` method.

`DefaultPolicyBuilder::build` returns `Arc<dyn LoadBalancingPolicy>`, probably for the same reason.

Lorak-mmk (changes requested):
Could you write an example (in `examples/` folder) for setting the custom coordinator?

Given that we had some users ask for it it makes sense to make it more discoverable.

scylla/src/client/pager.rs
muzarski: Previously it was `unwrap_or_else(|| Arc::clone(...))`. But @wprzytula asked me to refactor this a bit (current state) and I forgot to change it to `unwrap_or`.

scylla/src/statement/batch.rs
```diff
@@ -134,6 +135,21 @@ impl Batch {
...
+    ) {
+        self.config.load_balancing_policy = load_balancing_policy;
+    }
+
+    /// Get the load balancing policy set for the batch.
+    #[inline]
+    pub fn get_load_balancing_policy(&self) -> Option<&Arc<dyn LoadBalancingPolicy>> {
+        self.config.load_balancing_policy.as_ref()
+    }
+
```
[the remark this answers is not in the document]
wprzytula: > This applies to all get_load_balancing_policy methods: Doc comment should imo state directly that this method returns only the LBP that was explicitly overriden on a statement - it does not take into consideration the execution policy set on a statement.

You mean: _it does not take into consideration the execution policy set on **an execution profile**._, right?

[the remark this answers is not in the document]
Lorak-mmk: I meant "it does not take into consideration LBP set on execution profile that is set on this statement".

scylla/src/policies/load_balancing/single_target.rs
```diff
@@ -0,0 +1,90 @@
...
+            NodeIdentifier::Node(node) => Some(node),
+            NodeIdentifier::HostId(node_id) => cluster
+                .all_nodes
+                .iter()
+                .find(|node| node.host_id == *node_id),
+            NodeIdentifier::NodeAddress(addr) => cluster
+                .all_nodes
+                .iter()
+                .find(|node| SocketAddr::new(node.address.ip(), node.address.port()) == *addr),
+        };
```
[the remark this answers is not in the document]
muzarski: Should I replace `all_nodes` with `known_peers` in node address case.as well?

muzarski:
v2.2: Addressed @Lorak-mmk comments.

An idea: maybe we should introduce `SingleTargetLoadBalancingPolicyBuilder` or sth like that. I believe such solution is more flexible as it would allow us to add `SingleTargetLoadBalancingPolicy` fields without introducing additional constructors. As @Lorak-mmk mentioned, we could for example request the policy not to enforce the target, but prefer it. In such case, we could expose `with_fallback_policy(fallback_policy: Option<Arc<dyn LBP>>)` method on a builder or something similar. Notice that this would not be so pretty with current version (`new` constructor) - we would need to introduce new constructor to handle the new field

WDYT? cc: @wprzytula

scylla/src/statement/prepared.rs
```diff
@@ -437,6 +437,10 @@ impl PreparedStatement {
     }
 
     /// Set the load balancing policy for this statement, overriding the one from execution profile if not None.
+    ///
+    /// This method returns the retry policy that is **overriden** on this statement.
+    /// In other words, it returns the retry policy set using [`PreparedStatement::set_retry_policy`].
+    /// This does not take the retry policy from the set execution profile into account.
     #[inline]
     pub fn set_load_balancing_policy(
         &mut self,
```
Lorak-mmk: Commit: "statement: enrich get_retry_policy documentation"

I think you meant to put this comment on a get_ method, not a set_ method, right?

Lorak-mmk (changes requested):

scylla/src/statement/prepared.rs
muzarski: Correct. I'll fix it.

Lorak-mmk (dismissed):

wprzytula (dismissed):

Issue #1313, opened 2025-04-03
Make pending request channel size configurable


muzarski:
This would directly correspond to [`cass_cluster_set_queue_size_io`](https://github.com/scylladb/cpp-rust-driver/blob/v0.3.0/include/cassandra.h#L1747-L1761).

We even have a `TODO` in our code regarding this: https://github.com/scylladb/scylla-rust-driver/blob/v1.1.0/scylla/src/network/connection.rs#L453-L454.

By default we currently set the size to 1024. OTOH, cpp-driver sets this value to 8192 by detault. Maybe it is something to consider.

dhvll:
Hey I would like to contribute to this issue.

Lorak-mmk:
Great! It should not be too difficult. If you need any help feel free to ask.

Lorak-mmk:
> OTOH, cpp-driver sets this value to 8192 by detault. Maybe it is something to consider.

@muzarski is it also per-connection queue there?

muzarski:
> > OTOH, cpp-driver sets this value to 8192 by detault. Maybe it is something to consider.
> 
> [@muzarski](https://github.com/muzarski) is it also per-connection queue there?

Not really. The architecture is a bit different in cpp-driver. The queue is per-`RequestProcessor`. There are as many `RequestProcessors` as specified by [`cass_cluster_set_threads_io`](https://github.com/scylladb/cpp-rust-driver/blob/v0.3.0/include/cassandra.h#L1731-L1745). In rust-driver we have as many tokio tasks as there are connections - (`Connection::writer` task per connection).

So the semantics will not be exactly the same. This will need to be mentioned in cpp-rust-driver API documentation (same for `cass_cluster_set_threads_io` - we plan to configure the number of threads used by the tokio runtime using this, and not the number of tokio tasks handling the requests).

PR #1328, opened 2025-04-19
feat: add pending request channel size configuration to ConnectionConfig

dhvll:
# Make pending request channel size configurable
## Changes
1. Added [pending_request_channel_size] configuration option to [ConnectionConfig]
2. Set a new default value of 2048 (up from 1024)
3. Added comprehensive documentation explaining the configuration
4. Updated connection creation to use the configurable channel size

## Rationale
- The previous fixed size of 1024 was too conservative for many workloads
- cpp-driver uses 8192, but we chose 2048 as a more balanced default due to our per-connection queue architecture
- This change provides flexibility while maintaining reasonable defaults

## Usage
```rust
let session = SessionBuilder::new()
    .connection_config(
        ConnectionConfig::new()
            .with_pending_request_channel_size(4096)  // Customize as needed
    )
    .build()
    .await?;
```

## Pre-review checklist

### DIFF
--- a/scylla/src/network/connection.rs
+++ b/scylla/src/network/connection.rs
@@ -287,6 +287,7 @@ pub(crate) struct ConnectionConfig {
     pub(crate) authenticator: Option<Arc<dyn AuthenticatorProvider>>,
     pub(crate) address_translator: Option<Arc<dyn AddressTranslator>>,
     pub(crate) write_coalescing_delay: Option<WriteCoalescingDelay>,
+    pub(crate) pending_request_channel_size: Option<usize>,
 
     pub(crate) keepalive_interval: Option<Duration>,
     pub(crate) keepalive_timeout: Option<Duration>,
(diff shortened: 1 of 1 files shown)


wprzytula:
@dhvll CI is failing.
@Lorak-mmk Do you see value in adding such a config parameter?

muzarski:
> @Lorak-mmk Do you see value in adding such a config parameter?

It's not that important feature, but cpp-rust-driver requires it. See the issue: https://github.com/scylladb/scylla-rust-driver/issues/1313

dhvll:
I'll look into it.

scylla/src/network/connection.rs
```diff
@@ -322,17 +323,44 @@ impl ConnectionConfig {
...
+    ///
+    /// ```
+    /// let session = SessionBuilder::new()
+    ///     .connection_config(
+    ///         ConnectionConfig::new()
+    ///             .with_pending_request_channel_size(4096)
+    ///     )
+    ///     .build()
+    ///     .await?;
+    /// ```
```
Lorak-mmk, line 357: I think it is worth describing what a "pending request" is. I think users may think that it is a request that was sent, but we did not yet receive response to (while in reality it is a request that has not yet been sent).

scylla/src/network/connection.rs
```diff
@@ -451,7 +480,7 @@ impl Connection {
         }
 
         // TODO: What should be the size of the channel?
-        let (sender, receiver) = mpsc::channel(1024);
+        let (sender, receiver) = mpsc::channel(config.pending_request_channel_size.unwrap_or(1024));
         let (error_sender, error_receiver) = tokio::sync::oneshot::channel();
```
Lorak-mmk, line 484: Please remove the TODO comment. 
Also you are doing `unwrap_or(1024)`, but you wrote in PR description that you changes the default to 2048 - it looks like a mistake.

scylla/src/network/connection.rs
```diff
@@ -382,6 +418,7 @@ impl Default for HostConnectionConfig {
             tablet_sender: None,
 
             identity: SelfIdentity::default(),
+            pending_request_channel_size: Some(2048),
         }
     }
 }
```
Lorak-mmk, line 424: The fact that in connection you wrote `.unwrap_or(1024)` (which is a bug) tells me that this approach is too error-prone.

Why use an option here? I don't see anything optional - we need to create the channel, and it needs to have some size. Let's make it just usize.

scylla/src/network/connection.rs
```diff
@@ -411,19 +448,11 @@ impl Default for ConnectionConfig {
             tablet_sender: None,
 
             identity: SelfIdentity::default(),
+            pending_request_channel_size: Some(2048),
         }
     }
```
Lorak-mmk, line 453: ditto

Lorak-mmk (changes requested):

PR #1329, opened 2025-04-19
Construct `PreparedStatement` only where needed

wprzytula:
## Motivation

Constructing `PreparedStatement` involves allocations:
1. `ResultMetadata` must be put onto heap, behind an `Arc`;
2. `StatementConfig` must be cloned along with all `Arc`s inside;
3. Statement string must be cloned.

I noticed that in a number of places we materialise `PreparedStatement` even though it's not used afterwards.

### Introduced `RawPreparedStatement`

`RawPreparedStatement` contains everything that is needed to construct the `PreparedStatement`, but avoid all 3 allocations that I listed above. Those allocations are delayed until one calls `into_prepared_statement()`, if materialised `PreparedStatement` is really needed.

### Extracted `Connection::prepare_raw()`

`Connection::prepare_raw()` extracts most of the logic that was put in `Connection::prepare()` before; yet, it returns `RawPreparedStatement`. `Connection::prepare()` is reimplemented as a wrapper over `Connection::prepare_raw()` that calls `RawPreparedStatement::into_prepared_statement()`.

### Actual optimisation

Call sites that used to call `Connection::prepare()` and later ignore the resulting `PreparedStatement` (or at least not require it in a full materialised form) were adjusted to call `Connection::prepare_raw()`.
Those involve:
1. Repreparation logic. Repreparation only needs to make the server reprepare the statement; the returned metadata  etc. are not needed, as the driver already has `PreparedStatement` built during the original preparation.
2. Preparation on many nodes/shards. Only one `PreparedStatement` needs to be materialised; subsequent results of preparation on other nodes/shards can be ignored (or, more precisely, verified, but for this the raw form suffices).

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -30,7 +30,9 @@ use crate::policies::retry::{RequestInfo, RetryDecision, RetrySession};
 use crate::policies::speculative_execution;
 use crate::policies::timestamp_generator::TimestampGenerator;
 use crate::response::query_result::{MaybeFirstRowError, QueryResult, RowsError};
-use crate::response::{NonErrorQueryResponse, PagingState, PagingStateResponse, QueryResponse};
+use crate::response::{
+    NonErrorQueryResponse, PagingState, PagingStateResponse, QueryResponse, RawPreparedStatement,
+};
 use crate::routing::partitioner::PartitionerName;
 use crate::routing::{Shard, ShardAwarePortRange};
 use crate::statement::batch::batch_values;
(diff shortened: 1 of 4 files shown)


muzarski (approved):
Is `Connection::prepare` used anywhere after this PR?

wprzytula:
> Is `Connection::prepare` used anywhere after this PR?

Yes, in at least three cases:
1) tests
2) `ControlConnection`
3) transparent autopreparation on `query` with parameters.

Lorak-mmk (approved):
Is this PR correct? Yes
Is it really useful, and a good way to allocate our time? I highly doubt it. Preparing a statement has never been a performance-important part of the driver. Why would we introduce additional code ( = additional complexity ) to reduce a number of allocations there is truly beyond me.

wprzytula:
> Is this PR correct? Yes Is it really useful, and a good way to allocate our time? I highly doubt it. Preparing a statement has never been a performance-important part of the driver. Why would we introduce additional code ( = additional complexity ) to reduce a number of allocations there is truly beyond me.

While I agree that preparation is by no means _that_ performance critical as prepared execution, you're missing at least a few details:
1) there exist dynamic use cases, where preparation is not a fully separate phase that can be ignored when thinking about performance;
2) this PR optimises **re**preparation as well, which happens on the execution path.

Issue #1336, opened 2025-04-29
Enable derive (de)serialization of `Box<str>`


Sokom141:
With a struct like the one in the following snippet:
```rust
struct Foo {
  field: Box<str>,
}
```
It is not possible to automatically derive `SerializeValue` and `DeserializeValue` but I do not see any reason way that is the case.

To me, after a very fast read of the relevant source code ([Serialize](https://docs.rs/scylla-cql/1.1.0/src/scylla_cql/serialize/value.rs.html#399), [Deserialize](https://docs.rs/scylla-cql/1.1.0/src/scylla_cql/deserialize/value.rs.html#340)), it seems like there is just a small code addition to support [this use case](https://users.rust-lang.org/t/use-case-for-box-str-and-string/8295/4).

If there is the need I would be open to implement this small change myself and open a PR. I think it would be really useful for my use case.

Thanks!

wprzytula:
You're welcome to submit such PR :)

Lorak-mmk:
If you do, maybe you could also cover `Arc<str>`? I see no reason to have support for Box, but not Arc.

Lorak-mmk:
This turns out to be a bit more tricky. To sum up my [comment](https://github.com/scylladb/scylla-rust-driver/pull/1339#issuecomment-2863131941):
- `SerializeValue` should be implemented for `str` instead of `&str`.
- `SerializeValue` should be implemented for `Arc<T>`
- `DeserializeValue` should be implemented for `Box<T>` and `Arc<T>` - we need to verify that it allows using `Box<str>`
The above is tightly coupled with solving https://github.com/scylladb/scylla-rust-driver/issues/1343

Lorak-mmk:
@Sokom141 The other issues I found were fixed in those PRs:
https://github.com/scylladb/scylla-rust-driver/pull/1378
https://github.com/scylladb/scylla-rust-driver/pull/1379

so now I opened the PR that implements ser/deser of Box<T>, Arc<T>, Box<str>, Arc<str>, you're welcome to take a look: https://github.com/scylladb/scylla-rust-driver/pull/1382

PR #1339, opened 2025-05-04
Add the ability to serialize/deserialize `Arc<str>` and `Box<str>`

Sokom141:
# Description
This PR add the ability to automatically derive `SerializeValue` and `DeserializeValue` on structs that contain `Box<str>` and `Arc<str>`, enabling particular patterns to reduce memory footprint.

Along with the implementation I have added tests where I thought it made sense. Everything seems to work but I had troubles with the Makefile and wasn't able to bring up the 3-node cluster, so please double check (`cargo t -p scylla-cql` passed every test anyway). 

Thanks!

## Pre-review checklist

Fixes: #1336

### DIFF
--- a/scylla-cql/src/deserialize/value_tests.rs
+++ b/scylla-cql/src/deserialize/value_tests.rs
@@ -96,11 +96,7 @@ fn test_deserialize_ascii() {
         // Nonempty string
         assert_ser_de_identity(typ, &ASCII_TEXT, &mut Bytes::new());
         assert_ser_de_identity(typ, &ASCII_TEXT.to_owned(), &mut Bytes::new());
-        assert_ser_de_identity(
-            typ,
-            &ASCII_TEXT.to_owned().into_boxed_str(),
-            &mut Bytes::new(),
-        );
+        assert_ser_de_identity(typ, &Box::from(ASCII_TEXT), &mut Bytes::new());
         assert_ser_de_identity(typ, &Arc::from(ASCII_TEXT), &mut Bytes::new());
     }
 }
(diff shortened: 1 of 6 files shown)


scylla-cql/src/deserialize/value_tests.rs
```diff
@@ -107,8 +119,11 @@ fn test_deserialize_text() {
         deserialize::<&str>(&ColumnType::Native(NativeType::Text), &unicode).unwrap();
     let decoded_text_string =
         deserialize::<String>(&ColumnType::Native(NativeType::Text), &unicode).unwrap();
+    let decoded_text_boxed_str =
+        deserialize::<Box<str>>(&ColumnType::Native(NativeType::Text), &unicode).unwrap();
     assert_eq!(decoded_text_str, UNICODE_TEXT);
     assert_eq!(decoded_text_string, UNICODE_TEXT);
+    assert_eq!(decoded_text_boxed_str.as_ref(), UNICODE_TEXT);
```
Lorak-mmk, line 127: In `test_deserialize_text` you only added tests for boxed str, but not Arced str. Could you also add them?

scylla-cql/src/serialize/value_tests.rs
```diff
@@ -1635,6 +1635,24 @@ fn text_serialization() {
...
+
+#[test]
+fn arc_str_serialization() {
+    let val: Arc<str> = "abc".into();
+    assert_eq!(
+        do_serialize(val, &ColumnType::Native(NativeType::Text)),
+        vec![0, 0, 0, 3, 97, 98, 99]
+    );
+}
+
```
Lorak-mmk, line 1655: Could you move those cases into `text_serialization`? This test already covers both str and String, so I think it can also cover Box and Arc.

Lorak-mmk (changes requested):
Thanks for the contribution! I left some minor comments regarding tests.
I also think that docs should be adjusted. 
- `docs/source/data-types/text.md` - it should mention that Box<str> and Arc<str> are supported.
- `docs/source/data-types/data-types.md` mapping should also be updated.

Sokom141:
Let me know if everything looks good now.
Thank you

wprzytula (dismissed):

Sokom141:
Nice catch. This is fine for me, I am happy that this has uncovered some underlying problems!

wprzytula:
Closing in favour of merged #1382.

wprzytula:
Thanks for raising the issue @Sokom141!

Issue #1178, opened 2025-01-23
`CachingSession` does not allow using `SKIP_METADATA`


wprzytula:
## Problem statement

As `CachingSession` manages preparation internally, it does not expose its `PreparedStatement`s in any way. Therefore, it's not possible for a user of `CachingSession` to call `PreparedStatement::set_use_cached_result_metadata` in order to take advantage of that optimisation.

There are use cases when not having possibility to turn the optimisation on may result in significant inefficiency. As an example, the experimental NodeJS-over-Rust driver's API suits `CachingSession`'s API best, but lack of the mentioned optimisation makes the experimental driver fall short compared to the original NodeJS driver (which, most likely, has the optimisation turned on).

## Possible solution

Expose a way to set up `CachingSession` so that it turns the optimisation on for all `PreparedStatement`s it manages.
Of course, write a big `BEWARE` disclaimer mentioning what are the possible pitfalls when using the optimisation (`SELECT *`, etc.).

PR #1340, opened 2025-05-05
CachingSession: Allow skipping result metadata

Lorak-mmk:
As `CachingSession` manages preparation internally, it does not expose its `PreparedStatement`s in any way. Therefore, it's not possible for a user of `CachingSession` to call
`PreparedStatement::set_use_cached_result_metadata` in order to take advantage of that optimisation.

There are use cases when not having possibility to turn the optimisation on may result in significant inefficiency. As an example, the experimental NodeJS-over-Rust driver's API suits `CachingSession`'s API best, but lack of the mentioned optimisation makes the experimental driver fall short compared to the original NodeJS driver (which, most likely, has the optimisation turned on).

New method is added to recently introduced CachingSessionBuilder, allowing user to enable caching result metadata.
Additionally, I added the possibility to share the same Session between multiple CachingSessions.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1178

## Pre-review checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode
+.zed
 target/
 Cargo.lock
 /book/book
--- a/scylla/src/client/caching_session.rs
+++ b/scylla/src/client/caching_session.rs
@@ -299,8 +299,12 @@ impl CachingSessionBuilder<RandomState> {
     /// which can be used to create a new [CachingSession].
     ///
     pub fn new(session: Session) -> Self {
+        Self::new_shared(Arc::new(session))
+    }
+
+    pub fn new_shared(session: Arc<Session>) -> Self {
         Self {
-            session: Arc::new(session),
+            session,
             max_capacity: DEFAULT_MAX_CAPACITY,
             hasher: RandomState::default(),
             use_cached_metadata: false,
--- a/scylla/src/client/caching_session.rs
+++ b/scylla/src/client/caching_session.rs
@@ -37,7 +37,7 @@ pub struct CachingSession<S = RandomState>
 where
     S: Clone + BuildHasher,
 {
-    session: Session,
+    session: Arc<Session>,
     /// The prepared statement cache size
     /// If a prepared statement is added while the limit is reached, the oldest prepared statement
     /// is removed from the cache
@@ -65,7 +65,7 @@ where
 {
     pub fn from(session: Session, cache_size: usize) -> Self {
         Self {
-            session,
+            session: Arc::new(session),
             max_capacity: cache_size,
             cache: Default::default(),
             use_cached_metadata: false,
@@ -81,7 +81,7 @@ where
     /// and a [`BuildHasher`], using a customer hasher.
     pub fn with_hasher(session: Session, cache_size: usize, hasher: S) -> Self {
         Self {
-            session,
+            session: Arc::new(session),
             max_capacity: cache_size,
             cache: DashMap::with_hasher(hasher),
             use_cached_metadata: false,
@@ -288,7 +288,7 @@ pub struct CachingSessionBuilder<S = RandomState>
 where
     S: Clone + BuildHasher,
 {
-    session: Session,
+    session: Arc<Session>,
     max_capacity: usize,
     hasher: S,
     use_cached_metadata: bool,
@@ -300,7 +300,7 @@ impl CachingSessionBuilder<RandomState> {
     ///
     pub fn new(session: Session) -> Self {
         Self {
-            session,
+            session: Arc::new(session),
             max_capacity: DEFAULT_MAX_CAPACITY,
             hasher: RandomState::default(),
             use_cached_metadata: false,
--- a/scylla/src/client/caching_session.rs
+++ b/scylla/src/client/caching_session.rs
@@ -43,6 +43,7 @@ where
     /// is removed from the cache
     max_capacity: usize,
     cache: DashMap<String, RawPreparedStatementData, S>,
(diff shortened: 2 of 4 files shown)


muzarski (approved):
The test is missing `#[cfg_attr(scylla_cloud_tests, ignore)]`.

scylla/tests/integration/caching_session.rs
```diff
@@ -0,0 +1,94 @@
...
+            // It should also not be present for statement executed before setting skip_metadata
+            verify_statement_metadata(&caching_session, REQUEST_A, false, &mut feedback_rx).await;
+
+            caching_session.set_use_cached_result_metadata(false);
+
+            // Now we disabled skip_metadata. Metadata should be present for request executed before.
+            verify_statement_metadata(&caching_session, REQUEST_B, true, &mut feedback_rx).await;
+
+            running_proxy
+        },
```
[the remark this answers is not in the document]
Lorak-mmk: I don't see how this test could be considered complex. We have a very simple invariant (metadata is requested, for all requests, iff `set_use_cached_metadata` is false). We need to test both values. We also need to test both a new statement and already cached one (this is true also when we set the parameter at construction time). Me testing requests that were cached before parameter change is just me being overly cautious, and I don't think that those 2 function calls contribute that much to complexity.

Even if we agree that this is too complex, what is the solution?
- Introducing more constructors - we would need to add 2, it would be ugly and naming would be hard
- Introducing a builder - it is out of scope of this issue / PR (which is about introducing new parameter, not a builder), similarly to how https://github.com/scylladb/scylla-rust-driver/pull/1246#discussion_r2049276663 was declared out of scope.

So I see 3 ways forward, please tell me which one you prefer.
- Keeping the current solution
- Declaring that refactors such as introducing builder are not out of scope. Implementing builder in this PR (or a separate prerequisite PR) and implementing https://github.com/scylladb/scylla-rust-driver/pull/1246#discussion_r2049276663 (possibly also in a separate prerequisite PR).
- Keeping the approach that such refactors are out of scope, creating a separate task of introducing caching session builder and waiting with this PR for this task completion (which I understand is not optimal for javascript-rust driver).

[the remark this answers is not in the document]
wprzytula: > I don't see how this test could be considered complex. We have a very simple invariant (metadata is requested, for all requests, iff `set_use_cached_metadata` is false). We need to test both values (true and false). We also need to test both a new statement and already cached one (this is true also if we set the parameter at construction time). Me testing requests that were cached before parameter change is just me being overly cautious, and I don't think that those 2 function calls contribute that much to complexity.

The very need of considering runtime  (i.e., once the `CachingSession` has been constructed) changes of this (or any other) configuration parameter comes with mental burden that mutability involves. I'm against needless mutability in general, thence I deduce that we shouldn't introduce it here.
> 
> Even if we agree that this is too complex, what is the solution?
> 
>   * Introducing more constructors - we would need to add 2, it would be ugly and naming would be hard

Agreed that this is ugly and scales bad with future-added options.
> 
>   * Introducing a builder - it is out of scope of this issue / PR (which is about introducing new parameter, not a builder), similarly to how [Proxy: compresssion support #1246 (comment)](https://github.com/scylladb/scylla-rust-driver/pull/1246#discussion_r2049276663) was declared out of scope.

I believe the builder is the way to go.
> 
> 
> So I see 3 ways forward, please tell me which one you prefer.
> 
>     * Keeping the current solution
> 
>     * Declaring that refactors such as introducing builder are not out of scope. Implementing builder in this PR (or a separate prerequisite PR) and implementing [Proxy: compresssion support #1246 (comment)](https://github.com/scylladb/scylla-rust-driver/pull/1246#discussion_r2049276663) (possibly also in a separate prerequisite PR).

These two cases are by no means equivalent. The proxy is our private testing API, which means we are free to refactor it at any point, including breaking its API. The opposite is true for public API, which `CachingSession` is part of. We must be more careful and cautious about introducing new public APIs now, once we got API-stable.

Therefore, I believe we can still merge #1246 first and introduce the builder you requested in a follow-up, which is not the case for this PR. 

> 
>     * Keeping the approach that such refactors are out of scope, creating a separate task of introducing caching session builder and waiting with this PR for this task completion (which I understand is not optimal for javascript-rust driver).

I think this is my preferred approach - let's first add a `CachingSession` builder in a separate PR, then base this one on it.

Lorak-mmk:
Addressed the remaining review comment.

.zed/settings.json
```diff
@@ -0,0 +1,17 @@
...
+                        "enforce": true
+                    }
+                },
+                "group": {
+                    "enable": true
+                }
+            }
+        }
+    }
+}
```
muzarski, line 17: commit: CachingSession: Test for result metadata cache 

File `.zed/settings.json` should probably be added to `.gitgnore`.

muzarski (approved):

.zed/settings.json
Lorak-mmk: I added it by accident, sorry, I'll remove it tomorrow.

Although I would not mint if all contributors used those rust-analyzer settings

Lorak-mmk: Done, I also added it to gitignore.

muzarski (approved):

wprzytula (approved):

Lorak-mmk:
Addressed @wprzytula comments

muzarski (approved):

wprzytula (approved):

Issue #1342, opened 2025-05-07
Introduce CachingSession builder


Lorak-mmk:
Introducing more configuration options to CachingSession is currently not feasible.
It requires either:
- Mutating the session after creation (which we do not want)
- Introducing an exponential number of constructors.

The solution is to introduce a builder for CachingSession.

PR #1345, opened 2025-05-09
caching_session: introduce CachingSessionBuilder

wprzytula:
We can see that CachingSession requires more config options than there are currently. Not to cause exponential growth of number of constructors, a builder is introduced.
A unit test using proxy comes together, too.

Fixes: #1342

## Pre-review checklist

### DIFF
--- a/scylla/src/client/caching_session.rs
+++ b/scylla/src/client/caching_session.rs
@@ -247,9 +247,131 @@ where
     }
 }
 
+/// The default cache capacity set on the [CachingSessionBuilder].
+/// Can be changed using [CachingSessionBuilder::max_capacity].
+pub const DEFAULT_MAX_CAPACITY: usize = 128;
+
+/// [CachingSessionBuilder] is used to create new [CachingSession] instances.
+///
+/// # Example
+///
+/// ```
+/// # use scylla::client::session::Session;
+/// # use scylla::client::session_builder::SessionBuilder;
+/// # use scylla::client::caching_session::{CachingSession, CachingSessionBuilder};
+/// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+/// let session: Session = SessionBuilder::new()
+///     .known_node("127.0.0.1:9042")
+///     .build()
+///     .await?;
+/// let caching_session: CachingSession = CachingSessionBuilder::new(session)
+///     .max_capacity(2137)
+///     .build();
+/// # Ok(())
+/// # }
+/// ```
+pub struct CachingSessionBuilder<S = RandomState>
+where
+    S: Clone + BuildHasher,
+{
+    session: Session,
+    max_capacity: usize,
+    hasher: S,
+}
+
+impl CachingSessionBuilder<RandomState> {
+    /// Wraps a [Session] and creates a new [CachingSessionBuilder] instance,
+    /// which can be used to create a new [CachingSession].
+    ///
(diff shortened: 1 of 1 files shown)


scylla/src/client/caching_session.rs
```diff
@@ -247,9 +247,129 @@ where
...
+    /// which can be used to create a new [CachingSession].
+    ///
+    pub fn new(session: Session) -> Self {
+        Self {
+            session,
+            max_capacity: DEFAULT_MAX_CAPACITY,
+            hasher: RandomState::default(),
+        }
+    }
+}
```
Lorak-mmk, line 291: So the flow to create session with custom hasher using this is to first create builder with `RandomState` and then change it to a builder with different hasher using`with_hasher`?

scylla/src/client/caching_session.rs
```diff
@@ -247,9 +247,129 @@ where
...
+impl<S> CachingSessionBuilder<S>
+where
+    S: Clone + BuildHasher,
+{
+    /// Configures maximum capacity of the prepared statements cache.
+    pub fn max_capacity(mut self, max_capacity: usize) -> Self {
+        self.max_capacity = max_capacity;
+        self
+    }
+
```
Lorak-mmk, line 302: My one worry is that max capacity no longer has to be stated by the user explicitly, which may bite users unaware of this who use lots of different statements.

WDYT of holding Option<usize> and making build() return Result (Err if capacity was not specified)?

scylla/src/client/caching_session.rs
wprzytula: Yes, with small difference: the method is called `hasher()`.

scylla/src/client/caching_session.rs
wprzytula: Nah, this is ugly. At most, I could agree to accept `max_capacity` as a mandatory parameter in `CachingSessionBuilder::new()`.
But still - users should not expect that default configuration will work perfectly in their particular settings. Instead, we can warn them in a doc comment that they may want to tune `max_capacity` option.

Lorak-mmk: > Instead, we can warn them in a doc comment that they may want to tune max_capacity option.

Ok let's do that.

wprzytula: Done.

scylla/src/client/caching_session.rs
```diff
@@ -247,9 +247,135 @@ where
...
+const DEFAULT_MAX_CAPACITY: usize = 128;
+
+/// [CachingSessionBuilder] is used to create new [CachingSession] instances.
+///
+/// **NOTE:** The builder specifies a default capacity of the prepared statement cache
+/// that may be too low for use cases running lots of different prepared statements.
+/// If you expect to run a large number of different prepared statements (more than
+/// `DEFAULT_MAX_CAPACITY`), consider increasing the capacity with
+/// [CachingSessionBuilder::max_capacity].
+///
```
Lorak-mmk, line 259: Can we make this const pub? Otherwise users won't see it in docs, forcing them to check the source for the value.

wprzytula: I think we can.

wprzytula: Done.

Lorak-mmk (approved):

muzarski (approved):

PR #1351, opened 2025-05-10
Reorganise integration tests - part 2

wprzytula:
This is a follow-up of #1350. Motivation is the same.

This PR, in contrast with the previous one, dissolves many large test files and splits them into more granular, tree-structured modules.

### Sanity check

I deleted one test, which I believe to be redundant and pointless (`test_await_schema_agreement()`). With this in mind, the number of tests before and after is correct, which means I didn't lose any test.

Before:
```
Running tests/integration/main.rs (target/debug/deps/integration-6bde174c2b23de7e)

running 108 tests
```

After:
```
Running tests/integration/main.rs (target/debug/deps/integration-6bde174c2b23de7e)

running 107 tests
```

## Pre-review checklist


scylla/tests/integration/types/valuelists.rs
Lorak-mmk, line 1: Commit: "integration: extract valuelists.rs"

This test was only named that because the serialization trait from old framework was called ValueList.
There is no reason to keep this name, so the test should be renamed, and put with other SerializeRow tests instead of its own file.

scylla/tests/integration/metadata/keyspace_filtering.rs
Lorak-mmk, line 1: Commit: "integration: extract keyspace_filtering.rs "

I see that in the final version there are the following 4 files in the `metadata` folder:
- `keyspace_filtering.rs`
- `metadata_custom_timeouts.rs`
- `turn_off_schema_fetching.rs`
Each of them contains only one test. I think that it is a bit too much granular, especially that those 3 tests have something in common: they test configuration options of metadata fetching. Maybe we could put them in e.g. `configuration.rs`?
As a bonus `refresh_upon_schema_agreement.rs` could fit there too. Even more if we write a test that verifies the opposite (that if we disable automatic schema agreement, then there is no refresh after DDL).

scylla/tests/integration/statements/prepared/prepared_inherits_config.rs
Lorak-mmk, line 1: Commit: `integration: extract test_prepared_config()`

Looking at final `statements/prepared` I again feel there is too much granularity and chaos.
- test in basic.rs also looks into the cols_specs, so it is weird that we have `col_specs.rs`
- Last test case in the test in `basic.rs` verifies macros - I don't see what does it have to do with prepared statements. Maybe it should be extracted?
- In general I'm starting to feel like we went from one extreme (gigantic session.rs with many unrelated tests) to another (each test in its own file). I don't like either of those extreme - can we do it in more balanced way? The whole `prepared` folder could be a `prepared.rs` file.

Lorak-mmk (changes requested):
What I wrote about prepared I think also applies to batch and unprepared - they also could be a files instead of folders.

scylla/tests/integration/types/valuelists.rs
wprzytula: I understood the name _value list_ as a valid name independently of past existence of `ValueList` trait. If you believe there's no such notion in CQL glossary, then I can change the name - what new name do you suggest?

scylla/tests/integration/metadata/mod.rs
```diff
@@ -1,3 +1,4 @@
 mod contents;
 mod metadata_custom_timeouts;
 mod refresh_upon_schema_agreement;
+mod turn_off_schema_fetching;
```
muzarski, line 5: commit: integration: extract turn_off_schema_fetching.rs

I think the only test that this file contains could be put in `metadata/contents.rs`. WDYT?

scylla/tests/integration/metadata/contents.rs
```diff
@@ -495,6 +496,7 @@ async fn test_views_in_schema_info() {
     )
 }
 
+/// This test case indicates that we support enough CQL types to parse schema keyspace information.
```
muzarski, line 499: commit: metadata: fix test_fetch_system_keyspace() 

Good catch!

scylla/tests/integration/metadata/mod.rs
muzarski: I did not see @Lorak-mmk comment - makes sense to have sth like `configuration.rs`. I agree that currently `metadata` module is too granular.

scylla/tests/integration/statements/prepared/prepared_inherits_config.rs
muzarski: I agree - `prepared.rs` sounds most reasonable.

muzarski (changes requested):
Nice job.

PR #1356, opened 2025-05-15
Bump MSRV to 1.81 + minor fixes

Lorak-mmk:
This PR bumps the MSRV to 1.81 and also fixes some small Cargo.toml issues (redundant deps, old Criterion).
Previously this PR also transitioned driver to use `expect` lint level where possible, but I decided to do that in a separate PR.

Old cover letter:

We want to use `#[expect(...)]` in https://github.com/scylladb/scylla-rust-driver/pull/1296 which requires Rust 1.81
Bumping MSRV and transitioning from allow to expect where possible is big enough change that it imo warrants a separate PR.

I also extracted other smaller changes from #1296 here: Cargo.toml cleanup, Criterion bump, Makefile changes.

I noticed that at some point Clippy stopped checking naming of our pub items. This is because it has an option `avoid-breaking-exported-api`, enabled by default.
This is an absurd and terrible default because it makes the checks in PRs (when it is perfectly fine to break API) disabled as well.
For this reason we accumulated a lot of wrongly named items which we will need to fix in 2.0.
I added `#[expect(...)]` and TODO to each one.

Additionally I removed `Node::is_down` and all related code. There is no use for it and it was triggering a check, so instead of using `allow` or `cfg_attr` I decied to remove it totally.

## Pre-review checklist


Lorak-mmk:
I don't understand this failure in MSRV check. 
```
 error: this lint expectation is unfulfilled
Error:    --> scylla/tests/integration/macros/hygiene.rs:179:22
    |
179 |             #[expect(unused_imports)]
    |                      ^^^^^^^^^^^^^^
...
335 |     test_crate!(scylla);
    |     ------------------- in this macro invocation
    |
    = note: `-D unfulfilled-lint-expectations` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unfulfilled_lint_expectations)]`
    = note: this error originates in the macro `test_crate` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Those importa are obviously unused, why is the check not triggered?
```rust
        // The purpose of this test is to verify that the important types are
        // correctly re-exported in `scylla` crate.
        #[test]
        fn test_types_imports() {
            #[expect(unused_imports)]
            use _scylla::frame::response::result::{CollectionType, ColumnType, NativeType, UserDefinedType};
            #[expect(unused_imports)]
            use _scylla::value::{
                Counter, CqlDate, CqlDecimal, CqlDecimalBorrowed, CqlDuration, CqlTime, CqlTimestamp,
                CqlTimeuuid, CqlValue, CqlVarint, CqlVarintBorrowed, MaybeUnset, Row, Unset, ValueOverflow
            };
        }
```

scylla/src/network/tls.rs
```diff
@@ -60,7 +60,7 @@ impl TlsProvider {
     pub(crate) fn make_tls_config(
         &self,
         // Currently, this is only used for cloud; but it makes abstract sense to pass endpoint here
-        // also for non-cloud cases, so let's just allow(unused).
+        // also for non-cloud cases, so let's just expect(unused).
         #[allow(unused)] endpoint: &UntranslatedEndpoint,
```
wprzytula, line 64: :wrench: You updated the comment, but not the actual occurrence.

wprzytula (changes requested):

Lorak-mmk:
Removed the commits that change lint levels and remove unused lints. I'll do this work in a separate PR.

muzarski (approved):

wprzytula (approved):

PR #1357, opened 2025-05-19
Address Clippy's lints introduced in Rust 1.87

Lorak-mmk:
New Rust version, new Clippy lints enabled by default. I decided to silence the lints about too large errors. Those are our standard error types, it would not be ergonomical to Box them. In some cases the Ok variant is larger, so it would make no sense to Box errors. If it ever becomes an issue, we can reconsider.

## Pre-review checklist

### DIFF
--- a/scylla/src/response/query_result.rs
+++ b/scylla/src/response/query_result.rs
@@ -207,6 +207,8 @@ impl QueryResult {
     /// Ok(())
     /// # }
     /// ```
+    // `allow(clippy::result_large_err)` is fine. `QueryRowsResult` is larger than `IntoRowsResultError`.
+    #[allow(clippy::result_large_err)]
     pub fn into_rows_result(self) -> Result<QueryRowsResult, IntoRowsResultError> {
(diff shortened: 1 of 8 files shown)


scylla/src/response/query_result.rs
```diff
@@ -207,6 +207,8 @@ impl QueryResult {
     /// Ok(())
     /// # }
     /// ```
+    // `allow(clippy::result_large_err)` is fine. `QueryRowsResult` is smaller than `IntoRowsResultError`.
+    #[allow(clippy::result_large_err)]
```
wprzytula, line 211: :question: Didn't you mean: `is larger`?

wprzytula: Otherwise the explanation is counter-convincing.

wprzytula (approved):

scylla/src/response/query_result.rs
Lorak-mmk: Right, sorry.

muzarski (approved):

Issue #675, opened 2023-03-22
Enable filtering down nodes in `scylla::transport::load_balancing::DefaultPolicy`


havaker:
Our node up/down markers do not work reliably yet (https://github.com/scylladb/scylla-rust-driver/pull/607), but are required for a true `is_alive` check:

https://github.com/scylladb/scylla-rust-driver/blob/adb18f7312dfcf8f200e79a4789d59ff9bf6fe8c/scylla/src/transport/load_balancing/default.rs#L350-L354

Lorak-mmk:
@wprzytula @muzarski In user API we replaced event-based node status (`is_down`  method) with checking if any connections are open (`is_connected` method). I think this check is good enough to use in `DefaultPolicy`, do you agree?

wprzytula:
> [@wprzytula](https://github.com/wprzytula) [@muzarski](https://github.com/muzarski) In user API we replaced event-based node status (`is_down` method) with checking if any connections are open (`is_connected` method). I think this check is good enough to use in `DefaultPolicy`, do you agree?

Definitely. Scheduling this for a release soon.

PR #1364, opened 2025-05-28
Default policy: filter out disconnected nodes

Lorak-mmk:
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/675

The main change is simple: in `DefaultPolicy::is_alive` we switch to use `Node::is_connected` instead of `Node::is_enabled`.

This of course breaks unit tests, because nodes there are not connected. To work around this I introduced a test-only field in Node, that allows using result of `is_enabled` in `is_connected`.
I chose this solution, because it did not require a lot of changes, and works well with existing tests. For example, second part of `test_default_policy_with_lwt_statements` performs a test with disabled node. It did not require (nearly) any changes.
I am open to suggestions about alternative approaches here.

## Pre-review checklist


scylla/src/cluster/node.rs
```diff
@@ -78,6 +80,9 @@ pub struct Node {
 
     // If the node is filtered out by the host filter, this will be None
     pool: Option<NodeConnectionPool>,
+
+    #[cfg(test)]
+    enabled_as_connected: AtomicBool,
```
wprzytula, line 85: :wrench: Please take the explanation written in the commit message and put in a docstring here.

scylla/src/cluster/node.rs
```diff
@@ -357,7 +370,13 @@ mod tests {
...
                 rack,
                 pool: None,
+                enabled_as_connected: AtomicBool::new(false),
             }
         }
+
+        #[allow(unused)]
+        pub(crate) fn force_enabled_as_connected(&self) {
+            self.enabled_as_connected.store(true, Ordering::SeqCst);
+        }
```
wprzytula: :pick: `force` -> `use`?

wprzytula (approved):
Though this looks like a tiny change, it's actually a massive potential performance improvement in case of node outages. Thanks! :100:

wprzytula (approved):

muzarski (approved):

Issue #1369, opened 2025-06-03
"use of unresolved module or unlinked crate `openssl`" when running tests for scylla package


Lorak-mmk:
When running tests in the following way:
```
 SCYLLA_URI=172.42.0.2:9042 SCYLLA_URI2=172.42.0.3:9042 SCYLLA_URI3=172.42.0.4:9042 cargo test --package scylla
```
it fails in the following way:

```
failures:

---- scylla/src/client/session_builder.rs - client::session_builder::GenericSessionBuilder<DefaultMode>::tls_context (line 333) stdout ----
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `openssl`
 --> scylla/src/client/session_builder.rs:339:5
  |
9 | use openssl::ssl::{SslContextBuilder, SslVerifyMode, SslMethod, SslFiletype};
  |     ^^^^^^^ use of unresolved module or unlinked crate `openssl`
  |
  = help: if you wanted to use a crate named `openssl`, use `cargo add openssl` to add it to your `Cargo.toml`

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0433`.
Couldn't compile the test.

failures:
    scylla/src/client/session_builder.rs - client::session_builder::GenericSessionBuilder<DefaultMode>::tls_context (line 333)
```

We should probably to something like this https://users.rust-lang.org/t/doctests-that-require-a-non-default-feature-is-it-possible/29529/3 to only include this example when the corresponding feature is enabled. Or include it always, but ignore it if it is disabled.

PR #1370, opened 2025-06-04
doctests: correct configuration wrt crate features

wprzytula:
This patch fixes the unmet requirements for doctests wrt features:
1. The doctest for `SessionBuilder::tls_context()` is compiled conditionally, only when the required `openssl-010`  feature is turned on;
2. `scylla-macros`, `scylla-cql`, and `scylla-proxy` now have set `all-features` flag for docs.rs, which will make optional entities appear in those crates' respective docs. Before, `#[cfg(feature)]`'d entities would not show up in, e.g., `scylla-cql` crate docs on docs.rs.

Fixes: #1369

## Pre-review checklist

### DIFF
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -10,6 +10,9 @@ keywords = ["database", "scylla", "cql", "cassandra"]
 categories = ["database"]
 license = "MIT OR Apache-2.0"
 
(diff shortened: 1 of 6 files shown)


Lorak-mmk (approved):
Thanks for the quick fix!

PR #1373, opened 2025-06-05
default LBP: Warn on misconfigured local DC

wprzytula:
Ref: #1371

## What's done
`DefaultPolicy` now issues respective warnings:
1. if the preferred datacenter is not present in the cluster;
2. if all nodes in the preferred datacenter are disabled by the `HostFilter`.

This helps avoid confusion when the user expects requests to be sent to a specific datacenter, but it is not available.

### Note
As warnings are going to be emitted on every request (because the configuration is most likely session-wide), the messages will quickly flood the logs and the user should notice.

Alternatively, we could implement some frequency limiting of those warnings, but we prefer to:
- keep it simple for now,
- let the user notice the misconfiguration as quickly as possible.

## Pre-review checklist

### DIFF
--- a/scylla/src/policies/load_balancing/default.rs
+++ b/scylla/src/policies/load_balancing/default.rs
@@ -589,8 +589,9 @@ impl DefaultPolicy {
     }
 
     /// Checks for misconfiguration and warns if any is discovered.
-    fn pick_sanity_checks(&self, routing_info: &ProcessedRoutingInfo, _cluster: &ClusterState) {
-        if let Some(_preferred_dc) = self.preferences.datacenter() {
+    fn pick_sanity_checks(&self, routing_info: &ProcessedRoutingInfo, cluster: &ClusterState) {
+        if let Some(preferred_dc) = self.preferences.datacenter() {
+            // Preferred DC + no datacenter failover + SimpleStrategy is an anti-pattern.
             if let Some(ref token_with_strategy) = routing_info.token_with_strategy {
                 if !self.permit_dc_failover
                     && matches!(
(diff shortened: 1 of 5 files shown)


scylla/src/policies/load_balancing/default.rs
```diff
@@ -602,6 +588,77 @@ impl DefaultPolicy {
...
+                }
+            }
+
+            // Verify that the preferred datacenter is actually present in the cluster.
+            // If not, shout a warning.
+            if !cluster.all_nodes.iter().any(|node| {
+                node.datacenter
+                    .as_deref()
+                    .is_some_and(|dc| dc == preferred_dc)
+            }) {
```
Lorak-mmk, line 617: This iterates over potentially all nodes (even in a happy case). I don't think it is a good idea performance-wise to have such checks in the hot path.

scylla/src/policies/load_balancing/default.rs
```diff
@@ -602,6 +588,77 @@ impl DefaultPolicy {
...
+                    );
+                }
+            }
+            // Verify that there exist any enabled nodes in the preferred datacenter.
+            // If not, shout a warning.
+            else if !cluster.all_nodes.iter().any(|node| {
+                node.is_enabled()
+                    && node
+                        .datacenter
+                        .as_deref()
```
Lorak-mmk, line 640: ditto

Lorak-mmk (changes requested):

scylla/src/policies/load_balancing/default.rs
Lorak-mmk: Can't you use ReplicaLocator for that? I see it has `ReplicationInfo` which has:
```rust
    /// We keep a separate token ring for each datacenter.
    /// Each datacenter's ring contains only nodes from this datacenter.
    /// It makes it simpler and more efficient to calculate the NetworkTopologyStrategy replicas.
    datacenters: HashMap<String, DatacenterNodes>,
```

wprzytula: Good catch!

wprzytula:
Rebased on main.

Lorak-mmk (changes requested):
I think the API of the newly introduced RateLimiter could be better. My idea:
- Duration should be stored inside the limiter, because I find it unlikely to that we'll need to use multiple durations with single limiter.
- Instead of having a method returning bool, we can have `maybe_execute` taking a closure that will be called if timer passed. 
- After the above, the macros won't really give much value over having the limiter stored explicitly inside DefaultPolicy (or having it explicitly as a static). Let's avoid macros unless they really do help in something.
- Can you also move timestamp generator to use the new rate limiter?

Lorak-mmk:
> The idea I had with the macros there was to make rate-limited warnings maximally welcoming, so that there's minimum effort and code bloat involved when employing them. Keep in mind that the way macro encapsulates the static instance of the RateLimiter makes it immune to accidental sharing of the same limiter for two separate warnings.

In case of DefaultPolicy I don't see any reason to have a separate limiters for those 4 warnings. Each limiter is another global object that is always present. Macros also hide this overhead.

wprzytula:
> > This is limiting. This will, for instance, disallow for executing async code.
> 
> This can be solved by having also an async version of this method. What are the other limitations?

WDYM by the async version?
To have one method accepting a closure and one accepting an `impl Future`?

This seems bloaty. Why not just give users a flexible API that lets them utilize the `RateLimiter` as they want?

wprzytula:
> Each limiter is another global object that is always present. Macros also hide this overhead.

This overhead is justified: each warning string is, similarly, another static object that is always present. As the (I assume you mean the memory) overhead is tiny and static, I believe it's not anything to bikeshed about.

Lorak-mmk:
> This seems bloaty. Why not just give users a flexible API that lets them utilize the RateLimiter as they want?

Because it introduces macros for something that can be done without them, without much more verbosity.

Lorak-mmk:
> These warnings are different. We do want the user to get each of them, not only one. If we rate limit them together, then if multiple of them are triggered in close time, only the first one is printed.

It is extremely unlikely for multiple of those warnings to be triggered, I don't see much sense in caring about this case.

dkropachev:
> > Each limiter is another global object that is always present. Macros also hide this overhead.
> 
> This overhead is justified: each warning string is, similarly, another static object that is always present. As the (I assume you mean the memory) overhead is tiny and static, I believe it's not anything to bikeshed about.

What bothers me is that on every request driver runs these checks, which is not realy necessary.
What if we would do that only when driver get into undesired state, say empty query plan or failover to another DC ?

wprzytula:
> > This seems bloaty. Why not just give users a flexible API that lets them utilize the RateLimiter as they want?
> 
> Because it introduces macros for something that can be done without them, without much more verbosity.

You seem to mix two things. I'm arguing about exposing a boolean from the `RateLimiter`'s method, in contrast to only accepting a closure/Future. You are arguing that macros are unneeded. These are two distinct things.

Lorak-mmk:
> I can agree with this reasoning, but I still believe that there's no generic enough way to represent "an action" to create such API with just one method.

I really don't see the problem with having 2 - try_execute and try_execute_async.
Rust doesn't have effect system / keyword generics, and afaik this is the most typical way to deal with having sync / async version of something.

wprzytula:
> > > Each limiter is another global object that is always present. Macros also hide this overhead.
> > 
> > 
> > This overhead is justified: each warning string is, similarly, another static object that is always present. As the (I assume you mean the memory) overhead is tiny and static, I believe it's not anything to bikeshed about.
> 
> What concerns me is that the driver performs these checks on every request, even though they’re usually unnecessary. In effect, we’re sacrificing performance in the common case just to show a helpful warning when the user misconfigures the driver. What if we would do that only when driver get into undesired state, say empty query plan or failover to another DC ?

I understand your concerns.

However, if think these checks should be cheap. @Lorak-mmk do you agree?
-

Issue #1384, opened 2025-06-30
statements: Allow unsetting (serial) consistency


wprzytula:
## Problem statement

In the current implementation, it is not possible to unset the (serial) consistency level for a statement (`Statement`, `PreparedStatement`, `Batch`. This can be problematic when you want to revert to the default consistency level after having set it explicitly. The problem manifests in CPP-Rust Driver, where the user can specify `CASS_CONSISTENCY_UNKNOWN` to unset the consistency (and fall back to the consistency specified in the execution profile), but the corresponding Rust API does not allow this.

## Proposed solution

Implement a set of methods to unset (clear) overriden the consistency level for statements. This will allow users to revert to the default (exec profile-specified) consistency level after having set it explicitly.

### Set of methods:
`{Statement,PreparedStatement,Batch}::{clear/unset}(_serial)_consistency()`, so this gives us:

- `Statement::unset_consistency()`
- `Statement::unset_serial_consistency()`
- `PreparedStatement::unset_consistency()`
- `PreparedStatement::unset_serial_consistency()`
- `Batch::unset_consistency()`
- `Batch::unset_serial_consistency()`

PR #1385, opened 2025-06-30
statements: expose API to unset (serial) consistency

wprzytula:
This is required by the CPP Rust Driver, and also useful for Rust Driver direct users. See #1384 for details.

Bonus: fixed codewide typo "overriden" -> "overridden".

Fixes: #1384

## Pre-review checklist


wprzytula:
> Now that you modify this code, could you also change `serial_consistency` type in `StatementConfig`, and other similar places, so that they are not using nested Options? Those are confusing to me - its not clear which Option means what. We can introduce en enum like ConfigValue::Set/Unset.

OK, but this is not related to this PR. Either I or you can do this in a follow-up.

PR #1386, opened 2025-06-30
crates: allow `clippy::uninlined_format_args`

wprzytula:
Clippy has started to complain about format strings that are not inlined. In order to satisfy clippy's demands, we would have to introduce a lot of changes. At the same time, I don't see any value in avoiding use of format strings that are not inlined.

Thus, to calm clippy down, we allow this lint in all crates.

## Pre-review checklist

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -4,6 +4,9 @@ name = "examples"
 publish = false
 version = "0.0.0"
 
+[lints.clippy]
+uninlined_format_args = "allow"
+
 [dev-dependencies]
 anyhow = "1.0.33"
 futures = "0.3.6"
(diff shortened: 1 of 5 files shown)


piodul:
> For some reason clippy only started to warn about others after I fixed the first 3.

It probably produced errors for scylla-macros crate and didn't compile it, so it couldn't attempt to compile scylla-cql and scylla.

Lorak-mmk:
> It probably produced errors for scylla-macros crate and didn't compile it, so it couldn't attempt to compile scylla-cql and scylla.

This is most likely what happened, but its still a weird behavior imo.
For compiler this makes sense - if compilation of create X fails, then we have no way to compile any crate dependent on X.
For linting this is not so clear - lints failing in one create does not inherently prevent this crate from being indexed and used in analysis of its dependents.

Issue #643, opened 2023-02-05
Deal with the new clippy lint: `uninlined_format_args`


cvybhu:
New version of clippy introduced the uninlined_format_args lint which recommends putting variable names directly into format strings.

In #639 we decided that the new lint should be disabled for now, because rust-analyzer has limited support for the code that clippy proposes.
The lint was disabled in CI and Makefile by adding a flag, but manually running `cargo clippy` would still throw a ton of errors.
Having to always add this flag is an unneeded annoyance for developers.

Maybe we could disable the lint on the level of the whole crate?
I think there's a special syntax for macros that are supposed to cover the whole crate, something like `#![allow(clippy::new_ret_no_self)]` (saw it [here](https://github.com/EmbarkStudios/rust-ecosystem/issues/22)).

I also created this issue to track the problem and remeber to make the code clippy compliant once rust-analyzer is able to handle everything.

piodul:
> New version of clippy introduced the uninlined_format_args lint which recommends putting variable names directly into format strings.
> 
> In #639 we decided that the new lint should be disabled for now, because rust-analyzer has limited support for the code that clippy proposes. The lint was disabled in CI and Makefile by adding a flag, but manually running `cargo clippy` would still throw a ton of errors. Having to always add this flag is an unneeded annoyance for developers.

You can set up the RUSTFLAGS environment variable and not have to worry about it:
`export RUSTFLAGS="-Aclippy::uninlined_format_args"`

> Maybe we could disable the lint on the level of the whole crate? I think there's a special syntax for macros that are supposed to cover the whole crate, something like `#![allow(clippy::new_ret_no_self)]` (saw it [here](https://github.com/EmbarkStudios/rust-ecosystem/issues/22)).

I originally didn't decide to do it like this because we actually have multiple crates: scylla, scylla-cql, scylla-proxy, every integration test and example - they all count as separate crates and need a separate `allow` attribute. It's not super hard work, but it sounds annoying to do something like this just to disable a single lint, and you need to remember to disable it in new examples/integration tests. Usually we avoid such problems by just fixing the new issues pointed out by clippy, but this time I think we should wait until support for renaming in rust-analyzer gets better.

That said, I'm not strongly opposed to introducing the annotations if it's just for this single lint. If you feel it will make life easier to developers, then please send a PR.

[...]

PR #1389, opened 2025-07-03
Fix clippy::uninlined_format_args

Lorak-mmk:
Replaces: https://github.com/scylladb/scylla-rust-driver/pull/1386

`clippy::uninlined_format_args` warns about format strings that have its variables listed after format string, instead of having it directly in curly braces.
Advantages of that? Shorter code that plays better with `cargo fmt`, and no need to jump with your eyes between string and values to understand the format.

There are a lot of changes, but they are fully mechanical, performed by clippy itself using `cargo clippy --all-targets --fix --allow-dirty` after I fixed first three occurrences myself.

## Pre-review checklist

### DIFF
--- a/examples/allocations.rs
+++ b/examples/allocations.rs
@@ -126,7 +126,7 @@ async fn measure(
 #[tokio::main]
 async fn main() -> Result<()> {
     let args = Args::parse();
-    println!("{:?}", args);
+    println!("{args:?}");
 
     println!("Connecting to {} ...", args.node);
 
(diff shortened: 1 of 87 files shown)


piodul (approved):
Heh, the lint has made a comeback - we had to deal with it in the past when it was enabled by default: https://github.com/scylladb/scylla-rust-driver/issues/643. It was made non-default back again shortly afterwards.

Last time, my main concern was inlined format args didn't have good support in rust-analyzer. I checked and now they are properly supported, which is nice.

Lorak-mmk:
> Last time, my main concern was inlined format args didn't have good support in rust-analyzer. I checked and now they are properly supported, which is nice.

Maybe this is the reason this lint was removed (from defaults) before, and added again now?

Lorak-mmk:
Looks like this: https://github.com/rust-lang/rust-clippy/pull/14160

piodul:
Yeah, looks like it was not only my concern - in fact, I probably learned about the problem after reading the issue, and only then it became my main concern.

Issue #1395, opened 2025-07-21
Consider adding cratewide `#![warn(missing_docs)]`


wprzytula:
I've noticed that `#![warn(missing_docs)]` is present at module level of `scylla-cql/serialize/mod.rs`. This seems to be a reasonable way to prevent us from overlooking some items (modules, functions, types) with regard to writing doc comments.

Unless @Lorak-mmk disagrees, I'll enable `#![warn(missing_docs)]` in all our crates and eliminate all offenders by adding missing doc comments.

PR #1398, opened 2025-07-23
scylla-cql: Write missing docstrings - part 1

wprzytula:
Ref: #1395

This is the first part of the effort to fill missing docstrings, on the way to enable `missing-docs` project-wide.

1. `scylla-macros` were easily fixed to pass the `missing-docs` lint, so it's now enabled there.
2. `scylla-cql` had over 550 warnings when I enabled the lint. After this PR, the number went down to ~200.

I paid attention to introduce meaningful documentation, not just repeat the items' names.

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/frame/response/mod.rs
+++ b/scylla-cql/src/frame/response/mod.rs
@@ -19,16 +19,65 @@ use crate::frame::TryFromPrimitiveError;
 use super::frame_errors::CqlResponseParseError;
 
 /// Possible CQL responses received from the server
+// Why is it distinct from [ResponseOpcode]?
+// TODO(2.0): merge this with `ResponseOpcode`.
 #[derive(Debug, Copy, Clone)]
 #[non_exhaustive]
 pub enum CqlResponseKind {
+    /// Indicates an error processing a request.
     Error,
+
+    /// Indicates that the server is ready to process queries. This message will be
+    /// sent by the server either after a STARTUP message if no authentication is
+    /// required (if authentication is required, the server indicates readiness by
+    /// sending a AUTH_RESPONSE message).
     Ready,
+
+    ///  Indicates that the server requires authentication, and which authentication
+    /// mechanism to use.
+
+    /// The authentication is SASL based and thus consists of a number of server
+    /// challenges (AUTH_CHALLENGE) followed by client responses (AUTH_RESPONSE).
+    /// The initial exchange is however bootstrapped by an initial client response.
+    /// The details of that exchange (including how many challenge-response pairs
+    /// are required) are specific to the authenticator in use. The exchange ends
+    /// when the server sends an AUTH_SUCCESS message or an ERROR message.
+    ///
+    /// This message will be sent following a STARTUP message if authentication is
+    /// required and must be answered by a AUTH_RESPONSE message from the client.
     Authenticate,
+
+    /// Indicates which startup options are supported by the server. This message
+    /// comes as a response to an OPTIONS message.
     Supported,
+
+    /// The result to a query (QUERY, PREPARE, EXECUTE or BATCH messages).
+    /// It has multiple kinds:
+    /// - Void: for results carrying no information.
+    /// - Rows: for results to select queries, returning a set of rows.
+    /// - Set_keyspace: the result to a `USE` statement.
+    /// - Prepared: result to a PREPARE message.
+    /// - Schema_change: the result to a schema altering statement.
     Result,
+
+    /// An event pushed by the server. A client will only receive events for the
+    /// types it has REGISTER-ed to. The valid event types are:
+    /// - "TOPOLOGY_CHANGE": events related to change in the cluster topology.
+    ///   Currently, events are sent when new nodes are added to the cluster, and
+    ///   when nodes are removed.
+    /// - "STATUS_CHANGE": events related to change of node status. Currently,
+    ///   up/down events are sent.
+    /// - "SCHEMA_CHANGE": events related to schema change.
+    ///   The type of changed involved may be one of "CREATED", "UPDATED" or
+    ///   "DROPPED".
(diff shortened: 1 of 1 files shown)


scylla-cql/src/value.rs
```diff
@@ -812,210 +847,271 @@ impl TryInto<time_03::Time> for CqlTime {
...
+    /// if the value is null. The length of the tuple is part of its CQL type.
     Tuple(Vec<Option<CqlValue>>),
+    /// Universally unique identifier (UUID) of any version.
     Uuid(Uuid),
+    /// Arbitrary-precision integer.
     Varint(CqlVarint),
+    /// A vector of CQL values of the same type.
+    /// The length of the vector is part of its CQL type.
     Vector(Vec<CqlValue>),
 }
```
Lorak-mmk, line 933: Unrelated to this PR: Vector type is a good use case for array deserialization that we discussed not long ago.

scylla-cql/src/utils/mod.rs
```diff
@@ -1 +1,4 @@
+//! Various utilities.
+// TODO(2.0): Deconstruct this module by extracting submodules to the top level.
+
```
Lorak-mmk, line 3: Why? And what modules? There is only parse here.

scylla-cql/src/frame/protocol_features.rs
```diff
@@ -1,22 +1,53 @@
+//! Implementation and negotiation of extensions to the CQL protocol.
+
 use std::borrow::Cow;
 use std::collections::HashMap;
 
 const RATE_LIMIT_ERROR_EXTENSION: &str = "SCYLLA_RATE_LIMIT_ERROR";
+/// The extension used to add metadata for LWT optimization.
 pub const SCYLLA_LWT_ADD_METADATA_MARK_EXTENSION: &str = "SCYLLA_LWT_ADD_METADATA_MARK";
```
Lorak-mmk, line 8: You could describe what this metadata contains, or link to docs of this extension in scylla repo.

scylla-cql/src/frame/response/event.rs
```diff
@@ -1,74 +1,117 @@
...
 // Check triggers because all variants end with "Change".
 // TODO(2.0): Remove the "Change" postfix from variants.
 #[expect(clippy::enum_variant_names)]
 pub enum Event {
+    /// Topology changed.
     TopologyChange(TopologyChangeEvent),
+    /// Status of a node changed.
     StatusChange(StatusChangeEvent),
+    /// Schema changed.
     SchemaChange(SchemaChangeEvent),
```
Lorak-mmk, line 21: You could explain what is a topology change, and what is a status change (UP/DOWN vs ADD/REMOVE).

scylla-cql/src/frame/types.rs
```diff
@@ -16,25 +16,45 @@ use std::str;
...
+    /// This CL allows for some level of failure.
     Quorum = 0x0004,
+    /// _All_ replicas in the cluster must respond. May cause performance issues.
     All = 0x0005,
+    /// Same as [QUORUM](Consistency::Quorum), but confined to the same datacenter as the coordinator.
     #[default]
     LocalQuorum = 0x0006,
+    /// **Write-only**. A simple majority in each datacenter must respond.
     EachQuorum = 0x0007,
+    /// Same as [ONE](Consistency::One), but confined to the local datacenter.
```
Lorak-mmk, line 50: What happens if you use write-only CL on a read request?

Lorak-mmk (changes requested):

scylla-cql/src/value.rs
wprzytula: 100%.

scylla-cql/src/utils/mod.rs
wprzytula: Exactly. I fail to see how the `parse` module is an "util" more than any other module. That's why I suggest getting read of the `utils` module altogether.

scylla-cql/src/deserialize/row.rs
```diff
@@ -12,8 +15,14 @@ use crate::value::{CqlValue, Row};
...
 #[non_exhaustive]
 pub struct RawColumn<'frame, 'metadata> {
+    /// Index of the column in the row.
     pub index: usize,
+
+    /// Specification of the column, including its name and type.
     pub spec: &'metadata ColumnSpec<'metadata>,
+
+    /// Slice of the frame that contains the serialized value of the column.
+    /// None means that the column is null.
```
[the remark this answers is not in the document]
wprzytula: Ehm, it depends if you're actually asking about `Unset` or `Empty`. They can be distinguished from each other and from `null` by the length field in the binary form.
`match len`:
- `0` -> `Empty`,
- `-1` -> `null`,
- `-2` -> `Unset` (serialization-only).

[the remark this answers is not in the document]
Lorak-mmk: Ok, thanks. I am somehow never able to remember this.

scylla-cql/src/value.rs
```diff
@@ -812,210 +847,271 @@ impl TryInto<time_03::Time> for CqlTime {
 /// Represents a CQL Duration value
 #[derive(Clone, Debug, Copy, PartialEq, Eq)]
 pub struct CqlDuration {
+    /// Number of months.
     pub months: i32,
+    /// Number of days.
     pub days: i32,
+    /// Number of nanoseconds.
```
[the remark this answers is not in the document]
Lorak-mmk: Leave it as is.

scylla-cql/src/frame/response/event.rs
Lorak-mmk: If it is documented elsewhere then it is enough

Issue #1177, opened 2025-01-22
Don't wait for remaining speculative retries if plan is exhausted and there is no running fibers


muzarski:
This issue is related to `speculative_execution::execute` function - especially this part in current version of the driver: https://github.com/scylladb/scylla-rust-driver/blob/4a9367c1e1671773a704670de3c6acd089dc70d0/scylla/src/policies/speculative_execution.rs#L201-L214

There is a small issue with this code. If fiber returns  `None` (i.e., plan is exhausted) there still may be some remaining retries. The driver will wait for remaining retries. It is expected behaviour for most scenarios - there may be some other fibers still running, thus we do not return from this function prematurely. However, we could safely return from the function if both of the following conditions are met:
- fiber returns `None` -> plan is empty/exhausted
- there are no other running fibers in the meantime. If there are some, we should wait for their completion.

Second condition could be checked by introducing additional variable that keeps track of number of currently running fibers. From our experience, `speculative_execution::execute`'s logic seems to be error-prone. This is why, in addition to the change mentioned in the issue, we should think of a way to rewrite the logic in a safer manner.

Lorak-mmk:
> Second condition could be checked by introducing additional variable that keeps track of number of currently running fibers. 

No need for new variable, this is just `async_tasks.len()`


If we do this, a bit more tricky scenario is where we first exhaust the plan, but some executions are still pending.
In such case:
- We should not start new executions
- When running executions finish we should return.

I think all of this can be achieved just by setting `retries_remaining` to 0. 

```rust
            res = async_tasks.select_next_some() => {
                if let Some(r) = res {
                    if !can_be_ignored(&r) {
                        return r;
                    } else {
                        last_error = Some(r)
                    }
                } else {
                    retries_remaining = 0
                }
                if async_tasks.is_empty() && retries_remaining == 0 {
                    return last_error.unwrap_or({
                        Err(EMPTY_PLAN_ERROR)
                    });
                }
            }
```

PR #1401, opened 2025-07-24
Fix plan exhaustion handling in SE and introduce SE unit tests.

Lorak-mmk:
Up until now if execution plan was exhausted, we would still wait until we spawned and finished all SE fibers (unless one succeeds or finishes with non-ignorable error).
This is wasteful - if plan was exhausted by one fiber, then any new fiber will also see exhausted plan and fail, since plan is shared.
This PR fixes this behavior. I also introduced tests for it, using clock simulation facilities from Tokio. Thanks to that, those tests run instantly despite `sleep`s in tested code.

As a bonus. I wrote a regression test for an issue we fixed quite some time ago: #1085 
This issue was considered theoretical - we never tried to write a test / reproducer, because it would execute too long (thanks to sleeps), and could be flaky.
Using clock simulation removes those obstacles, and makes writing such test easy.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1177

## Pre-review checklist

### DIFF
--- a/scylla/src/policies/speculative_execution.rs
+++ b/scylla/src/policies/speculative_execution.rs
@@ -180,6 +180,11 @@ where
                     } else {
                         last_error = Some(r)
                     }
+                } else {
+                    // The only case where None is returned is when execution plan was exhausted.
+                    // If so, there is no reason to start any more fibers.
+                    // We can't always return - there may still be fibers running.
+                    retries_remaining = 0;
                 }
                 if async_tasks.is_empty() && retries_remaining == 0 {
                     return last_error.unwrap_or({
--- a/scylla/src/policies/speculative_execution.rs
+++ b/scylla/src/policies/speculative_execution.rs
@@ -144,7 +144,7 @@ const EMPTY_PLAN_ERROR: RequestError = RequestError::EmptyPlan;
 pub(crate) async fn execute<QueryFut, ResT>(
     policy: &dyn SpeculativeExecutionPolicy,
     context: &Context,
-    query_runner_generator: impl Fn(bool) -> QueryFut,
+    mut query_runner_generator: impl FnMut(bool) -> QueryFut,
 ) -> Result<(ResT, Coordinator), RequestError>
 where
     QueryFut: Future<Output = Option<Result<(ResT, Coordinator), RequestError>>>,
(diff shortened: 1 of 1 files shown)


Lorak-mmk:
Note to reviewers: if you have ideas for additional useful tests, I can implement them.

scylla/src/policies/speculative_execution.rs
```diff
@@ -180,6 +180,11 @@ where
                     } else {
                         last_error = Some(r)
                     }
+                } else {
+                    // The only case where None is returned is when execution plan was exhausted.
+                    // If so, there is no reason to start any more fibers.
+                    // We can't always return - there may still be fibers running.
+                    retries_remaining = 0;
```
wprzytula, line 187: :warning: Naming is extremely confusing. Retries are something that occurs within a single speculative fiber, whereas this should be named `spawnable_fibers_remaining` or similarly.
:wrench: I encourage you to adjust naming where possible without breaking changes. Unfortunately, `SpeculativeExecutionPolicy::{max_retry_count,retry_interval}` are set in stone until 2.0. For them, I suggest adding TODOs for 2.0 to fix their naming to avoid name clash with `RetryPolicy` abstractions.

scylla/src/policies/speculative_execution.rs
```diff
@@ -180,6 +180,11 @@ where
...
                 }
                 if async_tasks.is_empty() && retries_remaining == 0 {
                     return last_error.unwrap_or({
```
wprzytula, line 190: :question: Does it make sense to just finish this block of code if `async_tasks.is_empty()` but `retries_remaining > 0`? In such case we're going to unproductively wait for a new fiber to be spawn once the timer fires, or am I mistaken?
:thought_balloon: Actually, `async_tasks.is_empty()` means that there are no other running fibers. So if there are any more fibers that could be started, I guess we should spawn one immediately. WDYT?

scylla/src/policies/speculative_execution.rs
```diff
@@ -195,3 +195,132 @@ where
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    // Important to start tests with paused clock. If starting unpaused, and calling `tokio::time::pause()`, then
+    // things like `sleep` will advance the timer not fully accurately (I have no idea why), causing
+    // few ms added clock advancement at the end of the test.
+    // Starting paused is done with `#[tokio::test(flavor = "current_thread", start_paused = true)]`.
```
wprzytula, line 204: :wrench: Please mention that clock pausing is only doable with `current_thread` executor.
There's no need to specify it explicitly, as tests are by default run with a `current_thread` executor; explicitness may help here notice this fact, though.

:seedling: Actually, I'm now a bit afraid that our tests may avoid some problematic interleavings by limiting execution to a single thread. Perhaps we should run more sophisticated tests on `multi_thread` runtime to increase likelihood of catching some of them in some runs?

scylla/src/policies/speculative_execution.rs
```diff
@@ -195,3 +195,132 @@ where
...
+                        }
+                    }
+                };
+                counter += 1;
+                future
+            }
+        };
+
+        let now = tokio::time::Instant::now();
+        let res = execute(&policy, &EMPTY_CONTEXT, generator).await;
```
wprzytula, line 260: :pick: This function would be more readable if qualified by its module. The mere `execute` sounds vague.

scylla/src/policies/speculative_execution.rs
Lorak-mmk: We can certainly discuss fully changing the naming in 2.0. For now, variable name is consistent with methods on the trait, so I don't see much point in changing it. This is internal driver code, in speculative execution module - it is imo clear that retries here refer to speculative retries, not normal ones.

Also, not related to this PR at all.

scylla/src/policies/speculative_execution.rs
Lorak-mmk: Waiting may make sense - for example if cluster is overloaded. We even have an issue open about adding waiting functionality to RetryPolicy, and we had in-person discussions about that in summit. Imo removing waiting here may be problematic in that context, and also surprising to users (driver would not respect retry_interval).
In any case, it is not related to this PR. If my arguments are not convincing, please open an issue and let's discuss it further - maybe your idea is good and I just don't see it yet.

scylla/src/policies/speculative_execution.rs
Lorak-mmk: Possibly better testing strategy is to utilize time pausing (which makes clocks deterministic) + setting rng seed on Tokio runtime. We can test with multiple seeds to have more interleavings.
For integration tests, this won't be fully deterministic, because of Scylla. We could create a test-util that simulates Scylla, and make driver "connect" to it directly, eliminating the need for IO. Then the tests should become fully deterministic.

wprzytula: What you've written is the opposite of my point. I want the tests to be *less deterministic* in order to catch errors that could arise in (nondeterministic in nature) normal operating of the driver. I want to achieve that by enabling multithreaded tokio runtime. Does your suggestion address this problem and my idea, or is it completely parallel? I lack glue between our approaches.

scylla/src/policies/speculative_execution.rs
Lorak-mmk: Is `super::execute` fine?

Lorak-mmk: Changed to `super::execute`, lmk if its fine.

scylla/src/policies/speculative_execution.rs
Lorak-mmk: The request is retried (speculatively, so without waiting for the result of previous attempt).

scylla/src/policies/speculative_execution.rs
Lorak-mmk: Unrelated to this PR, please open a separate issue and continue the discussion there.

scylla/src/policies/speculative_execution.rs
wprzytula: > Will you add the TODOs I mentioned?
ping @Lorak-mmk

scylla/src/policies/speculative_execution.rs
wprzytula: > 🔧 Please mention that clock pausing is only doable with `current_thread` executor.

ping @Lorak-mmk

wprzytula (changes requested):

Lorak-mmk:
- Added a todo in SE trait
- Mentioned that clock pausing only works in current thread executor.

wprzytula (approved):

Issue #1402, opened 2025-07-25
Scylla Causing Panic in itertools


chris-pulseprime:
I am running scylla 1.3.0, and currently getting a panic from scylla's use of itertools:

WARN  tablets.rs:418  Nodes (71ec3177-6d5d-47ee-a6d6-e5e8af1ca6f0) listed as replicas for a tablet {ks: global_ks, table: mktdata_log_candle, range: [-9223372036854775807, -4611686018427387905]} are not present in ClusterState.known_peers, despite topology refresh. Removing problematic tablet.

ERROR runtime/lib.rs:178  thread 'tokio-runtime-worker' panicked at 'FormatWith: was already formatted once', /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/itertools-0.14.0/src/format.rs:95:21

I am not using or importing itertools directly. So far I have gone through my workspace with cargo tree to ensure all versions of scylla are 1.3.0 and all versions of itertools are 0.14.0, by updating other libraries using itertools. However, the panic still occurs.  

The panic in itertools is caused by attempting to take the inner value of the iterator more than once. This scenario looks to be occurring either at this format call, or something trying to run format again on the same iterator before it's released:
https://github.com/scylladb/scylla-rust-driver/blob/main/scylla/src/routing/locator/tablets.rs#L418-L420

I am unable to identify where the second take could be happening, but this is currently panicking every application start up and prevents us from running the application. I've tried downgrading scylla versions to 0.14 but that didn't change the behavior.

Lorak-mmk:
We do have a test that causes this warning to be triggered: look at `table_maintenance_tests`. I just make sure that it runs, and that it triggers the error:
```
RUST_LOG=trace cargo test  -p scylla table_maintenance_tests -- --no-capture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.08s
     Running unittests src/lib.rs (target/debug/deps/scylla-c62e0f13a760950c)

running 1 test
2025-07-25T15:36:49.304661Z  INFO scylla::routing::locator::tablets::tests: Test case 0
2025-07-25T15:36:49.304699Z  INFO scylla::routing::locator::tablets::tests: Test case 1
2025-07-25T15:36:49.304713Z  INFO scylla::routing::locator::tablets::tests: Test case 2
2025-07-25T15:36:49.304720Z  INFO scylla::routing::locator::tablets::tests: Test case 3
[...]

PR #1404, opened 2025-07-27
Fix panic on double formatting in traces

Lorak-mmk:
`Itertools::format` has a dangerous footgun: the result panics when it is formatted more than once.
It can happen when, for example, someone uses 2 tracing subscribers (each will format the message).
This PR:
- introduces new alternative, `safe_format` using extension trait. This alternative requires `Clone` bound on the iterator,
which allows multiple formatting.
- Replaces all other usages of `format`.
- Changes `setup_tracing` to perform double formatting, and thus prevent regeressions.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1402

## Pre-review checklist

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,4 @@
 avoid-breaking-exported-api = false
+disallowed-methods = [
+    { path = "itertools::Itertools::format", reason = "Footgun: panics on double formatting.", replacement = "crate::utils::safe_format::IteratorSafeFormatExt::safe_format" }
+]
--- a/scylla-cql/src/utils/safe_format.rs
+++ b/scylla-cql/src/utils/safe_format.rs
@@ -12,6 +12,7 @@ where
     <I as Iterator>::Item: Display,
 {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        #[allow(clippy::disallowed_methods)]
         self.cloneable_iter.clone().format(self.sep).fmt(f)
     }
 }
@@ -22,6 +23,7 @@ where
     <I as Iterator>::Item: Debug,
 {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        #[allow(clippy::disallowed_methods)]
         self.cloneable_iter.clone().format(self.sep).fmt(f)
     }
 }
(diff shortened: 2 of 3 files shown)


Lorak-mmk:
@wprzytula Maybe this bug warrants releasing 1.3.1, wdyt?

scylla/src/observability/driver_tracing.rs
```diff
@@ -176,5 +177,5 @@ fn partition_key_displayer<'ps, 'res, 'spec: 'ps>(
         Ok(c) => Either::Left(c),
         Err(_) => Either::Right("<decoding error>"),
     })
-    .format(", ")
+    .safe_format(", ")
 }
```
wprzytula, line 181: :question: This is the only spot that I've found where the cloning of the iterator may allocate: some `CqlValue`s are allocated (e.g., collections and UDTs), so `Clone`ing them requires allocations and is quite heavyweight. Can we solve this particular spot some other way, so that we don't pay this cost?

:thought_balloon:  One idea that I had was to make use of a lending iterator here (think `CassRowIterator` in CPP-Rust Driver): for each value, it deserializes it to a single-slot buffer it holds, and then returns a (mutable) reference to the held value. Of course, this would not work with `Itertools::format`, as it requires `Iterator`, which this lending iterator could not implement.

:seedling: I've thought that maybe this is not important, because displaying partition key values is obviously something optional, but... **no!** `partition_key_displayer()` is called on the critical path, in `Session::execute()` and `QueryPager::new_for_prepared_statement()`, so actually this code **is important**.
And this makes me wonder why we don't have driver tracing functionalities opt-in behind a feature. I'll open an issue about this.

piodul: > some `CqlValue`s are allocated (e.g., collections and UDTs), so `Clone`ing them requires allocations and is quite heavyweight. Can we solve this particular spot some other way, so that we don't pay this cost?

Just a nitpick: I don't see how cloning the iterator would clone some CqlValues (unless `pk_values_iter` keeps some CqlValues internally, but considering the `pk_values_iter`'s type this sounds very unlikely). What will cause more allocations is iterating - with some double-formatting subscribers you would have to clone the iterator and consume both the original and the clone instead of just the original.

piodul (commented):
If `Itertools::format` is such a footgun, perhaps you could forbid its use in the project? https://rust-lang.github.io/rust-clippy/master/#disallowed_methods

piodul (approved):

PR #1413, opened 2025-08-01
Added serde Derive Traits to Scylla Value Types.

### LINKED ISSUE #1331: Add serde feature to allow reuse of some value types directly.
Hello,

It would be very useful to have serde's Serialize and Deserialize traits derived for some Cql value types like CqlDuration and Counter.

I could make a pull request with an implementation. I would would like to hear other opinions on this. Maybe I'm the only one seeing this as an issue.

didoloan:
## Changes

Added serde's Serialize and Deserialize derive to some scylla value types.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1331

### DIFF
--- a/scylla-cql/src/value.rs
+++ b/scylla-cql/src/value.rs
@@ -4,10 +4,10 @@
 use std::net::IpAddr;
 use std::result::Result as StdResult;
 
-use thiserror::Error;
-use uuid::Uuid;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
+use thiserror::Error;
+use uuid::Uuid;
 
 use crate::deserialize::value::DeserializeValue;
 use crate::deserialize::value::{
--- a/scylla-cql/src/value.rs
+++ b/scylla-cql/src/value.rs
@@ -35,7 +35,7 @@ pub struct ValueOverflow;
 pub struct Unset;
 
 /// Represents an counter value
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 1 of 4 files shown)


scylla-cql/src/value.rs
```diff
@@ -33,6 +35,7 @@ pub struct ValueOverflow;
 pub struct Unset;
 
 /// Represents an counter value
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
wprzytula, line 38: :pick: you've imported both traits, so no need for the `serde` prefix.

wprzytula (changes requested):

wprzytula:
Please follow the procedure described in `CONTRIBUTING.md` regarding MSRV to fix the CI check.

scylla-cql/src/value.rs
didoloan: Sorry about that. Fixed.

wprzytula (approved):

Lorak-mmk (changes requested):
Sorry for the delay! Generally looks good. There are some changes needed.
- Please rebase on main. We no longer have Cargo.lock.msrv thanks to Rust 2024, so your last commit becomes redundant.
- Every commit should be an atomic change that passes checks and tests, and has proper description. Your third and fourth commit are fixing issues introduced by eralier commits. Instead of having new commits for that, the earlier commits should be edited to not introduce those issues in the first place. This approach also makes the review easier - we can review each commit separately, and not see issues that are fixed in some later commit.
- I'm not familiar with serde, so it is not obvious to me if the functionality you introduced works properly. Please add unit tests that demonstrate what this PR allows the user to do.

Lorak-mmk:
Continued in https://github.com/scylladb/scylla-rust-driver/pull/1912

Issue #1420, opened 2025-08-12
Potential race condition in CachingSession


Lorak-mmk:
Getting and removing elements from the cache (done in `CachingSession::add_prepared_statement_owned`) goes roughly as follows:
1. Try to get element from the map
2. If it is not present, prepare statement
3. Check if cache is full
4. If cache is full, get first statement from it
5. Remove the statement
6. Insert new statement

I think the following race might be possible:
- There are 2 tasks
- Both try to use different queries that are not in the cache
- Both notice that, and prepare the statement
- Both notice that cache is full
- Both get the same statement string to remove from the cache
- Both remove this statement from the cache
- Both insert their queries. Now there are `max_capacity + 1` elements in the cache

Potential consequence: we check if `self.max_capacity == self.cache.len()`, so if cache overflows we will never again remove from it, causing a leak.
Switching to `>=` will make the impact much less severe. TBH this is not very serious - I doubt that any reasonable app makes so many different requests that this leak would be problematic.

PR #1423, opened 2025-08-13
CachingSession: Mostly fix a race condition

Lorak-mmk:
A race can happen when multiple calls with different statements (none of which are in cache) are executed at the same time, with cache full.

It is possible that all of them will remove the same entry from the cache, and then add their own entries. Then size of the cache will exceed self.max_capacity, so we will never remove from cache again.

First fix that I thought of is to change "==" to "<=". With that fix, we would still be able to remove elements from the map after race happens. Note that we only remove an element when inserting a new element, so this would still never decrease the cache overflow unless 2 cache misses with the same query happen at the same time. So the memory leak would still happen, but much slower.

So in addition to that I changed  "if cache full: remove" to "while cache full: remove". In some cases it could lead to some queries getting evicted for no reason, hurting performance. There is also slim possibility of some thread getting starved for a bit. Those issues seem not very probable, so maybe it is better to risk them than the memory leak? I'm honestly not sure. Maybe we should remove the loop and only keep "<=".

I also thought of checking cache size in the happy path, but len() is not a simple operation in dashmap - it goes through shards and sums their lengths. I didn't benchmark it, but I'm afraid it could hurt performance.

Maybe in the future we should research how dedicated cache libraries are approaching this?

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1420

## Pre-review checklist

- [ ] ~~I added relevant tests for new features and bug fixes.~~ - Writing tests for that seems quite difficult...

### DIFF
--- a/scylla/src/client/caching_session.rs
+++ b/scylla/src/client/caching_session.rs
@@ -224,7 +224,15 @@ where
                 stmt
             };
 
-            if self.max_capacity == self.cache.len() {
+            // This loop was added to prevent a race condition (+ memory leak).
+            // When 2 threads enter this because cache is full, they may remove the same element,
+            // but add different ones. Then we get cache overflow.
+            // If we don't have a loop here, then this overflow would never disappear during typical
+            // operation of caching session.
+            // The loop has downsides: it could evict more entries than strictly necessary, or starve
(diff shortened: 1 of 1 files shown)


dkropachev:
Probably it worth to throw some warning when it cache is full.

PR #1427, opened 2025-09-07
Remove unnecessary `'static` bounds

cstyles:
As far as I can tell these bounds are unnecessary.

I ran the static checks and they "failed" but they're also failing on main for me so I don't think I introduced the errors. CI is also failing but I see the same failures on other branches so again, I don't think my changes caused them.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -702,7 +702,7 @@ impl QueryPager {
     /// It only allows deserializing owned types, because [Stream] is not lending.
     /// Begins with performing type check.
     #[inline]
-    pub fn rows_stream<RowT: 'static + for<'frame, 'metadata> DeserializeRow<'frame, 'metadata>>(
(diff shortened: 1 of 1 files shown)


cstyles:
> So if I understand correctly, _even though the deserialized data will always be owned_ (because `TypedRowStream` doesn't support borrowed deserialization yet), **there may still be benefit in allowing non-`'static` `RowT`**, because it may be an enum that has a non-`'static` variant apart from the `'static` one that is being deserialized into.

Exactly, that's what I was trying to show with the `Cow` example. I don't think the benefit is all that big but the alternative (leaving the unnecessary `'static` bound in) has no benefit that I can discern. So we might as well remove it.

wprzytula:
@cstyles Please put the above code example (the second one, with `Cow`) in the commit message for future readers.

wprzytula (approved):
Please rebase on main.

PR #1428, opened 2025-09-08
Fix clippy 1.89

Lorak-mmk:
As usual, clippy started to fail because of newly stabilized lints.
This PR fixes that.

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/macros/hygiene.rs
+++ b/scylla/tests/integration/macros/hygiene.rs
@@ -190,6 +190,7 @@ macro_rules! test_crate {
             _scylla::DeserializeValue, _scylla::SerializeValue, PartialEq, Debug,
         )]
         #[scylla(crate = _scylla)]
+        #[allow(dead_code)] // TODO: Change to expect after bumping MSRV to 1.89
         struct TestStructByName {
             a: ::core::primitive::i32,
             #[scylla(allow_missing)]
@@ -208,6 +209,7 @@ macro_rules! test_crate {
             _scylla::DeserializeValue, _scylla::SerializeValue, PartialEq, Debug,
         )]
         #[scylla(crate = _scylla, forbid_excess_udt_fields)]
+        #[allow(dead_code)] // TODO: Change to expect after bumping MSRV to 1.89
         struct TestStructByNameStrict {
             a: ::core::primitive::i32,
             #[scylla(allow_missing)]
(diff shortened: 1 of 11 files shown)


scylla/tests/integration/macros/hygiene.rs
```diff
@@ -334,6 +334,32 @@ macro_rules! test_crate {
...
+            use _scylla::deserialize::row::DeserializeRow;
+            use ::std::todo;
+            #[allow(clippy::diverging_sub_expression)]
+            fn deserialize_value<'x, T: DeserializeValue<'x, 'x>>() {
+                let _ = <T as DeserializeValue<'x, 'x>>::deserialize(todo!(), todo!());
+            }
+            #[allow(clippy::diverging_sub_expression)]
+            fn deserialize_row<'x, T: DeserializeRow<'x, 'x>>() {
+                let _ = <T as DeserializeRow<'x, 'x>>::deserialize(todo!());
+            }
```
wprzytula, line 351: :question: Why `allow` instead of `expect`?

scylla/tests/integration/macros/hygiene.rs
```diff
@@ -334,6 +334,32 @@ macro_rules! test_crate {
...
+            deserialize_value::<TestStructByName>();
+            deserialize_value::<TestStructByNameStrict>();
+            deserialize_value::<TestStructOrdered>();
+            deserialize_value::<TestStructOrderedStrict>();
+            deserialize_value::<TestStructOrderedSkipped>();
+            deserialize_value::<TestStructOrderedStrictSkipped>();
+            deserialize_value::<TestStructOrderedStrictSkipped>();
+            // For some reson, TestRowByName doesn't trigger the warning.
```
wprzytula, line 359: :pick: `reson` -> `reason`

wprzytula: :thought_balloon: `TestRowByName` is the only struct that derives `SerializeRow` in `match_by_name` flavor. My guess is that the problem is caused by creating a `pub struct` as part of the derive process. This somehow hints the compiler that the struct is used, as it shows up in the public API (by some trait impl that binds that public struct with the struct `SerializeRow` is being derived for).

Lorak-mmk: > My guess is that the problem is caused by creating a pub struct as part of the derive process. 

I changed the macro so that the generated struct is not pub. It did not fix the issue.

wprzytula (approved):

PR #1434, opened 2025-09-10
Workaround for cargo-semver-checks bug

Lorak-mmk:
Fixing the bug in upstream is difficult - author is reluctant that the patch will impose maintenance burden. For now I fixed in in a fork, let's use it to get rid of the breaking label from all new PRs.

## Pre-review checklist

### DIFF
--- a/.github/workflows/semver_checks.yml
+++ b/.github/workflows/semver_checks.yml
@@ -29,7 +29,7 @@ jobs:
     # could do us a lot of harm. This is also the reason that there are
     # 2 jobs - it's so that it's not possible to take over a job that
     # has permissions.
-    permissions: {} 
+    permissions: {}
     timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


.github/workflows/semver_checks.yml
```diff
@@ -58,9 +58,7 @@ jobs:
     - name: Setup rust toolchain
       uses: actions-rust-lang/setup-rust-toolchain@v1
     - name: Install semver-checks
-      uses: taiki-e/install-action@v2
-      with:
-        tool: cargo-semver-checks
+      run: cargo install --git https://github.com/Lorak-mmk/cargo-semver-checks.git cargo-semver-checks --branch main
```
wprzytula, line 61: :wrench: Please add a comment that explains what we are doing and why (temporary solution, fork because ...).

wprzytula (changes requested):

wprzytula (approved):

PR #1442, opened 2025-10-02
PreparedStatement: put id and lwt in shared state

Lorak-mmk:
Small possible improvement that I noticed when working on metadata id extension.

PreparedStatementSharedData was initially introduced for performance reasons, specifically to make cloning PreparedStatement cheaper. Now that we have it, I think we should put all shared state there. If we don't:
- It is confusing. If I see shared state, and I see that e.g. `id` is not a part of it, then I automatically begin to wonder if its not shared, of if its somehow mutable.
- (Minor) More stuff in shared -> cheaper cloning. Difference, if any, should be tiny.

prepare_tracing_ids unfortunately can't be moved because it is pub.

## Pre-review checklist

### DIFF
--- a/scylla/src/statement/prepared.rs
+++ b/scylla/src/statement/prepared.rs
@@ -96,6 +96,7 @@ use crate::routing::partitioner::{Partitioner, PartitionerHasher, PartitionerNam
 pub struct PreparedStatement {
     pub(crate) config: StatementConfig,
     /// Tracing IDs of all queries used to prepare this statement.
+    // TODO(2.0): Unpub this, move this to PreparedStatementSharedData, and expose a getter.
     pub prepare_tracing_ids: Vec<Uuid>,
 
     shared: Arc<PreparedStatementSharedData>,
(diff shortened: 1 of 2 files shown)


scylla/src/statement/prepared.rs
```diff
@@ -98,30 +98,28 @@ pub struct PreparedStatement {
     /// Tracing IDs of all queries used to prepare this statement.
     pub prepare_tracing_ids: Vec<Uuid>,
```
dkropachev, line 99: Is this good idea to give full access to `prepare_tracing_ids` ? don't see good reason for giving a write access to it.
Does it make sence to have a getter for it and mark it as depricated, to be removed in next major/minor ?

Lorak-mmk: Yes it would be good to move it to shared data, and not have it as public field. I'll add a todo.

dkropachev: https://github.com/scylladb/scylla-rust-driver/issues/1443

Lorak-mmk:
I've confirmed using github action logs that failing run has 1.89 preinstalled, while succeeding run has 1.90 preinstalled. Why that impacts anything I still have no idea. 
rustup version on both is the same (1.28.2)

Lorak-mmk:
The job passed with Rust 1.89 installed on the runner, so the workaround works.

dkropachev (approved):

wprzytula (approved):

Issue #1448, opened 2025-10-15
query/execute_iter can panic (or worse) if there is schema change between page queries


Lorak-mmk:
TypedRowStream performs type check once, on creation. Even though we don't support Scylla's metadata extension yet, it is still possible for this to not be enough.
If we disable result metadata caching (and it is the default), and the schema changes at some point, the requests will start returning result metadata that reflects the new schema.
TypeRowStream will then call DeserializeRow::deserialize on rows of such responses, without doing type check, which can have all kinds of bad consequences, like panic or even data corruption.

The solution is to type check each page.

Reproducer:
```rust
#[tokio::test]

async fn test_iter_methods_when_altering_table() {
    let session = create_new_session_builder().build().await.unwrap();
    let ks = unique_keyspace_name();

    session.ddl(format!("CREATE KEYSPACE IF NOT EXISTS {ks} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}")).await.unwrap();
    session
        .ddl(format!(
            "CREATE TABLE IF NOT EXISTS {ks}.t (a int, b int, d int, primary key (a, b))"
        ))
        .await
        .unwrap();

    let insert_stmt = session
        .prepare(format!("INSERT INTO {ks}.t (a, b, d) VALUES (?, ?, ?)"))
        .await
        .unwrap();
    // First lets insert some data
    for a in 0..10 {
        for b in 0..10 {
            session
                .execute_unpaged(&insert_stmt, (a, b, 1337))
                .await
                .unwrap();
        }
    }

    let mut select_stmt = session
        .prepare(format!("SELECT * FROM {ks}.t",))
        .await
        .unwrap();
    select_stmt.set_page_size(10);
    select_stmt.set_use_cached_result_metadata(false);
    let pager = session.execute_iter(select_stmt, &[]).await.unwrap();
    let mut stream = pager.rows_stream::<(i32, i32, Option<i32>)>().unwrap();

    // Let's fetch a few pages, but not all.
    for _ in 0..50 {
        let _row = stream.next().await.unwrap().unwrap();
    }

    session
        .query_unpaged(format!("ALTER TABLE {ks}.t ADD c text"), &())
        .await
        .unwrap();

    // With the bug, the code panics!
    // At some point, requests will return pages with new schema.
    // It contains new column, and the new schema was not type checked.
    // DeserializeRow::deserialize impl will panic because invariants that should
    // be enforce by type check are violated.
[...]

Issue #1374, opened 2025-06-11
Close prepared select statements invalidation loopholes


wprzytula:
Quoting the umbrella drivers issue:
> Prepared statements invalidation used to be completely broken, you can read details on it [here](https://github.com/scylladb/scylladb/issues/20860) Recently core merged [PR](https://github.com/scylladb/scylladb/pull/23292) that fixes problems for select statements. After that PR, if driver supply `SCYLLA_USE_METADATA_ID` at the startup, when statements is prepared server hands out result queries metadata hash. When driver executes this query it sends metadata id alongside with the request. Server checks driver metadata id and what it has on its side, if there is any difference it will send new metadata and metadata id. Driver suppose to pick it up and update metadata and metadata id on it's side, before deserializing response, solving issue with having outdated metadata on driver side. Unfortunately it is solved only for SELECT statements, UPDATE/INSERT will be solved on separate occasion.
> 
> Core changes are in, we need to start working on driver side.

[This commit from the Core](https://github.com/scylladb/scylladb/pull/23292/commits/086df245558b15b4058680ee6b6d6e2fc7912527) and [this from Python Driver](https://github.com/scylladb/python-driver/pull/457/commits/8aba164564678c133267fd24b8590c68aa13f5a2) can serve as a reference.

## To Be Done:

- parse `SCYLLA_USE_METADATA_ID` protocol extension from SUPPORTED frame;
- if present in SUPPORTED, send it in the STARTUP frame;
- if the extension is negotiated:
  - store the metadata ID in ResultMetadata
  - read `METADATA_CHANGED` (`=3`) flag in result metadata;
  - if the flag is set, read the new metadata ID (short bytes), which is put between the paging state and the global tables spec;
  - use the new metadata for deserialization instead of the old one;
  - [*] replace the old metadata with the new one, if received.

### Metadata mutability

If the [*] is not done, we will have a correct implementation that no longer faces the risk of using outdated metadata for deserialization.
However, we will have to receive the updated metadata upon every prepared statement execution, which may be costly for SELECTs with little data returned. It needs to be considered if we can (and should) make result metadata mutable in the `PreparedStatement`.

Considerations:
- `ResultMetadata` is already kept behind an `Arc`. We could use `ArcSwap` instead to get cheaply-synchronized mutability.
- repreparation could (and should) update the `ResultMetadata`.

## Prepared Metadata Issue

[...]

PR #1449, opened 2025-10-15
Typecheck each page in execute/query_iter

Lorak-mmk:
As described in the issue, type check should be performed on each page.

This is done by:
1. Changing ColumnIterator to allow type checking it
2. Returning a bool from QueryPager::next that indicates if new page was received.
3. Using this bool and ColumnIterator::next to perform type check on each page in TypedRowStream.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1448
Related to: https://github.com/scylladb/scylla-rust-driver/issues/1374

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -695,6 +695,12 @@ impl QueryPager {
     /// This is automatically called upon transforming [QueryPager] into [TypedRowStream].
     // Can be used with `next()` for manual deserialization.
     #[inline]
+    #[deprecated(
+        since = "1.4.0",
+        note = "Type check should be performed for each page, which is not possible with public API.
+Also, the only thing user can do (rows_stream) will take care of type check anyway.
+If you are using this API, you are probably doing something wrong."
+    )]
     pub fn type_check<'frame, 'metadata, RowT: DeserializeRow<'frame, 'metadata>>(
         &self,
     ) -> Result<(), TypeCheckError> {


Lorak-mmk:
Changed the approach. I added type checking method to ColumnIterator, which allows to keep QueryPager::next free of typechecking. I like this approach much more.
Thanks @wprzytula for the suggestion.

scylla/tests/integration/session/pager.rs
```diff
@@ -149,3 +149,67 @@ async fn test_iter_methods_with_modification_statements() {
...
+            "CREATE TABLE IF NOT EXISTS {ks}.t (a int, b int, d int, primary key (a, b))"
+        ))
+        .await
+        .unwrap();
+
+    let insert_stmt = session
+        .prepare(format!("INSERT INTO {ks}.t (a, b, d) VALUES (?, ?, ?)"))
+        .await
+        .unwrap();
+    // First lets insert some data
```
wprzytula: :pick:

wprzytula:
```suggestion
    // First let's insert some data
```

wprzytula (changes requested):
Good job! This indeed fixes an important bug, and typecheck on `ColumnIterator` makes sense in binding layers of other drivers.

scylla/src/client/pager.rs
```diff
@@ -691,6 +695,12 @@ impl QueryPager {
     /// This is automatically called upon transforming [QueryPager] into [TypedRowStream].
     // Can be used with `next()` for manual deserialization.
     #[inline]
+    #[deprecated(
+        since = "1.4.0",
+        note = "Type check should be performed for each page, which is not possible with public API.
+        Also, the only thing user can do (rows_stream) will take care of type check anyway.
+        If you are using this API, you are probably doing something wrong."
```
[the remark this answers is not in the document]
Lorak-mmk: Removing whitespaces at line beginnings makes sense. Adding "\n" doesn't - newlines are part of the string.
Deprecation currently looks like this:
image

PR #1455, opened 2025-10-25
integration: enable vector type tests for ScyllaDB

wprzytula:
ScyllaDB now supports vector type, so we can enable the vector type tests that were previously ignored for ScyllaDB.

**However**, there's a bug in Vector implementation (https://github.com/scylladb/scylladb/issues/26704), which makes the tests fail. This PR shall then be merged once the bug is fixed and a new stable release with the fix is available at docker, so the CI gets it.

## Pre-review checklist


wprzytula:
Rebased on main.

Lorak-mmk (approved):

PR #1463, opened 2025-10-30
Support result metadata id protocol extension

Lorak-mmk:
This PR introduces support for Scylla's [result metadata id extension](https://github.com/scylladb/scylladb/blob/master/docs/dev/protocol-extensions.md#negotiate-sending-metadata-id)
This extension is a CQLv4 backport of CQLv5 mechanism of providing drivers with new result metadata when the one they use becomes stale.
Changes in the protocol:
- Result metadata (send in Rows and Prepared response) got a new flag (`Metadata_changed`). If this flag is set, a new field `new_metadata_id` is present.
- PREPARED response and EXECUTE request also contain new obligatory `new_metadata_id` field. 

The idea is that driver sends its currently known result metadata id in execute, server checks if it is stale, and if so sends a new one in Results:Rows.

Support for this protocol extension is added. For that, changes in scylla-cql were required, because now parsing / serializing protocol messages depends on ProtocolFeatures, and we did not support that.
Because of backwards compatibility, I could not remove structs from scylla-cql. Instead I had to duplicate some of them.
Driver-level support utilizes [recently added](https://github.com/scylladb/scylla-rust-driver/pull/1445) capability to update result metadata on PreparedStatement.

Commit that are important and should be reviewed more carefully:
- `scylla-cql: Parsing metadata id` adds protocol support on Response side, to both `Prepared` and `Result:Rows`
- `scylla-cql: Add ExecuteV2` adds protocol support on the sending side by introducing new 
- `Connection: Update PreparedStatement metadata`, `Update PreparedStatement result id after reprepare` and `Connection: Send metadata id when necessary` enable driver-side support for the extension.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1374

## Pre-review checklist

Closes [DRIVER-149](https://scylladb.atlassian.net/browse/DRIVER-149)

[DRIVER-149]: https://scylladb.atlassian.net/browse/DRIVER-149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scylla/tests/integration/statements/prepared_metadata_extension.rs
+++ b/scylla/tests/integration/statements/prepared_metadata_extension.rs
@@ -1,0 +1,455 @@
+use std::collections::HashMap;
+use std::net::SocketAddr;
+use std::sync::Arc;
+use std::task::Poll;
+
+use futures::StreamExt;
+use itertools::Itertools;
+use scylla::client::session::Session;
+use scylla::client::session_builder::SessionBuilder;
+use scylla::cluster::metadata::{ColumnType, NativeType};
+use scylla::errors::DbError;
+use scylla::frame::response::result::{ColumnSpec, TableSpec};
+use scylla::statement::prepared::PreparedStatement;
+use scylla_cql::frame::protocol_features::ProtocolFeatures;
+use scylla_cql::frame::request::DeserializableRequest;
+use scylla_cql::frame::request::execute::ExecuteV2;
+use scylla_cql::frame::response::result::Result;
+use scylla_cql::frame::response::{Response, Supported};
+use scylla_cql::frame::{parse_response_body_extensions, types};
+use scylla_proxy::{
+    Condition, ProxyError, Reaction, RequestFrame, RequestReaction, RequestRule, ResponseFrame,
+    ResponseReaction, ResponseRule, RunningProxy, WorkerError,
+};
+
+use crate::utils::{
+    PerformDDL, fetch_negotiated_features, setup_tracing, test_with_3_node_cluster,
+    unique_keyspace_name,
+};
+
+async fn prepare_schema_and_data(session: &Session, ks: &str) {
+    tracing::info!("Preparing schema and data");
+    session
+        .ddl(format!(
+            "CREATE KEYSPACE IF NOT EXISTS {ks} WITH REPLICATION =
+        {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}"
+        ))
+        .await
+        .unwrap();
+    session
+        .ddl(format!(
+            "CREATE TABLE IF NOT EXISTS {ks}.t (a int, b Text, primary key (a))"
+        ))
+        .await
+        .unwrap();
+    session
+        .query_unpaged(
+            format!("INSERT INTO {ks}.t (a, b) VALUES (?, ?)"),
+            (1, "abc"),
+        )
+        .await
+        .unwrap();
+    tracing::info!("Preparing schema and data finished");
+}
+
+async fn drop_schema(session: &Session, ks: &str) {
+    session.ddl(format!("DROP KEYSPACE  {ks}")).await.unwrap();
+}
+
+fn assert_old_schema(stmt: &PreparedStatement, ks: &str) {
+    use ColumnType::*;
+    use NativeType::*;
+    let guard = stmt.get_current_result_set_col_specs();
+    let table_spec = TableSpec::borrowed(ks, "t");
+
+    assert_eq!(
+        guard.get().as_slice()[0],
+        ColumnSpec::borrowed("a", Native(Int), table_spec.clone())
+    );
+    assert_eq!(
+        guard.get().as_slice()[1],
+        ColumnSpec::borrowed("b", Native(Text), table_spec.clone())
+    );
+    assert_eq!(guard.get().as_slice().len(), 2);
+}
+
+fn assert_new_schema(stmt: &PreparedStatement, ks: &str) {
+    use ColumnType::*;
+    use NativeType::*;
+    let guard = stmt.get_current_result_set_col_specs();
+    let table_spec = TableSpec::borrowed(ks, "t");
+
+    assert_eq!(
+        guard.get().as_slice()[0],
+        ColumnSpec::borrowed("a", Native(Int), table_spec.clone())
+    );
+    assert_eq!(
+        guard.get().as_slice()[1],
+        ColumnSpec::borrowed("b", Native(Text), table_spec.clone())
+    );
+    assert_eq!(
+        guard.get().as_slice()[2],
+        ColumnSpec::borrowed("c", Native(Text), table_spec.clone())
+    );
+    assert_eq!(guard.get().as_slice().len(), 3);
+}
+
+// Creates two sessions, and one `SELECT *` statement for each session.
+// Tests that:
+// - Statements initially have the same metadata.
+// - After alter, statements still have same metadata.
+// - After executing statement from session 1, its metadata is updated.
+// - After that, statement 2 still has old metadata.
+// - For executing statement 2 after that, reprepare is not needed and new metadata is sent.
+async fn perform_test_for_proxy(
+    proxy_uris: &[String; 3],
+    translation_map: &HashMap<SocketAddr, SocketAddr>,
+    mut running_proxy: RunningProxy,
+    features: ProtocolFeatures,
+    apply_prepare_rules: impl FnOnce(&mut RunningProxy),
+    apply_execution_rules: impl FnOnce(&mut RunningProxy),
+) -> RunningProxy {
+    use Condition::*;
+    use scylla_proxy::RequestOpcode::Prepare;
+
+    let session_1 = SessionBuilder::new()
+        .known_node(proxy_uris[0].as_str())
+        .address_translator(Arc::new(translation_map.clone()))
+        .build()
+        .await
+        .unwrap();
+    let session_2 = SessionBuilder::new()
+        .known_node(proxy_uris[0].as_str())
+        .address_translator(Arc::new(translation_map.clone()))
+        .build()
+        .await
+        .unwrap();
+    let ks = unique_keyspace_name();
+    prepare_schema_and_data(&session_1, &ks).await;
+
+    apply_prepare_rules(&mut running_proxy);
+    let query_str = format!("SELECT * FROM {ks}.t WHERE a = ?");
+    tracing::info!("Preparing statement 1");
+    let statement_1 = session_1.prepare(query_str.as_str()).await.unwrap();
+    tracing::info!("Preparing statement 2");
+    let statement_2 = session_2.prepare(query_str.as_str()).await.unwrap();
+
+    running_proxy
+        .running_nodes
+        .iter_mut()
+        .for_each(|node| node.change_request_rules(Some(vec![])));
+
+    // Prepare again, to make sure statement is in cache on all nodes.
+    tracing::info!("Preparing statements again");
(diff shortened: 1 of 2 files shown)


Lorak-mmk:
@wprzytula Doc changes are done, so the only thing missing are tests. I don't think it should be a problem for review.

scylla-cql/src/frame/response/result.rs
```diff
@@ -1027,6 +1087,13 @@ fn deser_result_metadata(
 
     let paging_state = PagingStateResponse::new_from_raw_bytes(raw_paging_state);
 
+    let new_metadata_id = metadata_changed
+        .then(|| {
+            types::read_short_bytes(buf).map_err(ResultMetadataParseError::NewMetadataIdParseError)
+        })
+        .transpose()?
+        .map(|x| cow_bytes::CowBytes::from(x).into_owned());
```
wprzytula: :thought_balloon: So here we allocate if only new metadata id is present. Am I correct that it should happen rarely?

Even if that's the case, I'd love to introduce a new flavor of at least `read_short_bytes` which would operate on a `FrameSlice` instead of plain byte slices. Then the allocation would never be needed here.

scylla-cql/src/frame/response/result.rs
```diff
@@ -1241,10 +1324,22 @@ fn deser_prepared(
         as usize;
     let id: Bytes = buf[0..id_len].to_owned().into();
     buf.advance(id_len);
+    let result_metadata_id = features
+        .scylla_metadata_id_supported
```
wprzytula: :warning: Fallible indexing spotted! Shouldn't this have been fixed in a previous PR? We risk a panic due to invalid indexing IIUC.

scylla-cql/src/frame/request/execute.rs
```diff
@@ -52,7 +54,77 @@ impl DeserializableRequest for Execute<'_> {
...
+        let id = CowBytes::from(types::read_short_bytes(buf)?).into_owned();
+        let result_metadata_id = CowBytes::from(types::read_short_bytes(buf)?).into_owned();
+        let parameters = QueryParameters::deserialize(buf)?;
+
+        Ok(Self {
+            id,
+            result_metadata_id: Some(result_metadata_id),
+            parameters,
+        })
+    }
```
wprzytula: :thinking: What about lifetimes here?
1. You're using an anonymous lifetime parameter. This AFAIU means that `deserialize` must satisfy the arbitrary anonymous lifetime specified in the `impl` block; in particular, `Self` must not borrow from `buf`. Do you agree?
2. If so, then it follows that `'_` must actually be `'static`. Then let's write `'_static` there.
3. Otherwise - if you wanted `deserialize` to allow borrowed deserialization (why not BTW?), then let's make the returned value borrow from `buf`.

scylla-cql/src/frame/request/execute.rs
```diff
@@ -21,6 +21,7 @@ use super::{
 /// CQL protocol-level representation of an `EXECUTE` request,
 /// used to execute a single prepared statement.
 #[cfg_attr(test, derive(Debug, PartialEq, Eq))]
+#[deprecated(since = "1.4.0", note = "Use ExecuteWithResultMetadataId instead")]
```
wprzytula: :wrench: Outdated naming. `ExecuteV2` should be here instead.

scylla/tests/integration/statements/prepared.rs
```diff
@@ -554,8 +555,10 @@ async fn test_skip_result_metadata() {
         }
 
         // Verify that server sends metadata when driver doesn't send SKIP_METADATA flag.
+        // If the result metadata extension is enabled, then driver ignores the setting on PreparedStatement
+        // and always sends the SKIP_METADATA flag, so NO_METADATA should be present iff extension is enabled.
```
wprzytula: :memo: Oh, that's a big thing (ignoring the setting on `PreparedStatement`). I'm leaving this here not to forget about this later. Perhaps you document the change well enough in the following commits.

scylla/src/network/connection.rs
```diff
@@ -900,6 +902,30 @@ impl Connection {
...
+        prepared_statement: &PreparedStatement,
+        query_response: &QueryResponse,
+    ) {
+        if let ResponseWithDeserializedMetadata::Result(ResultWithDeserializedMetadata::Rows(
+            rows_result,
+        )) = &query_response.response
+        {
+            if rows_result.0.metadata().id().is_some()
+                && rows_result.0.metadata().id()
+                    != prepared_statement.get_current_result_metadata().id()
```
wprzytula: :wrench: Can we use `match` with guard for clarity here?

wprzytula: Can `id()` be `None` on `PreparedStatement`? When can this happen? Is it possible that we're here and it's `None`?

scylla/src/network/connection.rs
```diff
@@ -900,6 +902,30 @@ impl Connection {
...
+            {
+                // Metadata in response is either cached metadata extracted from statement, or a new extracted from response.
+                // If the id differs from the one in prepared statment, I see 2 possibilities:
+                // 1. Metadata changed, and was already updated on PreparedStatement by another execution.
+                // 2. Metadata changed, and was not yet updated on PreparedStatement.
+                // We could distinguish those two cases by checking id on cached_metadata, but it is not necessary,
+                // because in both cases we should, or at least may, update the metadata on PreparedStatement.
+                let metadata = rows_result.0.metadata().clone().into_owned();
+                prepared_statement.update_current_result_metadata(Arc::new(metadata));
```
wprzytula: :question: Why should we update the metadata anyway?
Don't we prefer to elide another allocation here in case of a race with another execution?

scylla/src/network/connection.rs
```diff
@@ -985,6 +1013,8 @@ impl Connection {
                     }
                 }
 
+                Self::handle_result_metadata_new_id(prepared_statement, &new_response);
```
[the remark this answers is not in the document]
wprzytula: One idea is to use a `loop` with at most one repetition.

scylla/src/network/connection.rs
```diff
@@ -697,14 +697,21 @@ impl Connection {
...
                 expected_id: previous_prepared.get_id().clone().into(),
-            })
-        } else {
-            Ok(())
+            });
+        }
+
+        let response = raw_prepared.into_response();
+        if response.result_metadata.id().is_some()
+            && previous_prepared.get_current_result_metadata().id() != response.result_metadata.id()
```
wprzytula: ditto: can we use a `match` for clarity?

scylla/src/network/connection.rs
```diff
@@ -933,6 +933,30 @@ impl Connection {
...
+            // Statement may have been prepared on connection that does not have the extension (think:
+            // cluster during upgrade, with mixed versions), and then used on this connection.
+            // What to do in this case? We have to send some id, because it is not optional in the protocol,
+            // so let's send empty id.
+            (Some(metadata), true) => Some(metadata.id().unwrap_or(&[])),
+            (Some(_), false) => None,
+            (None, true) => unreachable!("metadata extension enabled, but no cached metadata."),
+            (None, false) => None,
+        }
+    }
```
wprzytula: :question: Why the `assert` if `unreachable` handles the same case?

scylla/src/network/connection.rs
```diff
@@ -933,6 +933,30 @@ impl Connection {
...
+            // In the first branch we know that metadata extension is enabled, and we have extracted cached metadata.
+            // Can we `unwrap()` the id? I think not, because I see one scenario where it can be None.
+            // Statement may have been prepared on connection that does not have the extension (think:
+            // cluster during upgrade, with mixed versions), and then used on this connection.
+            // What to do in this case? We have to send some id, because it is not optional in the protocol,
+            // so let's send empty id.
+            (Some(metadata), true) => Some(metadata.id().unwrap_or(&[])),
+            (Some(_), false) => None,
+            (None, true) => unreachable!("metadata extension enabled, but no cached metadata."),
+            (None, false) => None,
```
wprzytula: :pick: Let's merge both `false` cases to one, to highlight that absence of the extension makes things simple.

scylla/src/network/connection.rs
```diff
@@ -952,24 +976,32 @@ impl Connection {
             .get_timestamp()
             .or_else(get_timestamp_from_gen);
 
+        let has_metadata_extension = self.features.protocol_features.scylla_metadata_id_supported;
+
+        // If the metadata id extension is supported, there is no point in not caching metadata,
+        // so we ignore the setting on prepared statement.
+        let skip_metadata =
+            prepared_statement.get_use_cached_result_metadata() || has_metadata_extension;
```
wprzytula: :question: Once again: did you remember to update the docs on `PreparedStatement`?

scylla/src/statement/prepared.rs
```diff
@@ -468,7 +491,8 @@ impl PreparedStatement {
     /// after statement preparation and will use it
     /// to deserialize the results of statement execution.
     ///
-    /// This option is false by default.
+    /// This option is false by default for DBs that don't support result metadata id extension.
+    /// For DBs that do (modern Scylla) this option is true, and can't be disabled.
```
wprzytula: :wrench:

scylla-cql/src/frame/response/result.rs
```diff
@@ -516,6 +536,13 @@ impl ResultMetadataHolder {
         }
     }
 
+    pub fn make_owned_arced(&self) -> Arc<ResultMetadata<'static>> {
+        match self {
+            Self::SelfBorrowed(_) => todo!(),
+            Self::SharedCached(metadata) => Arc::clone(metadata),
+        }
+    }
```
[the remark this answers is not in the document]
wprzytula: 💥  It's still there!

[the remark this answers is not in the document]
wprzytula: 💥  It's still there!

scylla-cql/src/frame/request/execute.rs
Lorak-mmk: I did not put much effort into those impls because they are only used by proxy, so I just did the same thing that impl for `Execute` does.
I'll add explicit static. I dont see how to allow borrowed deserialization - if we tie output lifetime to input lifetime, then we need a static slice to get a static ExecuteV2.

scylla/src/network/connection.rs
Lorak-mmk: In the second mentioned case we should update metadata. In the first we don't have to, but we may (it just updates metadata to be the same).
Those code paths are not critical for performance, schema changes are one-off events. I don't want to complicate the code just to optimize non-hot-path, I prefer to keep ti simpler to understand.

scylla/src/network/connection.rs
Lorak-mmk: Removed assert, rewrote the comment above it.

scylla/tests/integration/statements/prepared_metadata_extension.rs
```diff
@@ -0,0 +1,453 @@
...
+                            RequestReaction::forge_with_error(DbError::IsBootstrapping),
+                        )]))
+                    });
+            }
+
+            let (tx_execute, mut rx_execute) = tokio::sync::mpsc::unbounded_channel();
+
+            let apply_execute_rules = move |running_proxy: &mut RunningProxy| {
+                // We want to EXECUTE only on nodes with extension,
+                // to be force to send empty metadat id.
```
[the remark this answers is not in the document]
wprzytula: `forced`

Issue #1362, opened 2025-05-27
Test for SchemaAgreementError::RequiredHostAbsent


Lorak-mmk:
https://github.com/scylladb/scylla-rust-driver/pull/1355 introduced this variant, but test was not written for it, since it is difficult.
See the discussion: https://github.com/scylladb/scylla-rust-driver/pull/1355#discussion_r2107589683

PR #1475, opened 2025-11-14
Add test for RequiredHostAbsent error

Lorak-mmk:
Based on https://github.com/scylladb/scylla-rust-driver/pull/1473 (which was merged) and https://github.com/scylladb/scylla-rust-driver/pull/1479
[The above PR](https://github.com/scylladb/scylla-rust-driver/pull/1473) changes the semantics in a way that allows us to write the test for RequiredHostAbsent error.
The test adds proxy rules on node 1(which is coordinator for DDL)  that drop connection when:
- New connection is created (specifically, on STARTUP message)
- Schema check query is received (for the purpose of this test, this means any non-CC EXECUTE).

Now schema awaiting will perform check twice:
- First check will return broken connection error for coordinator, and drop all connections
- In second (and each subsequent) check checking schema agreement will require RequiredHostAbsent

Test needs to wait until schema agreement timeout - the semantics change means that it won't return error earlier.
I set the timeout to 1s, so there is no way the test will be flaky, and it should not be too slow.
If we need to make this test faster, then we need to make `check_schema_agreement_with_required_node` public and call it directly.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1362

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/session/schema_agreement.rs
+++ b/scylla/tests/integration/session/schema_agreement.rs
@@ -314,3 +314,85 @@ async fn test_schema_await_with_transient_failure() {
         Err(err) => panic!("{}", err),
     }
 }
+
+// Test that produces SchemaAgreementError::RequiredHostAbsent to prove that
+// such condition is possible, and handled correctly.
+#[tokio::test]
+async fn test_schema_await_required_host_absent() {
+    setup_tracing();
+
+    let res = test_with_3_node_cluster(
+        ShardAwareness::QueryNode,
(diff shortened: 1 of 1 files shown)


Lorak-mmk:
Rebased on current version of base PR, addressed comments.

Lorak-mmk:
Rebased on main. Added a check that uuid in error is correct.
Let's see if the error is reproducible.

wprzytula (dismissed):

Issue #872, opened 2023-12-08
CI: update or get rid of custom Scylla containers


piodul:
In addition to the main GitHub workflow which runs the driver tests with a 3-node cluster with the newest Scylla, we also have workflows that run Scylla with some custom configuration (e.g. enabled TLS or password authentication). For reasons that are unclear to me, the tests use custom docker images which include the custom configuration. Those containers were created very long time ago (3 years ago I think) so they are running versions that are not supported anymore.

We should fix this situation and perhaps change the tests not to use those images but rather configure the cluster differently.

mykaul:
If those configs can be shared across all drivers, manager, monitoring (and possibly one day, with Scylla itself!) that'd be great. No need to reinvent the wheel in every repo.

Issue #1116, opened 2024-11-06
Support for hostname verification


dimakr:
This issue is more of a feature request.

Currently the driver supports peer verification, which can be enabled by setting `openssl::ssl::SslVerifyMode` of SslContext to `PEER`. This allows to ensure that certificate that the server presents is signed by the trusted authority, but it doesn't guarantee that the certificate is issued exactly for the server/domain in question.
This could potentially be a cause of a MITM threat.

It should be possible to enable hostname verification in the driver, for encrypted communication, so that during TLS handshake we can ensure that the certificate was issued for server in question, i.e. that the certificate that servers presents contains its name/IP in the certificate's SAN extension (I'm making this assumption from example of how scylla-go-driver behaves during TLS handshake, when host/peer verification is enabled - it checks that both, certificate is signed by the correct authority and certificate IP/name in the certificate's SAN extension corresponds to the IP/name of the server).

Seems like the [openssl::ssl:SslConnector](https://docs.rs/openssl/0.10.32/openssl/ssl/struct.SslConnector.html) has `set_verify_hostname` method out of the box, but is probably not a suitable solution for the driver, as the driver uses already `openssl::ssl::SslContext` for more flexible configuration of SSL/TLS contexts. But `openssl::ssl::SslContext` doesn't expose any 'verify_hostname'- kind of API.

**Example scenario**

I'm not familiar with scylla-rust-driver itself, so latte stress tool (https://github.com/scylladb/latte) is used as an example application that uses the driver to communicate with Scylla server, with TLS encryption enabled:
- a certificate config was prepared for server with IP 52.51.61.135, but some other IPs/Names were intentionally set in SAN extension of the config, so that the certificate would fail hostname verification
```
[req]
default_bits = 4096
prompt = no
default_md = sha256
distinguished_name = dn
req_extensions = req_ext

[dn]
CN = PR-provision-test-dmitriy-db-node-c048f7e9-111
O = ScyllaDB
L = Herzelia
ST = Tel Aviv
C = IL

[req_ext]
subjectAltName = @alt_names
basicConstraints = critical,CA:FALSE
subjectKeyIdentifier = hash

[alt_names]
DNS.1 = PR-provision-test-dmitriy-db-node-c048f7e9-111
DNS.2 = ec2-52-51-61-111.eu-west-1.compute.amazonaws.com
DNS.3 = ip-10-4-3-111.eu-west-1.compute.internal
IP.1 = 10.4.3.111
[...]

PR #1482, opened 2025-11-21
Add CCM TLS tests, replacing TLS workflow

Lorak-mmk:
Until now we didn't test TLS very well. We only had examples in `examples/` folder, which were run against cluster with predefined certs and config.
There was no way to, for example, test client auth or hostname verification. 
This PR introduces proper TLS test infrastructure using CCM, adds TLS tests using that and removes old TLS workflow and files.
I noticed this when looking into https://github.com/scylladb/scylla-rust-driver/issues/1116 - there was no way to write a reproducer / regression test for that.

Introduces reproducer for https://github.com/scylladb/scylla-rust-driver/issues/1116
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/872

## Pre-review checklist


scylla/tests/integration/ccm/lib/cluster.rs
```diff
@@ -102,7 +102,10 @@ impl NodeList {
...
     }
 
-    #[expect(dead_code)]
+    #[cfg_attr(
+        any(not(feature = "openssl-010"), not(feature = "rustls-023")),
+        expect(dead_code)
+    )]
     pub(crate) fn len(&self) -> usize {
         self.0.len()
     }
```
wprzytula: :thinking: Shouldn't this be `all` quantifier?

wprzytula (dismissed):

Lorak-mmk:
Rebased on main

Lorak-mmk:
Addressed review comment

PR #1493, opened 2025-11-28
.gitignore: Add .DS_Store

piodul:
Those files are automatically created by the built-in MacOS file manager program when opening a directory. It's harmless, but annoying, so ignore those files in .gitignore.


wprzytula (approved):

piodul:
v2: rebased on current main branch to fix the issue with the docs-rs check

Issue #1480, opened 2025-11-19
Add basic event listener to allow cpp-rs-driver to implement host listeners


Lorak-mmk:
cpp-driver has a feature called listener callback. It allows users to register a function that will be called when some node status changes (added / removed / up / down).
As far as we can tell, it is based on events (+ sends artificial add+up events when connecting), without fully considering issues with dropped control connection.

To be able to implement this functionality in cpp-rs-driver, Rust Driver first needs to expose it.
Ultimately, we would like to have such channel exposed to user to:
- Expose more information than topology (see e.g. https://github.com/scylladb/scylla-rust-driver/issues/1405 )
- Have "good" semantics: for example handle CC dropping and send artificial events to account for that after reconnecting. Without this, such callback is not really reliable. See https://github.com/scylladb/scylla-rust-driver/issues/1409

For now, it will be enough to just have a simple api guarded by cpp-rust-unstable, so that the cpp-rs-driver feature can be implemented. cpp-driver also doesn't consider the above issues, so it should not be a regression.

Lorak-mmk:
"filters out" means that no connections are opened to those nodes?

wprzytula:
> Remind we, are we following this behavior? I think there were some discussions about it

I think we are doing one thing differently: we are **not** implicitly turning DC awareness on with assumption that the DC of the first contact point that we connected to is the local DC.

Apart from that, we are following the semantics. @muzarski implemented relevant logic that scans all LBPs (following CPP Driver's solution) and if it detects that all LBPs ban a node (as being remote), it tells `HostFilter` to filter out that node.

PR #1496, opened 2025-12-01
unstable: Implement HostListener mechanism for CPP-RS Driver

wprzytula:
### Introduced `unstable-host-listener` feature

The feature gates the whole `host_listener.rs` module. However, even with feature disabled, driver now emits logs about nodes changing their connectivity status, as well as joining/leaving the cluster.

### Added `HostListener` API

`HostListener` is a trait which allow subscribing to node-related (topology and status) events. It involves one method, `on_event`, which takes the context (for now: host ID and SocketAddr, easily extensible in the future due to lack of public fields and getter-based API) and the event kind (ADD, REMOVE, UP, or DOWN).

### Implemented connectivity change notification mechanism in `PoolRefiller`

`PoolRefiller` now notifies `ClusterWorker` through a channel that an endpoint is again reachable or no longer reachable. This is done by sending a pair `(<host_id>, <event_kind>)` (with event represented by `ConnectivityChangeEvent` enum) into the channel.

### Added node status tracking map to `Cluster`, based on connectivity

Now `ClusterWorker` stores a connectivity status marker for every node in the cluster, which is updated by `ClusterWorker` each time the connection pool that corresponds to the given node notifies it about the connectivity change.

### `ClusterWorker` notifies `HostListener` each time status marker in the tracker map changes

This is a notification deduplication mechanism: should any `PoolRefiller` somehow send two same notifications in a row, only the first will result in `HostListener::on_event` callback fired.

### `ClusterWorker` notifies `HostListener` about every node added/removed from the cluster metadata

Now, upon each metadata refresh:
1. every node that was present in old `ClusterState` but is absent from new `ClusterState` is reported as REMOVED.
2. every node that was absent from old `ClusterState` but is present in new `ClusterState` is reported as ADDED.

Additionally:
1. nodes are reported as DOWN before REMOVED (if hadn't been reported as DOWN earlier).
2. nodes are reported as UP after ADDED.
    - If the pool fails to open any connections to the new node, it will later be marked as DOWN. This mimics CPP Driver's behaviour.
3. nodes filtered out by `HostFilter` have no events reported at all. This mimics CPP Driver's behaviour.

### Manual testing

I built CPP-RS Driver on top of this branch and wrote the bridge to make `host_listener.c` example run. Then, I ran both CPP Driver and CPP-RS Driver in parallel and played with the cluster: stopped and started nodes, added and removed nodes. Broke control connection or not in all of these scenarios. In all situations the Rust implementation behave the same _or better_ than the CPP Driver, which serves as a reference.

### Automated CCM test

I wrote an integration test using CCM bridge (thank you @dkropachev and @Lorak-mmk for making this awesome testing technique possible!). It includes a test cases for quite convoluted cluster dynamics (nodes stopped and started, control connection broken and reestablished on another node, nodes removed and added), which makes me quite confident about correctness of the implementation.

### More Test Cases

There likely are corner cases that we haven't been testing yet. Ideas include:
- node changing its IP
- ...

Fixes: #1480

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/utils.rs
+++ b/scylla/tests/integration/utils.rs
@@ -450,7 +450,13 @@ pub(crate) async fn execute_unprepared_statement_everywhere(
 
 /// Checks that the session is connected to a cluster with expected number of nodes,
 /// that all nodes are connected (UP) and that a simple query can be executed everywhere.
-#[cfg(all(feature = "openssl-010", feature = "rustls-023"))]
+#[cfg_attr(
+    not(any(
+        all(scylla_unstable, feature = "unstable-host-listener"),
+        all(feature = "openssl-010", feature = "rustls-023")
+    )),
+    expect(dead_code)
+)]
 pub(crate) async fn check_session_works_and_fully_connected(
     expected_nodes: usize,
     session: &Session,
--- a/scylla/tests/integration/ccm/host_listener.rs
+++ b/scylla/tests/integration/ccm/host_listener.rs
@@ -1,0 +1,629 @@
+use std::fmt::Display;
+use std::net::{IpAddr, SocketAddr};
+use std::sync::{Arc, Mutex};
+
+use scylla::client::session::Session;
+use scylla::policies::host_filter::DcHostFilter;
+use scylla::policies::host_listener::{HostEvent, HostEventContext, HostListener};
+use tracing::info;
+use uuid::Uuid;
+
+use crate::ccm::lib::cluster::{Cluster, ClusterOptions};
+use crate::ccm::lib::node::{Node, NodeStartOptions};
+use crate::ccm::lib::{CLUSTER_VERSION, run_ccm_test_with_configuration};
+use crate::utils::{check_session_works_and_fully_connected, setup_tracing};
+
+/* LoggingListener HostListener implementation */
+
+fn ip4_last_octet(ip: IpAddr) -> u8 {
+    match ip {
+        std::net::IpAddr::V4(ipv4) => ipv4.octets()[3],
+        std::net::IpAddr::V6(_) => panic!("Expected IPv4 address, got IPv6: {ip}"),
+    }
+}
+
+#[derive(PartialEq, Eq)]
+enum EventKind {
+    Added,
+    Removed,
+    Up,
+    Down,
+    Unknown,
+}
+
+impl std::fmt::Debug for EventKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            EventKind::Added => f.write_str("ADDED"),
+            EventKind::Removed => f.write_str("REMOVED"),
+            EventKind::Up => f.write_str("UP"),
+            EventKind::Down => f.write_str("DOWN"),
+            EventKind::Unknown => f.write_str("<UNKNOWN>"),
+        }
+    }
+}
+impl Display for EventKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Debug::fmt(self, f)
+    }
+}
+
+#[derive(Debug, Default)]
+struct EventLog {
+    events: Vec<(Uuid, SocketAddr, EventKind)>,
+}
+
+impl EventLog {
+    /// Validates the event log for consistency.
+    ///
+    /// This performs a set of checks, which must hold true for any valid sequence of events.
+    fn validate(&self) {
+        self.validate_state_graph();
+
+        // The below checks are likely already covered by `validate_state_graph`,
+        // but they are simpler and more direct, so I keep them for clarity.
+        // We can remove them later if reviewers think they are redundant.
+
+        #[derive(Debug, Clone, Copy)]
+        enum SeenState {
+            Up,
+            Down,
+            Unknown,
+        }
+
+        // No DOWN/UP event may be duplicated.
+        let mut next_event_per_node = [SeenState::Unknown; 4];
+
+        for (i, (_, addr, event)) in self.events.iter().enumerate().rev() {
+            let node_index = ip4_last_octet(addr.ip()) as usize - 1;
+            match (event, next_event_per_node[node_index]) {
+                (EventKind::Down, SeenState::Down) => {
+                    panic!("Duplicated DOWN event for address {} at index {}", addr, i);
+                }
+                (EventKind::Up, SeenState::Up) => {
+                    panic!("Duplicated UP event for address {} at index {}", addr, i);
+                }
+                (EventKind::Down, _) => {
+                    next_event_per_node[node_index] = SeenState::Down;
+                }
+                (EventKind::Up, _) => {
+                    next_event_per_node[node_index] = SeenState::Up;
+                }
+                _ => {}
+            }
+        }
+
+        // No DOWN event may come before an UP event for the same node.
+        for (node_id, next_event) in next_event_per_node.iter().enumerate() {
+            if let SeenState::Down = next_event {
+                panic!("DOWN event came before UP event for node {}", node_id + 1);
+            }
+        }
+    }
+
+    /// Validates the event log for consistency, by following the state graph of each node
+    /// and updating it according to the events received.
+    ///
+    /// This panics for any invalid transition found.
+    fn validate_state_graph(&self) {
+        #[derive(Debug, Clone, Copy)]
+        enum NodeState {
+            Unknown,
+            JustAdded,
+            Up,
+            Down,
+            Removed,
+        }
+
+        let mut node_state = [NodeState::Unknown; 4];
+
+        for (i, (_, addr, event)) in self.events.iter().enumerate() {
+            let node_index = ip4_last_octet(addr.ip()) as usize - 1;
+            let state = &mut node_state[node_index];
+            match (*state, event) {
+                (_, EventKind::Unknown) => panic!("Received event of unknown kind"),
+
+                // ADDED
+                (NodeState::Unknown, EventKind::Added) => *state = NodeState::JustAdded,
+                (NodeState::JustAdded, EventKind::Added) => {
+                    panic!("Node {} ADDED duplicated, at index {}", node_index, i)
+                }
+                (NodeState::Up, EventKind::Added) => panic!(
+                    "Node {} ADDED when being in UP state, at index {}",
+                    node_index, i
+                ),
+                (NodeState::Down, EventKind::Added) => {
+                    panic!(
+                        "Node {} ADDED when being in DOWN state, at index {}",
+                        node_index, i
+                    );
+                }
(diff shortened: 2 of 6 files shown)


scylla/src/policies/host_listener.rs
```diff
@@ -0,0 +1,81 @@
...
+    Added,
+
+    /// Signalled when the driver learns that a known host is removed from the cluster.
+    ///
+    /// This happens in the following cases:
+    /// 1. When the control connection discovers that a known peer is no longer in the cluster.
+    /// 2. When a CQL event notifies about a new node leaving the cluster.
+    /// 3. When a CQL event notifies about a known node being added. Then this event is fired
+    ///    right before the `host_added` event.
+    Removed,
```
Lorak-mmk: Huh, so this comment now tells me that events really are taken into consideration. Interesting, I definitely need to read the implementation.

scylla/src/policies/mod.rs
```diff
@@ -17,6 +17,8 @@
 
 pub mod address_translator;
 pub mod host_filter;
+// #[cfg(cpp_rust_unstable)]
+pub mod host_listener;
```
Lorak-mmk: Why is this commented out?

scylla/src/client/session.rs
```diff
@@ -268,6 +268,10 @@ pub struct SessionConfig {
     /// re-establishing the control connection.
     pub host_filter: Option<Arc<dyn HostFilter>>,
 
+    // #[cfg(cpp_rust_unstable)]
+    /// Optional listener for host events (ADD, REMOVE, UP, DOWN).
+    pub host_listener: Option<Arc<dyn crate::policies::host_listener::HostListener>>,
```
Lorak-mmk: ditto

scylla/src/network/connection_pool.rs
```diff
@@ -1187,6 +1187,16 @@ struct OpenedConnectionEvent {
...
+/// Signals that connectivity to a node has changed.
+#[derive(Debug)]
+pub(crate) enum ConnectivityChangeEvent {
+    /// A new connection to the node was established, while there were no working connections.
+    Established { address: NodeAddr },
+
+    /// The last working connection to the node was lost.
+    Lost { address: NodeAddr },
+}
+
```
Lorak-mmk: Commit: " connection_pool: introduce ConnectivityChangeEvent "

It is surprising that NodeAddr is used for identification, given all the issues and ambiguities we have with identifying nodes this way. Maybe the reason for that will become clear in later commits, idk.

scylla/src/client/session.rs
```diff
@@ -902,6 +902,7 @@ impl Session {
             connection_config,
             pool_size: config.connection_pool_size,
             can_use_shard_aware_port: !config.disallow_shard_aware_port,
+            connectivity_events_sender: None, // This will be set by `Cluster::new`.
         };
```
Lorak-mmk: This becomes yet another way signal out of connection pool - we already had `pool_updated_notify` and `pool_empty_notifier`. Instead of introducing third one, can we do some unification?

`pool_empty_notifier` sounds like something really close to what we need - its just that we need to send both when it becomes empty, and starts being empty. Current users of `pool_empty_notifier` could ignore the "stops being empty" case.

scylla/src/network/connection_pool.rs
```diff
@@ -816,6 +821,23 @@ impl PoolRefiller {
...
+                    // notify that the host is alive again.
+                    debug!(
+                        "[{}] Connection pool is no longer empty, notifying listeners",
+                        endpoint,
+                    );
+
+                    // Ignore failure to send. If there are no listeners, it's fine.
+                    let event = ConnectivityChangeEvent::Established { address: endpoint };
+                    let _ = connectivity_notifier.send(event);
+                }
```
Lorak-mmk: I wonder, why are you performing the notification here?
I see that below there are some checks which could eliminate the connection from being added.
In the case it is accepted, `update_shared_conns` will be called, which also has notification logic added by you. Why duplicate it?

scylla/src/policies/host_listener.rs
```diff
@@ -0,0 +1,81 @@
...
+/// Kind of host event being signalled by [HostListener].
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum HostEvent {
+    /// Signalled when the driver learns that a new host is added to the cluster.
+    ///
+    /// This happens in the following cases:
+    /// 1. When the control connection discovers a new peer.
+    /// 2. When a CQL event notifies about a new node joining the cluster.
+    Added,
```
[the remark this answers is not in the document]
wprzytula: It should. I forgot to update the docstring.

scylla/src/policies/host_listener.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use uuid::Uuid;
+
+/// Context provided to [HostListener] callbacks.
+#[derive(Debug)]
+pub struct HostEventContext {
+    pub(crate) host_id: Uuid,
+    pub(crate) addr: SocketAddr,
+}
+
```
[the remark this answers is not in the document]
wprzytula: OK, will do. I first added it, but then removed due to `pub(crate)` visibility.

scylla/src/policies/host_listener.rs
wprzytula: This is as well outdated.

scylla/src/network/connection_pool.rs
wprzytula: I use it because this is an internal API, so I preferred to keep the additional information that NodeAddr has compared to SocketAddr. It's up to the upper layer to convert it to SocketAddr, which it btw does.

scylla/src/network/connection_pool.rs
wprzytula: `update_shared_conns` only has logic for DOWN, not for UP.

scylla/src/network/connection_pool.rs
```diff
@@ -816,6 +821,23 @@ impl PoolRefiller {
...
+
```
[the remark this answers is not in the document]
Lorak-mmk: Ok

scylla/src/cluster/worker.rs
```diff
@@ -456,6 +464,131 @@ impl ClusterWorker {
...
+        /// because the host listener mechanism identifies nodes by their SocketAddr only.
+        #[derive(Eq)]
+        struct NodeByAddress<'a>(NodeRef<'a>);
+
+        impl std::hash::Hash for NodeByAddress<'_> {
+            fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+                self.0.address.into_inner().hash(state);
+            }
+        }
+
```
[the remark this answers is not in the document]
wprzytula: I fear that it will be dysfunctional. I can't even properly reason about such scenarios. For now, I'm relying on the fact that there's no address translation in CPP-RS driver.

wprzytula:
v1.2:
- improved the integration test.
- fixed bug in Node: `inherit_with_ip_changed` must copy the last UP/DOWN marker state. This is irrelevant, though, because I'm planning to remove the marker from Node as discussed with @Lorak-mmk in person.

scylla/src/client/session.rs
wprzytula: As discussed, possible unification of these notification channels is out of scope of this PR.

wprzytula:
v1.4:
- removed UP/DOWN deduplication marker from Node; instead, ClusterWorker now owns a node status map for this purpose.
- moved ADDED/REMOVED condition handling to `ClusterData::new` (although made it call ClusterWorker's code back for better responsibility separation).
- fixed handling of two corner cases:
  - when a node changes its IP,
  - when one node is removed and another is added with the same IP.

scylla/src/network/connection_pool.rs
wprzytula: Shall we consider this resolved?

scylla/src/cluster/worker.rs
```diff
@@ -46,11 +47,27 @@ impl std::fmt::Debug for ClusterNeatDebug<'_> {
...
 struct ClusterWorker {
     // Cluster state to keep updated:
     cluster_state: Arc<ArcSwap<ClusterState>>,
 
+    /// Node status map.
+    ///
+    /// Maps host_id to connectivity status.
+    /// Used to track node status in order to deduplicate HostListener events.
+    node_status: HashMap<Uuid, (NodeAddr, NodeConnectivityStatus)>,
+
```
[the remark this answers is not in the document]
Lorak-mmk: ConnectivityChangeEvents no longer identify nodes by Uuid. Is there any reason now to have addr in this map?

Lorak-mmk (changes requested):
The tests look great!

scylla/src/cluster/worker.rs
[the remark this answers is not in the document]
wprzytula: Good observation. Will remove this.

scylla/src/cluster/worker.rs
```diff
@@ -467,6 +502,218 @@ impl ClusterWorker {
...
+    ///
+    /// Emit respective events to the [HostListener], if configured.
+    fn handle_topology_changes(
+        known_peers: &HashMap<Uuid, Arc<Node>>,
+        new_known_peers: &HashMap<Uuid, Arc<Node>>,
+        #[cfg(all(scylla_unstable, feature = "unstable-host-listener"))] host_listener: Option<
+            &dyn HostListener,
+        >,
+        node_status: &mut HashMap<Uuid, NodeConnectivityStatus>,
+    ) {
```
Lorak-mmk: I'm not a fan of those `#[cfg(...)]` now being everywhere now. Idea: we can make HostListener exist unconditionally, but without the feature it will not be present in SessionConfig, and it will not be pub.
To make it non-pub, you can make the module containing it private, the items themselves pub, and add a pub re-export guarded by cfg.

scylla/tests/integration/utils.rs
```diff
@@ -450,7 +450,13 @@ pub(crate) async fn execute_unprepared_statement_everywhere(
...
 /// Checks that the session is connected to a cluster with expected number of nodes,
 /// that all nodes are connected (UP) and that a simple query can be executed everywhere.
-#[cfg(all(feature = "openssl-010", feature = "rustls-023"))]
+#[cfg_attr(
+    not(any(
+        all(scylla_unstable, feature = "unstable-host-listener"),
+        all(feature = "openssl-010", feature = "rustls-023")
+    )),
+    expect(dead_code)
+)]
```
Lorak-mmk: When I see such declarations, I wonder if we can just `allow(dead_code)` for whole `ccm` module, and utils like this one. Probably not a good idea, idk.

wprzytula: The problem of such declarations will disappear once we start using CCM for testing stable core features of the driver. Let's do it :)

wprzytula:
> Uh, I thought it was going to be an approve, but I found one more thing, sorry :(

Done.

Lorak-mmk (approved):

PR #1497, opened 2025-12-03
Add nodejs-rs driver endpoint for raw rows.

adespawn:
Deserialization on the rust side, for the NodeJS driver has significant overhead, compared to the deserialization done by the JS driver from raw rows.

This commit introduces the endpoints that allow nodejs-rs driver to access raw rows parts of the query result.

## Pre-review checklist

~~- [ ] I added relevant tests for new features and bug fixes.~~
~~- [ ] I have adjusted the documentation in `./docs/source/`.~~
~~- [ ] I added appropriate `Fixes:` annotations to PR description.~~

### DIFF
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -836,6 +836,13 @@ impl DeserializedMetadataAndRawRows {
         );
         TypedRowIterator::new(raw)
     }
+
+    /// Allows to retrieve raw rows, without the need for deserialization
+    /// Intended to be used only in nodejs-rs driver only.
+    #[cfg(nodejs_rs_unstable)]
+    pub fn raw_rows(&self) -> &Bytes {
+        &self.raw_rows
+    }
 }
 
 /// Represents the result of a CQL `RESULT` response.
--- a/scylla-cql/Cargo.toml
+++ b/scylla-cql/Cargo.toml
@@ -101,6 +101,9 @@ harness = false
 [lints.rust]
 unnameable_types = "warn"
 unreachable_pub = "warn"
+unexpected_cfgs = { level = "warn", check-cfg = [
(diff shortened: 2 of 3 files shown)


scylla-cql/src/frame/response/result.rs
```diff
@@ -836,6 +836,13 @@ impl DeserializedMetadataAndRawRows {
         );
         TypedRowIterator::new(raw)
     }
+
+    /// Allows to retrieve raw rows, without the need for deserialization
+    /// Intended to be used only in nodejs-rs driver only.
+    #[cfg(nodejs_rs_unstable)]
+    pub fn get_raw_rows(&self) -> &Bytes {
+        &self.raw_rows
+    }
```
wprzytula: :wrench: https://rust-lang.github.io/api-guidelines/naming.html#c-getter

wprzytula: Actually, I think this function could be added without any `cfg` attached. It won't be accessible to driver's users anyway, as the very `DeserializedMetadataAndRawRows` is not accessible by them.
@Lorak-mmk WDYT?

wprzytula (changes requested):

Lorak-mmk:
Could you explain a bit more why do you need those? It is not obvious to me why you can't use, for example, our DeserializeRow impl for `ColumnIterator`, and understanding the use case more can help us make better API for it in the future.

Lorak-mmk:
That means every object created this way will keep the whole frame alive, right?
This is imo reasonable in Rust where you either need to use lifetimes to borrow from frame, or make your own types explicitly use `Bytes`.
In JS it may be a possible footgun if its not easily notceable.

Lorak-mmk (approved):

wprzytula (approved):

PR #1504, opened 2025-12-09
macros: Add support for C-style enum serialization

### LINKED ISSUE #923: Serialization / Deserialization derive macros should work for enums without data
This looks like it can be a common case that users want to use structs like in this forum question: https://forum.scylladb.com/t/can-we-make-an-enum-serializable-so-it-can-go-to-and-from-a-cqlvalue-in-rust/1242
While I don't see any sensible way to derive for arbitrary enums, we could add support for enums without data.
I think this shouldn't be too hard. The main question is how to specify the type to serialize to (i32, u8 etc)?

ziara1:
## Motivation
Users often map C-style enums (enums with unit variants and integer discriminants) to integer columns in ScyllaDB (e.g., `tinyint`, `smallint`, `int`). Currently, users must manually implement `SerializeValue` and `DeserializeValue` for these enums, which involves tedious boilerplate for matching variants to integers and vice versa.

## Solution
This commit extends `SerializeValue` and `DeserializeValue` macros to support C-style enums directly.
Users can now annotate their enum with `#[scylla(repr = "TYPE")]` (e.g., `"i8"`, `"i32"`), and the macro will generate code that:
- Serializes the enum variant as its integer discriminant.
- Deserializes an integer from the DB and matches it back to the corresponding enum variant.

**Behavior details:**
- If a discriminant is not specified explicitly, standard Rust rules apply (starts at 0, increments by 1).
- Deserialization fails with a proper error if the integer value from the DB does not match any known variant.
- Type checking delegates to the underlying integer type (e.g., `i32` checks against `Int` column).

### Refactored DeserializeValue derive
- Refactored `deserialize_value_derive` to handle `Data::Enum` separately.
- Added parsing for the `#[scylla(repr = "...")]` attribute via `EnumAttrs`.
- Implemented `deserialize_value_derive_enum`, which generates a `match` statement mapping integers from the database back to enum variants.

### Refactored SerializeValue derive
- Refactored `serialize_value_derive` to handle `Data::Enum`.
- Implemented `derive_serialize_value_enum`, which casts the enum variant to the representation type (e.g., `*self as i32`) before serialization.

### Tests
Added comprehensive tests covering:
- **Integer sizes:** Support for `i8` (TinyInt), `i16` (SmallInt), `i32` (Int), and `i64` (BigInt).
- **Discriminants:** Tested standard auto-incrementing discriminants and manual discriminants with gaps.
- **Error Handling:** Verified that deserialization fails with a clear error when the database value doesn't match any enum variant.
- **Type Checking:** Verified that using an enum wrapper incorrectly (e.g., against a `Text` column) produces a type check error.

Fixes: #923 

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/types_tests.rs
+++ b/scylla-cql/src/types_tests.rs
@@ -302,3 +302,233 @@ mod derive_macros_integration {
         }
     }
 }
+
+mod enum_serialization_tests {
+    use crate::_macro_internal::DeserializationError;
+    use crate::deserialize::FrameSlice;
+    use crate::deserialize::value::BuiltinDeserializationError;
+    use crate::deserialize::value::DeserializeValue;
+    use crate::frame::response::result::{ColumnType, NativeType};
+    use crate::serialize::value::SerializeValue;
+    use crate::serialize::writers::CellWriter;
+    use bytes::Bytes;
+    use scylla_macros::{DeserializeValue, SerializeValue};
+
+    // --- TEST HELPERS ---
+
+    fn assert_round_trip<T>(value: T, typ: ColumnType, expected_bytes: &[u8])
+    where
+        T: SerializeValue
+            + for<'f, 'm> DeserializeValue<'f, 'm>
+            + PartialEq
+            + std::fmt::Debug
+            + Copy,
+    {
+        // Serialization Test
+        let mut data = Vec::new();
+        let writer = CellWriter::new(&mut data);
+        value.serialize(&typ, writer).unwrap();
+
+        // Verify length header (4 bytes big endian) + data
+        let len = expected_bytes.len() as i32;
+        let mut expected_full = len.to_be_bytes().to_vec();
+        expected_full.extend_from_slice(expected_bytes);
+
+        assert_eq!(data, expected_full, "Serialization failed (byte mismatch)");
+
+        // Deserialization Test
+        // Simulate reading from a frame (skip 4 bytes length header)
+        let payload = Bytes::copy_from_slice(expected_bytes);
+        let slice = FrameSlice::new(&payload);
+
+        let deserialized = T::deserialize(&typ, Some(slice)).expect("Deserialization failed");
+        assert_eq!(
+            value, deserialized,
+            "Deserialized value does not match input"
+        );
+    }
+
+    fn assert_deser_error<T, F>(typ: ColumnType, bad_bytes: &[u8], check_error: F)
+    where
+        T: for<'f, 'm> DeserializeValue<'f, 'm> + std::fmt::Debug,
+        F: FnOnce(DeserializationError),
+    {
+        let payload = Bytes::copy_from_slice(bad_bytes);
+        let slice = FrameSlice::new(&payload);
+        let result = T::deserialize(&typ, Some(slice));
+
+        match result {
+            Ok(v) => panic!("Expected error, but deserialization succeeded: {:?}", v),
+            Err(e) => check_error(e),
+        }
+    }
+
+    // --- ENUM DEFINITIONS ---
+
+    #[derive(SerializeValue, DeserializeValue, Debug, PartialEq, Copy, Clone)]
+    #[scylla(crate = "crate")]
+    #[repr(i32)]
+    enum DefaultInt {
+        A = 0,
+        B = 1,
+        C = 100,
+    }
+
+    #[derive(SerializeValue, DeserializeValue, Debug, PartialEq, Copy, Clone)]
+    #[scylla(crate = "crate")]
+    #[repr(i8)]
+    enum TinyEnum {
+        Small = 10,
+        Max = 127,
+    }
+
+    #[derive(SerializeValue, DeserializeValue, Debug, PartialEq, Copy, Clone)]
+    #[scylla(crate = "crate")]
+    #[repr(i16)]
+    enum SmallEnum {
+        Kilo = 1000,
+        BigKilo = 30000,
+    }
+
+    #[derive(SerializeValue, DeserializeValue, Debug, PartialEq, Copy, Clone)]
+    #[repr(i64)]
+    #[scylla(crate = "crate")]
+    enum BigEnum {
+        Big = 5_000_000_000,
+    }
+
+    #[derive(SerializeValue, DeserializeValue, Debug, PartialEq, Copy, Clone)]
+    #[scylla(crate = "crate")]
+    #[repr(i32)]
+    enum Gaps {
+        First = 1,
+        Jump = 5,
+        FarAway = 99,
+    }
+
+    #[derive(SerializeValue, DeserializeValue, Debug, PartialEq, Copy, Clone)]
+    #[scylla(crate = "crate")]
+    #[repr(i32)]
+    enum NegativeEnum {
+        MinusOne = -1,
+        MinusHundred = -100,
+    }
+
+    #[derive(SerializeValue, DeserializeValue, Debug, PartialEq, Copy, Clone)]
+    #[scylla(crate = "crate")]
+    #[repr(i32)]
+    enum ImplicitEnum {
+        First,
+        Second,
+        Third,
+    }
+
+    // --- TESTS ---
+
+    #[test]
+    fn test_i32_standard() {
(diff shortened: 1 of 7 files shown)


docs/source/data-types/udt.md
```diff
@@ -76,3 +76,33 @@ while let Some((my_type_value,)) = iter.try_next().await? {
...
+```rust
+# extern crate scylla;
+# async fn check_only_compiles() {
+use scylla::{SerializeValue, DeserializeValue};
+
+// Defines an enum that will be stored as an 'int' in ScyllaDB.
+#[derive(SerializeValue, DeserializeValue, PartialEq, Debug, Clone, Copy)]
+#[scylla(repr = "i32")]
+#[repr(i32)]
+enum Status {
```
wprzytula: :thought_balloon: With current implementation, one needs to duplicate the `repr` attribute. Perhaps it makes more sense to parse `#[repr(<typ>)]` attributes in the macros code instead? Is this even possible?

scylla-macros/src/deserialize/value.rs
```diff
@@ -83,26 +90,195 @@ pub(crate) fn deserialize_value_derive(
...
+        syn::Data::Enum(data_enum) => deserialize_value_derive_enum(&input, data_enum),
+        syn::Data::Struct(_) => {
+            let implemented_trait: syn::Path = parse_quote!(DeserializeValue);
+            let implemented_trait_name = implemented_trait
+                .segments
+                .last()
+                .unwrap()
+                .ident
+                .unraw()
+                .to_string();
```
wprzytula: :wrench: Justify the `unwrap()`. Why can we be sure that there is at least one segment?

docs/source/data-types/udt.md
Lorak-mmk: Strongly agreed, if possible we should not introduce new attribute.

scylla/tests/integration/macros/enum_serialization.rs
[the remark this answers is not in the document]
Lorak-mmk: Please place such tests in `scylla-cql/src/`, either in `types_tests.rs` or in `serialize/value_tests.rs` and `deserialize/value_tests.rs`

ziara1:
@wprzytula  done

scylla-macros/src/serialize/value.rs
```diff
@@ -136,6 +140,60 @@ pub(crate) fn derive_serialize_value(
...
+        .map_err(|_| syn::Error::new_spanned(input, "Failed to parse repr type"))?;
+
+    let implemented_trait: syn::Path = parse_quote!(#crate_path::SerializeValue);
+
+    let res = parse_quote! {
+        #[automatically_derived]
+        impl #impl_generics #implemented_trait for #struct_name #ty_generics
+        where
+            #struct_name #ty_generics: ::std::marker::Copy,
+            #where_clause
```
Lorak-mmk: Is this a valid use of `where_clause`? Perhaps its not important for C-style enums (because they won't have lifetimes / generics anyway), but I think `where_clause` also contains the `where` keyword, and would just fail to compile here. The approach that is afaik correct (using `make_where_clause` and extending it) can be seen in my PR that fixes generics: https://github.com/scylladb/scylla-rust-driver/pull/1546

Lorak-mmk (changes requested):

scylla-cql/src/types_tests.rs
```diff
@@ -302,3 +302,210 @@ mod derive_macros_integration {
...
+    {
+        // Serialization Test
+        let mut data = Vec::new();
+        let writer = CellWriter::new(&mut data);
+        value.serialize(&typ, writer).unwrap();
+
+        // Verify length header (4 bytes big endian) + data
+        let len = expected_bytes.len() as i32;
+        let mut expected_full = len.to_be_bytes().to_vec();
+        expected_full.extend_from_slice(expected_bytes);
```
wprzytula:
```suggestion
        let mut expected_full = len.to_be_bytes();
        expected_full.extend_from_slice(expected_bytes);
```

scylla-macros/src/enum_attrs.rs
```diff
@@ -0,0 +1,71 @@
...
+                    variant,
+                    "Trait can only be derived for enums with unit variants (no data fields).",
+                ));
+            }
+        }
+    } else {
+        return Err(syn::Error::new_spanned(input, "Not an enum"));
+    }
+    Ok(())
+}
```
wprzytula: :wrench: Instead of `DeriveInput`, this function should accept [`DataEnum`](https://docs.rs/syn/latest/syn/struct.DataEnum.html). Then it couldn't even receive an invalid argument (non-enum definition).

wprzytula (changes requested):
`hygiene.rs` needs test cases for enums.

scylla-cql/src/types_tests.rs
wprzytula: You can easily take a slice from a fixed-size array. Use `as_slice()` or slicing syntax.

scylla-cql/src/types_tests.rs
```diff
@@ -302,3 +302,210 @@ mod derive_macros_integration {
...
+    {
+        let payload = Bytes::copy_from_slice(bad_bytes);
+        let slice = FrameSlice::new(&payload);
+        let result = T::deserialize(&typ, Some(slice));
+        assert!(
+            result.is_err(),
+            "Expected error, but deserialization succeeded: {:?}",
+            result
+        );
+    }
```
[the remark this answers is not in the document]
wprzytula: Can you elaborate?
If the error is dynamic (`dyn Error`), then syntactic pattern matching is indeed not viable. What we can do, however, is to downcast the error into the concrete error type we are expecting here, and panic if the cast fails.

wprzytula (changes requested):
Looks good! Just two nits.

scylla-macros/src/deserialize/value.rs
```diff
@@ -82,29 +89,123 @@ impl DeserializeCommonFieldAttrs for Field {
...
+
+                        impl ::std::error::Error for UnknownEnumVariant {}
+
+                        ::std::result::Result::Err(
+                            #crate_path::DeserializationError::new(
+                                UnknownEnumVariant(format!("Enum value does not match any variant: {:?}", val))
+                            )
+                        )
+                    }
+                }
```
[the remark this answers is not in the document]
ziara1: Do you mean the new commits aren't showing up for you, or does my new implementation still not resolve the core issue?

wprzytula (approved):

scylla-macros/src/deserialize/value.rs
```diff
@@ -82,29 +89,123 @@ impl DeserializeCommonFieldAttrs for Field {
...
+
+    for variant in &data_enum.variants {
+        let variant_ident = &variant.ident;
+
+        match_arms.push(quote::quote! {
+            v if v == Self::#variant_ident as #repr_type => {
+                ::std::result::Result::Ok(Self::#variant_ident)
+            }
+        });
+    }
```
Lorak-mmk, line 149: @wprzytula Before merging, we'll need to investigate the code generated by this.
The actual comparison is done in `if` - so each match arm matches, and is then verified in the condition.

Compiler may be smart enough to optimize it the same way it would a normal match, but it may not. If it is not, then we need to do it differently (for example requiring that enum variants are manually assigned values, and using those).

scylla-macros/src/deserialize/value.rs
```diff
@@ -82,29 +89,123 @@ impl DeserializeCommonFieldAttrs for Field {
...
+                                ::std::write!(f, "Enum value does not match any variant: {:?}", self.0)
+                            }
+                        }
+
+                        impl<T: ::std::fmt::Debug> ::std::error::Error for UnknownEnumVariant<T> {}
+
+                        ::std::result::Result::Err(
+                            #crate_path::DeserializationError::new(
+                                UnknownEnumVariant(val)
+                            )
```
Lorak-mmk, line 199: @wprzytula Perhaps `UnknownEnumVariant` should not be some local type unnameable by the user, but a normal error in our builtins? Right now there will be no way to match this error.

Issue #1293, opened 2025-03-24
Support `#[scylla(transparent)]` attribute by derive macros


wprzytula:
# Motivation

Users seems to often employ the newtype pattern (i.e., define single-field structs) to wrap the driver's types that implement `{De,S}erialize{Value,Row}`. In such case, they are forced to manually implemenent (trivially) the respective de/ser traits for those new types.

## Solution

Add support for a new item-level attribute, `#[scylla(transparent)]`:
1. on single-field structs (the original use case),
2. on single-variant enums (because why not; technically, it's the same).

The semantics is simple: delegate the trait method's implementation to the only field/variant of the struct/enum.

PR #1506, opened 2025-12-11
macros: Add support for #[scylla(transparent)] attribute

ziara1:
## Motivation
Users often employ the newtype pattern (i.e., define single-field structs) to wrap the driver's types that implement `{De,S}erialize{Value,Row}`. In such cases, they are currently forced to manually implement (trivially) the respective de/ser traits for those new types, which results in unnecessary boilerplate code.

## Solution
Add support for a new item-level attribute, `#[scylla(transparent)]`.
It is supported on:
- single-field structs (the original use case),
- single-variant enums (technically identical to single-field structs).

The semantics are simple: delegate the trait method's implementation (serialization, deserialization, and type checking) to the only field/variant of the struct/enum.

### Implemented transparent logic in DeserializeValue derive
Refactored `deserialize_value_derive` to delegate to a new `derive_transparent` function when the attribute is present.
The implementation ensures that:
- `type_check` delegates to the inner type.
- `deserialize` delegates to the inner type and wraps the result in the newtype constructor.
- Generic `where` clauses are handled correctly by merging generated predicates with existing user predicates using `make_where_clause()`.

### Implemented transparent logic in SerializeValue derive
Modified `serialize_value_derive` to handle the `transparent` attribute.
- **For structs:** Serializes the single field directly.
- **For enums:** Matches the single variant and serializes its inner field.
- Added validation to ensure the attribute is only used on items with exactly one field.

### Updated Attribute Parsing
Updated `StructAttrs` and `Attributes` definitions in both deserialize and serialize modules to parse the `#[scylla(transparent)]` boolean flag via `darling`.

### Tests
I added a comprehensive suite of tests covering:
- **Tuple Structs:** `struct Wrapper(i32)`
- **Named Structs:** `struct Wrapper { val: i32 }`
- **Enums:** Single variant tuple and named enums.
- **Nested Composition:** `Outer(Inner(i32))` to verify deep delegation.
- **Type Safety:** Verified that `type_check` correctly propagates errors (e.g., trying to read an `int` column as `text` through a wrapper fails).
- **Error Propagation:** Verified that deserialization errors (e.g., not enough bytes) are correctly propagated from the inner type through the wrapper.

Fixes: #1293 

## Pre-review checklist

### DIFF
--- a/docs/source/data-types/udt.md
+++ b/docs/source/data-types/udt.md
@@ -76,3 +76,53 @@ while let Some((my_type_value,)) = iter.try_next().await? {
 # Ok(())
 # }
 ```
+
+## Transparent Wrappers (Newtype Pattern)
+
+Sometimes it is desirable to wrap a driver's type (like `i32` or `String`) in a custom struct for type safety, a pattern often called the "newtype pattern".
+To avoid writing manual implementations of `SerializeValue` and `DeserializeValue`, you can use the `#[scylla(transparent)]` attribute.
+
+This attribute delegates the serialization, deserialization, and type checking logic to the single field contained within the struct or enum. The wrapper will be stored in the database exactly as the inner type is stored (e.g., as an `Int` or `Text` column), not as a UDT (unless the inner type is a UDT, of course).
+
+**Requirements:**
+* The struct must have exactly one field.
+* The enum must have exactly one variant with exactly one field.
+
+```rust
+# extern crate scylla;
+# async fn check_only_compiles() {
+use scylla::{DeserializeValue, SerializeValue};
+
+// This struct will be serialized as a simple Int (4 bytes),
+// strictly matching the inner `i32` type.
+#[derive(Debug, SerializeValue, DeserializeValue)]
+#[scylla(transparent)]
+struct UserId(i32);
+
+// Works with named fields too
+#[derive(Debug, SerializeValue, DeserializeValue)]
+#[scylla(transparent)]
+struct OrderId {
+    val: i64,
+}
+
+// This will be serialized as a CQL list<text>
+#[derive(Debug, SerializeValue, DeserializeValue)]
+#[scylla(transparent)]
+struct Tags(Vec<String>);
+
+// A standard UDT
+#[derive(Debug, SerializeValue, DeserializeValue)]
+struct Address {
+    street: String,
+    city: String,
+}
+
+// A transparent wrapper around the UDT.
+// This matches the same CQL UDT layout as `Address`.
+#[derive(Debug, SerializeValue, DeserializeValue)]
+#[scylla(transparent)]
+struct HomeAddress(Address);
+
+# }
+```
--- a/scylla-cql/src/deserialize/value.rs
+++ b/scylla-cql/src/deserialize/value.rs
@@ -682,7 +682,7 @@ where
 // collections
 
 make_error_replace_rust_name!(
-    pub(crate),
+    pub,
     typck_error_replace_rust_name,
     TypeCheckError,
     BuiltinTypeCheckError
--- a/scylla-macros/src/deserialize/value.rs
+++ b/scylla-macros/src/deserialize/value.rs
@@ -31,6 +31,9 @@ struct StructAttrs {
     // they will be ignored. With true, an error will be raised.
     #[darling(default)]
     forbid_excess_udt_fields: bool,
+
+    #[darling(default)]
+    transparent: bool,
 }
 
 impl DeserializeCommonStructAttrs for StructAttrs {
@@ -82,7 +85,12 @@ impl DeserializeCommonFieldAttrs for Field {
 pub(crate) fn deserialize_value_derive(
     tokens_input: TokenStream,
 ) -> Result<syn::ItemImpl, syn::Error> {
-    let input = syn::parse(tokens_input)?;
+    let input: syn::DeriveInput = syn::parse(tokens_input)?;
+
+    let attrs = StructAttrs::from_attributes(&input.attrs)?;
+    if attrs.transparent {
+        return derive_transparent(&input, &attrs);
+    }
 
     let implemented_trait: syn::Path = parse_quote!(DeserializeValue);
     let implemented_trait_name = implemented_trait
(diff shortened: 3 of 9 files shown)


scylla-macros/src/serialize/value.rs
```diff
@@ -96,13 +99,88 @@ pub(crate) fn derive_serialize_value(
...
+    if attributes.transparent {
+        if let syn::Data::Struct(data) = &input.data {
+            if data.fields.len() != 1 {
+                return Err(syn::Error::new_spanned(
+                    data.fields.clone(),
+                    "#[scylla(transparent)] requires exactly one field",
+                ));
+            }
+
+            let field = data.fields.iter().next().unwrap();
```
Lorak-mmk: In deserialization you extracted transparent case into a separate function. Can you do the same here?

scylla-macros/src/deserialize/value.rs
```diff
@@ -105,6 +113,94 @@ pub(crate) fn deserialize_value_derive(
...
+                Some(ident) => quote::quote! { |inner| Self { #ident: inner } },
+                None => quote::quote! { |inner| Self(inner) },
+            };
+            (&field.ty, ctor)
+        }
+        syn::Data::Enum(data) => {
+            if data.variants.len() != 1 {
+                return Err(syn::Error::new_spanned(
+                    input,
+                    "#[scylla(transparent)] enum requires exactly one variant",
```
Lorak-mmk: Why not pass `data.variants` as a span instead of input?

scylla-macros/src/serialize/value.rs
```diff
@@ -96,13 +99,88 @@ pub(crate) fn derive_serialize_value(
...
+                impl #impl_generics #implemented_trait for #struct_name #ty_generics #where_clause {
+                    fn serialize<'b>(
+                        &self,
+                        typ: &#crate_path::ColumnType,
+                        writer: #crate_path::CellWriter<'b>,
+                    ) -> ::std::result::Result<#crate_path::WrittenCellProof<'b>, #crate_path::SerializationError> {
+                        #crate_path::SerializeValue::serialize(&self.#field_accessor, typ, writer)
+                    }
+                }
+            });
```
Lorak-mmk: Should error be mapped to correct name here too?

scylla-macros/src/deserialize/value.rs
```diff
@@ -105,6 +113,94 @@ pub(crate) fn deserialize_value_derive(
...
+    Ok(parse_quote! {
+        #[automatically_derived]
+        impl<#frame_lifetime, #metadata_lifetime, #impl_generics_params> #crate_path::DeserializeValue<#frame_lifetime, #metadata_lifetime> for #struct_name #ty_generics
+        where #(#predicates),* #where_clause
+        {
+            fn type_check(
+                typ: &#crate_path::ColumnType,
+            ) -> ::std::result::Result<(), #crate_path::TypeCheckError> {
+                <#inner_type as #crate_path::DeserializeValue<#frame_lifetime, #metadata_lifetime>>::type_check(typ)
+            }
```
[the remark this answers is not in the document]
ziara1: `scylla-cql` does not export a helper function to replace the Rust name in `TypeCheckError` (like `value_deser_error_replace_rust_name` for `DeserializationError`).

wprzytula:
@ziara1 Please squash fixes into corresponding commits.

scylla-macros/src/deserialize/value.rs
```diff
@@ -82,7 +85,12 @@ impl DeserializeCommonFieldAttrs for Field {
 pub(crate) fn deserialize_value_derive(
     tokens_input: TokenStream,
 ) -> Result<syn::ItemImpl, syn::Error> {
-    let input = syn::parse(tokens_input)?;
+    let input: syn::DeriveInput = syn::parse(tokens_input)?;
+
+    let attrs = StructAttrs::from_attributes(&input.attrs)?;
+    if attrs.transparent {
+        return derive_transparent(&input, &attrs);
+    }
```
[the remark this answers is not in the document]
Lorak-mmk: In that case we have to modify StructDesc to support this case. Note that you will need it both here, and in tuple-struct PR. I suggest opening separate PR for this.

In general we want to avoid introducing new independent code paths, because they will make future maintenance much harder. Wherever possible you should use (and possible extend) existing infrastructure.

For example, serialization / deserialization with flavor `enforce_order` and `skip_name_checks` attribute is pretty much equivalent to what needs to be done in tuple-struct PR. The whole code path could be reused (after some attribute validation).

Lorak-mmk (changes requested):

docs/source/data-types/udt.md
```diff
@@ -76,3 +76,34 @@ while let Some((my_type_value,)) = iter.try_next().await? {
 # Ok(())
 # }
 ```
+
+## Transparent Wrappers (Newtype Pattern)
+
+Sometimes it is desirable to wrap a driver's type (like `i32` or `String`) in a custom struct for type safety, a pattern often called the "newtype pattern".
+To avoid writing manual implementations of `SerializeValue` and `DeserializeValue`, you can use the `#[scylla(transparent)]` attribute.
+
+This attribute delegates the serialization, deserialization, and type checking logic to the single field contained within the struct or enum. The wrapper will be stored in the database exactly as the inner type is stored (e.g., as an `Int` or `Text` column), not as a UDT.
```
wprzytula: :wrench: Slightly misleading: the wrapper may be stored as a UDT if its inner type is a UDT.

scylla-cql/src/deserialize/value.rs
```diff
@@ -682,7 +682,7 @@ where
 // collections
 
 make_error_replace_rust_name!(
-    pub(crate),
+    pub,
     typck_error_replace_rust_name,
     TypeCheckError,
     BuiltinTypeCheckError
```
wprzytula: :thinking: We purposely made this `pub(crate)`. We don't want for this to be part of `scylla-cql`'s public API.
Are similar helpers (e.g., for serialization) exposed as `pub` or is something else done to make them available in macros?

scylla-macros/src/deserialize/value.rs
```diff
@@ -895,3 +955,31 @@ impl DeserializeUnorderedGenerator<'_> {
...
+where
+    I: IntoIterator,
+{
+    let mut iter = iter.into_iter();
+    let item = iter.next()?;
+    if iter.next().is_some() {
+        return None;
+    }
+    Some(item)
+}
```
wprzytula:
```suggestion
fn get_exactly_one<I>(iter: I) -> Option<I::Item>
where
    I: IntoIterator,
{
    let mut iter = iter.into_iter();
    let item = iter.next()?;
    iter.next().is_none().then_some(item)
}
```

scylla-macros/src/lib.rs
```diff
@@ -88,6 +86,22 @@ mod serialize;
 /// }
 /// ```
 ///
+/// # Transparent Attribute
```
wprzytula:
```suggestion
/// # `transparent` Attribute
```

scylla-macros/src/lib.rs
```diff
@@ -88,6 +86,22 @@ mod serialize;
...
+///
+/// If you are using the newtype pattern (wrapping a type in a struct), you can use
+/// `#[scylla(transparent)]` to delegate serialization to the inner type.
+///
+/// This is supported for:
+/// * Structs with exactly one field.
+/// * Enums with exactly one variant (which has exactly one field).
```
wprzytula: :wrench: Leftover:

```suggestion
/// This is supported for:
/// * Structs with exactly one field.
```

scylla-macros/src/lib.rs
```diff
@@ -405,9 +424,9 @@ pub fn deserialize_row_derive(tokens_input: TokenStream) -> TokenStream {
 
 /// Derive macro for the [`DeserializeValue`](./deserialize/value/trait.DeserializeValue.html)
 /// trait that generates an implementation which deserializes a User Defined Type
-/// with the same layout as the Rust struct.
+/// with the same layout as the Rust struct, or deserializes a transparent wrapper.
 ///
-/// At the moment, only structs with named fields are supported.
+/// At the moment, only structs with named fields are supported (unless `transparent` is used).
```
wprzytula: :+1: This comment should be present in serialization, too, instead of removing it altogether.

scylla-macros/src/lib.rs
```diff
@@ -433,6 +452,25 @@ pub fn deserialize_row_derive(tokens_input: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
+/// # Transparent Attribute
```
wprzytula:
```suggestion
/// # `transparent` Attribute
```

scylla-macros/src/serialize/value.rs
```diff
@@ -136,6 +143,94 @@ pub(crate) fn derive_serialize_value(
...
+        syn::Data::Enum(data) => {
+            let variant = match get_exactly_one(&data.variants) {
+                Some(v) => v,
+                None => {
+                    return Err(syn::Error::new_spanned(
+                        &data.variants,
+                        "#[scylla(transparent)] enums requires exactly one variant",
+                    ));
+                }
+            };
```
wprzytula: :question: :exclamation: This is a massive leftover. In serialization, you removed support for single-variant enums.

scylla-macros/src/deserialize/value.rs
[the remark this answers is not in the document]
ziara1: I will have to create a separate PR for this.

scylla-cql/src/deserialize/value.rs
ziara1: I'm not sure how to do it without changing it to pub.

Issue #1409, opened 2025-07-30
Insufficient guarantees regarding ordering of events and control connection breakage.


Lorak-mmk:
Cluster Worker processes messages from a few channels. Two of them are:
- server events channel
- control connection broken channel

The issue is that there are no guarantees that can be currently made about the relative ordering of messages  on those channels.
What I mean is that if Cluster Worker started to put messages from both of those channels into a new one (like we want to do in https://github.com/scylladb/scylla-rust-driver/issues/1405 ), then the following order of messages would be possible:
[ <CC A broken>, <Event X>, <Event Y>, <Event Z>, <CC B broken>]
we would like to have a guarantee that all 3 events are coming from control connection B in such case. Unfortunately it is possible for those events to come from CC A, or CC connected after B.

Apart from  #1405, it is also a problem for https://github.com/scylladb/scylla-rust-driver/issues/1280

Possible solution: instead of using ConnectionPool, we should be using Connection, and listen on its event and broken channel. Then a new connection means new channels, so no reordering possible.

PR #1526, opened 2026-01-07
First batch of metadata improvements

Lorak-mmk:
This PR performs some refactors in metadata code (mostly in MetadataReader).
The biggest changes:
- Split `metadata.rs` into 3 files, because it was too big and I couldn't comprehend it.
- Removed usage of `ConnectionPool`. See commit message for in-depth reasoning about that.
- Introduced `ControlConnectionState`, which is now a source of truth about status of CC, and stores all the stuff related to a single connection (event channels, endpoint etc).
- `ControlConnection` is now created for duration of connection lifetime, not once per each fetch. This makes https://github.com/scylladb/scylla-rust-driver/issues/417 easier.

Unblocks: https://github.com/scylladb/scylla-rust-driver/issues/417
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1409

## Pre-review checklist

- [ ] ~~I added relevant tests for new features and bug fixes.~~ - no new features / bug fixes, so nothing to test.
- [ ] ~~I have provided docstrings for the public items that I want to introduce.~~ - no new items
- [ ] ~~I have adjusted the documentation in `./docs/source/`.~~ - nothing to adjust

### DIFF
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -1,3 +1,19 @@
+//! Metadata fetching and processing module.
+//!
+//! This module is responsible for querying cluster metadata from the control connection,
+//! including information about peers (nodes in the cluster), keyspaces, tables, views,
+//! and user-defined types (UDTs).
+//!
+//! The main entry point is [`ControlConnection::query_metadata`], which fetches all
+//! metadata needed to maintain an up-to-date view of the cluster topology and schema.
+//!
+//! Key functionality includes:
+//! - Querying peer information from `system.peers` and `system.local` tables
+//! - Fetching keyspace metadata including replication strategies
+//! - Loading table and materialized view schemas from `system_schema` tables
+//! - Processing user-defined types with proper dependency resolution via topological sort
+//! - Parsing CQL type strings into structured type representations
+
 use std::borrow::BorrowMut;
 use std::cell::Cell;
 use std::collections::HashMap;
--- a/scylla/src/cluster/metadata/reader.rs
+++ b/scylla/src/cluster/metadata/reader.rs
@@ -1,3 +1,18 @@
+//! This module contains the [`MetadataReader`] struct, which is responsible for
+//! fetching and maintaining cluster metadata through a control connection.
+//!
+//! The control connection is a dedicated connection to one of the cluster nodes
+//! that is used to:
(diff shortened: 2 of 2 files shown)


scylla/src/errors.rs
```diff
@@ -778,6 +778,13 @@ pub enum BrokenConnectionErrorKind {
         The connection was already broken for some other reason."
     )]
     ChannelError,
+
+    /// Specific to Control Connection. We tried to poll recevier of error oneshot channel,
```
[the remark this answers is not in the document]
wprzytula: I don't see this resolved.

scylla/src/cluster/metadata_reader.rs
```diff
@@ -102,6 +110,49 @@ impl MetadataReader {
...
+                                // We could panic, but we have a way to recover (by dropping a connection),
+                                // so let's do that (and of course log an error).
+                                error!("Error sender of control connection unexpectedly dropped. This is a bug in the driver, please open an issue!");
+                                // TODO: Should this be a new variant of BrokenConnectionErrorKind?
+                                // This doesn't really fit well because its specific to CC.
+                                // OTOH if we make it a new type, then we lose a guarantee that BrokenConnectionError is
+                                // backed by BrokenConnectionErrorKind.
+                                ConnectionError::from(BrokenConnectionError::from(BrokenConnectionErrorKind::ConnectionUnexpectedlyDropped))
+
+                            },
```
wprzytula: I don't really care about the exact error kind returned in scenarios caused by an undebatable driver bug.
The currently used error is OK.

wprzytula (changes requested):

scylla/src/errors.rs
```diff
@@ -778,6 +779,13 @@ pub enum BrokenConnectionErrorKind {
         The connection was already broken for some other reason."
     )]
     ChannelError,
+
+    /// Specific to Control Connection. We tried to poll receiver of error oneshot channel,
+    /// but it was closed. This should not be possible if we own the Connection.
+    #[error(
+        "Error sender of control connection unexpectedly dropped. This is a bug in the driver, please open an issue!"
+    )]
+    ConnectionUnexpectedlyDropped,
```
wprzytula: :wrench: You replaced use of this variant with `panic!` as I suggested, but the variant is still introduced here!

PR #1531, opened 2026-01-08
Add DeserializeRow support for tuple structs

ziara1:
## Motivation
Users often define data structures using tuple structs (structs with unnamed fields), such as `struct Point(f64, f64)` or `struct UserId(i32)`. Currently, the `DeserializeRow` derive macro only supports structs with named fields, where it maps database columns to Rust fields by name. This limitation forces users to either use named structs (which might not fit their domain model) or manually implement the deserialization traits, resulting in unnecessary boilerplate.

## Solution
This commit extends the `DeserializeRow` derive macro to support tuple structs.

When deriving `DeserializeRow` for a tuple struct, the macro generates an implementation that maps database columns to struct fields based on their **order** (index).
- The 1st column in the result set maps to the 0th field of the struct.
- The 2nd column maps to the 1st field, and so on.

## What's done
* **Refactored `deserialize_row_derive`**
    * Modified the entry point to detect `syn::Fields::Unnamed` (tuple structs).
    * Delegated logic to a new `deserialize_tuple_struct_derive` function for clear separation from named struct logic.

* **Implemented Tuple Logic**
    * **Type Checking:** The generated `type_check` method iterates through the provided `ColumnSpec` slice and the struct fields. It verifies that the database column type matches the expected Rust type for that position. It also verifies that the number of columns matches the number of fields.
    * **Deserialization:** The generated `deserialize` method consumes the `ColumnIterator` sequentially, deserializing each column into the corresponding field of the tuple struct.

* **Validation**
    * Ensured that strict column count checks are in place (errors are returned if the database returns too few or too many columns).

### Tests
Added a suite of unit tests covering:

* **Single-field wrappers:** `struct Wrapper(i32)` (Newtype pattern).
* **Multi-field tuples:** `struct MultiTuple(i32, String, f64)` to verify correct ordering.
* **Type Safety:** Verified that `type_check` correctly returns an error if the column type (e.g., `Text`) does not match the field type (e.g., `Int`).
* **Error Handling:** Verified that mismatched column counts (e.g., empty row for a struct requiring fields) return appropriate errors.

Fixes: #852 

## Pre-review checklist

### DIFF
--- a/scylla-macros/src/deserialize/row.rs
+++ b/scylla-macros/src/deserialize/row.rs
@@ -16,14 +16,24 @@ struct StructAttrs {
     crate_path: Option<syn::Path>,
 
     #[darling(default)]
-    flavor: Flavor,
+    flavor: Option<Flavor>,
 
     // If true, then the type checking code won't verify the column names.
     // Columns will be matched to struct fields based solely on the order.
     //
     // This annotation only works if `enforce_order` is specified.
     #[darling(default)]
-    skip_name_checks: bool,
+    skip_name_checks: Option<bool>,
+}
+
+impl StructAttrs {
+    fn flavor(&self) -> Flavor {
+        self.flavor.unwrap_or(Flavor::MatchByName)
+    }
+
+    fn skip_name_checks(&self) -> bool {
+        self.skip_name_checks.unwrap_or(false)
+    }
 }
 
 impl DeserializeCommonStructAttrs for StructAttrs {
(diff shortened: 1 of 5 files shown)


scylla-macros/src/deserialize/row.rs
```diff
@@ -69,7 +69,15 @@ impl DeserializeCommonFieldAttrs for Field {
...
-    let input = syn::parse(tokens_input)?;
+    let input: syn::DeriveInput = syn::parse(tokens_input)?;
+
+    if let syn::Data::Struct(syn::DataStruct {
+        fields: syn::Fields::Unnamed(fields),
+        ..
+    }) = &input.data
+    {
+        return deserialize_tuple_struct_derive(&input, fields);
+    }
```
Lorak-mmk: Here you are skipping all of normal deserialization implementation. 
I get that `StructDesc` doesn't support unnamed right now, but the solution is to fix that, not skip it altogether. Note that it even has a TODO about it! See lines 60-65 in `deserialize/mod.rs`. Otherwise you are introducing a lot of duplication, that we will need to handle with every future change to this code, and it also makes testing / review much more difficult.

scylla-macros/src/deserialize/row.rs
```diff
@@ -92,6 +100,127 @@ pub(crate) fn deserialize_row_derive(
...
+            }
+            if meta.input.peek(syn::Token![=]) {
+                let _ = meta.value()?.parse::<syn::Expr>()?;
+            }
+
+            Ok(())
+        })?;
+    }
+    Ok(())
+}
```
Lorak-mmk: Why does this manually parse attrs instead of relying on `StructAttrs`?

Lorak-mmk:
Also, we have one more file with macros tests. See `scylla/tests/integration/macros/hygiene.rs`. It verifies that we are not depending on specific paths being imported.

scylla-macros/src/deserialize/row.rs
Lorak-mmk: Thanks to this PR I started to investigate generics support on our macros, and discovered a lot of bugs :D

wprzytula (changes requested):
Although the issue does not specify it (sorry for that), support for unnamed structs should be added both for `(De)SerializeRow` and `(De)SerializeValue`.

PR #1545, opened 2026-01-14
Allow (de)serialization of Cow<'_, T> values

nrxus:
There is already precedence for doing this for Arc<T> and Box<T> so this PR merely extends the same functionality to a another stdlib smart pointer. 

In particular this is useful for me when doing a maybe borrowed deserialization in serde where the string may or may not need to be owned (depending on whether the string was escaped in the input). This maybe borrowed deserialization is done by storing it in a `Cow<'_, str>`. I want to then store that `Cow<'_, str>` as `TEXT` in Scylla without too much work but prior to this PR that was not doable. 

While I personally only care about serializing `Cow<'_, str>` I figured the driver should be more general and allow for any T and for both serialization and deserialization as it is done for Arc and Box. The main difference is that I didn't need to special case Cow<str> as a generic implementation works just fine. 

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/deserialize/value.rs
+++ b/scylla-cql/src/deserialize/value.rs
@@ -1618,6 +1618,25 @@ impl<'frame, 'metadata, T: DeserializeValue<'frame, 'metadata>> DeserializeValue
     }
 }
 
+impl<'frame, 'metadata, T: 'frame + ToOwned + ?Sized> DeserializeValue<'frame, 'metadata>
+    for Cow<'frame, T>
+where
+    &'frame T: DeserializeValue<'frame, 'metadata>,
+{
+    fn type_check(typ: &ColumnType) -> Result<(), TypeCheckError> {
+        <&T>::type_check(typ).map_err(typck_error_replace_rust_name::<Self>)
+    }
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        <&T>::deserialize(typ, v)
+            .map(Cow::Borrowed)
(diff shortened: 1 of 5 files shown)


scylla-cql/src/deserialize/value.rs
```diff
@@ -1618,6 +1618,25 @@ impl<'frame, 'metadata, T: DeserializeValue<'frame, 'metadata>> DeserializeValue
...
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        <&T>::deserialize(typ, v)
+            .map(Cow::Borrowed)
+            .map_err(deser_error_replace_rust_name::<Self>)
+    }
+}
```
Lorak-mmk: It's unfortunate that Rust doesn't have lifetime specialization (yes, I know there are big soundness issues with it). It would be great to use the `Owned` variant when deserializing to `Cow<'static, T>`, but that's not possible (at least for now).

scylla-cql/src/serialize/value_tests.rs
```diff
@@ -2332,6 +2333,15 @@ fn arc_serialization() {
...
 
+#[test]
+fn cow_serialization() {
+    let x: Cow<str> = "123".into();
+    assert_eq!(
+        do_serialize(x, &ColumnType::Native(NativeType::Text)),
+        vec![0, 0, 0, 3, 49, 50, 51]
+    );
+}
+
```
Lorak-mmk: Ditto, see test_box_errors and test_arc_errors

Lorak-mmk (approved):
Thanks!

Issue #1566, opened 2026-02-03
Failure parsing custom types (i.e., Vectors) with frozen types inside


wprzytula:
## Problem

@Akvear noticed the following misbehaviour:

Having a column of type `vector<frozen<set<int>>, 3>`, preparing a statement with a bind variable for that column fails:

`Error: RESULT:Prepared response deserialization failed: Invalid prepared metadata: Invalid column spec: Column spec deserialization failed, column index: 1, error: Invalid column type: Failed to parse custom CQL type: Malformed complex custom type name: FrozenType`

<details><summary>Reproducer by @Akvear</summary>
<p>

```rust
let session: Session = SessionBuilder::new()
        .known_node("172.17.0.2:9042")
        .build()
        .await?;

session
    .query_unpaged(
        "CREATE KEYSPACE IF NOT EXISTS vector_test WITH REPLICATION = \
         {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}",
        &[],
    )
    .await?;

session.use_keyspace("vector_test", false).await?;

session
    .query_unpaged(
        "CREATE TABLE IF NOT EXISTS vector_set_test (
            id int PRIMARY KEY,
            data vector<frozen<set<int>>, 3>
        )",
        &[],
    )
    .await?;

let mut set1 = HashSet::new();
set1.insert(1);
set1.insert(2);
set1.insert(3);

let mut set2 = HashSet::new();
set2.insert(4);
set2.insert(5);

let mut set3 = HashSet::new();
set3.insert(6);
set3.insert(7);
set3.insert(8);

let test_data: Vec<HashSet<i32>> = vec![set1, set2, set3];

session
    .query_unpaged(
        "INSERT INTO vector_set_test (id, data) VALUES (?, ?)",
        (1, &test_data),
    )
    .await?;

#[derive(Debug, DeserializeRow)]
struct Row {
    id: i32,
    data: Vec<HashSet<i32>>,
}

let mut iter = session
    .query_iter("SELECT id, data FROM vector_set_test WHERE id = 1", &[])
    .await?
    .rows_stream::<Row>()?;
```
</p>
</details> 

## Cause

Frozen types are not handled correctly by the custom type parser. Actually, `frozen` flag is always set to `false` there, and `FrozenType` is not recognized as a supported type name.

## Solution

Handle `FrozenType` correctly in custom types.

PR #1567, opened 2026-02-03
custom_type_parser: fix parsing frozen types

wprzytula:
Fixes: #1566

Frozen types were not handled correctly in the custom type parser. When a type was wrapped in `FrozenType(...)`, the parser did not set the `frozen` flag on the inner type. Even worse: `FrozenType` was not recognized at all, leading to parse errors.

This commit introduces a `frozen_context` field in the parser state, which tracks whether we are currently parsing a frozen type. When `FrozenType(...)` is encountered, this field is set to true for the duration of parsing the inner type, and then set back to false.

This commit also adds unit test cases for parsing frozen collections inside vectors, as well as extends the integration tests for vectors with a test case for vectors of frozen collections.

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/types/cql_collections.rs
+++ b/scylla/tests/integration/types/cql_collections.rs
@@ -387,40 +387,47 @@ async fn test_vector_type_unprepared() {
     let ks = unique_keyspace_name();
 
     session.ddl(format!("CREATE KEYSPACE IF NOT EXISTS {ks} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}")).await.unwrap();
+
     session
         .ddl(
             format!(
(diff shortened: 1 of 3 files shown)


scylla/tests/integration/types/cql_collections.rs
```diff
@@ -396,50 +400,94 @@ async fn test_vector_type_unprepared() {
...
+
+    // ######################## VECTORS WITH COLLECTIONS ################################
+    {
+        session
+            .ddl(format!(
+                "CREATE TABLE IF NOT EXISTS {ks}.vector_frozen_collections (
+                id int PRIMARY KEY,
+                data vector<frozen<list<set<int>>>, 3>
+            )"
+            ))
```
Lorak-mmk: No need to create a second table, it will just increase duration of the test.
Change the first table definition to have columns needed for both tests.

Issue #1575, opened 2026-02-26
[scylla 2025.3+] The `LIST ROLES OF reader` admin query returns list of empty dictionaries if we use prepared statement mechanism


vponomaryov:
Latte version:  `0.43.1-scylladb`
Driver version: `1.4.1`

Test runs with the failure using affected scylla versions:
- [scylla-2026.1/rolling-upgrade/upgrade_custom/rolling-upgrade-custom-d2-w1-latency-regression-test#4](https://argus.scylladb.com/tests/scylla-cluster-tests/b90d81cd-8e4a-4e84-afe3-e7b309b4cef9)
- [scylla-staging/valerii/vp-rolling-upgrade-custom-d2-w1-latency-regression#51](https://argus.scylladb.com/tests/scylla-cluster-tests/58f50680-c1b5-4e0c-b8a6-21642a674641)
- [scylla-staging/valerii/vp-rolling-upgrade-custom-d2-w1-latency-regression#50](https://argus.scylladb.com/tests/scylla-cluster-tests/3cc0acb0-6b94-483a-8990-44723147f6fb)

Working test run with the Scylla 2025.1:
- [scylla-2026.1/rolling-upgrade/upgrade_custom/rolling-upgrade-custom-d2-w1-latency-regression-test#4](https://argus.scylladb.com/tests/scylla-cluster-tests/bb533f57-e5a1-4b4b-87d3-b011f7249744)

Latte uses following code for query processing: https://github.com/scylladb/latte/blob/ab3cc567990830ed570e4d43a74c1736af421a67/src/scripting/context.rs#L973-L985

Rune script (`private schema`): https://github.com/scylladb/scylla-qa-internal/blob/78f35a25b66d77d447faae36c970e3f3a5f852f8/custom_d2/workload1/latte/custom_d2_workload1.rn

The issue is reproduced only using Scylla version `2025.3`  and newer.
It is working in Scylla version `2025.2` and older.

### Description:

If we execute `LIST ROLES OF reader` admin CQL command then we get list of empty dicts where we have one dict per existing record.
Error from the referenced rune script:
```
E               164 │         if rec.role == role {
E                   │            ^^^^^^^^ Missing index by static string slot `1`
E
E               Backtrace:
E               /scylla-qa-internal/custom_d2/workload1/latte/custom_d2_workload1.rn:164:12:
E                       if rec.role == role {
[...]

PR #1581, opened 2026-03-04
Fix handling of empty result metadata

Lorak-mmk:
In https://github.com/scylladb/scylla-rust-driver/issues/1575 it was found that driver can't deserialize results of prepared `LIST ROLES OF <role>`.
This issue is a result of two problems:
- Bug in Rust Driver. It doesn't handle the case where result metadata is not returned in PREPARED. This is what this PR fixes. I introduced full handling of empty result metadata, also in reprepare, and added tests for it.
- Weird behavior in Scylla: for such requests, it uses the same metadata id both in PREPARED, and in EXECUTE. So in PREPARED we receive id X, and empty metadata. In EXECUTE we send id X, and server doesn't send us result metadata even though it is different than the (empty) one we received in PREPARED alongside id X. The spec is not very precise about this, but I would consider this a bug, or at least a bad footgun that should be changed.

This PR fixes the driver-side problem.
Now the driver:
- Will send empty result metadata id in EXECUTE if it has empty result metadata locally
- Will no replace non-empty local metadata with an empty one received from server in reprepare
- Will replace empty local metadata with received non-empty one, even if they have the same id.

I introduced a lot of comments with reasoning about the new behavior, and also some tests for edge cases.

AI usage: main changes were done by me, because they consist of small amount of code, that handles tricky edge cases. Tests were written by AI, and then verified and modified by me.
Opus 4.6 is surprisingly good at using our proxy and ccm integration. Produced test logic was also reasonable, but it required some manual adjustments.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1575

## Pre-review checklist


wprzytula (commented):
❓ You added:
- `integration/ccm/result_metadata_extension.rs‎`,
- `integration/statements/prepared_metadata_extension.rs‎`.

Is there a difference between result and prepared metadata extension?

Issue #1535, opened 2026-01-09
Make it possible to make breaking changes in scylla-cql


Lorak-mmk:
When releasing driver 1.0, we had to also release `scylla-cql` 1.0, because it is part of driver public API. The problem is that while some part of `scylla-cql` is present in our public API (like `values` module, ser/deser traits and related structs), the rest is not.
We would like to be able to make breaking changes to those other parts, but it currently isn't possible.

Why?

When making a breaking change in scylla-cql, we of course need to bump the major version, according to Rust semver rules.
Now, what happens if a user has both scylla and scylla-cql im their dependencies and imports some struct using scylla re-exports, and some directly from scylla-cql? Those will be incompatible, because they come from different crates (different versions).

What can we do about it?
My idea is to split scylla-cql into at least 2 crates. One will contain stuff that is part of `scylla` public API, the other all other stuff (like the actual protocol implementation).

Let's call the first crate `scylla-cql-values` (or maybe `scylla-cql-core`?), the other will be called `scylla-cql`. To prevent `scylla-cql` from spilling into public API, we can call `cargo-public-api` in CI, and grep it.

The biggest issue will be isolating those 2 APIs. What is currently re-exported in `scylla`:

- Stuff from `value` module, so types like `CqlDate`, `Counter` etc. This is of course fine.
- Lot of stuff from `serialize` and `deserialize` module. Note, that not everything we want public in `scylla-cql` is re-exported. For example, we would like to publish `serialize_vector` because it will be useful for `python-rs-driver`. But then it would be part of public API of `scylla-cql-values`! Not sure what to do about that. We can't put it in new `scylla-cql`, becuse it has to depend on `scylla-cql-values`. Perhaps we need more than 2 crates to break such cycles.
- Lots of stuff from frame module!
```rust
pub mod frame {
    //! Abstractions of the CQL wire protocol.

    pub use scylla_cql::frame::{Authenticator, Compression, frame_errors};
    pub(crate) use scylla_cql::frame::{
        FrameParams, SerializedRequest, parse_response_body_extensions, protocol_features,
        read_response_frame, request, server_event_type,
    };

    pub mod types {
        //! CQL binary protocol in-wire types.

        pub use scylla_cql::frame::types::{Consistency, SerialConsistency};
    }

    pub mod response {
        //! CQL responses sent by the server.

        pub(crate) use scylla_cql::frame::response::*;

        pub mod result {
[...]

PR #1604, opened 2026-04-13
scylla-cql-split, part 1: New public APIs

Lorak-mmk:
Final PR, if a reviewer wants to look into the bigger picture: https://github.com/scylladb/scylla-rust-driver/pull/1593

## Recap of the problem

We want to split scylla-cql into two crates, introducing a new crate: scylla-cql-core.
Why? scylla-cql is present in scylla public API. This means we cannot make major changes to scylla-cql without also bumping major version of scylla crate.
Unfortunately, not all of scylla-cql API is truly stable, and we need the ability to make breaking changes to it. We were bitten by this twice already, when implementing private link, and when implementing result metadata id.

The idea is to split off a new scylla-cql-core crate that contains parts of scylla-cql that really should be public. At the same time, definitions of those types in scylla-cql would become just re-exports from scylla-cql-core.
This is fine from a semver perspective.
After that, scylla-cql becomes a private dependency (public types in scylla are from scylla-cql-core), so we can introduce 2.0. Users depending on scylla-cql 1.x will keep compiling, and be able to interoperate with scylla crate, because the common parts of the API are present in
scylla-cql-core crate.
This is a bit similar to the https://github.com/dtolnay/semver-trick

## Approach

This is a big change. A lot of commits, a lot of changes, a lot of opportunities for things to go bad.
In order to be sure myself that this change is actually fine, and to allow reviewers to verify it, I plan to introduce it in parts:
- First PR (this one): Publishing some new scylla-cql APIs
- Second PR: Various internal refactors 
- Third PR: Introducing new crate and moving code.

My idea is to make it as obvious as possible that there are no breaking changes.
First PR will publish some APIs, and make other minor changes. Published APIs may be controversial, but there should be no room for a breaking change.
Second PR will move some code around, but mostly test related, and only inside scylla-cql. Its backwards compatibility should not be more difficult to judge than for any other PR.

Third PR will introduce the new crate, and move a lot of code into it. Because of first 2 PRs, it will require basically no other changes, and a diff viewer that highlights moved code differently than added / removed, should make it quite obvious that nothing is breaking there as well.
Note: I had to move much more stuff into core than I initially wanted, because they turned out to be unexpected parts of scylla public API, because of `From` impls, error types, or some niche constructors. We'll be able to fix it in 2.0. For now, the core will contain some unnecessary stuff, but I think it will be a win anyway.

## This PR

In order to split the crates we need to publish some new APIs.
Why? In many places, there is some type / module we want to move to core, but some other code in scylla-cql uses its functions / methods that are currently pub(crate).
We can either move the calling code as well, or publish the methods. Imo publishing the methods is almost always the right choice. In each commit I explained why I think its the correct choice in the given place.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/1535

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/deserialize/frame_slice.rs
+++ b/scylla-cql/src/deserialize/frame_slice.rs
@@ -82,9 +82,8 @@ impl<'frame> FrameSlice<'frame> {
     /// For correctness in an unlikely case that someone calls `to_bytes()` on such
     /// a deficient slice, a special treatment is added there that copies
     /// the slice into a new-allocation-based Bytes.
-    /// This is pub(crate) for the above reason.
     #[inline]
-    pub(crate) fn new_borrowed(frame_subslice: &'frame [u8]) -> Self {
+    pub fn new_borrowed(frame_subslice: &'frame [u8]) -> Self {
         Self {
             frame_subslice,
             original_frame: &EMPTY_BYTES,
--- a/scylla-cql/src/deserialize/frame_slice.rs
+++ b/scylla-cql/src/deserialize/frame_slice.rs
@@ -76,15 +76,16 @@ impl<'frame> FrameSlice<'frame> {
 
     /// Creates a new FrameSlice from a reference to a slice.
     ///
+    /// This should be avoided!
+    ///
     /// This method creates a not-fully-valid FrameSlice that does not hold
     /// the valid original frame Bytes. Thus, it is intended to be used in
     /// legacy code that does not operate on Bytes, but rather on borrowed slice only.
     /// For correctness in an unlikely case that someone calls `to_bytes()` on such
     /// a deficient slice, a special treatment is added there that copies
     /// the slice into a new-allocation-based Bytes.
-    /// This is pub(crate) for the above reason.
     #[inline]
-    pub(crate) fn new_borrowed(frame_subslice: &'frame [u8]) -> Self {
+    pub fn new_borrowed(frame_subslice: &'frame [u8]) -> Self {
         Self {
             frame_subslice,
             original_frame: &EMPTY_BYTES,
--- a/scylla-cql/src/deserialize/mod.rs
+++ b/scylla-cql/src/deserialize/mod.rs
@@ -104,8 +104,6 @@ impl DeserializationError {
 // - BEFORE an error is cloned (because otherwise the Arc::get_mut fails).
 macro_rules! make_error_replace_rust_name {
     ($privacy: vis, $fn_name: ident, $outer_err: ty, $inner_err: ident) => {
-        // Not part of the public API; used in derive macros.
-        #[doc(hidden)]
         #[allow(clippy::needless_pub_self)]
         $privacy fn $fn_name<RustT>(mut err: $outer_err) -> $outer_err {
             let rust_name = std::any::type_name::<RustT>();
--- a/scylla-cql/src/deserialize/mod.rs
+++ b/scylla-cql/src/deserialize/mod.rs
@@ -45,6 +45,15 @@ impl TypeCheckError {
     pub fn downcast_ref<T: std::error::Error + 'static>(&self) -> Option<&T> {
         self.0.downcast_ref()
     }
+
+    /// Try to get a mutable reference to the inner error by downcasting.
+    ///
+    /// This will only succeed if the `Arc` holding the error has no other
+    /// outstanding clones (i.e. the strong count is 1 and there are no weak
+    /// references).
+    pub fn try_downcast_mut<T: std::error::Error + 'static>(&mut self) -> Option<&mut T> {
+        Arc::get_mut(&mut self.0)?.downcast_mut()
+    }
 }
 
 /// An error indicating that a failure happened during deserialization.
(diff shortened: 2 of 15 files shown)


Lorak-mmk:
Not ready after all, found some problems.

scylla-cql/src/serialize/row.rs
```diff
@@ -594,8 +594,7 @@ impl SerializedValues {
     }
 
     /// Creates value list from the request frame
-    /// This is used only for testing - request deserialization.
-    pub(crate) fn new_from_frame(buf: &mut &[u8]) -> Result<Self, LowLevelDeserializationError> {
+    pub fn new_from_frame(buf: &mut &[u8]) -> Result<Self, LowLevelDeserializationError> {
```
wprzytula, line 607: 🔧 While I'm OK with making this `pub`, I think the docstring should be more descriptive then.
For example: what part of the frame should be passed to this function to correctly extract a value list?

scylla-cql/src/serialize/row.rs
```diff
@@ -559,7 +559,8 @@ impl SerializedValues {
         self.serialized_values.len()
     }
 
-    pub(crate) fn write_to_request(&self, buf: &mut impl BufMut) {
+    /// Writes the serialized values to the request buffer, including the preceding u16 element count.
+    pub fn write_to_request(&self, buf: &mut impl BufMut) {
         buf.put_u16(self.element_count);
         buf.put(self.serialized_values.as_slice())
     }
```
wprzytula, line 569: 🔧 Same request: please enhance the docstring.

scylla-cql/src/frame/request/query.rs
```diff
@@ -264,7 +264,7 @@ impl PagingStateResponse {
         matches!(*self, Self::NoMorePages)
     }
 
-    pub(crate) fn new_from_raw_bytes(raw_paging_state: Option<&[u8]>) -> Self {
+    pub fn new_from_raw_bytes(raw_paging_state: Option<&[u8]>) -> Self {
         match raw_paging_state {
             Some(raw_bytes) => Self::HasMorePages {
                 state: PagingState::new_from_raw_bytes(raw_bytes),
```
wprzytula, line 276: 🔧 Missing docstring.

scylla-cql/src/deserialize/frame_slice.rs
```diff
@@ -82,9 +82,8 @@ impl<'frame> FrameSlice<'frame> {
     /// For correctness in an unlikely case that someone calls `to_bytes()` on such
     /// a deficient slice, a special treatment is added there that copies
     /// the slice into a new-allocation-based Bytes.
-    /// This is pub(crate) for the above reason.
     #[inline]
-    pub(crate) fn new_borrowed(frame_subslice: &'frame [u8]) -> Self {
+    pub fn new_borrowed(frame_subslice: &'frame [u8]) -> Self {
         Self {
             frame_subslice,
             original_frame: &EMPTY_BYTES,
```
wprzytula, line 91: 🔧 ❗ Important: please be very clear in the earlier part of this docstring that use of this is discouraged and only left for legacy reasons. I mean, before the wall of text that we have now, put "**Use is discouraged!**".
💡 Maybe let's make this `#[deprecated()]`?

scylla-cql/src/deserialize/mod.rs
```diff
@@ -104,8 +104,6 @@ impl DeserializationError {
 // - BEFORE an error is cloned (because otherwise the Arc::get_mut fails).
 macro_rules! make_error_replace_rust_name {
     ($privacy: vis, $fn_name: ident, $outer_err: ty, $inner_err: ident) => {
-        // Not part of the public API; used in derive macros.
-        #[doc(hidden)]
         #[allow(clippy::needless_pub_self)]
         $privacy fn $fn_name<RustT>(mut err: $outer_err) -> $outer_err {
```
wprzytula, line 109: 🔧 Missing docstrings for now public functions.

scylla-cql/src/frame/request/batch.rs
```diff
@@ -51,8 +51,7 @@ where
 }
 
 /// The type of a batch.
-#[derive(Clone, Copy)]
-#[cfg_attr(test, derive(Debug, PartialEq, Eq))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum BatchType {
```
wprzytula, line 55: 💭 Maybe let's derive `Hash` too?

wprzytula (changes requested):
Looks good. Only minor documentation-related issues.

Lorak-mmk:
Rebased on main

scylla-cql/src/deserialize/frame_slice.rs
Lorak-mmk: I'd rather not make it deprecated, there may be some uses cases for it. I added a warning

Lorak-mmk:
Addressed review comments

wprzytula (approved):
Apart from the failing documentation build, looks good!

wprzytula (approved):
💯

PR #1609, opened 2026-04-16
scylla-cql-split, part 2: Internal refactors

Lorak-mmk:
Final PR, if a reviewer wants to look into the bigger picture: https://github.com/scylladb/scylla-rust-driver/pull/1593

## Recap of the problem

We want to split scylla-cql into two crates, introducing a new crate: scylla-cql-core.
Why? scylla-cql is present in scylla public API. This means we cannot make major changes to scylla-cql without also bumping major version of scylla crate.
Unfortunately, not all of scylla-cql API is truly stable, and we need the ability to make breaking changes to it. We were bitten by this twice already, when implementing private link, and when implementing result metadata id.

The idea is to split off a new scylla-cql-core crate that contains parts of scylla-cql that really should be public. At the same time, definitions of those types in scylla-cql would become just re-exports from scylla-cql-core.
This is fine from a semver perspective.
After that, scylla-cql becomes a private dependency (public types in scylla are from scylla-cql-core), so we can introduce 2.0. Users depending on scylla-cql 1.x will keep compiling, and be able to interoperate with scylla crate, because the common parts of the API are present in
scylla-cql-core crate.
This is a bit similar to the https://github.com/dtolnay/semver-trick

## Approach

This is a big change. A lot of commits, a lot of changes, a lot of opportunities for things to go bad.
In order to be sure myself that this change is actually fine, and to allow reviewers to verify it, I plan to introduce it in parts:
- First PR (this one): Publishing some new scylla-cql APIs
- Second PR: Various internal refactors 
- Third PR: Introducing new crate and moving code.

My idea is to make it as obvious as possible that there are no breaking changes.
First PR will publish some APIs, and make other minor changes. Published APIs may be controversial, but there should be no room for a breaking change.
Second PR will move some code around, but mostly test related, and only inside scylla-cql. Its backwards compatibility should not be more difficult to judge than for any other PR.

Third PR will introduce the new crate, and move a lot of code into it. Because of first 2 PRs, it will require basically no other changes, and a diff viewer that highlights moved code differently than added / removed, should make it quite obvious that nothing is breaking there as well.
Note: I had to move much more stuff into core than I initially wanted, because they turned out to be unexpected parts of scylla public API, because of `From` impls, error types, or some niche constructors. We'll be able to fix it in 2.0 (and some even earlier). For now, the core will contain some unnecessary stuff, but I think it will be a win anyway.

## This PR

This PR performs several refactors in scylla-cql that will allow to move code into scylla-cql-core gradually.
Most important ones include splitting macros tests into a separate module, and making ser/deser tests use public APIs.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/1535

## Pre-review checklist


Lorak-mmk:
closing, I want to try making it a stacked PR

Lorak-mmk:
Ah, stacked PRs cannot be created from forks: https://github.github.com/gh-stack/faq/#can-stacks-be-created-across-forks , so they are basically useless.

scylla-cql/src/serialize/mod.rs
```diff
@@ -43,4 +43,13 @@ impl SerializationError {
     pub fn downcast_ref<T: Error + 'static>(&self) -> Option<&T> {
         self.0.downcast_ref()
     }
+
+    /// Try to get a mutable reference to the inner error by downcasting.
+    ///
+    /// This will only succeed if the `Arc` holding the error has no other
+    /// outstanding clones (i.e. the strong count is 1 and there are no weak
+    /// references).
```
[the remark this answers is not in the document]
Lorak-mmk: This is false, Arc::get_mut requires no weak refs as well.

scylla-cql/src/deserialize/mod.rs
```diff
@@ -45,6 +45,15 @@ impl TypeCheckError {
     pub fn downcast_ref<T: std::error::Error + 'static>(&self) -> Option<&T> {
         self.0.downcast_ref()
     }
+
+    /// Try to get a mutable reference to the inner error by downcasting.
+    ///
+    /// This will only succeed if the `Arc` holding the error has no other
+    /// outstanding clones (i.e. the strong count is 1 and there are no weak
+    /// references).
```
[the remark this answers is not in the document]
Lorak-mmk: This is false, Arc::get_mut requires no weak refs as well.

scylla-cql/src/deserialize/mod.rs
```diff
@@ -77,6 +86,15 @@ impl DeserializationError {
     pub fn downcast_ref<T: Error + 'static>(&self) -> Option<&T> {
         self.0.downcast_ref()
     }
+
+    /// Try to get a mutable reference to the inner error by downcasting.
+    ///
+    /// This will only succeed if the `Arc` holding the error has no other
+    /// outstanding clones (i.e. the strong count is 1 and there are no weak
+    /// references).
```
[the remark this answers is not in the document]
Lorak-mmk: This is false, Arc::get_mut requires no weak refs as well.

scylla-cql/src/value.rs
```diff
@@ -1360,171 +1355,18 @@ impl std::fmt::Display for CqlValue {
...
-                    raw.map(|v| CqlValue::deserialize(typ, Some(FrameSlice::new_borrowed(v))))
-                        .transpose()
-                })
-                .collect::<StdResult<_, _>>()?;
-            CqlValue::Tuple(t)
-        }
-    })
+) -> std::result::Result<CqlValue, DeserializationError> {
+    CqlValue::deserialize(typ, Some(FrameSlice::new_borrowed(buf)))
 }
```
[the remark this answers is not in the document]
Lorak-mmk: Huh, this is actually true - all other paths used FrameSlice, but tuple actually advanced the buffer. Imo this is not a behavior we need to preserve, but please confirm @wprzytula .

scylla-cql/src/deserialize/value_tests.rs
wprzytula, line 1: 🤔 The changes here are different than commit message (of `Remove test methods from CqlValue`) describes.
This commit changes tests that operate on `CqlValue` into tests that don't! Tests used to call `deser_cql_value()`; now, they don't.

scylla-cql/src/macros_doctests.rs
wprzytula, line 1: 🏕️ As we're here, could you add a module-level docstring describing the purpose, as well as what goes here and when?

scylla-cql/src/serialize/row_tests.rs
wprzytula, line 1: ♻️  commit `ser/row_tests: Move macros tests to macros_tests.rs`:
I dislike that tests for both (Serialize)Value and Row derive macros now land in one module. I'd prefer to keep them in respective submodules of `serialize` and `deserialize` modules.

scylla-cql/src/deserialize/mod.rs
wprzytula, line 1: 📝 Commit `deser tests: Use only public APIs`:
I'll review this commit once you resolve the TODO:
> TODO: Describe the ugly hack with storing typck err in DeserializationError that this commit removes

scylla-cql/src/value.rs
```diff
@@ -1361,171 +1355,18 @@ impl std::fmt::Display for CqlValue {
...
 
 /// Deserializes any CQL value from a byte slice according to the provided CQL type.
+///
+/// This delegates to the [`DeserializeValue`] implementation for [`CqlValue`].
+///
+/// The `new_borrowed` version of `FrameSlice` is deficient in that it does not hold
+/// a `Bytes` reference to the frame, only a slice.
+/// This is not a problem here, fortunately, because none of `CqlValue` variants contain
+/// any `Bytes` - only exclusively owned types - so we never call `FrameSlice::to_bytes()`.
 pub fn deser_cql_value(
```
wprzytula, line 1366: 💡 In 2.0, we could make `CqlValue::Blob` contain `Bytes` instead of `Vec<u8>`, enabling zero-copy blob deserialization in dynamic use cases employing CqlValue.

wprzytula: Also, in 2.0 we will be able to delete this function.

wprzytula: WDYT about making it deprecated right now? I don't see why someone would call this instead of using DeserializeValue impl.

wprzytula (changes requested):

Lorak-mmk:
Rebased on main

scylla-cql/src/value.rs
Lorak-mmk: If we don't have to do something in cql-split PRs, let's keep it for later please. Changing and rebasing those commits is already painful enough.

scylla-cql/src/serialize/row_tests.rs
Lorak-mmk: Not sure what you mean. macros_tests.rs has `value` and `row` submodules. Each of them has `serialize` and `deserialize` submodules. I don't see how I could split it further.

scylla-cql/src/deserialize/value_tests.rs
Lorak-mmk: Yes, I guess the commit message is not great. I'll improve it, but let's first agree on the changes.
My goal was removal of pub(crate) methods from CqlValue.
I could either change the tests to match on CqlValue directly without using methods, or change them to not use CqlValue.

Those tests didn't look like a comprehensive testing of CqlValue. I assume they are mostly leftovers from old ser/deser frameworks, and I thought we would get better coverage by removing the usage of CqlValue from them.
I even introduced code to test 2 ways of deserialization: destination type (like NaiveDate) and via intermediate type like CqlDate. 

I guess we should, at some point in the future, properly go trough those tests and organize them, so that they are easier to understand, don't have leftovers etc.

Lorak-mmk:
Addressed some review comments, responded to others.

scylla-cql/src/value.rs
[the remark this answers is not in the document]
wprzytula: I confirm that we don't want to preserve this.

For a moment I thought that changing the parameter type from `&mut T` to `&T` is not a breaking change, but it was too naive.

Anyway, let's document that this `&mut` there is a legacy and users should not expect any buffer advances.

scylla-cql/src/deserialize/value_tests.rs
wprzytula: I agree the changes here are good direction in general. I only fear that we reduce test coverage for the CqlValue itself with this. Do you agree?

scylla-cql/src/serialize/row_tests.rs
wprzytula: I likely meant that they land in a single _file_, even though there are respective submodules. In that case nvm.

scylla-cql/src/value.rs
wprzytula: https://github.com/scylladb/scylla-rust-driver/issues/1643

wprzytula (commented):
I have no more comments for now. Once part 1 is merged and this is rebased, I'll revisit this to make sure I didn't overlook anything.

scylla-cql/src/deserialize/value_tests.rs
Lorak-mmk: Reduce? Sure, but it was already quite bad.

Lorak-mmk:
- Rebased on main
- Edited the commit message for commit that removes test methods from CqlValue
- Edited the comment on deser_cql_value to mention that the buffer is not advanced.

@wprzytula This is now ready for review I think.

scylla-cql/src/deserialize/value_tests.rs
wprzytula: Then let's have an issue to enhance it. WDYT?

wprzytula (approved):
Please investigate the Counter-related Rabbit's comment.

Lorak-mmk:
The comment may be valid, but its not a problem I introduced in this PR. For that reason I think it deserves a separate issue and fix. That way we won't forget about in in changelog.
I'll open an issue. Please confirm this PR can be merged.

wprzytula (approved):

PR #1582, opened 2026-03-09
Support client-routes routing

wprzytula:
Fixes: #1570

Note to reviewers: Both unit and integration tests were created by Claude Opus 4.6 and reviewed by me. It solved that incredibly well!

## Background

PrivateLink routing is based on the recently introduced `system.client_routes` table. Driver needs to fetch it and perform address translation based on its contents.

We were told that some nodes in the cluster may be reachable via ClientRoutes and others directly, so the translator should behave as identity function in case `system.client_routes` misses entry for given node.
However, these scenarios involve inherent races, which result in driver nondeterministically opening connections either directly or via ClientRoutes. Therefore, **we decided to not support such scenarios**. All nodes must be reachable via ClientRoutes.

### New cargo feature: `"unstable-client-routes"`
It guards driver's inner structures and logic regarding fetching and storing `system.client_routes` entries, as well as reacting to `CLIENT_ROUTES_CHANGE` CQL event. Moreover, it guards driver's public API than enables configuring the driver to connect to client routes-dependent architecture, for instance AWS PrivateLink or GCP Private Service Connect.

### Even more mess in scylla-cql

I needed to add a new event type (`CLIENT_ROUTES_CHANGE`). The problem is, neither `Event` nor `EventType` were marked as `#[non_exhaustive]`...
Not to introduce breaking changes, I duplicated the whole family of entities:
- `EventV2`,
- `EventTypeV2`,
- `RegisterV2`,
- `ResponseWithDeserializedMetadataV2`,
- `NonErrorResponseWithDeserializedMetadataV2`.
Ugh. Let's hope it'll become better once we extract `scylla-cql-core`. For now, to reduce clutter in `scylla` crate, I imported new entities with aliases for older names, so that `EventV2` is still seen as `Event` by the code.

### Represent `ClientRoutes`
I introduced entities that represent `system.client_routes` entries. They are fetched in two situtations:
1. Upon metadata refresh - then they are only filtered by configured connection ids - entries for all host ids are fetched.
2. Upon `CLIENT_ROUTES_CHANGE` event - then they are filtered by both connection ids and host ids, those which are contained in the event payload.

### `ClientRoutesSubscriber` trait
[...]

Issue #1570, opened 2026-02-09
PrivateLink support


dkropachev:
Implement PrivateLink support for the Rust driver: dynamic address discovery via `system.client_routes`, `CLIENT_ROUTES_CHANGE` event handling, and Host ID-based address translation.

See scylladb/scylla-drivers#115 for full specification.

Migrated to Jira: [DRIVER-90](https://scylladb.atlassian.net/browse/DRIVER-90)

[DRIVER-90]: https://scylladb.atlassian.net/browse/DRIVER-90?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mykaul:
Team, this is a high priority - let's ensure we focus on this, this week please.

mykaul:
@wprzytula - status?

wprzytula:
> [@wprzytula](https://github.com/wprzytula) - status?

I'm just refining integration tests, then the PR will be fully ready. All other parts of the PR have already been approved by @Lorak-mmk.

PR #1610, opened 2026-04-16
New proxy features

wprzytula:
Refs: #1570

This PR add a few new features to the proxy. They will be useful for proper integration testing of client routes feature, in #1582.

The features include:
- `{RunningProxy,RunningNode}::wait_for_connection()` - enables waiting until there's at least one connection opened to the proxy.
- `RunningNode::inject_event_to_cc()` - injects the given event to all control connections opened to that proxy node.

Relevant unit tests are added for both features.

## Pre-review checklist

### DIFF
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -1,7 +1,8 @@
 use crate::actions::{EvaluationContext, RequestRule, ResponseRule};
 use crate::errors::{DoorkeeperError, ProxyError, WorkerError};
 use crate::frame::{
-    self, FrameOpcode, FrameParams, RequestFrame, ResponseFrame, read_response_frame, write_frame,
+    self, FrameOpcode, FrameParams, RequestFrame, ResponseFrame, ResponseOpcode,
+    read_response_frame, write_frame,
 };
 use crate::{RequestOpcode, TargetShard};
 use bytes::Bytes;
@@ -391,6 +392,7 @@ impl Proxy {
                     active_count: std::sync::atomic::AtomicUsize::new(0),
                     notify: tokio::sync::Notify::new(),
                 });
+                let cc_event_sender = Arc::new(Mutex::new(HashMap::new()));
                 let running = {
                     let (request_rules, response_rules) = match node {
                         InternalNode::Real {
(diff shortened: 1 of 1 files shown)


scylla-proxy/src/proxy.rs
```diff
@@ -444,6 +492,24 @@ impl RunningNode {
...
+            // Prepare the notification future BEFORE checking the count
+            // to avoid a race where a connection arrives between the check
+            // and the await.
+            let notified = self.connection_tracker.notify.notified();
+            if self.connection_tracker.active_count.load(Ordering::Relaxed) > 0 {
+                return;
+            }
+            notified.await;
+        }
+    }
```
Lorak-mmk, line 512: A useful property in tests is knowing that driver is fully connected to a node (connection pool is full, new connections won't be opened). I wanted to have that when writing schema agreement tests.
This function could be extended with a parameter `connections_num` that would be used instead of 0 to achieve that.
There would still be a problem with CC, not sure how to solve that well.

scylla-proxy/src/proxy.rs
```diff
@@ -1537,7 +1604,22 @@ impl ProxyWorker {
                         }
                         let _ = cluster_tx.send(request); // default action
                     }
-                    None => return Ok(()),
+                    None => {
+                        // Connection closed. If this was the control
+                        // connection (REGISTER was seen), remove only this
+                        // connection's sender from the shared map.
+                        if event_registered_flag.load(Ordering::Relaxed) {
```
Lorak-mmk, line 1611: General question to the PR: you use Relaxed everywhere. Are you sure this is correct?

Lorak-mmk (approved):

scylla-proxy/src/proxy.rs
wprzytula: Discussed in priv. Short answer: yes.

scylla-proxy/src/proxy.rs
wprzytula: This can be done in a separate follow-up, once needed.

scylla-proxy/src/proxy.rs
```diff
@@ -1379,7 +1604,22 @@ impl ProxyWorker {
...
+                            info!(
+                                "Control connection {} ({} →  {} ({})) closed; removed cc_event_sender",
+                                connection_no,
+                                driver_addr,
+                                DisplayableRealAddrOption(real_addr),
+                                DisplayableShard(shard),
+                            );
+                        }
+                        return Ok(());
+                    }
```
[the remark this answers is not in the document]
wprzytula: Discussion with Opus convinced me that this is irrelevant.

PR #1618, opened 2026-04-24
Refactor of TCP options

Lorak-mmk:
This PR:
- Puts all TCP options in a single struct instead of storing them inline in ConnectionConfig / HostConnectionConfig
- Moves to setting them before `connect` or `bind`
- Switches from using `SockRef` to builtin Tokio methods for setting them

This last point allowed me to discover that `set_linger` is actually discouraged and deprecated in Tokio because of the possibility of blocking the thread on drop. Because of that, I deprecated this method in the driver as well.
I added a new method to configure 0-wait linger, which mirrors what Tokio does (they also have a method for this, and it is not deprecated).

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -15,7 +15,9 @@ use crate::errors::{
 };
 use crate::frame::response::result;
 use crate::network::tls::TlsProvider;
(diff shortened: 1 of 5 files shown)


wprzytula (dismissed):

Lorak-mmk:
Rebased on main

wprzytula (approved):

PR #1622, opened 2026-04-24
Fix reuseaddr connection

### LINKED ISSUE #1619: Handling occupied port doesn't work when reuseaddr is enabled
Driver has the following logic when using advanced shard awareness:
- Produce an iterator of port numbers to try to bind locally to.
- Iterate over it.
- For each port:
    - Try to open connection using this local port
    - If it works, great. If not, check if the error indicates that the port is occupied
    - If it is, try next port
    - If it is not, return the error

Turns out that when reuseaddr is enabled, a different error (`AddrNotAvailable`) is returned for occupied port, making adv shard awareness not resilient to occupied ports.

Lorak-mmk:
Based on https://github.com/scylladb/scylla-rust-driver/pull/1618
Includes https://github.com/scylladb/scylla-rust-driver/pull/1621

When using REUSEADDR, the advanced shard awareness port iteration logic breaks. This is because a different error is returned in this case, which is not recognized by `is_address_unavailable_for_use`.
This PR slightly refactors test `many_connections` to allow testing with different configs, fixes the bug, and adds a regression test.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1619

## Pre-review checklist

### DIFF
--- a/scylla/src/errors.rs
+++ b/scylla/src/errors.rs
@@ -615,7 +615,9 @@ impl ConnectionError {
     pub fn is_address_unavailable_for_use(&self) -> bool {
         if let ConnectionError::IoError(io_error) = self {
             match io_error.kind() {
-                ErrorKind::AddrInUse | ErrorKind::PermissionDenied => return true,
+                ErrorKind::AddrInUse
+                | ErrorKind::PermissionDenied
+                | ErrorKind::AddrNotAvailable => return true,
                 _ => {}
             }
         }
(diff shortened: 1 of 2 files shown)


wprzytula (dismissed):

Lorak-mmk:
Rebased on main

wprzytula (dismissed):

wprzytula (approved):

PR #1624, opened 2026-04-28
fix(scylla-cql): decode Cassandra VectorType custom marshaller (CEP-30)

bkearns:
## Summary

Cassandra 5.0 (CEP-30) advertises `vector<inner, dim>` columns over the
native protocol as a `Custom`-typed column whose class string is
`org.apache.cassandra.db.marshal.VectorType(<inner_class>, <dim>)`.
The current `Custom` branch in `deser_cql_value` unconditionally errors
with `CustomTypeNotSupported`, so any read of a vector column from a
Cassandra-compatible server (Cassandra 5.x and CQL-compatible engines
such as ferrosa) fails with:

```
DeserializationError: Failed to deserialize Rust type
scylla_cql::frame::response::result::CqlValue from CQL type
Custom("org.apache.cassandra.db.marshal.VectorType(...FloatType, 768)"):
Support for custom types is not yet implemented
```

This patch parses the marshaller class, dispatches on the inner
fixed-size element type, and unpacks `dim` contiguous big-endian
primitives into `CqlValue::List`.

Inner types covered in this PR:

- `FloatType`  → `CqlValue::Float` (4-byte BE)
- `DoubleType` → `CqlValue::Double` (8-byte BE)
- `Int32Type`  → `CqlValue::Int` (4-byte BE)
- `LongType`   → `CqlValue::BigInt` (8-byte BE)
- `ShortType`  → `CqlValue::SmallInt` (2-byte BE)
- `ByteType`   → `CqlValue::TinyInt` (1 byte)

Variable-size element types (text, UDTs, and the VInt-prefixed
encoding) are intentionally out of scope for this minimal patch and
still surface as `CustomTypeNotSupported`. Same for unrelated custom
classes — the existing error contract is preserved.

## Design notes

- **No public API changes.** Vector cells deserialize as `CqlValue::List`,
  which keeps the patch surface small and lets downstream code that
  already knows it's reading an embedding column treat the result like
  any other list of primitives. A first-class `CqlValue::Vector` variant
  + `ColumnType::Vector` parsing path could be layered on top later;
  this PR avoids breaking-change debate.
- **Class-name parsing is whitespace-tolerant only where Cassandra emits
  whitespace.** The parser accepts `, ` between the inner class and the
  dimension because that's what Cassandra writes; the inner class name
  is `trim()`-ed defensively.
- **Truncated input is reported, never panicked.** Each per-element
  reader returns `Option`; a short buffer surfaces a
  `DeserializationError` carrying the original class string plus a
  `(truncated buffer reading vector element)` suffix.

## Test plan

- [x] New unit tests in `scylla-cql/src/frame/response/result.rs`
  covering each supported inner type, the production 768-dim case,
  truncated-buffer handling, and a regression for unrelated custom
  classes (still errors).
- [x] Full `scylla-cql` lib test suite green (227/227, no regressions).
- [x] Verified end-to-end against a Cassandra-compatible server with
  `vector<float, 768>` columns: previously-failing
  `SELECT ... entity_embedding ...` reads now return populated lists.


Lorak-mmk:
I think your change may be a bit stale.
1. You have a ton of conflicts, and merge commits on your branch. You should rebase on main
2. What you try to achieve is done for a long time now. See `get_complex_abstract_type`. Vector type is supported, class name (`org.apache.cassandra.db.marshal.VectorType`) is correctly parsed. What problem are you trying to solve? Are you sure you are using the current version of the driver?

Lorak-mmk:
Closing for now. Looks like misguided reimplementation of functionality we already have. Please reopen if you think I'm wrong.

Issue #1239, opened 2025-02-13
StatusChange events should be utilized as hints for keepaliver and refiller


Lorak-mmk:
We can't use those events as exact oracles for node status for multiple reasons:
- Events may be dropped
- Node being up does not always imply that client can connect to it (there may be network issues)
- I think there are other scenarios too

We can still utilize those events as hints.
- When we receive UP event, then we should tell the pool refiller to try to refill immediately instead of waiting (up to 10s) that way we will regain connections quicker
- When we receive DOWN event we should tell the keepaliver to send a keepalive - that way driver will realize quicker that the connection is broken.

Lorak-mmk:
Btw I think we should only use them as hints, and not store the status in `Node` at all, because it is not useful.

PR #1630, opened 2026-05-07
connection pool: introduce refill trigger and use it upon client routes change and STATUS_CHANGE UP

wprzytula:
Refs: #1615
Refs: #1239

## Solved problem
When client routes change (e.g., during rolling restarts or NLB port remaps), connection pools for affected nodes may be in exponential backoff due to failed connections to stale proxy addresses. The driver waits for the backoff timer to expire before retrying — up to 10 seconds — even though it already knows the new address.

Similarly, when a node comes back up after being down, the pool may be deep in exponential backoff. The driver receives a `STATUS_CHANGE UP` event but previously ignored it entirely, so reconnection was delayed until the next backoff-driven retry.

This PR makes the driver immediately retry connections in both cases.

## Changes

**Commit 1: `connection_pool: expose immediate refill trigger`**

Adds `trigger_immediate_refill()` to `PoolRefiller`, which sends a notification that bypasses the exponential backoff timer. `Node` gets a corresponding `trigger_pool_refill()` method.

**Commit 2: `client_routes: return updated host IDs from route methods`**

`replace_client_routes()` and `merge_client_routes_update()` now return `HashSet<Uuid>` — the host IDs whose routes were added or changed. Adds 13 unit tests covering all return-value cases (unchanged routes, changed hostname/port/tls\_port, new hosts, removed hosts, etc.). The returned values are not yet wired to callers.

**Commit 3: `trigger immediate pool refill upon client routes update`**

Wires the host IDs from commit 2 into pool refill triggering through two paths:
- **Full refresh**: `fetching.rs` → `Metadata.client_routes_updated_hosts` → `worker.rs perform_refresh()` → `ClusterState::trigger_pool_refills_for_hosts()`
- **Event**: `reader.rs fetch_client_route_updates_on_event()` returns `HashSet<Uuid>` → `worker.rs` → `ClusterState::trigger_pool_refills_for_hosts()`

Reduces `CONNECTION_WAIT_TIMEOUT` in integration tests from 10s to 3s, since pools now recover much faster.

**Commit 4: `connection_pool: trigger immediate refill on STATUS_CHANGE UP`**

When the driver receives a `STATUS_CHANGE UP` event for a node, triggers an immediate pool refill, resetting accumulated backoff. Previously, `STATUS_CHANGE` events were completely ignored; the existing TODO comment suggested treating UP events as hints to open new connections — this commit implements exactly that. DOWN events remain ignored: the pool notices the node is down when connections fail and enters backoff naturally.

## Testing
- 13 new unit tests for the host ID returning logic in `ClientRoutesAddressTranslator`
- Existing integration tests (`ccm::client_routes::*`) exercise the full path

## Future possibilities

As part of the ongoing metadata handling refactor, the refill triggering mechanism may be useful.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/client_routes.rs
+++ b/scylla/src/client/client_routes.rs
@@ -100,17 +100,23 @@ pub(crate) trait ClientRoutesSubscriber: Send + Sync {
     /// Replaces the old knowledge about client routes with a new full snapshot.
     /// The snapshot contains all entries that match any of connection ids yielded by
     /// [Self::get_connection_ids]. In particular, no filtering by host ids is done.
-    fn replace_client_routes(&self, client_routes: ClientRoutes);
+    ///
+    /// Returns the set of host IDs whose routes were added or changed
+    /// compared to the previous state (i.e., hosts that may need pool refill).
+    fn replace_client_routes(&self, client_routes: ClientRoutes) -> HashSet<Uuid>;
 
     /// Merges existing knowledge about client routes with a partial snapshot,
     /// fetched in response to a CLIENT_ROUTES_CHANGE:UPDATE_NODES event.
     /// The snapshot contains all entries that match connection ids and host ids
     /// present in the event.
+    ///
+    /// Returns the set of host IDs that were added or updated (i.e., hosts
+    /// that now have a route and may need pool refill).
     fn merge_client_routes_update(
         &self,
         event: &ClientRoutesChangeEvent,
         client_routes: ClientRoutes,
-    );
+    ) -> HashSet<Uuid>;
 }
 
 /// Translator for client routes: uses content of system.client_routes
@@ -202,8 +208,10 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
         &self.connection_ids
     }
 
-    fn replace_client_routes(&self, mut new_routes: ClientRoutes) {
+    fn replace_client_routes(&self, mut new_routes: ClientRoutes) -> HashSet<Uuid> {
         let mut guard = self.client_routes.write().unwrap();
+
+        let mut added_or_updated_host_ids: HashSet<Uuid> = HashSet::new();
 
         // Handle existing entries, i.e. ones for hosts that we already had routes to:
         // - remove dangling routes;
@@ -235,6 +243,13 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
                         }
                     };
 
+                    // Detect if the route actually changed — if so, the pool may
+                    // need an immediate refill (e.g. it could be in exponential
+                    // backoff due to a stale address).
+                    if known_host_routes.sticky_route != new_sticky_route {
+                        added_or_updated_host_ids.insert(*host_id);
+                    }
+
                     // We always replace the sticky route because it could get updated
                     // (wrt hostname or ports, for example).
                     known_host_routes.sticky_route = new_sticky_route;
@@ -255,6 +270,7 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
 
         // Add entries for hosts that weren't present in the map before.
         // Entries for previously present hosts have already been removed from the new entries map.
+        // These are newly routable hosts — their pools may need an immediate refill.
         guard.extend(
             new_routes
                 .routes
@@ -274,16 +290,19 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
                         sticky_route,
                         other_routes,
                     };
+                    added_or_updated_host_ids.insert(host_id);
                     Some((host_id, host_routes))
                 }),
         );
+
+        added_or_updated_host_ids
     }
 
     fn merge_client_routes_update(
         &self,
         event: &ClientRoutesChangeEvent,
         new_routes: ClientRoutes,
-    ) {
+    ) -> HashSet<Uuid> {
         #[deny(clippy::wildcard_enum_match_arm)]
         let (connection_ids, host_ids) = match event {
             ClientRoutesChangeEvent::UpdateNodes {
@@ -295,6 +314,7 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
         };
 
         let mut guard = self.client_routes.write().unwrap();
+        let mut added_or_updated_host_ids = HashSet::new();
 
         // Iterate over all entries listed in the event.
         // Filter only those with connection ids that the driver cares about.
@@ -337,6 +357,7 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
                 Some(new_route) => {
                     // Route specified in the event is present, which means that the event reported route creation or update.
                     // In both cases, we just insert it. Let's just handle the sticky route update separately.
+                    added_or_updated_host_ids.insert(host_id);
                     match guard.get_mut(&host_id) {
                         // There have already been routes for that host.
                         Some(host_routes) => {
@@ -363,6 +384,8 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
                 }
             }
         }
+
+        added_or_updated_host_ids
     }
 }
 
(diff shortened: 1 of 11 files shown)


scylla/src/network/connection_pool.rs
```diff
@@ -617,6 +638,17 @@ impl PoolRefiller {
...
                 }
+
+                _ = self.refill_now_notify.notified() => {
+                    debug!(
+                        "[{}] Immediate refill requested, resetting backoff",
+                        self.endpoint_description()
+                    );
+                    self.refill_delay_strategy.on_successful_fill();
+                    self.had_error_since_last_refill = false;
+                    self.start_filling();
```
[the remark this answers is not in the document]
Lorak-mmk: I agree that this and other related cases need to be considered. One way is to guard with `refill_scheduled`.  Is that fully correct? I don't know. But those things need to be considered and described - the code of the pool is quite tricky and I think it is easy to introduce bugs.

scylla/src/client/client_routes.rs
```diff
@@ -235,6 +243,17 @@ impl ClientRoutesSubscriber for ClientRoutesAddressTranslator {
...
 
+                    // Detect if the route actually changed — if so, the pool may
+                    // need an immediate refill (e.g. it could be in exponential
+                    // backoff due to a stale address).
+                    let old = &known_host_routes.sticky_route;
+                    if old.hostname != new_sticky_route.hostname
+                        || old.port != new_sticky_route.port
+                        || old.tls_port != new_sticky_route.tls_port
+                    {
+                        added_or_updated_host_ids.insert(*host_id);
```
Lorak-mmk, line 250: Why are you comparing fields manually, instead of comparing whole structs? Is it to intentionally skip connection_id?

scylla/src/cluster/worker.rs
```diff
@@ -361,7 +361,10 @@ impl ClusterWorker {
                                 Event::ClientRoutesChange(evt) => {
                                     let res = self.metadata_reader.fetch_client_route_updates_on_event(&evt).await;
                                     match res {
-                                        Ok(()) => continue, // Don't go to refreshing.
+                                        Ok(updated_hosts) => {
+                                            self.cluster_state.load().trigger_pool_refills_for_hosts(updated_hosts.iter().cloned());
+                                            continue; // Don't go to refreshing.
+                                        }
```
Lorak-mmk, line 368: Nit: `Uuid` is copy, so `copied()` is enough and better conveys that we don't perform expensive clone here.

scylla/src/cluster/worker.rs
```diff
@@ -487,6 +491,9 @@ impl ClusterWorker {
             .await,
         );
 
+        new_cluster_state
+            .trigger_pool_refills_for_hosts(client_routes_updated_hosts.iter().cloned());
+
```
Lorak-mmk, line 518: ditto

Lorak-mmk (changes requested):

wprzytula:
Rebased on main.

scylla/src/client/client_routes.rs
wprzytula: That was the idea, but now I concluded that it's not worth it. Let's derive `PartialEq`.

scylla/src/network/connection_pool.rs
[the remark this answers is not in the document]
wprzytula: Opus produced the following analysis in our lengthy discussion:

> ---
> > 1. Is the logic put in a proper place?
>
> Yes. The refill_now_notify branch at line 642 is inside the same tokio::select! loop as all other pool events. This is the correct place — it's the pool worker's main event loop, and adding a new event source here follows the established pattern (timer, ready connections, connection errors, keyspace requests). The Notify mechanism is also appropriate: it's a lightweight one-shot signal that doesn't need a channel.
> > 2. Does the logic conflict with the tokio::select!?
>
> There are two potential interactions to consider:
> a) Race with the scheduled refill timer (line 603):
> When refill_now_notify fires, it sets refill_scheduled = false, which disables the timer branch. This is correct — the immediate refill supersedes any pending scheduled refill. After the select!, the post-loop scheduling logic at line 658 will re-schedule a future refill if need_filling() is still true (i.e., the pool isn't full yet and no connections are in-flight).
> b) Calling start_filling() while connections are in-flight:
> Both the timer branch (line 605) and the refill_now_notify branch (line 649) call start_filling() without checking is_filling(). This is not new behavior — the timer path already does this. So the refill_now_notify branch is consistent with the existing pattern.
> The consequence is that start_filling() may launch duplicate connection attempts for slots that have in-flight connections (since it counts only established connections, not pending ones). But:
> - Excess connections are routed to self.excess_connections and cleaned up at line 612-618.
> - The pool size (self.conns) never exceeds the target.
> However, there's a subtlety worth noting: the refill_now_notify branch resets the backoff (on_successful_fill()) and clears had_error_since_last_refill. If the pool is currently in backoff due to actual node unavailability (not stale routes), and a spurious route update triggers refill_now_notify, the backoff gets reset unnecessarily. This seems acceptable because:
> - Route updates are infrequent.
> - If the node is truly down, the next fill will fail and re-enter backoff.
> One real concern: refill_scheduled = false on line 650 cancels a pending timer. But since start_filling() just ran, the post-loop logic at 658 will check need_filling() — which returns false while connections are in-flight (is_filling() == true). So a new timer won't be scheduled until all in-flight connections resolve. This is fine and consistent.
> ---
> Bottom line: The logic is in the right place and doesn't conflict with the select! loop. The only "issue" (duplicate in-flight connections) is pre-existing and applies equally to the timer branch.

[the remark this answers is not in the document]
wprzytula: I didn't give up:

> But wait — the concern I raised earlier was wrong. I said "the in-flight connections might be to the old address" — but that's not how it works. The address translation happens at connection time from the current translator state, which has already been updated by the time trigger_pool_refill() is called. The in-flight connections were started before the route update, so yes, they might target the old address. But launching new duplicate connections won't help either — they'll race with the old ones, and the old ones will either succeed (to the old address, which might still work) or fail (and get cleaned up).
> Actually, the key insight is: the in-flight connections to the old address will either succeed or fail on their own. If they fail, handle_ready_connection will handle the error, need_filling() will become true, and the post-loop scheduling at line 658 will schedule a new refill — now with the backoff already reset by the refill_now_notify handler. So the backoff reset is the important part, not the immediate start_filling().
> So the guard if self.need_filling() { self.start_filling(); } is actually fine:
> - It avoids duplicate in-flight connections.
> - The backoff reset still happens (line 647).
> - Once in-flight connections resolve, the post-loop logic (line 658) will schedule a refill with the reset backoff (short delay), using the new route.

> The only thing lost is the immediate start of new connections — instead there's a short delay (the reset backoff's base delay). But that's negligible.
> My revised take: the guard is a good idea. It's cleaner, avoids redundant work, and doesn't sacrifice correctness. The backoff reset on line 647 is the critical part; the start_filling() call is just an optimization for the case where the pool isn't currently filling.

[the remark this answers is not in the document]
wprzytula: I conclude it makes sense to put the guard. Also, I made the trigger branch schedule an immediate refill and do `continue`, to fall into the refill branch immediately.

wprzytula:
Addressed review comments.

wprzytula:
### Client routes tests on main:
```
        PASS [  24.830s] scylla::integration ccm::client_routes::test_client_routes_nlb_port_remap
        PASS [  34.014s] scylla::integration ccm::client_routes::test_client_routes_scale_out
        PASS [  43.870s] scylla::integration ccm::client_routes::test_client_routes_multi_dc_topology_change
        PASS [  44.418s] scylla::integration ccm::client_routes::test_client_routes_event_driven_reroute
        PASS [  59.672s] scylla::integration ccm::client_routes::test_client_routes_rolling_restart
```

### Client routes tests on this branch:
```
        PASS [  22.006s] scylla::integration ccm::client_routes::test_client_routes_nlb_port_remap
        PASS [  33.419s] scylla::integration ccm::client_routes::test_client_routes_scale_out
        PASS [  38.077s] scylla::integration ccm::client_routes::test_client_routes_event_driven_reroute
        PASS [  43.991s] scylla::integration ccm::client_routes::test_client_routes_multi_dc_topology_change
        PASS [  49.006s] scylla::integration ccm::client_routes::test_client_routes_rolling_restart
```

Visible gains for topology-changing tests.

scylla/src/network/connection_pool.rs
```diff
@@ -617,6 +638,26 @@ impl PoolRefiller {
...
+                    if self.need_filling() {
+                        debug!(
+                            "[{}] Pool needs filling, scheduling immediate refill",
+                            self.endpoint_description()
+                        );
+                        // Schedule the refill immediately, without waiting for the backoff delay.
+                        next_refill_time = tokio::time::Instant::now();
+                        refill_scheduled = true;
+                        continue;
+                    }
```
Lorak-mmk: I have to admit that the logic of connection pool is a bit unclear to me.

What we want to achieve here is bypassing the exponential wait.
In this case, doesn't it make sense to only do something if refill was already scheduled?
That way it is a bit more obvious that we are only bypassing the wait here.

It would look like this:
```rust
                _ = self.refill_now_notify.notified() => 
                    if refill_scheduled {
                        debug!(
                            "[{}] Immediate refill requested, resetting backoff",
                            self.endpoint_description()
                        );

                        // This resets the backoff, so the reconnect policy will go back to the initial delay after this.
                        // Why? Maybe we received UP a bit too early, and port will open very soon. It makes sense to not
                        // wait for a long time even if we fail now.
                        self.refill_delay_strategy.on_successful_fill();

                        // Here we do the same thing we would if the refill time arrived naturally.
                        self.had_error_since_last_refill = false;
                        self.start_filling();
                        refill_scheduled = false;
                    } else {
                       // We are not waiting for refill to be started, so there is no point in doing anything.
                        debug!("Ignoring immediate refill request, because no refill was scheduled");
                    }
                }
```

Does it really matter? I'm not sure. It looks to me like the `if` block after select (`if !refill_scheduled && self.need_filling() {`) maintains the invariant that when we enter the `select!`, we have `need_filling => refill_scheduled`. Because of that, the two approaches (checking `need_filling`, checking `refill_scheduled`) should be equivalent.

One argument for my approach is that it doesn't use `continue`, so we keep entering the invariant-maintaining block (`if !refill_scheduled && self.need_filling()`) after each `select!`, which makes the code a bit easier to reason about.

WDYT?

scylla/src/cluster/metadata/mod.rs
```diff
@@ -54,6 +54,11 @@ pub(crate) enum SingleKeyspaceMetadataError {
 pub(crate) struct Metadata {
     pub(crate) peers: Vec<Peer>,
     pub(crate) keyspaces: HashMap<String, Result<Keyspace, SingleKeyspaceMetadataError>>,
+
+    /// Host IDs whose client routes were added or updated during this metadata fetch.
+    /// Used to trigger immediate pool refills for nodes that may have been in backoff
+    /// due to `TranslationError::NoRuleForHost`.
+    pub(crate) client_routes_updated_hosts: HashSet<Uuid>,
 }
```
Lorak-mmk, line 63: It's really weird for me that the client route updates are not part of metadata struct, but a list of updated hosts returned from the translator is part of it.
I guess it can stay that way for now. When refactoring metadata, I plan to remove translator from CC anyway and make client route updates part of metadata update.

Lorak-mmk:
This PR needs `Ref: https://github.com/scylladb/scylla-rust-driver/issues/1239`

scylla/src/network/connection_pool.rs
wprzytula: Refactored as discussed offline.

scylla/src/network/connection_pool.rs
```diff
@@ -574,15 +574,30 @@ impl PoolRefiller {
...
+            // or convert specific date in the future to instant.
+            // 1000 years overflows on macOS, 100 years overflows on FreeBSD.
+            tokio::time::Instant::now() + std::time::Duration::from_secs(86400 * 365 * 30)
+        }
 
         loop {
             tokio::select! {
-                _ = tokio::time::sleep_until(next_refill_time), if refill_scheduled => {
+                // Note that `far_future()` is just to avoid `unwrap()` here; the guard anyway ensures that `scheduled_refill` is `Some`.
+                _ = tokio::time::sleep_until(scheduled_refill.as_ref().map_or(far_future(), |r| r.when)), if scheduled_refill.is_some() => {
```
Lorak-mmk: This makes no sense to me. Guard guarantees that is is `Some`, so it is perfectly fine to unwrap.
Sure it is not very elegant, but that is just a consequence of `tokio::select!` not having `if let`-like capabilities.
It is still better than introducing weird dead code.

Lorak-mmk (changes requested):

scylla/src/network/connection_pool.rs
wprzytula: :cry:

Lorak-mmk (approved):

scylla/src/network/connection_pool.rs
wprzytula: `tokio::select!`'s documentation:

> Additionally, each branch may include an optional if precondition. If the precondition returns false, then the branch is disabled. **The provided <async expression> is still evaluated** but the resulting future is never polled. This capability is useful when using select! within a loop.

`unwrap()` panics, hence the pool refiller panics and driver hangs.
Opus itself proposes to use `far_future()`, which means it has exactly the same fix idea as I had.

Lorak-mmk: Wow I hate this design. I still see no need for `far_future` - Future is not polled so it is dead code. You may as well use 0.

wprzytula: OK.

Lorak-mmk (approved):

PR #1646, opened 2026-05-25
cluster/state: add method to get node by host id

### LINKED ISSUE #1645: implement `ClusterState::get_node()`
> Nit: This is O(N^2) where N is number of nodes, because you have outer `iter` over nodes, and inner `any`. 
> Our `ClusterState` already contains `known_peers` which is a hashmap from uuid. We could expose it to the user with a method like `get_node(uuid: Uuid) -> Option<Node>`

_Originally posted by @Lorak-mmk in https://github.com/scylladb/scylla-rust-driver/pull/1631#discussion_r3299270668_

wprzytula:
This will avoid the need to iterate over all nodes when looking for a specific one by the host id, which is a common operation.

Fixes: #1645

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -290,6 +290,11 @@ impl ClusterState {
         &self.all_nodes
     }
 
+    /// Access details about specific node known to the driver, querying by host id.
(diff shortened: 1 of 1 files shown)


wprzytula:
Implemented Copilot's suggestion.

Lorak-mmk (approved):

PR #1651, opened 2026-05-27
Coordinator: Clarify connection_address semantics

### LINKED ISSUE #1513: Coordinator for a given node may be represented by two different addresses
When connecting to ScyllaDB, the request coordinator that will be present in the query metadata may contain an address with either a shard aware port (19042 by default) or non shard aware port (9042). This behaviour is non-deterministic, meaning the coordinator present in response may change between runs, despite the same node answering queries given query.

Lorak-mmk:
Connection address may be different for different connections to the same node. This was not obvious to users, who may try to identify the node using this address. This commit clarifies the semantics in the doc comment.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1513

## Pre-review checklist

### DIFF
--- a/scylla/src/response/coordinator.rs
+++ b/scylla/src/response/coordinator.rs
@@ -11,6 +11,11 @@ use crate::{
 #[derive(Debug, Clone)]
 pub struct Coordinator {
     /// Translated address, i.e., one that the connection is opened against.
+    ///
+    /// This may be different than [`Node::address`], and may be different on different
+    /// connections to the same node, because of various driver mechanisms (address translation,
+    /// shard-aware port, client routes). You should never use this
+    /// for node identification. Use [`Node::host_id`] of the [`Self::node`] for that.
     connection_address: SocketAddr,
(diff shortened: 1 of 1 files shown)


scylla/src/response/coordinator.rs
```diff
@@ -11,6 +11,11 @@ use crate::{
 #[derive(Debug, Clone)]
 pub struct Coordinator {
     /// Translated address, i.e., one that the connection is opened against.
+    ///
+    /// This may be different than [`Node::address`], and may be different on different
+    /// connections to the same node, because of various driver mechanisms (address translation,
+    /// shard-aware port, client routes a.k.a private link). You should never use this
```
wprzytula: 🔧 Let's not mention _private link_ here. It's a proprietary Amazon name for mechanism employed in AWS, while _client routes_ are a more general and more flexible mechanism.

wprzytula (changes requested):

wprzytula (approved):

PR #1653, opened 2026-05-27
Fix client-routes tests flakiness upon node decommission

wprzytula:
### Flakiness symptom
ConnectionPoolError(Broken) in `assert_queries_reach_all_nodes()`.

### Analysis
`assert_queries_reach_all_nodes` used to iterate over all nodes in the driver's ClusterState, which could include nodes that had just been decommissioned but whose removal the driver hadn't yet observed. This caused ConnectionPoolError(Broken) because the proxy chain for the decommissioned node was
already shut down.

### Core fix: only query active nodes in assertions

Only target nodes whose host_id is known to the ClientRoutesCluster (i.e., nodes with running proxies).
Assert that each such node is also present in the driver's cluster state to prevent false-positives.

## Pre-review checklist

- ~~[ ] I have provided docstrings for the public items that I want to introduce.

### DIFF
--- a/scylla/tests/integration/ccm/client_routes.rs
+++ b/scylla/tests/integration/ccm/client_routes.rs
@@ -24,9 +24,12 @@ use crate::ccm::lib::client_routes::{
 };
 use crate::ccm::lib::cluster::ClusterOptions;
 use crate::ccm::lib::node::NodeId;
-use crate::utils::{HEALTHCHECK_QUERY, execute_unprepared_statement_on_every_node, setup_tracing};
+use crate::utils::{HEALTHCHECK_QUERY, setup_tracing};
 
+use scylla::client::execution_profile::ExecutionProfile;
 use scylla::client::session::Session;
+use scylla::policies::load_balancing::{NodeIdentifier, SingleTargetLoadBalancingPolicy};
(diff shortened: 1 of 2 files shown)


scylla/tests/integration/ccm/client_routes.rs
```diff
@@ -61,37 +64,50 @@ fn cluster_2dc_2_2() -> ClusterOptions {
...
+            )
+        });
+        let policy = SingleTargetLoadBalancingPolicy::new(NodeIdentifier::Node(node.clone()), None);
+        let execution_profile = ExecutionProfile::builder()
+            .load_balancing_policy(policy)
+            .build();
+        let mut statement: Statement = HEALTHCHECK_QUERY.into();
+        statement.set_execution_profile_handle(Some(execution_profile.into_handle()));
+        session.query_unpaged(statement, &()).await.unwrap();
+    }
```
Lorak-mmk: Previous version called `execute_unprepared_statement_on_every_node` which executes everything concurrently. You version executes serially. Please fix that.

Lorak-mmk (changes requested):
Only one comment, rest looks fine.

Lorak-mmk (approved):

PR #1654, opened 2026-05-27
Handle node attribute changes correctly w.r.t. HostFilter

### LINKED ISSUE #1503: If HostFilter no longer filters out a node after it changed its IP, connections won't be opened anyway
I believe I found a bug in `ClusterState::new` logic.

Suppose a `Peer` was filtered out by the `HostFilter`. Then, once a corresponding `Node` was created, it had no connection pool associated.
Now suppose the node changes it IP. Then suppose it makes `HostFilter` accept the `Peer`, as it was only tough on with the previous IP and now it's OK. Then `Node::inherit_with_ip_changed` constructor will be called based on the old `Node` and... it will create a new `Node` with still no connection pool!

```rust
 for peer in metadata.peers {
  // Take existing Arc<Node> if possible, otherwise create new one
  // Changing rack/datacenter but not ip address seems improbable
  // so we can just create new node and connections then
  let peer_host_id = peer.host_id;
  let peer_address = peer.address;
  let peer_tokens;

  let node: Arc<Node> = match known_peers.get(&peer_host_id) {
      Some(node) if node.datacenter == peer.datacenter && node.rack == peer.rack => {
          let (peer_endpoint, tokens) = peer.into_peer_endpoint_and_tokens();
          peer_tokens = tokens;
          if node.address == peer_address {
              Arc::clone(node)
          } else {
              // If IP changes, the Node struct is recreated, but the underlying pool is preserved and notified about the IP change.
              Arc::new(Node::inherit_with_ip_changed(node, peer_endpoint))
          }
      }
      _ => {
          let is_enabled = host_filter.is_none_or(|f| f.accept(&peer));
          let (peer_endpoint, tokens) = peer.into_peer_endpoint_and_tokens();
          peer_tokens = tokens;
          Arc::new(Node::new(
              peer_endpoint,
              pool_config,
              used_keyspace.clone(),
              is_enabled,
              #[cfg(feature = "metrics")]
              Arc::clone(metrics),
          ))
      }
    };
...
```

Lorak-mmk:
HostFilter can make its decision based on any of `Peer` fields: `host_id`, `address`, `tokens`, `datacenter`, `rack`.
`ClusterState::new` did not take it into accocunt - when address or token list changed, HostFilter was not consulted again.

This PR fixes the logic, making it more robust and imo easier to comprehensively understand.
Now HostFilter is consulted always, and is a source of truth for Node being enabled.
Node object is reused only if all attributes match, and HostFilter didn't change its decision between old and new node.
Pool is reused if both old and new nodes are enabled, and differ at most by tokens and address.

Tests for this logic are added.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1503

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -484,6 +484,11 @@ impl ClusterState {
         // The returned iterator is nonempty, because it returns at least `first_working_pool`.
     }
 
+    #[cfg(test)]
+    fn known_peers(&self) -> &HashMap<Uuid, Arc<Node>> {
+        &self.known_peers
+    }
+
     pub(super) fn update_tablets(&mut self, raw_tablets: Vec<(TableSpec<'static>, RawTablet)>) {
         let replica_translator = |uuid: Uuid| self.known_peers.get(&uuid).cloned();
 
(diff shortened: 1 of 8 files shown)


wprzytula:
> Pool is reused if both old and new nodes are enabled, and differ at most by tokens and address.

It's interesting that according to this, pool is not reused if dc or rack has changed. Why so?

Lorak-mmk:
> It's interesting that according to this, pool is not reused if dc or rack has changed. Why so?

I'm not sure - it has been that way before my PR, and this decision was supported by a comment.
One reason I see for this is that surely node cannot be in two dcs/racks at the same time.
So if we shared a pool, DC-aware LBP routed (old) request to the node using old ClusterState, and some new request did the same using new ClusterState then surely one of them is wrong.

scylla/src/cluster/state.rs
```diff
@@ -170,15 +170,19 @@ impl ClusterState {
...
+                            pool_config,
+                            connectivity_events_sender.clone(),
+                            used_keyspace.clone(),
+                            #[cfg(feature = "metrics")]
+                            Arc::clone(metrics),
+                        )
+                    } else {
+                        Node::new_disabled(peer_endpoint)
+                    };
+                    Arc::new(node)
```
wprzytula: commit `Node: split construction for enabled and disabled`:
💯 I like this change a lot!

scylla/src/cluster/state.rs
```diff
@@ -152,24 +152,22 @@ impl ClusterState {
             // Changing rack/datacenter but not ip address seems improbable
             // so we can just create new node and connections then
             let peer_host_id = peer.host_id;
-            let peer_address = peer.address;
-            let peer_tokens;
+            let is_enabled = host_filter.is_none_or(|f| f.accept(&peer));
+            let (peer_endpoint, peer_tokens) = peer.into_peer_endpoint_and_tokens();
```
wprzytula, line 155: Commit `ClusterState: Small refactor creation`:
⛏️ This would be even more readable if `is_enabled` were hoisted in a separate commit from `peer_endpoint` and `peer_tokens`.

scylla/src/cluster/state.rs
```diff
@@ -484,6 +484,11 @@ impl ClusterState {
         // The returned iterator is nonempty, because it returns at least `first_working_pool`.
     }
 
+    #[cfg(test)]
+    fn known_peers(&self) -> &HashMap<Uuid, Arc<Node>> {
+        &self.known_peers
+    }
```
wprzytula, line 490: 🔧 The issue also suggests to rename `known_peers` to `known_nodes`.

https://github.com/scylladb/scylla-rust-driver/issues/1503#issuecomment-3627583713

wprzytula (approved):
🚀

PR #1656, opened 2026-05-27
Connection pool: Don't panic on out of range shard

### LINKED ISSUE #1560: Driver panics when custom loadbalancing policy returns non-existent shard.
# Problem statement
When implementing `LoadBalancingPolicy` trait for own type user can specify which Node, which Shard (if any) the query will go to.

When invalid Shard (out of range of available Shards) is specified the driver panics.

# Proposed solution

Check if specified Shard is in valid range and default to random Shard when otherwise.

Lorak-mmk:
This has been fixed before, but not well enough.
The panic could still occur in the part that tried to get connections from random shards.
It was not detected, because regression test asked for a shard just one past the end of the array, which was not enough to trigger the error.

This commit both fixes the test (I made sure it is no passing before fixing the connection pool), and fixes the crash in connection pool.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1560

## Pre-review checklist

### DIFF
--- a/scylla/src/network/connection_pool.rs
+++ b/scylla/src/network/connection_pool.rs
@@ -358,7 +358,10 @@ impl NodeConnectionPool {
         }
 
         // If this fails try getting any other in random order
-        let mut shards_to_try: Vec<u16> = (0..shard).chain(shard + 1..nr_shards.get()).collect();
+        // We may attempt the original shard again, but its not a problem.
+        // An iteration of the loop below should be cheap for a shard with
+        // no connections.
+        let mut shards_to_try: Vec<u16> = (0..nr_shards.get()).collect();
 
         let orig_shard = shard;
         while !shards_to_try.is_empty() {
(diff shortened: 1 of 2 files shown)


scylla/src/network/connection_pool.rs
```diff
@@ -358,7 +358,15 @@ impl NodeConnectionPool {
...
-        let mut shards_to_try: Vec<u16> = (0..shard).chain(shard + 1..nr_shards.get()).collect();
+        // `shard` is either out of bounds for `shard_conns`, or `shard_conns[shard]` is empty.
+        // We want to try all other shards, but we have to consider the first case!
+        let mut shards_to_try: Vec<u16> = if shard > nr_shards.get() {
+            (0..nr_shards.get()).collect()
+        } else {
+            (0..shard)
+                .chain(shard.saturating_add(1)..nr_shards.get())
+                .collect()
+        };
```
wprzytula: 💭 IIUC, this ifological dance is to create a Vec of all shard numbers that are not the shard that has already been tried. Correct?

wprzytula (approved):

Lorak-mmk:
Simplified the code a bit.

wprzytula (approved):

PR #1657, opened 2026-05-28
Fix shard aware port range crash (and other problems)

### LINKED ISSUE #1640: Shard-aware port selection panics on narrow valid ranges
[scylla/src/routing/sharding.rs#L121-L185](https://github.com/scylladb/scylla-rust-driver/blob/9226fe6c0fbcdc0965cc274c5552ec47da465d33/scylla/src/routing/sharding.rs#L121-L185)

`iter_source_ports_for_shard_from_range()` assumes every valid range has a port for every shard. A valid range like `49152..=49152` can make `random_range` panic instead of returning no candidates. This is reachable with a user-provided `ShardAwarePortRange`.

Regression example:

```rust
#[test]
fn test_iter_source_ports_for_shard_from_too_small_range_is_empty() {
    let sharder = Sharder::new(ShardCount::new(4).unwrap(), 12);
    let port_range = ShardAwarePortRange::new(49152..=49152).unwrap();

    let ports: Vec<u16> = sharder
        .iter_source_ports_for_shard_from_range(1, &port_range)
        .collect();

    assert!(ports.is_empty());
}
```

### LINKED ISSUE #1679: Try valid shard-aware source ports at unaligned range starts
The shard-aware source-port iterator can skip valid ports at the beginning of an unaligned allowed range.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/scylla/src/routing/sharding.rs#L179

The first candidate is computed from the next multiple of the shard count plus the shard id. If the lower bound already maps to the requested shard, or to a later residue, valid ports before that next multiple are skipped.

Example: with an allowed range of `21371..=42424`, 4 shards, and shard 3, port `21371` maps to shard 3, but the iterator starts later. This can make shard-aware connection opening report no source port without trying every allowed port.

Suggested direction: compute the first candidate by advancing from the lower bound to the first port in range whose residue maps to the requested shard.

Lorak-mmk:
When too small range was provided to `iter_source_ports_for_shard_from_range` it could panic, because `draw_source_port_for_shard_from_range` tried to take random element from an empty range.

Instead of just fixing that I decided to improve the logic of this code in general.
Switched many operations to checked ones, commented on why
asserts / unchecked operations are safe etc. Also we will now use all of
provided port range. Before, we would only use a subset after a first
port mapping to shard 0. Because of that, the test had to be slightly
adjusted.

I advise reviewers to read full new implementations of `draw_source_port_for_shard_from_range`, `calculate_lowest_port_for_shard_in_range` and `iter_source_ports_for_shard_from_range`. It should make more sense than reading diffs.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1640
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1679

## Pre-review checklist

### DIFF
--- a/scylla/src/routing/sharding.rs
+++ b/scylla/src/routing/sharding.rs
@@ -176,8 +176,7 @@ impl Sharder {
 
         // Choose smallest available port number to begin at after wrapping
         // apply the formula from draw_source_port_for_shard for lowest possible gen_range result
-        let first_valid_port =
-            range_start.div_ceil(self.nr_shards.get()) * self.nr_shards.get() + shard as u16;
+        let first_valid_port = range_start.next_multiple_of(self.nr_shards.get()) + shard as u16;
 
         let before_wrap = (starting_port..=*range_end).step_by(self.nr_shards.get().into());
         let after_wrap = (first_valid_port..starting_port).step_by(self.nr_shards.get().into());
@@ -318,7 +317,7 @@ mod tests {
             I: Iterator<Item = u16>,
         {
             let max_port_num = port_range.0.end();
-            let min_port_num = port_range.0.start().div_ceil(nr_shards) * nr_shards;
+            let min_port_num = port_range.0.start().next_multiple_of(nr_shards);
 
             let sharder = Sharder::new(ShardCount::new(nr_shards).unwrap(), 12);
 
--- a/scylla/src/routing/sharding.rs
+++ b/scylla/src/routing/sharding.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::num::NonZeroU16;
 use std::ops::RangeInclusive;
 
+use itertools::Either;
 use rand::Rng as _;
 use thiserror::Error;
 
@@ -126,6 +127,25 @@ impl Sharder {
             shard,
             &ShardAwarePortRange::EPHEMERAL_PORT_RANGE,
         )
+        .expect("Impossible to draw any port from provided range")
+    }
+
+    fn calculate_lowest_port_for_shard_in_range(
+        &self,
+        shard: u16,
+        port_range: &ShardAwarePortRange,
+    ) -> Option<u16> {
+        let (range_start, range_end) = (*port_range.0.start(), *port_range.0.end());
+        let nr_shards = self.nr_shards.get();
+
+        // Calculate how much we need to add to range_start to reach a port assigned to `shard`.
+        let shard_for_first_port = range_start % nr_shards;
+        // `as u16` is lossless, because the last operation is `% nr_shards` which is originally a u16.
+        let offset = ((u32::from(nr_shards) - u32::from(shard_for_first_port) + u32::from(shard))
+            % u32::from(nr_shards)) as u16;
+
+        let first_valid_port = range_start.checked_add(offset)?;
+        (first_valid_port <= range_end).then_some(first_valid_port)
     }
 
     /// Randomly choose a source port `p` such that `shard == shard_of_source_port(p)`.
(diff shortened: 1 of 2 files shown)


wprzytula:
Please rebase on main.

Lorak-mmk:
Rebased on main

Lorak-mmk:
Rebased on main

Lorak-mmk:
Rebased on main

scylla/src/routing/sharding.rs
```diff
@@ -169,19 +202,35 @@ impl Sharder {
...
+        };
 
-        // Choose smallest available port number to begin at after wrapping
-        // apply the formula from draw_source_port_for_shard for lowest possible gen_range result
-        let first_valid_port = range_start.next_multiple_of(self.nr_shards.get()) + shard as u16;
+        let make_valid_ports_iter =
+            || (first_valid_port..=*range_end).step_by(self.nr_shards.get().into());
+        let valid_ports = make_valid_ports_iter();
+        let valid_ports_count =
+            <StepBy<RangeInclusive<u16>> as ExactSizeIterator>::len(&valid_ports);
```
wprzytula: ⛏️ You can write `<_ as ExactSizeIterator>::len` to reduce verbosity.

scylla/src/routing/sharding.rs
```diff
@@ -135,14 +156,26 @@ impl Sharder {
...
+    ) -> Option<u16> {
         assert!(shard < self.nr_shards.get() as u32);
-        let (range_start, range_end) = (port_range.0.start(), port_range.0.end());
-        rand::rng().random_range(
-            (range_start + self.nr_shards.get() - 1)..(range_end - self.nr_shards.get() + 1),
-        ) / self.nr_shards.get()
-            * self.nr_shards.get()
-            + shard as u16
+        // Correct because of the above assert.
+        let shard: u16 = shard as u16;
```
wprzytula, line 161: 🤔 Here you take this approach, and in `iter_source_ports_for_shard_from_range` you use `try_into()` with another early return (compiler can deduce it's not possible because of the `assert!`).

scylla/src/routing/sharding.rs
```diff
@@ -365,4 +414,16 @@ mod tests {
...
+    fn test_iter_source_ports_for_shard_from_too_small_range_is_empty() {
+        let sharder = Sharder::new(ShardCount::new(4).unwrap(), 12);
+        let port_range = ShardAwarePortRange::new(49152..=49152).unwrap();
+
+        let ports: Vec<u16> = sharder
+            .iter_source_ports_for_shard_from_range(1, &port_range)
+            .collect();
+
+        assert!(ports.is_empty());
+    }
```
wprzytula, line 420: 🔧  Could you add a similar test for `draw_source_port_for_shard[_from_range]`? It's also a user-facing API (though not used by the driver itself).

wprzytula (commented):
The maths look correct. I'm unsure about panicking assertions instead of returning None / empty iterator.

scylla/src/routing/sharding.rs
Lorak-mmk: > The maths look correct. I'm unsure about panicking assertions instead of returning None / empty iterator.

The assertions about `shard < self.nr_shards.get() as u32` have been there before my changes, and I think this is a reasonable thing to assert.
The `try_into` in `iter_source_ports_for_shard_from_range` was a leftover - I replaced it with same `as` as in `draw_source_port_for_shard_from_range`

scylla/src/routing/sharding.rs
Lorak-mmk: `draw_source_port_for_shard_from_range` which could get exposed to a small port range is not public.
It is only called by `draw_source_port_for_shard` which uses ephemeral range.

I'll add a test for `draw_source_port_for_shard` using a big shard number: it should panic.

Lorak-mmk:
Addressed review feedback

scylla/src/routing/sharding.rs
wprzytula: You removed one occurrence but not the other.

wprzytula (approved):

Lorak-mmk:
Removed the other occurence of verbose ExactSizeIterator call

wprzytula (approved):

Issue #973, opened 2024-03-26
DefaultPolicy test: change test nodes to be multi shard


Lorak-mmk:
Currently tests for DefaultPolicy use nodes with 1 shard. This means we have no way to test if the policy returns correct shards, not only correct nodes.
We should add a constructor for Node (accessible only by tests) to allow us to specify shard amount and modify the tests to verify returned shards too.

Rayan-and-beyond:
I'd like to take this. I'll keep it focused on a test-only way to construct multi-shard nodes and extend the DefaultPolicy unit coverage to assert shard selection as well as node selection.

Lorak-mmk:
This is imo the kind of issue where reviewing and guiding external contribution would take more work than doing it ourselves.

Rayan-and-beyond:
makes sense, i'll leave this one with you. thanks.

PR #1658, opened 2026-05-28
Configure session rack

### LINKED ISSUE #1440: Add session config for local rack and local dc
Currently setting local DC and locak rack for DefaultPolicy is unergonomic. User needs to create execution profile, create policy (with local dc/rack set), and set the default execution profile on session builder.

Better way would be to have a method on SessionBuilder to set it. Then LBP could retrieve it from `RoutingInfo` / `ClusterState`.
Configuring it this way would also help with adding new parts of the driver that need to know local DC, like https://github.com/scylladb/scylla-rust-driver/issues/1279

Lorak-mmk:
This PR allows configuring preferred DC / Rack on a SessionBuilder level. Until now, it had to be configured on DefaultPolicy.
Why do this?
- Knowing a preferred DC / Rack will be useful to other parts of the driver. For example, we will now be able to make Control Connection prefer nodes in local rack / DC. It could potentially be useful with .e.g processing tablet hints for strongly consistent tables in local mode in the future. In 2.0 we could provide this information automatically to other parts of the driver, for example HostFilter, allowing writing a HostFilter that only connects to local DC, without the need to explicitly give it local dc name.
- Setting up preferred DC / rack is something that pretty much all users should do, so it should be as ergonomic as possible. Before, it required building ExecutionProfile and using DefaultPolicyBuilder:
```rust
        .known_node(uri)
        .default_execution_profile_handle(
            ExecutionProfile::builder()
                .load_balancing_policy(
                    DefaultPolicy::builder()
                        .prefer_datacenter_and_rack("dc1".to_string(), "rack1".to_string())
                        .build(),
                )
                .build()
                .into_handle(),
        )
        .build()
        .await?;
```
After the change this becomes:
```rust
        .known_node(uri)
        .prefer_datacenter_and_rack("dc1".to_string(), "rack1".to_string())
        .build()
        .await?;
```

The existing configuration on DefaultPolicyBuilder is of course preserved.
Now setting preference on DefaultPolicy will override Session-level preference. This means that existing code will continue to work as before: Session-level preference will be Any, possibly overwritten by policy-level preference.

`SessionBuilder` gets 3 new methods: `prefer_datacenter`, `prefer_datacenter_and_rack`, `prefer_no_datacenter`.
`DefaultPolicyBuilder` gets `prefer_no_datacenter` (overrider Session-level preference, makes policy non-dc-aware) and `inherit_location_preference` (fallback to Session-level preference). Note that the fallback is the default, so the latter method is only useful to set back the default changed by earlier call to e.g. `prefer_datacenter`.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1440

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -45,6 +45,7 @@ use crate::policies::load_balancing::{self, LoadBalancingPolicy, RoutingInfo};
 use crate::policies::retry::{RequestInfo, RetryDecision, RetrySession};
 use crate::response::query_result::ColumnSpecs;
 use crate::response::{Coordinator, NonErrorQueryResponse, QueryResponse};
+use crate::routing::NodeLocationPreference;
 use crate::statement::prepared::{PartitionKeyError, PreparedStatement};
 use crate::statement::unprepared::Statement;
 use tracing::{Instrument, error, trace, trace_span, warn};
@@ -1089,6 +1090,7 @@ pub(crate) struct PreparedPagerConfig {
     pub(crate) cluster_state: Arc<ClusterState>,
     #[cfg(feature = "metrics")]
     pub(crate) metrics: Arc<Metrics>,
+    pub(crate) location_preference: Arc<NodeLocationPreference>,
 }
 
 /// An intermediate object that allows to construct a stream over a query
@@ -1218,6 +1220,7 @@ If you are using this API, you are probably doing something wrong."
         execution_profile: Arc<ExecutionProfileInner>,
         cluster_state: Arc<ClusterState>,
         #[cfg(feature = "metrics")] metrics: Arc<Metrics>,
+        node_location_preference: Arc<NodeLocationPreference>,
     ) -> Result<Self, PagerExecutionError> {
         let (sender, receiver) = oneshot::channel::<ResultFirstPage>();
 
@@ -1236,12 +1239,6 @@ If you are using this API, you are probably doing something wrong."
             .map(PageQueryTimeouter::new);
 
         let page_size = statement.get_validated_page_size();
-
-        let routing_info = RoutingInfo {
-            consistency,
-            serial_consistency,
-            ..Default::default()
-        };
 
         let load_balancing_policy = Arc::clone(
             statement
@@ -1273,6 +1270,13 @@ If you are using this API, you are probably doing something wrong."
                         )
                         .await
                 }
+            };
+
+            let routing_info = RoutingInfo {
+                consistency,
+                serial_consistency,
+                node_location_preference: &node_location_preference,
+                ..Default::default()
             };
 
             let query_ref = &statement;
@@ -1373,6 +1377,7 @@ If you are using this API, you are probably doing something wrong."
                 token,
                 table: table_spec,
                 is_confirmed_lwt: config.prepared.is_confirmed_lwt(),
+                node_location_preference: &config.location_preference,
             };
 
             let page_query = |connection: Arc<Connection>,
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -1275,8 +1275,10 @@ If you are using this API, you are probably doing something wrong."
             let routing_info = RoutingInfo {
                 consistency,
                 serial_consistency,
+                token: None,
+                table: None,
+                is_confirmed_lwt: false,
                 node_location_preference: &node_location_preference,
-                ..Default::default()
             };
 
             let query_ref = &statement;
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1305,8 +1305,10 @@ impl Session {
                 .config
                 .serial_consistency
                 .unwrap_or(execution_profile.serial_consistency),
+            token: None,
+            table: None,
+            is_confirmed_lwt: false,
             node_location_preference: &self.node_location_preference,
-            ..Default::default()
         };
 
         let span = RequestSpan::new_query(&statement.contents);
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -36,6 +36,7 @@ use crate::response::query_result::{MaybeFirstRowError, QueryResult, RowsError};
 use crate::response::{
     Coordinator, NonErrorQueryResponse, PagingState, PagingStateResponse, QueryResponse,
 };
+use crate::routing::NodeLocationPreference;
 use crate::routing::partitioner::PartitionerName;
 use crate::routing::{Shard, ShardAwarePortRange};
 use crate::statement::batch::batch_values;
@@ -96,6 +97,7 @@ pub struct Session {
     tracing_info_fetch_attempts: NonZeroU32,
     tracing_info_fetch_interval: Duration,
     tracing_info_fetch_consistency: Consistency,
+    node_location_preference: Arc<NodeLocationPreference>,
     internal_statements: InternalStatements,
 }
 
@@ -139,6 +141,7 @@ impl std::fmt::Debug for Session {
             "tracing_info_fetch_consistency",
             &self.tracing_info_fetch_consistency,
         )
+        .field("node_location_preference", &self.node_location_preference)
         .finish()
     }
 }
@@ -176,6 +179,15 @@ impl From<Arc<rustls::ClientConfig>> for TlsContext {
 #[derive(Clone)]
 #[non_exhaustive]
 pub struct SessionConfig {
+    /// The preferred location of nodes to contact.
+    ///
+    /// When set to a specific datacenter (or datacenter and rack), the driver
+    /// treats matching nodes as "local" and prioritizes them in query plans.
+    /// This acts as a default that load balancing policies can use or override.
+    ///
+    /// Default: [`NodeLocationPreference::Any`].
+    pub node_location_preference: NodeLocationPreference,
+
     /// List of database servers known on Session startup.
     /// Session will connect to these nodes to retrieve information about other nodes in the cluster.
     /// Each node can be represented as a hostname or an IP address.
@@ -383,6 +395,7 @@ impl SessionConfig {
     /// ```
     pub fn new() -> Self {
         SessionConfig {
+            node_location_preference: NodeLocationPreference::Any,
             known_nodes: Vec::new(),
             local_ip_address: None,
             shard_aware_local_port_range: ShardAwarePortRange::EPHEMERAL_PORT_RANGE,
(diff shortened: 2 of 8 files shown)


scylla/src/client/session_builder.rs
```diff
@@ -19,6 +19,30 @@ use std::sync::Arc;
...
+
+    #[expect(unused)]
+    pub(crate) fn rack(&self) -> Option<&str> {
+        match self {
+            Self::Any | Self::Datacenter(_) => None,
+            Self::DatacenterAndRack(_, rack) => Some(rack),
+        }
+    }
+}
+
```
wprzytula: 💡 I'd put this either in `client/mod.rs` or `routing/mod.rs`.

scylla/src/client/session_builder.rs
```diff
@@ -19,21 +19,39 @@ use std::sync::Arc;
 use std::time::Duration;
 use tracing::warn;
 
+/// Describes the preferred location of nodes to contact when executing queries.
```
wprzytula: queries -> statements

scylla/src/policies/load_balancing/default.rs
```diff
@@ -891,7 +891,9 @@ impl DefaultPolicy {
     }
 
     fn effective_preference(&self) -> &NodeLocationPreference {
-        &self.preferences
+        self.preferences
+            .as_ref()
+            .unwrap_or(&NodeLocationPreference::Any)
     }
```
wprzytula: 🤔 How does this borrow check? Is it thanks to some lifetime extension?

wprzytula: https://chatgpt.com/share/6a1d96af-4990-83eb-a466-2f33ebcccbfe

wprzytula: > Conceptually, Rust treats it like:

> ```rust
> static ANY: NodeLocationPreference = NodeLocationPreference::Any;
> 
> fn effective_preference(&self) -> &NodeLocationPreference {
>    self.preferences.as_ref().unwrap_or(&ANY)
> }
> ```

scylla/src/client/session_builder.rs
```diff
@@ -582,6 +582,36 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
+        self.config.node_location_preference =
+            NodeLocationPreference::DatacenterAndRack(datacenter_name, rack_name);
+        self
+    }
+
+    /// Clears any node location preference, treating all nodes equally.
+    pub fn no_node_preference(mut self) -> Self {
+        self.config.node_location_preference = NodeLocationPreference::Any;
+        self
+    }
```
wprzytula, line 572:
```suggestion
    /// Clears any node location preference, treating all nodes equally.
    pub fn prefer_no_datacenter(mut self) -> Self {
        self.config.node_location_preference = NodeLocationPreference::Any;
        self
    }
```

scylla/src/policies/load_balancing/mod.rs
```diff
@@ -57,6 +58,22 @@ pub struct RoutingInfo<'a> {
...
+    fn default() -> Self {
+        Self {
+            consistency: types::Consistency::default(),
+            serial_consistency: None,
+            token: None,
+            table: None,
+            is_confirmed_lwt: false,
+            node_location_preference: &NodeLocationPreference::Any,
+        }
+    }
```
wprzytula, line 76: 🔧 I purposely removed `Default` impl from `RoutingInfo`, because it hid bugs. I'd prefer not to implement it, and instead require explicit construction at each creation spot.

wprzytula: Oh, I see it's not true - you are removing the implementation. Then I must have mistaken this struct with some other struct. And we can't deimpl `Default` now, because this would be a breaking change. Nvm then.

scylla/src/client/pager.rs
```diff
@@ -1237,12 +1240,6 @@ If you are using this API, you are probably doing something wrong."
 
         let page_size = statement.get_validated_page_size();
 
-        let routing_info = RoutingInfo {
-            consistency,
-            serial_consistency,
-            ..Default::default()
-        };
```
wprzytula, line 1244: Ah, this is what I meant in another comment. I did not remove `Default` impl from `RoutingInfo`, but I did add here overrides for `consistency` and `serial_consistency`. This fixes bugs in setting them in the driver.

wprzytula: Ugh, I really consider existence of `Default` impl for `RoutingInfo` and its use in the driver a footgun. I'd rather stop using it in our codepaths.

scylla/src/client/session.rs
```diff
@@ -1472,6 +1474,7 @@ impl Session {
             self.cluster.get_state(),
             #[cfg(feature = "metrics")]
             Arc::clone(&self.metrics),
+            Arc::clone(&self.node_location_preference),
         )
```
wprzytula, line 1480: Commit `Add &NodeLocationPreference to RoutingInfo`:
Commit message says:
> The pager paths default to Any for now.

It's clearly not true.

scylla/src/policies/load_balancing/default.rs
```diff
@@ -890,15 +911,19 @@ impl DefaultPolicy {
...
 
-    fn effective_preference(&self) -> &NodeLocationPreference {
-        self.preferences
-            .as_ref()
-            .unwrap_or(&NodeLocationPreference::Any)
+    fn effective_preference<'a>(
+        &'a self,
+        session_preference: &'a NodeLocationPreference,
+    ) -> &'a NodeLocationPreference {
+        self.preferences.as_ref().unwrap_or(session_preference)
```
wprzytula: ⛏️ I'd accept `&RoutingInfo` here, to prevent misuse by providing preference of different origin as the argument.

scylla/src/policies/load_balancing/default.rs
```diff
@@ -1059,6 +1059,19 @@ impl DefaultPolicyBuilder {
...
 
+    /// Clears the policy-level node location preference, so that the
+    /// session-level preference (from [`SessionConfig`](crate::client::session::SessionConfig))
+    /// is used instead.
+    ///
+    /// This is the default behaviour — calling this method is only needed
+    /// to undo a previous call to [`prefer_datacenter`](Self::prefer_datacenter),
+    /// [`prefer_datacenter_and_rack`](Self::prefer_datacenter_and_rack),
+    /// or [`prefer_any_node`](Self::prefer_any_node).
+    pub fn disable_preference_overwrite(mut self) -> Self {
```
wprzytula: 🔧 At minimum, this should be `override` (not `overwrite`), as it does not mutate what's overriden (which happens as part of the overwrite).

wprzytula: I'm considering `prefer_what_session_prefers` as an alternative name. Aligns much more with other `prefer_*` methods.
After all, the Default LBP always has a preference; sometimes the preference is to just do what Session tells to do.

scylla/src/policies/load_balancing/default.rs
```diff
@@ -2704,6 +2704,101 @@ mod tests {
...
+            let expected = ExpectedGroupsBuilder::new().group([1, 2, 3, 4, 5]).build();
+            test_default_policy_with_given_cluster_and_routing_info(
+                &policy,
+                &cluster,
+                &routing_info,
+                &expected,
+            );
+        }
+    }
 }
```
wprzytula, line 2781: 🔧 Do you think it's worth adding test cases for rack awareness, too?

wprzytula (changes requested):

scylla/src/policies/load_balancing/default.rs
Lorak-mmk: Fun fact: Opus initally didn't think it was possible, and tried introducing such ANY constant. I had to remove it manually.

scylla/src/policies/load_balancing/mod.rs
Lorak-mmk: Yes, I did not add Default impl, just changed it from derive to manual impl.
And I agree that it is a footgun - I initially made some bugs in the PR because of it. And yes, we unfortunately can't remove it now.

scylla/src/client/pager.rs
wprzytula: :camping:  Could you remove uses of `RoutingInfo::default()` in a precommit?

scylla/src/policies/load_balancing/default.rs
Lorak-mmk: Not really. Those tests are written to make sure we correctly fallback to session-level preference, not that we have correct implementation of dc/rack awareness. 
Checking DC is imo enough to verify the fallback.

Lorak-mmk: Also, we have a separate issue for testing racks in DefaultPolicy unit tests: https://github.com/scylladb/scylla-rust-driver/issues/973
I don't think it falls in the scope of this PR.

scylla/src/client/session_builder.rs
Lorak-mmk: I changed to requests, executing statements sounds weird to me.

Lorak-mmk:

wprzytula:
Please first address CodeRabbit's comments, then reask for my review.

scylla/src/client/session_builder.rs
```diff
@@ -541,6 +541,36 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
     }
 
+    /// Sets the preferred datacenter for this session.
+    ///
+    /// Nodes in the given datacenter will be treated as "local" and preferred
+    /// by load balancing policies that do not override this preference.
+    pub fn prefer_datacenter(mut self, datacenter_name: String) -> Self {
+        self.config.node_location_preference = NodeLocationPreference::Datacenter(datacenter_name);
+        self
+    }
```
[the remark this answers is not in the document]
Lorak-mmk: I don't want unnecessary generics. Also it is better for allocations to be explicit, not hidden.

scylla/src/client/session_builder.rs
```diff
@@ -541,6 +541,36 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
+    /// This preference is used by load balancing policies that do not override it.
+    pub fn prefer_datacenter_and_rack(
+        mut self,
+        datacenter_name: String,
+        rack_name: String,
+    ) -> Self {
+        self.config.node_location_preference =
+            NodeLocationPreference::DatacenterAndRack(datacenter_name, rack_name);
+        self
+    }
```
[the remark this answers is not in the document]
Lorak-mmk: I don't want unnecessary generics. Also it is better for allocations to be explicit, not hidden.

wprzytula (approved):

PR #1660, opened 2026-05-29
Tablets: Require first_tablet < last_tablet

### LINKED ISSUE #1641: Tablet payload parsing overflows at i64::MAX
[scylla/src/routing/locator/tablets.rs#L61-L103](https://github.com/scylladb/scylla-rust-driver/blob/9226fe6c0fbcdc0965cc274c5552ec47da465d33/scylla/src/routing/locator/tablets.rs#L61-L103)

`from_custom_payload()` advances the left-open tablet boundary with `first_token + 1`. That overflows on a valid payload when `first_token = i64::MAX` and panics.

Regression example:

```rust
#[test]
fn test_raw_tablet_deser_wraps_after_max_token() {
    let mut custom_payload = HashMap::new();
    let mut data = vec![];

    let value = CqlValue::Tuple(vec![
        Some(CqlValue::BigInt(i64::MAX)),
        Some(CqlValue::BigInt(5678)),
        Some(CqlValue::List(vec![])),
    ]);

    SerializeValue::serialize(&value, &RAW_TABLETS_CQL_TYPE, CellWriter::new(&mut data)).unwrap();

    custom_payload.insert(
        CUSTOM_PAYLOAD_TABLETS_V1_KEY.to_string(),
        Bytes::copy_from_slice(&data[4..]),
    );

    let tablet = RawTablet::from_custom_payload(&custom_payload).unwrap().unwrap();
    assert_eq!(tablet.first_token, Token::new(i64::MIN + 1));
}
```

Lorak-mmk:
Without this invariant we could accept invalid ranges, breaking the search and panicking on addition overflow.
With this invariant:
- `first_token + 1` is correct, because `first_token < last_token <= i64::MAX`
- There won't be possibility of invalid range being inserted into TableTablets, breaking search.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1641

## Pre-review checklist

### DIFF
--- a/scylla/src/routing/locator/tablets.rs
+++ b/scylla/src/routing/locator/tablets.rs
@@ -22,6 +22,10 @@ pub(crate) enum TabletParsingError {
     TypeCheck(#[from] TypeCheckError),
     #[error("Shard id for tablet is negative: {0}")]
     ShardNum(i32),
+    #[error(
+        "First element of tablet payload token range must be strictly smaller than the second, but the range is ({0}, {1}]"
+    )]
+    WrongTokenRange(i64, i64),
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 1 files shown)


wprzytula (approved):
Please verify my understanding.
- token ranges are `(beg, end]` - left-opened.
- token ranges mustn't be empty.
- token ranges mustn't wrap, which is what `beg < end` guards against.
- this means that `i32::MIN` can't belong to any range; all other values can.
- the largest possible range is `(i32::MIN, i32::MAX]` is valid and it contains all tokens but `i32::MIN`.
- the above is OK because `i32::MIN` is a special reserved token value.

Lorak-mmk:
All of that is correct if you switch i32 to i64

PR #1664, opened 2026-06-01
Implement / fix / optimize optional methods and traits for our iterators

### LINKED ISSUE #674: Implement `nth` method for `scylla::locator::ReplicaSetIterator` more efficiently
Manual implementation for `ChainedNTS` variant would be more efficient than just skipping nodes via `.next()` (https://github.com/scylladb/scylla-rust-driver/pull/612#discussion_r1043236397).

https://github.com/scylladb/scylla-rust-driver/blob/adb18f7312dfcf8f200e79a4789d59ff9bf6fe8c/scylla/src/transport/locator/mod.rs#L535-L550

### LINKED ISSUE #1627: FixedLengthBytesSequenceIterator is missing size_hint
When fixing that we should go over other iterators in our crates and check if they are also missing it.

Lorak-mmk:
In our driver we have various `Iterator` implementations. They all have `next` method of course, because it is mandatory.
There are however some optional functionalities that iterators can implement, most notably: `nth` method, `size_hint` method, and `ExactSizeIterator` trait. Our approach to those have been mixed. Various of our iterators are missing some (or all) of those, despite being implementable without problems. some implementations were wrong or suboptimal.

This PR fixes all the problems I could find. 

There are 2 things I'm not sure about (see relevant commit messages):
- Should we have precise type hint and ExactSizeIterator for UdtIterator? There is a TODO message that might affect it but I don't fully understand it
- Do we want `nth` for `VectorIterator`?

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1627
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/674

## Pre-review checklist

### DIFF
--- a/scylla-cql/src/deserialize/value.rs
+++ b/scylla-cql/src/deserialize/value.rs
@@ -1502,6 +1502,13 @@ where
     }
 }
 
+impl<'frame, 'metadata, K, V> ExactSizeIterator for MapIterator<'frame, 'metadata, K, V>
+where
+    K: DeserializeValue<'frame, 'metadata>,
+    V: DeserializeValue<'frame, 'metadata>,
+{
+}
+
 impl<'frame, 'metadata, K, V> DeserializeValue<'frame, 'metadata> for BTreeMap<K, V>
 where
     K: DeserializeValue<'frame, 'metadata> + Ord,
--- a/scylla-cql/src/deserialize/value.rs
+++ b/scylla-cql/src/deserialize/value.rs
@@ -1448,7 +1448,12 @@ where
     type Item = Result<(K, V), DeserializationError>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let raw_k = match self.raw_iter.next() {
+        // Note the lack of `?` - we want each call to `next()` to consume EXACTLY two items
+        // from the inner iterator. This is necessary for correct and precise
+        // `size_hint`, and thus the implementation of `ExactSizeIterator`.
+        let raw_k_result = self.raw_iter.next();
+        let raw_v_result = self.raw_iter.next();
+        let raw_k = match raw_k_result {
             Some(Ok(raw_k)) => raw_k,
             Some(Err(err)) => {
                 return Some(Err(mk_deser_err::<Self>(
@@ -1458,7 +1463,7 @@ where
             }
             None => return None,
         };
-        let raw_v = match self.raw_iter.next() {
+        let raw_v = match raw_v_result {
             Some(Ok(raw_v)) => raw_v,
             Some(Err(err)) => {
                 return Some(Err(mk_deser_err::<Self>(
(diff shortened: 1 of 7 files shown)


scylla-cql/src/deserialize/value.rs
```diff
@@ -1337,6 +1337,39 @@ where
...
+                        BuiltinDeserializationErrorKind::RawCqlBytesReadError(err),
+                    )));
+                }
+                self.remaining -= n;
+            }
+            self.next_constant_length_elem(element_length)
+        } else {
+            // Variable-length elements: fall back to the default sequential behaviour.
+            for _ in 0..n {
+                // Discard the item (which may be Ok or Err); only stop on exhaustion.
```
wprzytula, line 1365: > Is it useful? I don't know. No reasonable use case comes to my mind.
> Perhaps it is best to drop it - this is a question for reviewers.
`nth` has default implementation, so it is also not a problem to drop
this later.

You've already written it, so let's keep it.
But please introduce unit tests for this custom impl. Otherwise it's unwise to keep it.

scylla-cql/src/deserialize/value.rs
```diff
@@ -1814,7 +1814,11 @@ impl<'frame, 'metadata> Iterator for UdtIterator<'frame, 'metadata> {
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        self.raw_iter.size_hint()
+        // The iterator yields exactly one item per remaining field, regardless of whether
+        // the serialized form contains a value for that field or not (missing fields are
+        // reported as `Ok(None)`).
+        let len = self.remaining_fields.len();
+        (len, Some(len))
     }
```
wprzytula, line 1821: > I see `// TODO: Should we fail when there are too many fields?`, but I'm
not sure what it means tbh.

This TODO was originally written by @piodul in #665. As it was packed in the same PR as the deserialization macros, I believe it addressed the same matter as `forbid_excess_udt_fields` option now does.

scylla/src/routing/locator/mod.rs
```diff
@@ -637,13 +637,48 @@ impl<'a> Iterator for ReplicaSetIterator<'a> {
...
+                        *replicas_idx += remaining;
+                        break;
+                    }
+                    // Skip past the rest of the current datacenter.
+                    remaining -= left_in_current;
+                    if *datacenter_idx + 1 < locator.datacenters.len() {
+                        *datacenter_idx += 1;
+                        *replicas_idx = 0;
+                        let datacenter = &locator.datacenters[*datacenter_idx];
+                        let repfactor = *datacenter_repfactors.get(datacenter).unwrap_or(&0);
```
wprzytula, line 679: 1. This performs indexing, which may panic. Explain why it won't, similarly to how you'd justify an `unwrap()` or `expect()`.
2. You use `0` in case `datacenter_repfactors` do not contain given datacenter. Is it correct?

wprzytula: Ad 1.: Now I see you have a check before. Can you use `get()` and pattern match instead of a check, then indexing?

Lorak-mmk:
Rebased on main

scylla/src/routing/locator/mod.rs
Lorak-mmk: I did it this way because it is done this way in preexisting `next` implementation. I thought that keeping them similar will improve readability.
I can add new commit that changes them both to get rid of indexing.

Lorak-mmk: > You use 0 in case datacenter_repfactors do not contain given datacenter. Is it correct?

Yes. This is what `next` does as well.

Lorak-mmk:
- Added AI-generated unit tests for new `nth` implementations
- Added commit that removes direct indexing from `ReplicaSetIterator::next/nth`
-

scylla-cql/src/deserialize/value.rs
Lorak-mmk: I'll drop the ExactSizeIterator impl in case it collides with this.

Lorak-mmk:
- Dropped ExactSizeIterator for UDT Iterator
- Fixed a bug found by CodeRabbit (and added regression test) in a new commit.

wprzytula (approved):
> Dropped ExactSizeIterator for UDT Iterator

Why such decision?

Lorak-mmk:
>     Why such decision?

In case we want to solve the TODO there and it requires the iterator to return different number of items.

Lorak-mmk:
- Added back ExactSizeIterator impl for UdtIterator after discussion with @wprzytula 
- Removed TODO from UdtIterator

Lorak-mmk:
Rebased on main

wprzytula (approved):

PR #1665, opened 2026-06-01
    feat: add transparent dual-write support (unstable-dual-write fea…

tarzanek:
…ture)

    Adds opt-in transparent dual-write to a secondary Scylla/Cassandra
    cluster, gated behind the `unstable-dual-write` Cargo feature flag.

    ## Design

    All DML writes (INSERT, UPDATE, DELETE, BATCH) executed on the primary
    session are automatically mirrored to a secondary cluster in a
    fire-and-forget manner. Errors on the secondary are logged at WARN level
    and never propagated to the caller.

    A dedicated background OS thread owns the secondary Session and runs its
    own single-threaded Tokio runtime with a LocalSet. Tasks are sent over
    an unbounded mpsc channel (Send + 'static data only) and processed
    sequentially to preserve write ordering on the secondary.

    This architecture avoids the non-Send future limitation of the driver's
    generic BatchValues/SerializeRow traits while keeping the hot path on
    the primary thread free of any blocking work.

    ## New public API

    - `SessionBuilder::dual_write_config(config: SessionConfig)` — sets the secondary cluster configuration; the secondary session is connected transparently during `build()`.

    ## What is mirrored

    - `execute_unpaged` / `execute_single_page`: prepared DML writes including LWT (INSERT/UPDATE/DELETE … IF …)
    - `batch`: all-prepared batches (skipped with debug log if any statement is unprepared)
    - `query_unpaged` / `query_single_page`: unprepared DML with empty values (INSERT/UPDATE/DELETE without bound values)
    - `prepare`: eagerly prepares statements on the secondary so its cache is warm before the first execute mirror
    - `use_keyspace`: keyspace switches are mirrored after primary succeeds

    All mirrors are sent only **after** the primary write succeeds.

    ## What is not mirrored

    - SELECT / reads
    - Unprepared queries with non-empty bound values (logged at DEBUG)
    - Batches containing unprepared statements (logged at DEBUG)

    Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -78,6 +78,9 @@ unstable-reconnect-policy = []
 # Enables fetching contents of the `system.client_routes` table.
 # Needed for custom routing (AWS PrivateLink, GCP PSC, etc.).
 unstable-client-routes = []
+# Enables transparent dual-write support: mirrors DML writes (INSERT/UPDATE/DELETE/BATCH)
+# to a secondary cluster in a fire-and-forget manner.
+unstable-dual-write = []
 
 
 [dependencies]
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -50,7 +50,12 @@ use itertools::Itertools;
 use scylla_cql::frame::response::NonErrorResponseWithDeserializedMetadataV2 as NonErrorResponseWithDeserializedMetadata;
 use scylla_cql::frame::response::error::DbError;
 use scylla_cql::serialize::batch::BatchValues;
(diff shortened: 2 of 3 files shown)


Lorak-mmk:
This is not a functionality that any driver ever had AFAIK, and a complexity we are not ready to maintain unless it is absolutely criticial. 

Imo such use cases are better addressed with a proxy - https://github.com/scylladb/scylla-hermes , https://github.com/datastax/zdm-proxy

PR #1666, opened 2026-06-02
Expose cluster name in ClusterState

### LINKED ISSUE #1213: Expose cluster name in metadata
Java and Python drivers provide cluster name in metadata `get_cluster().metadata.cluster_name` in Python, and `sessionRule.session().getMetadata().getClusterName()` in Java.
Rust driver should also support that.

conorbros:
## Pre-review checklist

## Description
Fetches `cluster_name` from the `cluster_name` column in `system.local` during metadata refresh and exposes it via `ClusterState::cluster_name() -> &str`, bringing the Rust driver in line with the Java and Python drivers.

### Changes:

- Extended the `system.local` query to select `cluster_name` via a dedicated `LocalNodeInfoRow` deserialization struct
- Added `cluster_name: Option<String>` to `Metadata` and `ClusterState`
- Exposed `ClusterState::cluster_name() -> &str` as a public API, returning `""` when the cluster name is unavailable
- Added integration test asserting the cluster name is correctly read from the running cluster
- Updated schema documentation
- Added `--cluster-name TestCluster` to the ScyllaDB test cluster docker-compose
- Added `CASSANDRA_CLUSTER_NAME=TestCluster` to the Cassandra test cluster docker-compose

Fixes: #1213

### DIFF
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -49,6 +49,9 @@ pub struct ClusterState {
     /// for a given (token, replication strategy, table) tuple.
     /// It relies on both topology and schema metadata.
     pub(crate) locator: ReplicaLocator,
+
+    /// The name of the cluster, as reported by the `cluster_name` column in `system.local`.
+    pub(crate) cluster_name: Option<String>,
 }
 
 /// Enables printing [ClusterState] struct in a neat way, skipping the clutter involved by
@@ -286,7 +289,13 @@ impl ClusterState {
             known_nodes: new_known_nodes,
             keyspaces,
             locator,
+            cluster_name: metadata.cluster_name,
         }
+    }
+
+    /// Returns the name of the cluster, as reported by the `cluster_name` column in `system.local`.
+    pub fn cluster_name(&self) -> &str {
+        self.cluster_name.as_deref().unwrap_or("")
     }
 
     /// Access keyspace details collected by the driver.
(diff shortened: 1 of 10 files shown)


scylla/src/cluster/state.rs
```diff
@@ -286,9 +289,15 @@ impl ClusterState {
...
             locator,
+            cluster_name: metadata.cluster_name,
         }
     }
 
+    /// Returns the name of the cluster, as reported by the `cluster_name` column in `system.local`.
+    pub fn cluster_name(&self) -> Option<&str> {
+        self.cluster_name.as_deref()
+    }
+
```
Lorak-mmk, line 300: I don't think this should be an Option. I tested, and even if cluster name is not set, it will be an empty string in Scylla. Option will only allow us  to differentiate between empty string and NULL, which is most likely a niche case not appearing in practice. Let's make this just `&str`, with empty string returned if null appeared in DB.

test/cluster/docker-compose.yml
Lorak-mmk, line 1: You edited scylla docker compose, but not cassandra docker compose. This is likely a cause of test failure.

Lorak-mmk (changes requested):
Thanks for the contribution!
Just two minor things:
- Let's make public API non-Option
- Test is failing for Cassandra, please fix it.

When making changes please amend your existing commit instead of adding a new commit.

Lorak-mmk (approved):
Thanks!

PR #1667, opened 2026-06-04
metadata: Expose release_version on Node

### LINKED ISSUE #1286: Add server release version info to metadata

conorbros:
## metadata: Expose release_version on Node and Peer

Adds `release_version: Option<String>` to `Peer` and `Node`, populated from the `release_version` column in `system.peers` and `system.local`.

This mirrors the per-node version API available in other drivers:
- Python driver: `Host.release_version`
- Java driver: `Node.getCassandraVersion()`

## Changes

- Added `release_version` to `NodeInfoRow` and `LocalNodeInfoRow` deserialization structs and extended the `system.peers` / `system.local` queries accordingly
- Added `pub release_version: Option<String>` to `Peer` and `Node`
- Threaded the field through `PeerEndpoint` and all `Node` constructors
- Renamed `inherit_with_ip_changed` to `inherit_preserving_pool` and extended it to handle address and/or `release_version` changes while preserving the connection pool
- Updated node reuse guards in `ClusterState` to detect `release_version` changes and recreate the `Node` struct (while preserving the pool)
- Added assertion to `test_session_should_have_cluster_metadata` integration test

## Design decisions

**Per-node only.** A cluster-level version API was considered (as in the earlier PR #1307) but rejected — other drivers don't expose one, and the concept is semantically ambiguous during rolling upgrades.

**Node reuse on version change.** When `release_version` changes (e.g. during a rolling upgrade), the node reuse guards detect the change and recreate the `Node` struct via `inherit_preserving_pool`, which preserves the connection pool. A rolling upgrade stops the node first dropping all connections, but the pool may already have reconnected by the time metadata refresh detects the version change — recreating the pool unnecessarily is avoided.

**Known limitation.** The integration test validates `release_version` by comparing against `SELECT release_version FROM system.local`, which returns the coordinator node's version. In a homogeneous cluster this is correct; in a mixed-version cluster the assertion could give false positives for non-coordinator nodes.

Fixes: #1286

## Pre-review checklist

  > `schema.md` covers schema metadata (keyspaces, tables, columns, UDTs). Node topology metadata (datacenter, rack, release_version) accessible via `get_nodes_info()` is not documented anywhere in the docs. Given that #1666 recently added `cluster_name` to this page, it may be reasonable to extend it further to cover node metadata — happy to do so if the reviewer agrees.

### DIFF
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -217,6 +217,7 @@ struct NodeInfoRow {
     datacenter: Option<String>,
     rack: Option<String>,
     tokens: Option<Vec<String>>,
+    release_version: Option<String>,
 }
 
 #[derive(DeserializeRow)]
@@ -230,6 +231,7 @@ struct LocalNodeInfoRow {
     rack: Option<String>,
     tokens: Option<Vec<String>>,
     cluster_name: Option<String>,
+    release_version: Option<String>,
 }
 
 #[derive(Clone, Copy)]
@@ -254,7 +256,7 @@ impl ControlConnection {
     ) -> Result<(Vec<Peer>, Option<String>), MetadataError> {
         let peers_query_stream = self
             .query_iter(
-                "SELECT host_id, rpc_address, data_center, rack, tokens FROM system.peers",
+                "SELECT host_id, rpc_address, data_center, rack, tokens, release_version FROM system.peers",
                 &(),
             )
             .map(|pager_res| {
@@ -273,7 +275,11 @@ impl ControlConnection {
             .and_then(|row| future::ok((NodeInfoSource::Peer, row, None::<String>)));
 
         let local_query_stream = self
-            .query_iter("SELECT host_id, rpc_address, data_center, rack, tokens, cluster_name FROM system.local WHERE key='local'", &())
+            .query_iter(
+                "SELECT host_id, rpc_address, data_center, rack, tokens, \
+                 cluster_name, release_version FROM system.local WHERE key='local'",
+                &(),
+            )
             .map(|pager_res| {
                 let pager = pager_res?;
                 let rows_stream = pager.rows_stream::<LocalNodeInfoRow>()?;
@@ -295,6 +301,7 @@ impl ControlConnection {
                     datacenter: row.datacenter,
                     rack: row.rack,
                     tokens: row.tokens,
+                    release_version: row.release_version,
                 };
                 future::ok((NodeInfoSource::Local, node_row, cluster_name))
             });
@@ -349,6 +356,7 @@ impl ControlConnection {
             datacenter,
             rack,
             tokens,
+            release_version,
         } = row;
 
         let host_id = match host_id {
@@ -398,6 +406,7 @@ impl ControlConnection {
             tokens,
             datacenter,
             rack,
+            release_version,
         })
     }
 }
--- a/scylla/src/cluster/metadata/mod.rs
+++ b/scylla/src/cluster/metadata/mod.rs
@@ -76,6 +76,10 @@ pub struct Peer {
     pub datacenter: Option<String>,
     /// Rack this node is in, if known.
     pub rack: Option<String>,
+    /// The release version of the server running on this node, as reported in
+    /// `system.local` / `system.peers`. May be `None` if the column is absent
+    /// or the value is null.
+    pub release_version: Option<String>,
 }
 
 /// An endpoint for a node that the driver is to issue connections to,
@@ -115,6 +119,7 @@ pub(crate) struct PeerEndpoint {
     pub(crate) address: NodeAddr,
     pub(crate) datacenter: Option<String>,
     pub(crate) rack: Option<String>,
+    pub(crate) release_version: Option<String>,
 }
 
 impl Peer {
@@ -124,6 +129,7 @@ impl Peer {
             address: self.address,
             datacenter: self.datacenter.clone(),
             rack: self.rack.clone(),
+            release_version: self.release_version.clone(),
         }
     }
 
@@ -134,6 +140,7 @@ impl Peer {
                 address: self.address,
                 datacenter: self.datacenter,
                 rack: self.rack,
+                release_version: self.release_version,
             },
             self.tokens,
         )
@@ -341,6 +348,7 @@ impl Metadata {
                     datacenter: None,
                     rack: None,
                     host_id: Uuid::new_v4(),
+                    release_version: None,
                 }
             })
             .collect();
--- a/scylla/src/cluster/node.rs
+++ b/scylla/src/cluster/node.rs
@@ -78,9 +78,9 @@ impl Display for NodeAddr {
 
 /// Node represents a cluster node along with its data and connections
 ///
-/// Note: if a Node changes its broadcast address, then it is not longer
-/// represented by the same instance of Node struct, but instead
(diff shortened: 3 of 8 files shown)


Lorak-mmk:
> Per-node only. A cluster-level version API was considered (as in the earlier PR https://github.com/scylladb/scylla-rust-driver/pull/1307) but rejected — other drivers don't expose one, and the concept is semantically ambiguous during rolling upgrades.

I don't think it was rejected yet. It is not obvious what semantics such cluster-level API should have, but it does not mean we don't want it. It still needs discussion.

I see 3 options here:
- `ClusterState` version is `Some` only if all `Node` versions are the same
- `ClusterState` version is the version of ControlConnection node
- Not have cluster-level option at all

@wprzytula @conorbros WDYT?

> Added release_version to NodeInfoRow and LocalNodeInfoRow deserialization structs and extended the system.peers / system.local queries accordingly

One problem with doing it that way is that the version is basically useless for Scylla, because Scylla hardcodes its Cassandra version in system.local/system.peers to some ancient version (3.0.8 I think?).
Scylla versions can be retrieved from `system.versions` table (or something like that, I don't remember precisely).

Imo it is important the any new version API makes sense for both Scylla and Cassandra.
What options I see:
1. Expose both `scylla_version` and `cassandra_version` fields
2. For Scylla clusters ignore Cassandra version and store the Scylla version in the new field

1 is more "precise" - for Scylla clusters you can check both the Scylla version and the Cassandra version it is advertising.
2 is most likely more pleasant to use - you have one field, with a server version, regardless of which server you connected to.

Now I think there is also option 3: store both versions (so option 2), but have a method that returns the relevant version (so has semantics of option 1).

@wprzytula @conorbros  WDYT?

wprzytula:
> > Per-node only. A cluster-level version API was considered (as in the earlier PR #1307) but rejected — other drivers don't expose one, and the concept is semantically ambiguous during rolling upgrades.
> 
> I don't think it was rejected yet. It is not obvious what semantics such cluster-level API should have, but it does not mean we don't want it. It still needs discussion.
> 
> I see 3 options here:
> 
>   * `ClusterState` version is `Some` only if all `Node` versions are the same
> 
>   * `ClusterState` version is the version of ControlConnection node
> 
>   * Not have cluster-level option at all
> 
> 
> @wprzytula @conorbros WDYT?

> other drivers don't expose one, and the concept is semantically ambiguous during rolling upgrades.

This convinces me that such API is both needless and problematic. Let's not have it.

> > Added release_version to NodeInfoRow and LocalNodeInfoRow deserialization structs and extended the system.peers / system.local queries accordingly
> 
> One problem with doing it that way is that the version is basically useless for Scylla, because Scylla hardcodes its Cassandra version in system.local/system.peers to some ancient version (3.0.8 I think?). Scylla versions can be retrieved from `system.versions` table (or something like that, I don't remember precisely).
> 
> Imo it is important the any new version API makes sense for both Scylla and Cassandra. What options I see:
> 
>   1. Expose both `scylla_version` and `cassandra_version` fields
> 
>   2. For Scylla clusters ignore Cassandra version and store the Scylla version in the new field
> 
> 
> 1 is more "precise" - for Scylla clusters you can check both the Scylla version and the Cassandra version it is advertising. 2 is most likely more pleasant to use - you have one field, with a server version, regardless of which server you connected to.
> 
> Now I think there is also option 3: store both versions (so option 2), but have a method that returns the relevant version (so has semantics of option 1).
> 
> @wprzytula @conorbros WDYT?

Option 3 sounds best.
If it shows problematic, we can deprecate it later and implement option 1.

Lorak-mmk (commented):
See the above comment

Lorak-mmk:
> Option 3 sounds best.
> If it shows problematic, we can deprecate it later and implement option 1.

Just to make sure, because I made a small mistake in my comment:
Option 3 is to have both version fields (/methods) exposed to user (exactly the same as in option 1), but also have a method that returns the relevant version (scylla version for scylla clusters, cassandra version for cassandra clusters).
Is that fine for you?

conorbros:
**Cluster-level version API**

Agree with push back on adding a cluster-level API.
A cluster-level `release_version` implies that "the cluster has a version", but in a distributed system only nodes have versions, and they can differ. Whatever semantics we choose for a cluster-level field break in the scenarios where version information is needed:

- `Some` only when all nodes agree will become `None` mid-rolling-upgrade, leaving the consumer with no version information.
- Using the version of the `ControlConnection` might not be representative of all nodes in the cluster.

I think we should keep per-node `release_version`. A consumer that needs "what is the minimum version across the cluster?" or "are all nodes on the same version?" can derive that from `get_nodes_info()`. Some ClusterState convenience methods could be a good solution too, e.g. `nodes_minimum_release_version()` or `nodes_agreed_release_version()`. These would be explicit about what they compute rather than implying a single cluster version exists.

**Scylla vs Cassandra version**

`system.versions` is a local virtual table, it exposes the version of the node you're directly connected to. Not gossiped and there is no per-peer equivalent. 

All metadata in this driver is fetched through the control connection using `system.local` and `system.peers`, the metadata reader does not connect to each peer individually so `system.versions` would only give us the Scylla version for the control connection's node. 

The approach is a parallel query of system.versions per-node as part of metadata refresh. avoids the correctness issues but adds latency to the metadata refresh. Will implement the API as per option 3. This change is worthy of a separate PR which I will open shortly.

Lorak-mmk:
Metadata refresh should not query all nodes.

One option is for connection pool to query it on the first connection opened to the node. The problem with this approach is that a quick node restart may theoretically be undetected by the pool (first connection after restart created before last before-restart connection detected as dead). In this case, the version may become stale.

TBH I don't have a good idea on how to properly solve this, which is one of the reasons this issue was not addressed for so long.

I'm curious @conorbros, what is your use case for this? I assume you need it for something.

conorbros:
https://github.com/scylladb/scylla-rust-driver/pull/1706

Opened the new PR for scylla version.
My solution only queries the node when it actually needs the version, not every metadata refresh.

No use case, just picked an issue I thought I could solve.

I think it would be best to merge this PR first and then I will rebase the other.

Lorak-mmk:
This is (as you see now) not a trivial feature to implement or review, with various non-obvious problems, and edge cases. Performing thorough review would take more of our capacity that we can currently spare.
Thanks for providing this attempt at implementing it, but we decided to not merge your changes, but instead address it ourselves some other time. We will surely look at your approaches when discussing how to implement this.

conorbros:
I understand. Thanks for taking the time to look at it. 
My goal over the next few weeks is to make some open source contributions and I'm interested in this project because I am familiar with cassandra. 
Let me know if there is any issues I could take and you would have capacity to review.

Lorak-mmk:
You could look trough issues with "good first label" issue that don't have assignee nor linked PR. If one catches your eye, you can let me know before starting your work, I'll tell you if there are any un-obvious difficulties.

PR #1706, opened 2026-06-08
Expose `scylla_version` on `Node`

conorbros:
Adds a per-node ScyllaDB version, read from the `system.versions` virtual table and exposed as `Node::scylla_version() -> Option<&str>`. The version is only knowable after a node's connection is up, so it is stored in a write-once `OnceLock` on `Node` and filled in by the cluster worker. The driver re-queries it on reconnect to track rolling upgrades.

## Changes

| Area | Change |
|---|---|
| `cluster/node.rs` | New `scylla_version: OnceLock<String>` field (private). Accessor `scylla_version() -> Option<&str>` and crate-internal `set_scylla_version(String)` (write-once). |
| `cluster/state.rs` | `ClusterState::new` split into `build` (assemble state, no versions) + `populate_scylla_versions` (wait for pools, query, record). New `query_scylla_version` helper. New `reconnected_nodes` parameter. |
| `cluster/worker.rs` | New `reconnected_nodes: HashSet<Uuid>` accumulated from `Established` connectivity events, consumed + cleared each refresh. The separate post-refresh pool-wait calls are removed (the wait now lives in `populate_scylla_versions`). |
| tests | New integration suite `metadata/scylla_version.rs` (4 tests) + `build`-based classification unit tests in `state.rs`. |

## Design

The version can only be written after the `Node` is shared into the token ring and no longer uniquely owned. So it is modelled as write-once interior mutability: a `OnceLock<String>` on `Node`, empty at construction and set once after the pool comes up.

A node that reconnects or changes address is a brand-new `Node`, so the version is never rewritten. Readers see either the empty or the populated value.

## Design: build, then populate

`ClusterState::new()` runs in two steps: `build`, then `populate_scylla_versions`.

### build

The pre-existing `ClusterState::new` body, unchanged except for the reconnect rule in node classification:

- **new** (not in previous `known_nodes`): `Node::new` (empty version)
- **address or release_version changed**: `inherit_preserving_pool` (fresh `Node`, empty version)
- **reconnected** since last refresh: `inherit_preserving_pool` (fresh `Node`, empty version)
- **otherwise unchanged**: `Arc::clone` (reused, keeps its version)

A node is queried by `populate_scylla_versions` if its version is empty.

### populate_scylla_versions

1. Wait for every pool to initialize, so the queries below have a connection.
2. For every node whose version is still `None`, query `system.versions` (in parallel, 500 ms timeout each) and set the result on the node.

## Rolling-upgrade handling (`reconnected_nodes`)

When a node's pool re-establishes after being unreachable, the worker receives `ConnectivityChangeEvent::Established` and records the node's `host_id` in `reconnected_nodes`. The next refresh passes that set into `build`, which recreates those nodes with an empty version so the new version is queried.

## Degradation behavior

`query_scylla_version` returns `None` (never errors) on a query error, a 500 ms timeout, an empty result, or a Cassandra node with no `system.versions`. A `None` is simply not written, leaving the `OnceLock` empty, so the node is retried on the next refresh.

## note

The pool-init wait moved into `populate_scylla_versions`, so `trigger_pool_refills_for_hosts` now runs fire-and-forget instead of awaited. Seems safe because refill only resets backoff and the load balancer skips disconnected nodes but please confirm.

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -217,6 +217,7 @@ struct NodeInfoRow {
     datacenter: Option<String>,
     rack: Option<String>,
     tokens: Option<Vec<String>>,
+    release_version: Option<String>,
 }
 
 #[derive(DeserializeRow)]
@@ -230,6 +231,7 @@ struct LocalNodeInfoRow {
     rack: Option<String>,
     tokens: Option<Vec<String>>,
     cluster_name: Option<String>,
+    release_version: Option<String>,
 }
 
 #[derive(Clone, Copy)]
@@ -254,7 +256,7 @@ impl ControlConnection {
     ) -> Result<(Vec<Peer>, Option<String>), MetadataError> {
         let peers_query_stream = self
             .query_iter(
-                "SELECT host_id, rpc_address, data_center, rack, tokens FROM system.peers",
+                "SELECT host_id, rpc_address, data_center, rack, tokens, release_version FROM system.peers",
                 &(),
             )
             .map(|pager_res| {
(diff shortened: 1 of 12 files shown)


Lorak-mmk:
This is (as you see now) not a trivial feature to implement or review, with various non-obvious problems, and edge cases. Performing thorough review would take more of our capacity that we can currently spare.
Thanks for providing this attempt at implementing it, but we decided to not merge your changes, but instead address it ourselves some other time. We will surely look at your approaches when discussing how to implement this.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/1286 so that we see this PR when looking at the issue

PR #1708, opened 2026-06-09
Fix shorter tuple ser/deser (again)

### LINKED ISSUE #1625: Tuple deserialization in `deser_cql_value` doesn't allow missing fields
In https://github.com/scylladb/scylla-rust-driver/issues/1452 we discovered that tuples of size N can actually be encoded using a tuple of size M if M <= N.
For deserialization of Rust tuples it was fixed in https://github.com/scylladb/scylla-rust-driver/pull/1453/changes
This PR unfortunately missed `deser_cql_value`. Tuple deserialization there still looks like this:
```rust
        Tuple(type_names) => {
            let t = type_names
                .iter()
                .map(|typ| -> StdResult<_, DeserializationError> {
                    let raw = types::read_bytes_opt(buf).map_err(|e| {
                        mk_deser_err::<CqlValue>(
                            typ,
                            BuiltinDeserializationErrorKind::RawCqlBytesReadError(e),
                        )
                    })?;
                    raw.map(|v| CqlValue::deserialize(typ, Some(FrameSlice::new_borrowed(v))))
                        .transpose()
                })
                .collect::<StdResult<_, _>>()?;
            CqlValue::Tuple(t)
        }
```

For serialization the situation is exactly reversed!
`serialize_cql_value` allows smaller tuples:
```rust
        CqlValue::Tuple(t) => {
            // We allow serializing tuples that have less fields
            // than the database tuple, but not the other way around.
            let fields = match typ {
                ColumnType::Tuple(fields) => {
                    if fields.len() < t.len() {
                        return Err(mk_typck_err::<CqlValue>(
                            typ,
                            TupleTypeCheckErrorKind::WrongElementCount {
                                rust_type_el_count: t.len(),
                                cql_type_el_count: fields.len(),
                            },
                        ));
                    }
                    fields
                }
                _ => {
                    return Err(mk_typck_err::<CqlValue>(
                        typ,
                        TupleTypeCheckErrorKind::NotTuple,
                    ));
                }
            };
            serialize_tuple_like(typ, fields.iter(), t.iter(), writer)
        }
```

However, serializing Rust tuples requires an exact match:
```rust
                let ($($tidents,)*) = match typ {
                    ColumnType::Tuple(typs) => match typs.as_slice() {
                        [$($tidents),*] => ($($tidents,)*),
                        _ => return Err(mk_typck_err::<Self>(
                            typ,
                            TupleTypeCheckErrorKind::WrongElementCount {
                                rust_type_el_count: $length,
                                cql_type_el_count: typs.len(),
                            }
                        ))
                    }
                    _ => return Err(mk_typck_err::<Self>(
                        typ,
                        TupleTypeCheckErrorKind::NotTuple,
                    ))
                };
```

Lorak-mmk:
CQL tuple of size N can be stored in DB (and returned to the driver) as a tuple of size M, for M <= N.

Both DBs allow inserting such shorter tuples directly. In Scylla they can also appear if you ALTER column, increasing the field count of the tuple.

We allowed serializing `CqlValue::Tuple` to a column with larger tuple for a very long time now.
Some time ago (~9 months) we also allowed deserializing Rust tuples from such shorter representation: https://github.com/scylladb/scylla-rust-driver/pull/1453

The problem is that the PR above forgot about `CqlValue::Tuple`! Without adjusting it, it is possible for values to exist in DB that are not deserializable to CqlValue, which is a problem (it is supposed to be able to represent all values).
The first thing that this PR does is fixing it (and adding regression test).

After fixing it, I thought it is weird that we allow both ser and deser of CqlValue::Tuple, and deser of Rust tuples, but not ser of Rust tuples. For consistency I changed it as well (and added a test).

Note: technically, all of this is behavior, not API. So we could theoretically change the behavior again in 1.x - Rust semver rules don't prohibit it.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1625

## Pre-review checklist

### DIFF
--- a/scylla-cql-core/src/deserialize/value.rs
+++ b/scylla-cql-core/src/deserialize/value.rs
@@ -224,12 +224,19 @@ impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for CqlValue {
                     let t = type_names
                         .iter()
                         .map(|typ| -> Result<_, DeserializationError> {
-                            let raw = frame_slice.read_cql_bytes().map_err(|e| {
-                                mk_deser_err::<CqlValue>(
-                                    typ,
-                                    BuiltinDeserializationErrorKind::RawCqlBytesReadError(e),
-                                )
-                            })?;
+                            let raw = if frame_slice.is_empty() {
+                                // Special case: if there are no bytes left to read, consider the tuple element as null.
+                                // This is needed to handle tuples with fewer elements than expected
+                                // (see https://github.com/scylladb/scylla-rust-driver/issues/1452 for context).
+                                None
+                            } else {
+                                frame_slice.read_cql_bytes().map_err(|e| {
+                                    mk_deser_err::<CqlValue>(
+                                        typ,
+                                        BuiltinDeserializationErrorKind::RawCqlBytesReadError(e),
+                                    )
+                                })?
+                            };
                             raw.map(|v| CqlValue::deserialize(typ, Some(v))).transpose()
                         })
                         .collect::<Result<_, _>>()?;
@@ -1624,7 +1631,7 @@ macro_rules! impl_tuple {
                         {
                             let cql_bytes = if v.is_empty() {
                                 // Special case: if there are no bytes left to read, consider the tuple element as null.
-                                // This is needed to handle tuples with less elements than expected
+                                // This is needed to handle tuples with fewer elements than expected
                                 // (see https://github.com/scylladb/scylla-rust-driver/issues/1452 for context).
                                 None
                             } else {
--- a/scylla/tests/integration/types/cql_collections.rs
+++ b/scylla/tests/integration/types/cql_collections.rs
@@ -385,6 +385,18 @@ async fn test_cql_tuple_db_repr_shorter_than_metadata() {
             // Total: 31 bytes
             assert_eq!(buf.len(), 31);
         }
+
+        // Verify that CqlValue::Tuple can be shorter than the DB column type indicates.
+        let tuple_ser_cql = CqlValue::Tuple(vec![
+            Some(CqlValue::Tuple(vec![
+                Some(CqlValue::Int(1)),
+                Some(CqlValue::Text("Ala".to_owned())),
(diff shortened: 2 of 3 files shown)


Lorak-mmk:
Rebased on main

scylla-cql-core/src/deserialize/value.rs
```diff
@@ -224,12 +224,19 @@ impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for CqlValue {
...
                         .map(|typ| -> Result<_, DeserializationError> {
-                            let raw = frame_slice.read_cql_bytes().map_err(|e| {
-                                mk_deser_err::<CqlValue>(
-                                    typ,
-                                    BuiltinDeserializationErrorKind::RawCqlBytesReadError(e),
-                                )
-                            })?;
+                            let raw = if frame_slice.is_empty() {
+                                // Special case: if there are no bytes left to read, consider the tuple element as null.
+                                // This is needed to handle tuples with less elements than expected
```
wprzytula: ⛏️ less -> fewer

scylla/tests/integration/types/cql_collections.rs
```diff
@@ -362,6 +362,11 @@ async fn test_cql_tuple_db_repr_shorter_than_metadata() {
             None, // missing third field of outer tuple
         ]);
         insert_and_select(&session, table_name, &tuple_ser, &tuple_cql_deser).await;
+
+        // Test that shorter Rust tuple can also be successfully inserted, and is of correct form
+        // when selected.
+        let tuple_ser_raw = ((1, "Ala".to_owned()), 2);
+        insert_and_select(&session, table_name, &tuple_ser_raw, &tuple_deser).await;
```
wprzytula, line 370: 💭 This test does verify that ser composed with deser produces identity. What we still miss is verification of the produced serialized form - asserting that it looks correct.

Lorak-mmk:
Rebased on main, addressed review comments.

wprzytula (approved):

PR #1712, opened 2026-06-10
Reject out-of-range shard ids in sharding info

### LINKED ISSUE #1696: Validate shard id is less than shard count before indexing pools
The connection pool trusts the shard id reported by the server and later indexes the per-shard connection array with it.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/scylla/src/network/connection_pool.rs#L826

Shard metadata validation ensures the shard count is nonzero, but it does not reject a shard id greater than or equal to the shard count. A malformed or buggy server response can therefore panic the pool worker instead of being treated as invalid sharding information.

Suggested direction: validate that the reported shard id is strictly less than the reported shard count before resizing/indexing the per-shard pool state.

Photon101:
## Summary

- Reject server-reported shard IDs that are not lower than `SCYLLA_NR_SHARDS` while parsing sharding info.
- Preserve acceptance of the highest valid shard ID.
- Add regression coverage for `shard == nr_shards` and `shard > nr_shards`.

## Checks

- `rustup run stable cargo test -p scylla shard_info --lib`
- `rustup run stable cargo test -p scylla routing::sharding --lib`
- `rustup run stable cargo fmt --all -- --check`
- `rustup run stable cargo check -p scylla --all-features`
- `RUSTFLAGS='-Dwarnings --cfg scylla_unstable' rustup run stable cargo clippy --all-targets --all-features`
- `RUSTFLAGS='-Dwarnings' rustup run stable cargo clippy --all-features --tests --examples`
- `git diff --check`

## Not run

- `make test` / live-cluster and ccm suites; this change is covered by parser unit tests and I did not use a local Scylla/ccm cluster.

## Pre-review checklist

Fixes: #1696

### DIFF
--- a/scylla/src/routing/sharding.rs
+++ b/scylla/src/routing/sharding.rs
@@ -83,12 +83,23 @@ impl std::str::FromStr for Token {
 }
 
 impl ShardInfo {
-    pub(crate) fn new(shard: u16, nr_shards: ShardCount, msb_ignore: u8) -> Self {
-        ShardInfo {
+    pub(crate) fn new(
+        shard: u16,
+        nr_shards: ShardCount,
+        msb_ignore: u8,
+    ) -> Result<Self, ShardingError> {
+        if shard >= nr_shards.get() {
+            return Err(ShardingError::ShardIdOutOfRange {
+                shard,
+                nr_shards: nr_shards.get(),
+            });
+        }
+
+        Ok(ShardInfo {
             shard,
             nr_shards,
             msb_ignore,
-        }
+        })
     }
 
     pub(crate) fn get_sharder(&self) -> Sharder {
@@ -245,6 +256,12 @@ pub(crate) enum ShardingError {
     #[error("Sharding info contains an invalid number of shards (0)")]
     ZeroShards,
 
+    /// A bug in Scylla. Shard IDs must be lower than the total number of shards.
+    #[error(
+        "Sharding info contains an invalid shard id ({shard}); it must be lower than the number of shards ({nr_shards})"
+    )]
+    ShardIdOutOfRange { shard: u16, nr_shards: u16 },
+
     /// A bug in Scylla. Failed to parse string to number.
     #[error("Failed to parse a sharding info parameter's value: {0}")]
     ParseIntError(#[from] std::num::ParseIntError),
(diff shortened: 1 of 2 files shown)


scylla/src/routing/sharding.rs
wprzytula, line 103: 💭 Why not make the check occur at the construction time, by returning `Result<Self>::Err(ShardingError::ShardIdOutOfRange)` in case comparison fails?

wprzytula (changes requested):

scylla/src/routing/sharding.rs
wprzytula: This way we still have the check only once, but don't have to risk a panic in `ShardInfo::new()`.

Lorak-mmk:
The original issue mentioned that a connection vector is indexed with the shard-id in a panicking way, and this is what we need to solve. Example of such code:
`PoolSize::PerShard(target) => self.conns[shard_id].len() < target.get(),`

I think we can go with your approach of guaranteeing that shard is is always not greater than shard count in `Sharder`, but if we do this then we should also:
- Deduplicate `self.conns[shard_id]` accesses if possible
- On the one remaining access put a comment explaining why it won't panic.

scylla/src/routing/sharding.rs
Photon101: Thanks, addressed in `6e346e6e`. `ShardInfo::new` now returns `Result<Self, ShardingError>` and rejects out-of-range shard IDs at construction, with `TryFrom` delegating to it. I also reduced the pool indexing to a single `self.conns[shard_id]` access with a comment documenting why it is safe.

Checks: `cargo fmt --all -- --check`; `cargo test -p scylla routing::sharding --lib`; `cargo test -p scylla network::connection_pool --lib --no-run`; `cargo clippy -p scylla --lib -- -D warnings`.

wprzytula (approved):

wprzytula:
@Photon101 please rebase on main and squash your commits; even better, split your work into commits that reflect atomic logical changes.

Photon101:
Rebased on current main and squashed the branch to one atomic commit (`848184e2`).

Checks run locally:
- `cargo test -p scylla routing::sharding --lib`
- `cargo test -p scylla network::connection_pool --lib --no-run`
- `cargo clippy -p scylla --lib -- -D warnings`
- `git diff --check origin/main..HEAD`

wprzytula (approved):

Lorak-mmk (approved):

PR #1713, opened 2026-06-10
ci: pin GitHub Actions to commit SHAs

roydahan:
## Summary
- Pin all external GitHub Actions to full commit SHAs to reduce supply chain attack surface
- Upgrade outdated actions to their latest versions

**This PR was generated automatically. Please verify that GitHub Actions work as expected with these changes before merging.**

Reference: https://github.com/scylladb/scylladb/pull/29421

### DIFF
--- a/.github/workflows/authenticate_test.yml
+++ b/.github/workflows/authenticate_test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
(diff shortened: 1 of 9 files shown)


Lorak-mmk:
This is some AI slop, why do we have to waste time looking into such things.
Our actions are already pinned to shas, and requiring shas is enabled in repository settings.
This PR is based on `main` branch from February of 2025 (!!!), so the diffs make no sense.

WTF?!

roydahan:
I think it's a combination of AI issue with GH issue.
I had an agent that reviewed and opened 100+ PRs to set this.
At some point today GH service had issues and it started creating PRs based on old code.

PR #1717, opened 2026-06-16
conditionally publish APIs for C# RS Driver

wprzytula:
C# RS Driver needs access to some driver's APIs which are purposely not public. This PR:
- introduces `unstable-csharp-rs` feature, analogous to existing `unstable-nodejs-rs` etc.;
- publishes multiple APIs;
- adds some new APIs, which operate on private internals.

Please note that the last two commits were created by @phalat24.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -1032,6 +1032,19 @@ impl QueryPager {
         )
     }
 
+    /// This is essentially the same as `next()`, but it is *public* and only available
+    /// when the `unstable-csharp-rs` feature is enabled.
+    /// Rationale: I don't know a way to make a function conditionally public,
+    /// based on a compile-time flag, so I added an new wrapper function
+    /// that is conditionally compiled and is public.
+    #[cfg(all(scylla_unstable, feature = "unstable-csharp-rs"))]
+    #[inline]
+    pub async fn next_column_iterator(
+        &mut self,
+    ) -> Option<Result<(ColumnIterator<'_, '_>, bool), NextRowError>> {
+        self.next().await
+    }
+
     /// Tries to acquire a non-empty page, if current page is exhausted.
     /// Boolean value in `Some(Ok(r))` is true if a new page was fetched.
     fn poll_fill_page(
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1816,6 +1816,27 @@ impl Session {
         .await
     }
 
+    /// This is essentially the same as `execute_iter_nongeneric`, but it is *public*
+    /// and only available when the `unstable-csharp-rs` feature is enabled.
+    /// Rationale: I don't know a way to make a function conditionally public,
+    /// based on a compile-time flag, so I added a new wrapper function
+    /// that is conditionally compiled and is public.
+    ///
+    /// This function is intended to be used by the C#-RS Driver's internals
+    /// and it's potentially a footgun: if a user passes `SerializedValues`
+    /// that don't match the prepared statement, the driver will misbehave
+    /// (potentially leading to data corruption).
+    /// Therefore, this function must not be exposed to end users of the Rust driver.
+    #[cfg(all(scylla_unstable, feature = "unstable-csharp-rs"))]
+    #[inline]
+    pub async fn execute_iter_preserialized(
+        &self,
+        prepared: PreparedStatement,
+        values: SerializedValues,
+    ) -> Result<QueryPager, PagerExecutionError> {
+        self.execute_iter_nongeneric(prepared, values).await
+    }
+
     /// Prepares all statements within the batch and returns a new batch where every
     /// statement is prepared.
     ///
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -2449,3 +2449,17 @@ enum SchemaNodeResult {
     Success(Uuid),
     BrokenConnection(BrokenConnectionError),
 }
+
+/// Additional API for interop-based code.
+#[cfg(all(scylla_unstable, feature = "unstable-csharp-rs"))]
+impl Session {
+    /// Needed to bridge C#'s counterpart of `await_schema_agreement`
+    /// which requires a specified host to participate in agreement checking.
+    pub async fn await_schema_agreement_with_required_node_external(
+        &self,
+        required_node: Option<Uuid>,
+    ) -> Result<Uuid, SchemaAgreementError> {
+        self.await_schema_agreement_with_required_node(required_node)
+            .await
+    }
+}
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -18,7 +18,7 @@ use tokio::sync::mpsc;
 use tracing::{debug, warn};
 use uuid::Uuid;
 
-use super::metadata::{Keyspace, Metadata, Strategy};
+use super::metadata::{Keyspace, Metadata, Strategy, Table};
 use super::node::{Node, NodeRef};
 
 /// Represents the state of the cluster, including known nodes, keyspaces, and replica locator.
@@ -319,27 +319,44 @@ impl ClusterState {
         table: &str,
         partition_key: &dyn SerializeRow,
     ) -> Result<Token, ClusterStateTokenError> {
-        let Some(table) = self
-            .keyspaces
-            .get(keyspace)
-            .and_then(|k| k.tables.get(table))
-        else {
-            return Err(ClusterStateTokenError::UnknownTable {
-                keyspace: keyspace.to_owned(),
-                table: table.to_owned(),
-            });
-        };
+        let table_meta = self.lookup_table_meta(keyspace, table)?;
+
         let values = SerializedValues::from_serializable(
-            &RowSerializationContext::from_specs(table.pk_column_specs.as_slice()),
+            &RowSerializationContext::from_specs(table_meta.pk_column_specs.as_slice()),
             partition_key,
         )?;
+
+        // Delegate actual token calculation to centralized helper.
+        self.do_compute_token(table_meta, &values)
+    }
+
+    fn do_compute_token(
+        &self,
+        table: &Table,
+        serialized_partition_key: &SerializedValues,
+    ) -> Result<Token, ClusterStateTokenError> {
         let partitioner = table
             .partitioner
             .as_deref()
             .and_then(PartitionerName::from_str)
             .unwrap_or_default();
-        calculate_token_for_partition_key(&values, &partitioner)
+        calculate_token_for_partition_key(serialized_partition_key, &partitioner)
             .map_err(ClusterStateTokenError::TokenCalculation)
+    }
+
+    /// Helper: lookup table metadata or return `UnknownTable` error.
+    fn lookup_table_meta(
+        &self,
+        keyspace: &str,
+        table: &str,
+    ) -> Result<&Table, ClusterStateTokenError> {
+        self.keyspaces
+            .get(keyspace)
+            .and_then(|k| k.tables.get(table))
+            .ok_or_else(|| ClusterStateTokenError::UnknownTable {
+                keyspace: keyspace.to_string(),
+                table: table.to_string(),
+            })
     }
 
     /// Access to replicas owning a given token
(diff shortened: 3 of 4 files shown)


scylla/tests/integration/cluster/compute_token_preserialized.rs
```diff
@@ -0,0 +1,179 @@
...
+    // compute tokens
+    let token_preser = session
+        .get_cluster_state()
+        .compute_token_preserialized(&ks, "t1", &sv)
+        .unwrap();
+    let token_direct = session
+        .get_cluster_state()
+        .compute_token(&ks, "t1", &(&v.0,))
+        .unwrap();
+    assert_eq!(token_preser, token_direct);
```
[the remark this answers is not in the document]
wprzytula: @Lorak-mmk Rly? What about awaiting schema agreement automatically?

scylla/src/client/pager.rs
```diff
@@ -1138,6 +1138,19 @@ impl QueryPager {
...
+    /// based on a compile-time flag, so I added an new wrapper function
+    /// that is conditionally compiled and is public.
+    #[cfg(all(scylla_unstable, feature = "unstable-csharp-rs"))]
+    #[inline]
+    pub async fn next_column_iterator(
+        &mut self,
+    ) -> Option<Result<(ColumnIterator<'_, '_>, bool), NextRowError>> {
+        self.next().await
+    }
+
```
Lorak-mmk, line 1047: When initially trying to understand why this is necessary:
- First I thought that `ColumnIterator` implements `DeserializeRow`, so you could just use `rows_stream`
- Then I noticed you also return `bool` indicating wheter its a new page - perhaps you need to perform your own type check, and want to do it per page?
- Then I realized that ColumnIterator is not usable with `rows_stream` because `rows_stream` needs owned type.

Now, two observations:
- A comment would be nice, otherwise any future reader needs to repeat the reasoning I did above to understand why we need it, and if we still need it.
- Perhaps we could expose API to all users that allows iterating trough pages? It would mean slightly less ergonomic API when iterating over rows (nested loop required), but it would allow using tokio-task iteration with borrowed deserialization.

scylla/src/client/session.rs
```diff
@@ -1815,6 +1815,26 @@ impl Session {
...
+    #[cfg(all(scylla_unstable, feature = "unstable-csharp-rs"))]
+    #[inline]
+    pub async fn execute_iter_preserialized(
+        &self,
+        prepared: PreparedStatement,
+        values: SerializedValues,
+    ) -> Result<QueryPager, PagerExecutionError> {
+        self.execute_iter_nongeneric(prepared, values).await
+    }
+
```
Lorak-mmk: Again, a rationale for that would be nice.

Why do you need to preserialize (in other words, why can't C# serialization be called from SerializeRow impl)? Why can't you have a wrapper over SerializedValues with SerializeRow impl? 

Such descriptions would be helpful when designing APIs of Rust Driver, because we would know a use case we need to serve.

scylla/src/cluster/state.rs
```diff
@@ -529,6 +546,90 @@ impl ClusterState {
...
+
+        if serialized_partition_key.element_count() as usize != table_meta.partition_key.len() {
+            return Err(ClusterStateTokenError::PartitionKeyCountMismatch {
+                keyspace: keyspace.to_string(),
+                table: table.to_string(),
+                received: serialized_partition_key.element_count() as usize,
+                expected: table_meta.partition_key.len(),
+            });
+        }
+
```
Lorak-mmk: Why do we need this validation? User needs to provide correct values to get correct results, and in general we cannot check if the provided value is correct. What is the point of verifying this one case? It gives a false sense of safety, please get rid of it.

scylla/src/cluster/state.rs
```diff
@@ -529,6 +546,90 @@ impl ClusterState {
...
+    /// but the partitioner is known (or default is desired).
+    pub fn compute_token_preserialized_with_partitioner(
+        &self,
+        partitioner: &PartitionerName,
+        serialized_partition_key: &SerializedValues,
+    ) -> Result<Token, ClusterStateTokenError> {
+        calculate_token_for_partition_key(serialized_partition_key, partitioner)
+            .map_err(ClusterStateTokenError::TokenCalculation)
+    }
+}
```
Lorak-mmk, line 615: What is your use case for this, how can that happen? Is it some legacy C# API?

scylla/src/cluster/state.rs
```diff
@@ -736,4 +837,110 @@ mod tests {
...
+
+            // Serialized values with zero elements
+            let sv = SerializedValues::new();
+            let res = cs.compute_token_preserialized(ks, tbl, &sv);
+            assert!(matches!(
+                res,
+                Err(ClusterStateTokenError::PartitionKeyCountMismatch { .. })
+            ));
+        }
+    }
```
Lorak-mmk: The "interop" API is bascially the same as normal one, and uses the same code paths (`lookup_table_meta`, `do_compute_token`). Do you really believe we need 100 lines of tests for it? I know LLMs make writing tests easy, but it doesn't mean every test is worth having and maintaining.

scylla/src/errors.rs
```diff
@@ -1117,6 +1117,25 @@ pub enum ClusterStateTokenError {
...
+    PartitionKeyCountMismatch {
+        /// Keyspace name.
+        keyspace: String,
+        /// Table name.
+        table: String,
+        /// Number of values received.
+        received: usize,
+        /// Number of values expected.
+        expected: usize,
+    },
```
Lorak-mmk: If your validation needs a new error variant, then I'm even more against it,

scylla/src/cluster/state.rs
```diff
@@ -529,6 +546,90 @@ impl ClusterState {
...
+    /// table’s partition key.
+    /// A basic check is performed to ensure the number of provided values matches
+    /// the number of partition key columns.
+    pub fn compute_token_preserialized(
+        &self,
+        keyspace: &str,
+        table: &str,
+        serialized_partition_key: &SerializedValues,
+    ) -> Result<Token, ClusterStateTokenError> {
+        let table_meta = self.lookup_table_meta(keyspace, table)?;
```
Lorak-mmk, line 591: @wprzytula Can we get rid of "execute_iter_preserialized" and this thing when you finish bound statements?

scylla/tests/integration/cluster/compute_token_preserialized.rs
```diff
@@ -0,0 +1,179 @@
...
+    let ks = unique_keyspace_name();
+
+    session
+        .ddl(format!("CREATE KEYSPACE IF NOT EXISTS {ks} WITH REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}")).await.unwrap();
+    session.use_keyspace(ks.as_str(), true).await.unwrap();
+
+    // Single-column partition key
+    session
+        .ddl("CREATE TABLE IF NOT EXISTS t1 (a text primary key)")
+        .await
```
Lorak-mmk: Why do we need integration test for this API? Integration tests are expensive in terms of time spent on them - you need to create keyspace, table, drop them later. What is the point of having those 180 lines?

scylla/src/client/session.rs
```diff
@@ -2755,3 +2755,17 @@ enum SchemaNodeResult {
...
+    /// Needed to bridge C#'s counterpart of `await_schema_agreement`
+    /// which requires a specified host to participate in agreement checking.
+    pub async fn await_schema_agreement_with_required_node_external(
+        &self,
+        required_node: Option<Uuid>,
+    ) -> Result<Uuid, SchemaAgreementError> {
+        self.await_schema_agreement_with_required_node(required_node)
+            .await
+    }
+}
```
Lorak-mmk, line 2465: Nit: I suspect that original C# impl performs agreement check on specified node by querying both its `system.local` and `system.peers`, which is a different approach that Rust Driver takes (qyerying `system.local` on all nodes, optionally requiring the specified one to respond).
If you didn't already, you may want to describe that in C# driver docs.

Lorak-mmk (changes requested):
LLMs can now write tons of test code, which doesn't mean all this code is worth having, running, and maintaining,

scylla/src/client/pager.rs
wprzytula: > Now, two observations:
> 
>   * A comment would be nice, otherwise any future reader needs to repeat the reasoning I did above to understand why we need it, and if we still need it.

Sure, I'll add a comment.
>
>   * Perhaps we could expose API to all users that allows iterating trough pages? It would mean slightly less ergonomic API when iterating over rows (nested loop required), but it would allow using tokio-task iteration with borrowed deserialization.

Perhaps. We did consider this in the past; this however is not a Rust Driver's functionality related PR - it's about exposing an internal point of attachment for a wrapper driver. It requires minimal changes and doesn't influence driver's public API.

Lorak-mmk: Sure, I meant it as an idea for the future, not for this PR

wprzytula: > Sure, I meant it as an idea for the future, not for this PR

I suggest :seedling:/:pushpin: emoji for this to avoid confusion :))

scylla/src/client/session.rs
wprzytula: > Again, a rationale for that would be nice.
> 
> Why do you need to preserialize (in other words, why can't C# serialization be called from SerializeRow impl)?

See https://github.com/scylladb-zpp-2025-csharp-rs-driver/csharp-driver/blob/master/rust/src/pre_serialized_values.rs for how we do serialization:
- we use C# value-level serialization as a blackbox;
- we manually append C#-serialized values to `SerializedValues`;
- then we pass the ready `SerializedValues` without overhead to Rust Driver.

> Why can't you have a wrapper over SerializedValues with SerializeRow impl?

Because it would introduce overhead. No matter `SerializeRow` impl used, all existing driver's APIs call `PreparedStatement::serialize()`, which calls `SerializedValues::from_serializable`, which calls `SerializedValues::from_closure()`, which allocates a new `Vec` and serialized values there.
 
> Such descriptions would be helpful when designing APIs of Rust Driver, because we would know a use case we need to serve.

scylla/src/cluster/state.rs
wprzytula: @phalat24 is the author; let him respond.

scylla/src/cluster/state.rs
wprzytula: @phalat24 is the author; let him respond.

scylla/src/cluster/state.rs
wprzytula: @phalat24 is the author; let him respond.

scylla/src/errors.rs
wprzytula: @phalat24 is the author; let him respond.

scylla/src/cluster/state.rs
wprzytula: I don't think so. The purpose of this is boosting performance by avoiding needless allocation described by me in a comment above, and I doubt `BoundStatement` will help. The problem is with the `SerializeRow`, which does not accept ready `SerializedValues` without further allocations.

scylla/tests/integration/cluster/compute_token_preserialized.rs
wprzytula: @phalat24 is the author; let him respond.

scylla/src/client/session.rs
wprzytula: @phalat24 is the author; let him respond.

scylla/src/client/session.rs
Lorak-mmk: > See https://github.com/scylladb-zpp-2025-csharp-rs-driver/csharp-driver/blob/master/rust/src/pre_serialized_values.rs for how we do serialization:

But why do you need to do it this way? Why can't your serialization be called from `SerializeRow` and append to the writer?

wprzytula: Because we need the serialization to be done synchronously due to lifetime constraints, not asynchronously inside Rust future.

Lorak-mmk: Interesting, I'd like to understand it better in the future, because I still don't fully understand why that would be the case.

For now it is good enough.

scylla/src/cluster/state.rs
Lorak-mmk: But the purpose of BoundStatement is exactly what you do here: preserialize data, pass it around in owned form (like SerializedValues), cheaply compute token without double serialization.

In the other comment you mentioned the problem is with async; building BoundStatement should require no async.

scylla/src/client/session.rs
wprzytula: > Interesting, I'd like to understand it better in the future, because I still don't fully understand why that would be the case.

You mean you miss Rust context or C# context?

Lorak-mmk: C# context.

wprzytula: In C# context, we want to leverage the stack-bound variables to make memory management easier. If the whole serialization process happens synchronously, meaning that it does not exceed the scope of the C# caller, this opens up the possibility to use stack-allocated buffers etc., as well as allows cleaning up heap resources immediately before the caller finishes. If serialization happened asynchronously, we would have to:
- either prevent garbage collection of ALL C# values,
- or prevent garbage collection of ALL buffers with serialized values (some of them are currently stack-allocated - this would push them to the heap),
- or allocate the whole buffer of concatenated serialized values, essentially duplicating `SerializedValues` and involving more allocations and copying.

Lorak-mmk: > either prevent garbage collection of ALL C# values,

My guess was that preventing C# values from being collected for the duration of the request is a reasonable thing to do. If its not, then I understand the problem now. BoundStatement should take care of it.

Btw when using all those stack-pinned stuff do you take care not to overflow the stack? For example, only use stack for certain amount of values, and values up to certain size?

scylla/src/cluster/state.rs
wprzytula: OK, I think you're right.

scylla/src/cluster/state.rs
wprzytula: Deleted.

scylla/src/cluster/state.rs
wprzytula: Described in the docstring.

scylla/src/cluster/state.rs
wprzytula: Deleted.

scylla/tests/integration/cluster/compute_token_preserialized.rs
wprzytula: Deleted.

scylla/src/client/session.rs
wprzytula: ACK.

scylla/src/errors.rs
wprzytula: Deleted.

scylla/src/client/pager.rs
wprzytula: Opened https://github.com/scylladb-zpp-2025-csharp-rs-driver/csharp-driver/issues/161 not to forget fixing this bug/limitation.

Lorak-mmk (approved):

PR #1720, opened 2026-06-17
Fetch info about tablet keyspaces from system_schema.scylla_keyspaces

### LINKED ISSUE #1534: Fetch info about tablet keyspaces
Currently load balancing assumes initially that all keyspaces are vnode-based, and routes requests according to that.
When it receives tablet feedback from some requests, it puts the resulting tablet in replica locator, and from then onwards the keyspace will be recognized as tablet-based.

The alternative I'm proposing here is to enhance this mechanism. When fetching metadata (in MetadataReader) driver could read the `system_schema.scylla_keyspaces` table to learn if the keyspace is tablet-based.
We could then expose this info in `ClusterState`. IIRC it was requestes at least once (cc @piodul - I think you asked about this when working in scylla-cdc-rust).

One thing to be taken care of: We need to understand the semantics of this table and backwards compatibility guarantees.
I'm pretty sure that in the past all user-created keyspaces had a row in this table, and `initial_tablets` was null for vnode keyspaces.
When I created a vnode keyspace in latest scylla, it doesn't have a row in this table at all. While its not a problem to support both semantics, it makes me worry that the table could have more incompatible changes in the future.

Lorak-mmk:
Until now we based our knowledge of keyspaces being tablet based or not solely on tablet hints received in responses to queries.
Because of that, `ReplicaLocator` returned vnode-based replicas for tables where we had not yet received any hints. Also `ClusterState::new` performed precomputation of vnode-based replicas also for tablet-keyspaces replication strategies.

This is all generally acceptable for load balancing (because we will receive hints soon), but:
- It may be confusing to direct users of replica-calculating methods of ClusterState, and ReplicaLocator, because they would just return incorrect (vnode-based) replica sets.
- Precomputing replicas may be expensive, and we had to do this on each call to `ClusterState::new`.

This PR addresses those issues. `Keyspace` now contains `tablets: bool` field. Data in `TabletsInfo` is synchronized with ClusterState: now each tablet-based table will always have an entry in `TabletsInfo`. Replica precomputation can now skip tablet-based keyspaces.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1534

## Pre-review checklist


scylla/src/routing/locator/tablets.rs
```diff
@@ -592,13 +597,47 @@ impl TabletsInfo {
...
+        self.tablets.retain(|k, _| {
+            let Some(keyspace) = keyspaces.get(k.ks_name()) else {
+                return false;
+            };
+            if !keyspace.tablets {
+                return false;
+            }
+
+            keyspace.tables.contains_key(k.table_name())
+        });
```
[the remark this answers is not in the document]
Lorak-mmk: Huh, this is actually true. Important to note that this was also true before my PR, so we can fix it in a separate patch.

I was under the impression that even with schema fetching is disabled, tables will be populated anyway in order to fetch partitoners which the driver needs. This turns out to not be the case - with schema fetching disabled table list will be empty.

@wprzytula What would you propose to do? I think that we should fetch table list even with schema fetching disabled - just not fetch columns.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -1602,6 +1616,61 @@ impl ControlConnection {
...
+
+    /// Fetches a (possibly filtered) list of tablet-based keyspaces.
+    ///
+    /// A keyspace is tablet-based if its `initial_tablets` column in
+    /// `system_schema.scylla_keyspaces` is not NULL.
+    ///
+    /// On Cassandra and old ScyllaDB versions the `system_schema.scylla_keyspaces`
+    /// table does not exist. This is handled the same way as in
+    /// [`Self::query_table_partitioners`]: the resulting `DbError::Invalid` is caught
+    /// and an empty map is returned (so all keyspaces default to non-tablet-based).
```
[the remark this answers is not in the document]
wprzytula: @Lorak-mmk Agreed.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -1602,6 +1616,61 @@ impl ControlConnection {
...
+                let result = match row_result {
+                    Ok((keyspace_name, initial_tablets)) => {
+                        initial_tablets.map(|_| Ok(keyspace_name))
+                    }
+                    Err(e) => Some(Err(e)),
+                };
+                future::ready(result)
+            })
+            .try_collect::<HashSet<_, _>>()
+            .await;
```
wprzytula, line 1655: 💭 Why not reduce the network load by filtering out non-tablet keyspaces on the server side, with `ALLOW FILTERING` appended to the query string?

scylla/src/cluster/metadata/mod.rs
```diff
@@ -146,6 +146,12 @@ pub struct Keyspace {
     pub strategy: Strategy,
     /// Whether the keyspace has durable writes enabled.
     pub durable_writes: bool,
+    /// Whether the keyspace is tablet-based.
+    ///
+    /// This is determined based on whether `initial_tablets` is set in
+    /// `system_schema.scylla_keyspaces`. On Cassandra and old ScyllaDB versions,
+    /// where this table or column is not present, this is always `false`.
+    pub tablets: bool,
```
wprzytula: 💭 Why not `tablet_based`? `tablets` suggests something different than a boolean property.

scylla/src/routing/locator/tablets.rs
```diff
@@ -541,7 +542,10 @@ impl TabletsInfo {
     /// It goes through tablet info and adjusts it to topology changes, to prevent
     /// a situation where local tablet info and a real one are permanently different.
     /// What is updated:
-    /// 1. Info for dropped tables is removed.
+    /// 1. Info for dropped tables, and tables that are in non-tablet keyspaces
+    ///    according to fetched schemam. Empty tablet lists are added for tables
+    ///    in tablet-based keyspaces, to prevent ReplicaLocator from treating
+    ///    them as VNode-based.
```
wprzytula, line 548:
```suggestion
    /// 1. Info for dropped tables, and tables that are in non-tablet keyspaces
    ///    according to fetched schema. Empty tablet lists are added for tables
    ///    in tablet-based keyspaces, to prevent ReplicaLocator from treating
    ///    them as VNode-based.
```

wprzytula (commented):
Please address Copilot's comments.

scylla/src/cluster/metadata/fetching.rs
Lorak-mmk: I tried, but it seems `IS NOT NULL` is only allowed in Materialized View definition, not in ordinary select.

scylla/src/routing/locator/tablets.rs
```diff
@@ -592,13 +597,47 @@ impl TabletsInfo {
...
+            }
+
+            keyspace.tables.contains_key(k.table_name())
+        });
+
+        // Now we add empty entries for all tables in tablet-based keyspaces
+        // that don't already have an entry. This prevents ReplicaLocator
+        // from returning VNode-based replicas for such tables.
+        keyspaces
+            .iter()
```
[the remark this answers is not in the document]
Lorak-mmk: Not related to this PR.

scylla/src/routing/locator/tablets.rs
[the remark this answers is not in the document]
Lorak-mmk: Opened https://github.com/scylladb/scylla-rust-driver/issues/1741

Lorak-mmk:

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -1602,6 +1616,61 @@ impl ControlConnection {
...
+                error:
+                    MetadataFetchErrorKind::NextRowError(NextRowError::NextPageError(
+                        NextPageError::RequestFailure(RequestError::LastAttemptError(
+                            RequestAttemptError::DbError(DbError::Invalid, _),
+                        )),
+                    )),
+                ..
+            }) => Ok(HashSet::new()),
+            result => result,
+        }
```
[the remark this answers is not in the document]
Lorak-mmk: I don't want to rely on error messages. Consequenes of being to lenient here are better than of being too strict.

scylla/src/cluster/metadata/fetching.rs
wprzytula: Is comparison `initial_tablets > 0` viable here, or are `initial_tablets = null` and `initial_tablets = 0` distinct cases, with only the first indicating a non-tablet-based keyspace?

Lorak-mmk: They are distinct cases. In fact, in recent versions of scylla, all tablet keyspaces have `initial_tablets == 0`, at least by default. Only `null` indicates vnode-based keyspace.

wprzytula: This works in cqlsh:
`SELECT keyspace_name, initial_tablets FROM system_schema.scylla_keyspaces WHERE initial_tablets >= 0 ALLOW FILTERING;`

I believe this correctly discovers tablet-based keyspaces.

wprzytula: Actually, when I created a vnode-based keyspace:
`CREATE KEYSPACE vnodes WITH REPLICATION = { 'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1 } and TABLETS = { 'enabled': false} ;`
**it doesn't show in `system_schema.scylla_keyspaces` at all!**

Tablet-based keyspaces do show there.

Lorak-mmk: Ok, in that case can we keep the current version of the code? It works, and doesn't require `ALLOW FILTERING` (which I think can impact execution negatively). I'm also pretty sure that at some point non-tablet keyspaces did show up in this table - perhaps I'm wrong, or perhaps this behavior was removed. In any case, my approach seems quite safe to me.

wprzytula: 1. `ALLOW_FILTERING` exists solely to require a user to confirm that they understand that the query's complexity will not be _O(returned_rows)_. In this case, this will be _O(rows_in_scylla_keyspaces_table)_.
2. In this case, the alternative to `ALLOW_FILTERING` is to make the DB scan the same amount of rows, but send them all instead of just those that match our requirement. This is a waste that makes no sense.
3. Even if non-tablet keyspaces do show up in the table, I don't understand how your approach is _safer_ than mine.
   - both approaches recognize `initial_tablets` being nonnegative number as tablet keyspaces;
   - both approaches recognize `initial_tablets` being null as vnode keyspaces;
   - the only difference occurs for `initial_tablets` being a negative number. Your approach will correctly regard such keyspaces as tablet based, while mine will deem them vnode-based. Nonetheless, I believe this case is not worth considering, being an obvious server bug.

Lorak-mmk: Ok

Lorak-mmk: Hmm... We both agree that with current server implementations our approaches are the same. They only differ if server decides to put vnode keyspaces in the table.

Implementing your approach requires more changes. I'm using `query_filter_keyspace_name`, which we use in many places, and this is the part that adds `where` clause. Adding `initial_tablets >= 0 ALLOW FILTERING` would require more changes. I don't think it is worth it - we can do this when (if) the server behavior changes in the future.

wprzytula (approved):

PR #1739, opened 2026-06-22
Enhance `QueryPager`: optimise for single-page results and prevent plan exhaustion

### LINKED ISSUE #1736: Optimise `QueryPager` for single-page results
### Background

`QueryPager`-based APIs are the recommended ones for executing `SELECT` queries. This is because they provide convenient automatic paging, transparent page boundaries, and next page prefetching that boosts performance.

### Performance Problem

However, the `PagerWorker`'s machinery is quite heavyweight, as it involves spawning a dedicated tokio task to fetch consecutive pages (ahead of time of their consumption) and feed them into a channel; `QueryPager` just reads pages from the channel.
The overhead is acceptable when the size of the query is large, because it amortises well. A performance problem arises when the query returns a single page, not to mention a single row.

### Motivation

As single-row and other small-size queries are common in ScyllaDB, we want to optimise for this case.

#### Note on wrapper drivers

Note that although Rust Driver's users can choose non-`QueryPager` APIs (`query_unpaged()` or `query_single_page()`) for queries that are predicted to return small amount of rows, this is not the case for Rust Driver-based drivers' users. For instance, C# RS Driver executes **all** user statements through the `QueryPager`, incurring large overhead for non-`SELECT`s and small `SELECT`s.

### Key Idea

When using `QueryPager` APIs, the first page is always fetched synchronously (from the caller's perspective). This means we can perform the first execution on the caller's (tokio) task, not spawning the dedicated tokio task at this point at all.
Only after the paging state response of the first page indicates presence of additional pages, is the pager worker task spawned and starts fetching consecutive pages in the background.

### Additional Gains

This simplifies the interaction between `QueryPager`-`PagerWorker`: now the channel between them is only used for the second and further pages. There's no longer need for the _send proofs_, which were previously required to ensure that the worker sends at least the first page or an error into the channel. Altogether, the simplification is massive.

### LINKED ISSUE #1737: Premature plan exhaustion in multi-page `QueryPager` execution
## Problem Statement

`PagerWorker::work()` currently produces a single `Plan` for the whole paged execution. This means that once a target (node+shard) is given up by the `RetrySession` (having decided to `RetryNextTarget`), it will never be attempted for subsequent pages. The following problems arise:
1. **Reduced Availability:** If every node in the cluster _sometimes_ fails to serve the query, meaning that at least two targets will have to be queried for at least some of the pages, the plan will quickly get exhausted with a few such pages and the query will fail due to the plan exhaustion.
Example scenario, given a query of 100 pages and 3-node cluster:
    - node 1 gets rejected after it fails to serve page 21,
    - then node 2 gets rejected after it fails to serve page 37,
    - then node 3 gets rejected after it fails to serve page 42,
    - then the query fails due to plan exhaustion.
2. **Reduced Performance**: If all replicas for given query fail serving the query at least once (with an error that leads to `RetryNextTarget`), then the load balancing will stick to the remaining targets in the plan, which are non-replicas. This will impose additional server overhead (due to the coordinator not being a replica), meaning increased load on the cluster as well as increased latency.
3. **Increased Billing Costs**: If all local rack / local DC replicas fail serving the query at least once (similarly to the previous point), remaining pages will be served by remote nodes. Cloud networking costs for cross-AZ and cross-DC communication rise significantly.

## Solution

Produce a new `Plan` for each page execution. This will make all nodes tried for every page in a pessimistic case, preventing the reduced availability issue. As the new plan will again prioritise the preferred nodes (local-rack, local-DC replicas), performance and billing costs will also benefit.

### To Keep In Mind: Page Cache Optimisation
#1587 describes the  _"consistent target"_ optimisation:

> `PagerWorker`, used in `Session::*_iter()` APIs for automatic paging, has an optimisation that keeps the same target queried for consecutive > pages as long as no error arises. This optimisation is important for the server, because (citing ScyllaDB Core's recommendations):
> > otherwise you miss the querier cache and will have to rebuild the readers from scratch on the next page.

The optimisation is currently done with the single `Plan` by sticking to the same target until `RetrySession` decides to move.
With separate `Plan`s for each page, the optimisation will need to be reintroduced another way, likely by pushing the last queried target to the front of the newly produced `Plan`.

#### When to put the last successful coordinator on the front of the new `Plan` and when not?

1. When the last coordinator wasn't a replica, I think it should **not** be prioritised. The cost of inter-node communication for each subsequent page will likely be higher than a one-time cost of rebuilding the readers on a replica. @denesb Please confirm this.
2. When the last coordinator was a remote replica, I have no idea whether it should be prioritised. It may depend on the networking costs that cross-DC or cross-rack communication would involve. @Lorak-mmk @mykaul @dkropachev @piodul  I'll appreciate your feedback here.

#### Connection considerations

Should the same connection be reused for subsequent pages, or can possibly different connections to the same target (node+shard) be used? @denesb

If a connection broke inbetween, should I try a different one but to the same coordinator, or should I give up and rely on the new plan without special preference for the last coordinator?

wprzytula:
Fixes: #1736
Fixes: #1737 (for now, every successful coordinator is _sticky_, no matter if it's replica or not)

### Behavioral changes

1. Before, one `Plan` was used for all pages. Now, a fresh new `Plan` is produced for each page. This fixes #1737.
2. Before, the same target was queried until the pages are exhausted or the `RetrySession` decides to move on to the next target. Now, the previous target is reused if it has served the last page query successfully (_I've coined a term for this mechanism: **coordinator stickiness**_).
    - **TODO:** only apply _coordinator stickiness_ when the coordinator is a preferred target for the LBP. See [#1737](https://github.com/scylladb/scylla-rust-driver/issues/1737#issuecomment-4767346928) for the idea.
3. Before, the consistency would be stable across page queries. This meant that if the `RetrySession` decided to downgrade the consistency for one query, it would stay downgraded for the remaining page queries too. Now, the consistency is reset to the original one for each page.

### Performance changes

Before, the `PagerWorker` (for `Session::{query,execute}_iter()`) and `SingleConnectionPagerWorker` (for `ControlConnection::execute_iter()`) were spawned on a separate tokio task unconditionally. Now, they are spawned only for fetching second page and further pages, and only if the first page query returns `PagingStateResponse::HasMorePages`. This reduces overhead of single-paged queries, which are likely a common scenario.

### Internal refactors

- `PagerWorker` is now `'static`, allowing `Send`ing it from the main task to the new spawned task.
- `PagerConstructionError` is removed.
- `Coordinator` is no longer an `Option` in `FirstReceivedPage`, which removed the `expect()` and replaced that runtime assertion with type-system compile-time non-nullity assertion.
- The whole `ProvingSender` abstraction is removed, as the first page is no longer sent through the channel, but returned directly instead. 

### Testing

Tests are added for the new behaviour:
1. Verifies that a new Plan is created for each page.
2. Verifies that _coordinator stickiness_ is applied correctly (for now, for any target that responded successfully).
3. Verifies that consistency is reset for each page to the original value.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -102,7 +102,7 @@ enum FirstPageContent {
 struct FirstReceivedPage {
     content: FirstPageContent,
     tracing_id: Option<Uuid>,
-    request_coordinator: Option<Coordinator>,
+    request_coordinator: Coordinator,
 }
 
 type ResultNextPage = Result<NextReceivedPage, NextPageError>;
@@ -479,7 +479,7 @@ impl PagerWorker {
                                     rows: DeserializedMetadataAndRawRows::mock_empty(),
                                 },
                                 tracing_id: None,
-                                request_coordinator: Some(coordinator),
+                                request_coordinator: coordinator,
                             },
                             PagingStateResponse::NoMorePages,
                         ));
@@ -580,7 +580,7 @@ impl PagerWorker {
                     FirstReceivedPage {
                         content: FirstPageContent::Rows { rows },
                         tracing_id,
-                        request_coordinator: Some(coordinator),
+                        request_coordinator: coordinator,
                     },
                     paging_state_response,
                 ))
@@ -605,7 +605,7 @@ impl PagerWorker {
                     FirstReceivedPage {
                         content: FirstPageContent::SetKeyspace { set_keyspace },
                         tracing_id,
-                        request_coordinator: Some(coordinator),
+                        request_coordinator: coordinator,
                     },
                     PagingStateResponse::NoMorePages,
                 ))
@@ -624,7 +624,7 @@ impl PagerWorker {
                     FirstReceivedPage {
                         content: FirstPageContent::SchemaChange { schema_change },
                         tracing_id,
-                        request_coordinator: Some(coordinator),
+                        request_coordinator: coordinator,
                     },
                     PagingStateResponse::NoMorePages,
                 ))
@@ -645,7 +645,7 @@ impl PagerWorker {
                             rows: DeserializedMetadataAndRawRows::mock_empty(),
                         },
                         tracing_id,
-                        request_coordinator: Some(coordinator),
+                        request_coordinator: coordinator,
                     },
                     PagingStateResponse::NoMorePages,
                 ))
@@ -1505,11 +1505,8 @@ If you are using this API, you are probably doing something wrong."
         session: &Session,
     ) -> Result<Self, PagerExecutionError> {
         let tracing_ids = Vec::from_iter(first_page.tracing_id);
-        let coordinator_id = first_page
-            .request_coordinator
-            .as_ref()
-            .map(|coordinator| coordinator.node().host_id);
-        let request_coordinators = Vec::from_iter(first_page.request_coordinator);
+        let coordinator_id = first_page.request_coordinator.node().host_id;
+        let request_coordinators = vec![first_page.request_coordinator];
 
         let first_page = match first_page.content {
             FirstPageContent::Rows { rows } => RawRowLendingIterator::new(rows),
@@ -1548,20 +1545,7 @@ If you are using this API, you are probably doing something wrong."
                     warnings: Vec::new(),
                 };
                 session
-                    .handle_auto_await_schema_agreement(
-                        &response,
-                        // Making it impossible to pass None here on the type level would be possible,
-                        // but it would require heavy restructuring. The culprit is SingleConnectionPagerWorker,
-                        // which is used for ControlConnection::execute_iter, and which doesn't have enough
-                        // data to provide a Coordinator in its proof of sending the first page.
-                        //
-                        // I could duplicate data types and the proving sender with Coordinator for PagerWorker
-                        // and no Coordinator for SingleConnectionPagerWorker, but it's not worth it given that
-                        // this None here is an erroneous situation that can only happen due to a bug in the driver.
-                        //
-                        // Also, we hope to refactor the Pager soon [#1549](https://github.com/scylladb/scylla-rust-driver/issues/1549).
-                        coordinator_id.expect("PagerWorker always has Coordinator specified"),
-                    )
+                    .handle_auto_await_schema_agreement(&response, coordinator_id)
                     .await?;
                 // The stream will be empty.
                 RawRowLendingIterator::new(DeserializedMetadataAndRawRows::mock_empty())
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -293,13 +293,8 @@ impl PageSender {
 
 // PagerWorker works in the background to fetch pages
 // QueryPager receives them through a channel
-struct PagerWorker<'a, QueryFunc, SpanCreatorFunc> {
-    // Closure used to perform a single page query
-    // AsyncFn(Arc<Connection>, Option<Arc<[u8]>>) -> Result<QueryResponse, RequestAttemptError>
-    page_query: QueryFunc,
-
+struct PagerWorker {
     load_balancing_policy: Arc<dyn LoadBalancingPolicy>,
-    routing_info: RoutingInfo<'a>,
     query_is_idempotent: bool,
     query_consistency: Consistency,
     retry_session: Box<dyn RetrySession>,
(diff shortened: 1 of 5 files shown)


scylla/src/client/pager.rs
```diff
@@ -476,110 +590,23 @@ where
...
+                        return Ok((
+                            FirstReceivedPage {
+                                content: FirstPageContent::Rows {
+                                    rows: DeserializedMetadataAndRawRows::mock_empty(),
+                                },
+                                tracing_id: None,
+                                request_coordinator: None,
+                            },
+                            PagingStateResponse::NoMorePages,
+                        ));
```
Lorak-mmk, line 534: Commit: "pager: fetch first page without spawning"

Old code used `Some(coordinator.clone())`, new uses `None`, why?

scylla/src/client/pager.rs
```diff
@@ -365,38 +439,48 @@ where
...
-                    .query_pages(
-                        &connection,
-                        current_consistency,
-                        node,
-                        coordinator.clone(),
-                        sender,
-                    )
+                let fetch_result = self
+                    .fetch_one_page(&connection, consistency, &request_span, &page_query)
                     .instrument(span.clone())
```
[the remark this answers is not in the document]
Lorak-mmk: This looks important if true!

Also, need to decide about consistency-downgrade behavior between pages. I think pages should be considered separate requests, so consistency should reset between pages.

scylla/src/client/pager.rs
```diff
@@ -478,7 +478,7 @@ where
                                     rows: DeserializedMetadataAndRawRows::mock_empty(),
                                 },
                                 tracing_id: None,
-                                request_coordinator: None,
+                                request_coordinator: Some(coordinator),
                             },
                             PagingStateResponse::NoMorePages,
```
Lorak-mmk: Commit: "pager: set coordinator in IgnoreWriteError case"

Why do you have a commit that fixes that instead of doing it the right way in the first place?

scylla/src/client/pager.rs
```diff
@@ -204,16 +219,48 @@ where
...
+                    .map(|(node, shard)| {
+                        let span = create_execution_span(node, Some(shard));
+                        // For each node in the plan choose a connection to use.
+                        // This connection will be reused for same node retries to preserve paging cache on the shard.
+                        let connection_fut = itertools::Either::Right(
+                            node.connection_for_shard(shard).instrument(span.clone()),
+                        );
+                        (node, Some(shard), span, connection_fut)
+                    }),
+            );
```
Lorak-mmk, line 237: We now have target-deduplication in 3 different places: DefaultPolicy, Plan, and here. Can we have one code for it?

Lorak-mmk (changes requested):

scylla/src/client/pager.rs
wprzytula: I didn't notice this LLM artifact - I assumed it was `None` in the old code too.

wprzytula:
Rebased on main.

wprzytula:
Addressed some comments.

wprzytula:
Rebased on main.

scylla/src/client/pager.rs
wprzytula: Idk what you mean. Each of these 3 sites has a different semantics for choosing a target to be deduplicated. I don't know how I could make them one code.

wprzytula:
Addressed remaining comments. One requires further clarification.

scylla/src/client/pager.rs
```diff
@@ -1343,103 +1296,177 @@ If you are using this API, you are probably doing something wrong."
...
+                first_page_coordinator,
+                first_page_connection,
+            } => {
+                /* REMAINING PAGES */
+                let worker_task = async move {
+                    let partition_key = match config.prepared.extract_partition_key(&config.values)
+                    {
+                        Ok(res) => res,
+                        Err(err) => {
+                            let _ = sender
```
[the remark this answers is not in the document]
wprzytula: Good catch!

scylla/src/client/pager.rs
Lorak-mmk: All 3 have different semantics? Perhaps that needs some unification?

wprzytula:
Addressed another round of Copilot's comments.

wprzytula:
Fixed one test - it didn't work on Cassandra.
Added one more regression test for token-unaware queries.

wprzytula:
Addressed CodeRabbit's comments.

wprzytula:
Rebased on main.

scylla/src/client/pager.rs
```diff
@@ -1459,181 +1494,116 @@ If you are using this API, you are probably doing something wrong."
...
-                        );
-                        return futures::future::pending().await;
-                    }
-                }
+            .map_err(|RequestTimeoutError(timeout)| {
+                NextPageError::RequestFailure(RequestError::RequestTimeout(timeout))
+            })?
+            .map_err(|attempt_error| {
+                NextPageError::RequestFailure(RequestError::LastAttemptError(attempt_error))
+            })?;
```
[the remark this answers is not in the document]
wprzytula: Single-connection pager worker purposely supports only executing SELECTs.

wprzytula:
Fixed bugs found in an offline review.

The span structure is still bad, but I don't want to invest much in improving it here. Execution path deduplication is to be done first, then we will get a final good span tree.

scylla/tests/integration/session/pager.rs
```diff
@@ -815,3 +820,464 @@ async fn test_token_unaware_pager() {
...
+                .await
+                .unwrap();
+
+            let mut create_ks = format!(
+                "CREATE KEYSPACE IF NOT EXISTS {ks} WITH \
+                 REPLICATION = {{'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}}"
+            );
+            if scylla_supports_tablets(&session).await {
+                create_ks += " and TABLETS = { 'enabled': false}";
+            }
```
Lorak-mmk: Why?

scylla/tests/integration/session/pager.rs
```diff
@@ -815,3 +820,464 @@ async fn test_token_unaware_pager() {
...
+                .await
+                .unwrap()
+                .rows_stream::<(i32,)>()
+                .unwrap();
+
+            let mut count = 0;
+            while let Some(row) = row_stream.next().await {
+                row.unwrap();
+                count += 1;
+            }
```
Lorak-mmk: Possible functional impl: `let count = row_stream.map(|row| row.unwrap()).count().await;`, then `row_stream` doesn't need mut.

scylla/tests/integration/session/pager.rs
wprzytula: LLM artifact that has snipped through my review.

wprzytula:
Addressed review comments. As a bonus, made all newly added pager tests actually not use any DDLs.

Lorak-mmk (approved):

PR #1746, opened 2026-06-23
renovate: Don't auto update rand

Lorak-mmk:
It is part of the public API. Need to update in 2.0.

## Pre-review checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -46,6 +46,21 @@
                 "secrecy-08"
             ],
             "enabled": false
+        },
+        {
+            "description": "rand is part of public API; surface new versions on the Dependency Dashboard only (no auto-PR), to be addressed in 2.0.",
+            "matchManagers": [
+                "cargo"
+            ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -46,6 +46,21 @@
...
+            "matchDepNames": [
+                "rand",
+                "rand_pcg",
+                "rand_chacha"
+            ],
+            "matchFileNames": [
+                "scylla/Cargo.toml"
+            ],
+            "dependencyDashboardApproval": true
         }
```
wprzytula, line 64: ❓ How is rand part of the public API? Let's mention that here.

Lorak-mmk: It is explained in Cargo.toml, where the dependency is defined:
```
# Part of ReplicaSet::choose_filtered public API.
# Used internally in some other places (mostly in LBP / routing).
rand = "0.9.0"
```

Is that not enough?

wprzytula (approved):

PR #1749, opened 2026-06-23
Fix bug in `Metrics::get_latency_percentile_ms`

### LINKED ISSUE #1745: `Metrics::get_latency_percentile_ms` returns number of bucket observations instead of mean value
We had this exact same bug for snapshot before, and it was fixed here: https://github.com/scylladb/scylla-rust-driver/pull/1327

It is still present for get_latency_percentile_ms

Lorak-mmk:
It was returning bucket count, instead of median value.
This is the exact same bug that was fixed in https://github.com/scylladb/scylla-rust-driver/pull/1327 , but this time
in get_latency_percentile_ms.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1745

## Pre-review checklist

### DIFF
--- a/scylla/src/observability/metrics.rs
+++ b/scylla/src/observability/metrics.rs
@@ -337,7 +337,7 @@ impl Metrics {
     /// Returns latency from histogram for a given percentile
     /// # Arguments
     ///
-    /// * `percentile` - float value (0.0 - 100.0)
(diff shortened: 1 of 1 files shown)


Lorak-mmk:
Rebased on main

wprzytula (approved):

PR #1750, opened 2026-06-23
Metrics: `cfg` refactor

### LINKED ISSUE #1268: Encapsulate `metrics` feature-related conditional complilation in `metrics.rs`
@Lorak-mmk suggested limiting `metrics`-related `cfg(feature)`s to `metrics.rs` module in order to increase readability.

https://github.com/scylladb/scylla-rust-driver/pull/1147#discussion_r1963534663

Lorak-mmk:
Metrics are guarded by `metrics` feature. Unfortunately, they need to be passed around in many places in our codebase, leading to sprawl of `cfg` directives. I really don't like having this amount of conditional compilation, so here I implement a solution that is in my opinion cleaner: having compile-time backends for metrics.

In other words, when `metrics` feature is disabled we use a small no-op Metrics implementation, and when it is enabled we use the real one.
No-op backend has no public items, and only 0-sized, or uninhabited items, with empty methods, so it should have 0 performance overhead.
It has no public items, and no querying methods, so there is no way to accidently misuse it and get wrong results.

Public metrics-dependent stuff is still guarded by cfg: `metrics` field in `Context`, `Session::get_metrics`, `PercentileSpeculativeExecutionPolicy`.

I saw one posssible undesirable behavior: we pass around `Arc<Metrics>`, not just `Metrics`. This could cause some overhead even when feature is disabled. I fixed that by moving to just `Metrics` wherever possible, and storing the `Arc` inside.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1268

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1172,7 +1172,7 @@ impl Session {
             host_listener,
             config.cluster_metadata_refresh_interval,
             tablet_receiver,
-            Arc::clone(&metrics),
+            metrics.as_ref().clone(),
             client_routes_config,
         )
         .await?;
@@ -1474,7 +1474,7 @@ impl Session {
             statement,
             execution_profile,
             self.cluster.get_state(),
-            Arc::clone(&self.metrics),
+            self.metrics.as_ref().clone(),
             Arc::clone(&self.node_location_preference),
         )
         .await
@@ -1800,7 +1800,7 @@ impl Session {
                 values,
                 execution_profile,
                 cluster_state: self.cluster.get_state(),
-                metrics: Arc::clone(&self.metrics),
+                metrics: self.metrics.as_ref().clone(),
                 location_preference: Arc::clone(&self.node_location_preference),
             },
         )
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -22,7 +22,6 @@ use crate::network::{
 };
 use crate::observability::driver_tracing::RequestSpan;
 use crate::observability::history::{self, HistoryListener};
-#[cfg(feature = "metrics")]
 use crate::observability::metrics::Metrics;
 use crate::observability::tracing::TracingInfo;
 use crate::policies::address_translator::AddressTranslator;
@@ -88,7 +87,6 @@ pub struct Session {
     cluster: Cluster,
     default_execution_profile_handle: ExecutionProfileHandle,
     schema_agreement_interval: Duration,
-    #[cfg(feature = "metrics")]
     metrics: Arc<Metrics>,
     schema_agreement_timeout: Duration,
     schema_agreement_automatic_waiting: bool,
@@ -113,7 +111,6 @@ impl std::fmt::Debug for Session {
             )
             .field("schema_agreement_interval", &self.schema_agreement_interval);
 
-        #[cfg(feature = "metrics")]
         d.field("metrics", &self.metrics);
 
         d.field(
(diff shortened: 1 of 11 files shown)


scylla/src/client/session.rs
```diff
@@ -1140,7 +1137,6 @@ impl Session {
             reconnect_policy: Arc::new(ExponentialReconnectPolicy::new()),
         };
 
-        #[cfg(feature = "metrics")]
         let metrics = Arc::new(Metrics::new());
```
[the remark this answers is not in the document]
Lorak-mmk: I cannot, Context has an Arc. This single per-session allocation is not a problem.

Lorak-mmk:
Rebased on main

scylla/src/network/connection_pool.rs
```diff
@@ -532,8 +529,7 @@ struct UseKeyspaceRequest {
 }
 
 impl PoolRefiller {
-    #[allow(clippy::too_many_arguments)] // Not always triggered, because of the metrics, so
-    // I can't use `expect`.
+    #[allow(clippy::too_many_arguments)]
```
wprzytula: ⛏️ Now it should be always triggered, so you can switch to `expect`.

scylla/src/observability/metrics_noop.rs
```diff
@@ -0,0 +1,59 @@
+//! No-op metrics module used when the `metrics` feature is disabled.
+//!
+//! All recording methods are zero-cost no-ops;
+//! Nothing is public, to make sure we don't expose those no-op structs to the user.
+//! Query methods are removed to make sure we don't use non-functioning methods by accident.
+
+/// Error that occurred upon a metrics operation.
+#[non_exhaustive]
+pub(crate) enum MetricsError {}
```
wprzytula, line 8: ❓ Is this `non_exhaustive` just-in-case? The error is non `pub` anyway.

scylla/src/client/session.rs
```diff
@@ -1172,7 +1172,7 @@ impl Session {
             host_listener,
             config.cluster_metadata_refresh_interval,
             tablet_receiver,
-            Arc::clone(&metrics),
+            metrics.as_ref().clone(),
             client_routes_config,
```
wprzytula, line 1176: 🤔 Do we need an `Arc<Metrics>` here? Why not just have `Metrics` and cheaply clone them?

scylla/src/client/session.rs
```diff
@@ -1800,7 +1800,7 @@ impl Session {
                 values,
                 execution_profile,
                 cluster_state: self.cluster.get_state(),
-                metrics: Arc::clone(&self.metrics),
+                metrics: self.metrics.as_ref().clone(),
                 location_preference: Arc::clone(&self.node_location_preference),
             },
```
wprzytula, line 1805: 🤔 Same doubt. Why have `metrics` as `Arc<Metrics>` instead of `Metrics`?

scylla/src/observability/metrics_noop.rs
Lorak-mmk: Yes, I'll remove it.

scylla/src/client/session.rs
Lorak-mmk: Discussed offline. To prevent future confusion I improved commit message a bit.

scylla/src/client/session.rs
Lorak-mmk: ditto

Lorak-mmk:

wprzytula (approved):

PR #1758, opened 2026-06-30
pager: fix reported retry consistency and improve test coverage

wprzytula:
This PR improves test coverage of the `QueryPager` by introducing two regression tests.

One guards against a pre-existing bug - reporting a wrong consistency to `RetryPolicy` (the original consistency instead of the actual consistency) - and fixes the bug.

Another guards against a bug that I've introduced during my refactor (it has never landed on main). The bug involved an error thrown when executing token-unaware prepared statements.

I paid attention to keeping the tests lightweight: they execute no DDLs. Bows to @Lorak-mmk.

## Pre-review checklist

### DIFF
--- a/scylla/tests/integration/session/pager.rs
+++ b/scylla/tests/integration/session/pager.rs
@@ -701,3 +701,117 @@ async fn test_pager_retry_receives_current_consistency() {
         );
     }
 }
+
+/// Regression test: `execute_iter` and `query_iter` must not return a spurious
+/// `PartitionKeyError` for token-unaware queries.
+///
+/// The bug was in `new_for_prepared_statement`: it called
+/// `extract_partition_key_and_calculate_token` unconditionally, whereas
+/// previously (before the "fetch first page without spawning" refactor) that
+/// call was only reached inside the spawned worker task, which was guarded by
+/// `is_token_aware()`. For non-token-aware statements the call returns
+/// `Ok(None)` today, so there is no actual failure — but this test locks in
+/// the correct behaviour and will catch any future regression.
+///
+/// Three cases are exercised:
+///
+/// 1. Unprepared `Statement` + `query_iter` (token is always `None` here;
+///    baseline sanity check).
+/// 2. Token-unaware `PreparedStatement` with **no bind markers**
+///    (`SELECT peer FROM system.peers` — the PK column `peer` never appears
+///    as `?`, so the server returns `pk_indexes = []`).
+/// 3. Token-unaware `PreparedStatement` with **bind markers on non-PK
+///    columns** (`system_schema.columns WHERE keyspace_name = 'system' AND
+///    position >= ?` — the PK `keyspace_name` is a hardcoded literal, so
+///    again `pk_indexes = []`).
+///
+/// All three cases must succeed end-to-end and return at least one row, using
+/// `page_size = 1` to force the multi-page `work()` code path.
+#[tokio::test]
+async fn test_token_unaware_pager() {
+    setup_tracing();
+
+    let session = create_new_session_builder().build().await.unwrap();
+
+    // Drain a pager and assert it yields at least one row without error.
+    async fn assert_pager_succeeds(
+        mut stream: impl futures::Stream<Item = Result<Row, impl std::error::Error>> + Unpin,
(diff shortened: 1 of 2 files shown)


scylla/tests/integration/session/pager.rs
```diff
@@ -535,3 +536,282 @@ async fn test_pager_timeouts() {
...
+        Err(ProxyError::Worker(WorkerError::DriverDisconnected(_))) => (),
+        Err(err) => panic!("{}", err),
+    }
+
+    // Both retry sessions must have observed Consistency::Two (the downgraded
+    // CL), not Consistency::All (the original query CL).
+    for (label, rx) in [("first page", &mut rx1), ("second page", &mut rx2)] {
+        let observed = rx.recv().await.unwrap_or_else(|| {
+            panic!("[{label}] retry policy did not send an observed consistency")
+        });
```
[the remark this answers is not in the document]
wprzytula: Every test in the CI has a 60s timeout, which is enough.

scylla/tests/integration/session/pager.rs
```diff
@@ -535,3 +536,282 @@ async fn test_pager_timeouts() {
...
+    async fn assert_pager_succeeds(
+        mut stream: impl futures::Stream<Item = Result<Row, impl std::fmt::Debug>> + Unpin,
+        label: &str,
+    ) {
+        let mut count = 0_usize;
+        while let Some(row) = stream.next().await {
+            row.unwrap_or_else(|e| panic!("[{label}] unexpected stream error: {e:?}"));
+            count += 1;
+        }
+        assert!(count >= 1, "[{label}] expected at least one row, got none");
```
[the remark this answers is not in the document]
wprzytula: We already everywhere assume a 3-node cluster operating for our testing.

scylla/tests/integration/session/pager.rs
```diff
@@ -535,3 +536,168 @@ async fn test_pager_timeouts() {
...
+                            .and(Condition::TrueForLimitedTimes(1)),
+                        RequestReaction::noop(),
+                    ),
+                    // All subsequent Executes: forge error.
+                    RequestRule(
+                        execute_not_control.clone(),
+                        RequestReaction::forge().server_error(),
+                    ),
+                ]));
+            });
```
Lorak-mmk, line 673: Worth noting that this test assumes that stickiness works: if the second page is issues to a different node, it will not be errored-out because this new node `TrueForLimitedTimes` won't be used up yet.

scylla/tests/integration/session/pager.rs
```diff
@@ -701,3 +701,117 @@ async fn test_pager_retry_receives_current_consistency() {
...
+    setup_tracing();
+
+    let session = create_new_session_builder().build().await.unwrap();
+
+    // Drain a pager and assert it yields at least one row without error.
+    async fn assert_pager_succeeds(
+        mut stream: impl futures::Stream<Item = Result<Row, impl std::fmt::Debug>> + Unpin,
+        label: &str,
+    ) {
+        let mut count = 0_usize;
```
Lorak-mmk, line 741: Interesting constraint. Can we do `impl Error`, unless it doesn't work here? I think it is a bit more typical in such situations.

Also, why do you need `Unpin`?

scylla/tests/integration/session/pager.rs
wprzytula: You're right. Action item: temporarily disable this test in the transitional period in #1739 before the coordinator stickiness is restored.

scylla/tests/integration/session/pager.rs
wprzytula: Good questions. I accepted this as Sonnet produced, I can investigate if it can be simplified.

wprzytula: `StreamExt::next()` requires `Unpin`.

wprzytula: Changed to `impl Error`.

Lorak-mmk (approved):

wprzytula:
Fixed naming.

Lorak-mmk (approved):

PR #1766, opened 2026-07-04
scylla-cql-core: document CqlTimeuuid::lsb_signed and harden test_timeuuid_ordering

### LINKED ISSUE #1637: CqlTimeuuid low-half ordering uses the wrong signed transform
`lsb_signed()` in [value.rs#L208-L209](https://github.com/scylladb/scylla-rust-driver/blob/9226fe6c0fbcdc0965cc274c5552ec47da465d33/scylla-cql/src/value.rs#L208-L209) XORs every byte with `0x80` instead of flipping the 64-bit sign bit, and [`cmp()`](https://github.com/scylladb/scylla-rust-driver/blob/9226fe6c0fbcdc0965cc274c5552ec47da465d33/scylla-cql/src/value.rs#L252-L259) relies on that value. Two `timeuuid`s with equal `msb()` can sort incorrectly on the client.

```rust
#[test]
fn timeuuid_lsb_signed_ordering() {
    let x = CqlTimeuuid::from_str("00000000-0000-1000-8000-00000000007f").unwrap();
    let y = CqlTimeuuid::from_str("00000000-0000-1000-8000-000000000080").unwrap();
    assert!(x < y);
}
```

vismaytiwari:
## Pre-review checklist

- [x] I added appropriate `Refs:`/`Fixes:` annotations to PR description.

## Description

`CqlTimeuuid`'s `Ord` compares `msb()` and then `lsb_signed()`. The low 64 bits are compared as **signed bytes** — each of the 8 bytes independently, most-significant first — which Scylla/Cassandra implement as `lsb ^ 0x8080808080808080` followed by an unsigned compare (`timeuuid_tri_compare` in `utils/UUID.hh`). This is **not** the same as comparing the low bits as one signed 64-bit integer.

The existing behavior is correct (it matches Scylla). This PR only makes it harder to break by accident — there is **no functional change**:

- Adds a doc comment on `lsb_signed` explaining the byte-wise signed comparison and why the `^ 0x8080...` transform reproduces it.
- Strengthens `test_timeuuid_ordering`. Its previous values sorted the same whether the low bits were compared byte-wise-signed or as a single 64-bit signed integer, so the test could not catch a regression to the latter. It now includes `00000000-0000-1000-8000-000000000080` and `00000000-0000-1000-8000-00000000007f` (equal `msb`, differing only in the low byte), which sort `...080` before `...07f` **only** under the byte-wise comparison; a 64-bit signed compare would order them the other way.

Refs #1637 — after investigation this turned out not to be a bug: the reported ordering matches how Scylla sorts timeuuids.

### DIFF
--- a/scylla-cql-core/src/value.rs
+++ b/scylla-cql-core/src/value.rs
@@ -239,16 +239,19 @@ impl CqlTimeuuid {
         ])
     }
 
-    /// Returns the least significant bytes of the timeuuid transformed so that
-    /// an unsigned comparison of the result orders the values the same way a
-    /// signed (two's-complement `i64`) comparison of [`Self::lsb`] would.
+    /// Returns the least significant bytes transformed so that an unsigned
+    /// comparison of the result reproduces Scylla/Cassandra's ordering of those
+    /// bytes for two timeuuids with equal `msb` (used by the [`Ord`] impl).
     ///
-    /// Cassandra/Scylla compare the low 64 bits of a timeuuid as a signed
-    /// integer (see the [`Ord`] impl). Flipping the sign bit maps the signed
-    /// range onto the unsigned range while preserving order, which lets `cmp`
-    /// use a plain `u64` comparison.
+    /// Cassandra legacy compares the 8 low bytes as *signed bytes* — each byte
+    /// independently, most-significant byte first — not as one signed 64-bit
+    /// integer. Scylla implements this in `timeuuid_tri_compare` (utils/UUID.hh)
+    /// as `lsb ^ 0x8080808080808080` followed by an unsigned compare: flipping
+    /// the top bit of every byte maps each byte's signed order onto its unsigned
+    /// order, so a big-endian `u64` comparison of the XORed value matches the
+    /// byte-wise signed ordering.
     fn lsb_signed(&self) -> u64 {
-        self.lsb() ^ 0x8000000000000000
+        self.lsb() ^ 0x8080808080808080
     }
 }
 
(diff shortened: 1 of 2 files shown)


Lorak-mmk (changes requested):
The way I understand the problem, a sorting order was different locally (`Ord` impl) than how the server would sort them. Right?

The funny thing is we already have an integration test for timeuuid ordering!
See `test_timeuuid_ordering` in `scylla/tests/integration/types/cql_types.rs`.

It has an array of timeuuids, in specific order.
It verifies that Scylla sorts those timeuuids the same way, and it also verifies that our Ord impl sorts them the same way. So the question is, if our sorting is wrong then why doesn't the test fail? Are the values happen to be chosen in a way that don't trigger the bug? If so, could you change the values to make the test useful?

vismaytiwari:
Hey @Lorak-mmk 
Good catch, and you're right about why it slipped through: the values that were there sorted the same whether the low 64 bits were compared as signed or with the old byte-wise `^ 0x8080…80`, so they never exercised the signed path. I replaced two of them with `00000000-0000-1000-8000-00000000007f` and `00000000-0000-1000-8000-000000000080` — same msb, differing only below the top byte, which the old code ordered backwards. Kept the list at 9 so the 9! permutation check is unchanged. Thanks for the pointer to that test. Also if the changes makes sense
Pls run workflows, It seems only the maintainer  can run it

Lorak-mmk:
Interesting, the tests failed. I tried it locally as well, both for Scylla and Cassandra, and they failed. 
I tried reverting your first commit (the one that changes the comparison logic), leaving only the test change, and the test still fails!

So either the test change is wrong, or Scylla / Cassandra use even different comparison logic - or both.
The next step is to look into this more carefully. Comparison logis is defined in in scylla repo in `utils/UUID.hh`. Need to analyze it, check if it matches our semantics (either before or after your change), and fix our logic if necessary.

vismaytiwari:
I went through `utils/UUID.hh`. `timeuuid_tri_compare` compares the low 64 bits as `uuid_read_lsb(o) ^ 0x8080808080808080` — a byte-wise signed comparison (Cassandra legacy), not a full 64-bit signed one. The driver's `msb()`/`lsb()` already match Scylla's `timeuuid_read_msb`/`uuid_read_lsb` exactly, and the original `lsb_signed` used that same `^ 0x8080808080808080` — so it already matched Scylla. My change to `^ 0x8000000000000000` is what diverges.

In other words, #1637 assumes the low bits sort as a signed 64-bit integer, but Cassandra/Scylla order them per-byte-signed, so the existing behavior was actually correct.  sorry for the noise, and thanks for pushing on the tests, that's exactly what surfaced it.

I think its better to close this PR.

Lorak-mmk:
It sure looks like there is no bug after all.

I do think there are some changes worth implementing:
- As was noted in the issue, lsb_signed needs a comment explaining what it does and why
- After your semantics change, which introduced a bug, our test_timeuuid_ordering does not fail - so it is definitely needed to change the values there to detect such errors.

vismaytiwari:
Agreed, thanks for digging into it. I've reverted the semantics change — `lsb_signed` is back to the original `^ 0x8080808080808080` (byte-wise signed, matching `timeuuid_tri_compare` in Scylla).

Kept the two useful pieces you pointed out:
- Added a comment on `lsb_signed` explaining it's a byte-wise signed comparison (each of the 8 bytes independently), not a single signed 64-bit comparison, and why the `^ 0x8080...` transform reproduces it.
- Reworked `test_timeuuid_ordering` so it would catch a break like mine: the `...8000-000000000080` / `...8000-00000000007f` pair (equal msb, differ only in the low byte) is ordered `...080` before `...07f`, which only holds under the byte-wise comparison — a 64-bit signed compare would sort them the other way. The values that were there before sorted the same under either transform, so they couldn't detect it.

No functional change now; sorry again for the detour.

Lorak-mmk (approved):
Thanks! I'll squash-merge because of unclean history.

Issue #1549, opened 2026-01-16
Deduplicate request execution paths


Lorak-mmk:
Currently we have a lot of duplicated code between `query/execute_unpaged/single_page` and `query/execute_iter`. This is a root cause of https://github.com/scylladb/scylla-rust-driver/issues/1495 , https://github.com/scylladb/scylla-rust-driver/issues/1495 , https://github.com/scylladb/scylla-rust-driver/issues/1418 , their lack of use of speculative execution (for which we don't have an issue I think), and other differences of behavior. We could try to patch all that, but its an endless work.

We should have:
- one function (working name: `run_request_page`) that runs the full request (think: either single page, or whole unpaged request), including retries and speculative execution.
- One function (working name: `run_any_request`), that internally uses the above, but also has the capability to handle results like schema change, use keyspace etc
 
`run_any_request` should be used by `_unpaged` and `_single_page` APIs.
`_iter` APIs should use it for first page, and then use the `run_request_page` for subsequent pages (in a separate Tokio task). This is because handling special request types requires access to `Session`, which we don't want to require in a task spawned by `_iter`.
This has the additional advantage of not needing to spawn a Tokio task for non-select requests and selects that return single page, lowering the overhead of `_iter`.

PR #1770, opened 2026-07-07
session: small fixes & refactors

wprzytula:
Refs: #1549

## Motivation

While working on #1549, I decide to extract some noncontroversial prefix of the work to a separate PR, so here it is.

## Contents

- Fixes bug that `retry_policy::RequestInfo` was populated with the original consistency instead of the (possibly different) actual consistency used for the request. Added a regression test analogous to one already present in `pager.rs`.
- Updates obsolete naming, moves some code, fixes typo.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1016,11 +1016,7 @@ impl Session {
                 execution_profile,
                 |connection: Arc<Connection>,
                  consistency: Consistency,
-                 execution_profile: &ExecutionProfileInner| {
-                    let serial_consistency = batch
-                        .config
-                        .serial_consistency
-                        .unwrap_or(execution_profile.serial_consistency);
+                 _execution_profile: &ExecutionProfileInner| {
                     async move {
                         connection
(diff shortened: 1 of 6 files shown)


scylla/src/client/session.rs
Lorak-mmk, line 1: Commit: "session: fix bug with consistency in retry decision"

It would be best to open an issue for that, and maybe even have as separate PR (but I don't insist on either of those). Right now we risk missing this important change when writing the changelog.

wprzytula:
Rebased on main.

scylla/src/client/session.rs
Lorak-mmk: Actually let's at least have an issue.

wprzytula: https://github.com/scylladb/scylla-rust-driver/pull/1758 is visible for the changelog purposes and will remind us to describe that the issue was present for both `_iter` and non-`_iter` execution APIs. This is IMO good enough.

Lorak-mmk (approved):

Issue #1629, opened 2026-04-29
Expose some ergonomic tools for tuple serialization / deserialization


Lorak-mmk:
Users that implement their own ser/deser can do so easily for most collections.
For example:
- Vectors can be serialized from `Vec<T>` or `&[T]`, deserialized as `Vec<T>` or `VectorIterator<T>`
- Sets can be serialized from `Vec<T>`,  `&[T]`, `HashSet<T, S>`, and deserialized to `ListlikeIterator<T>`, `HashSet<T, S>`, `BTreeSet<T>` or `Vec<T>`

Tuples can be serialized / deserialized to Rust tuples, but it is not a great fit for more dynamic use cases like latte.
Then there is `CqlValue`, but we want to avoid it for any performance-sensitive work.
This means that right now users need to operate on raw bytes (`read_bytes_opt` etc) and handle e.g. use cases from differing tuple sizes.

I think we should:
- Allow ser/deser of tuples to `Vec<T>`
- Introduce deserializing iterator for tuples, similarly to what we have for vectors / lists / sets

PR #1771, opened 2026-07-07
Add public methods required for python-rs driver

patrycja-ziemkiewicz:
## Summary
Introduces a python-rs feature flag to conditionally expose internal methods required exclusively by the python-rs driver.
## Changes
-  Added python-rs feature flag.
- Applied #[cfg(feature = "python-rs")] to methods needed by the python-rs driver
- Made serialize_next_variable_length_elem public for python-rs.
- Made serialize_next_constant_length_elem public for python-rs.
- Made raw_rows() public
- Implement constructors for UntranslatedPeer, RequestInfo, RoutingInfo
- Add clone to Peer

## Pre-review checklist

### DIFF
--- a/scylla-cql-core/src/deserialize/frame_slice.rs
+++ b/scylla-cql-core/src/deserialize/frame_slice.rs
@@ -2,7 +2,10 @@
 
 use bytes::Bytes;
 
+use crate::deserialize::value::DeserializeValue;
+use crate::deserialize::{DeserializationError, TypeCheckError};
 use crate::frame::frame_errors::LowLevelDeserializationError;
+use crate::frame::response::result::ColumnType;
 use crate::frame::types;
 
 /// A borrowed reference to a part of the frame.
@@ -189,6 +192,23 @@ impl<'frame> FrameSlice<'frame> {
             frame_subslice: cql_bytes,
             original_frame: self.original_frame,
         }))
+    }
+}
+
+// What is the purpose of implementing DeserializeValue for Option<FrameSlice<'_>>?
+//
+// Sometimes users might be interested in operating on Option<FrameSlice<'_>> directly.
+// Implementing DeserializeValue for it allows us to simplify our interface.
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for Option<FrameSlice<'frame>> {
+    fn type_check(_typ: &ColumnType) -> Result<(), TypeCheckError> {
+        Ok(())
+    }
+
+    fn deserialize(
+        _typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        Ok(v)
     }
 }
 
--- a/scylla-cql-core/src/deserialize/value.rs
+++ b/scylla-cql-core/src/deserialize/value.rs
@@ -1592,6 +1592,39 @@ where
     }
 }
 
+/// A type that pairs a frame slice with its column metadata during deserialization.
+/// This is useful when you need both the raw serialized data and the type information
+/// to properly interpret or further deserialize the value.
+pub struct FrameSliceWithMetadata<'frame, 'metadata> {
+    pub frame_slice: Option<FrameSlice<'frame>>,
+    pub column_type: &'metadata ColumnType<'metadata>,
+}
+
+/// What is the purpose of implementing DeserializeValue for FrameSliceWithMetadata<'_, '_>?
+///
+/// Sometimes users might be interested in getting both the raw frame slice and the column
+/// metadata together during deserialization. This allows deferred or custom deserialization
+/// where the user needs access to the type information alongside the serialized bytes.
+/// Implementing DeserializeValue for FrameSliceWithMetadata<'_, '_> allows us to
+/// simplify our interface and provide this capability directly.
+impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata>
+    for FrameSliceWithMetadata<'frame, 'metadata>
+{
+    fn type_check(_typ: &ColumnType) -> Result<(), TypeCheckError> {
+        Ok(())
+    }
+
+    fn deserialize(
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        Ok(FrameSliceWithMetadata {
+            frame_slice: v,
+            column_type: typ,
+        })
+    }
+}
+
 // tuples
 
 // Implements tuple deserialization.
--- a/scylla-cql-core/src/serialize/value.rs
+++ b/scylla-cql-core/src/serialize/value.rs
@@ -980,7 +980,19 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
         .map_err(|_| mk_ser_err_named(rust_name, typ, BuiltinSerializationErrorKind::SizeOverflow))
 }
 
-fn serialize_next_constant_length_elem<'t, T: SerializeValue + 't>(
+/// Serializes an element with constant-length encoding
+#[cfg(all(scylla_unstable, feature = "unstable-python-rs"))]
+pub fn serialize_next_constant_length_elem_unstable<'t, T: SerializeValue + 't>(
+    rust_name: &'static str,
+    element_type: &ColumnType,
+    typ: &ColumnType,
+    builder: &mut CellValueBuilder,
+    element: &'t T,
+) -> Result<(), SerializationError> {
+    serialize_next_constant_length_elem(rust_name, element_type, typ, builder, element)
+}
+
+pub(crate) fn serialize_next_constant_length_elem<'t, T: SerializeValue + 't>(
     rust_name: &'static str,
     element_type: &ColumnType,
     typ: &ColumnType,
--- a/scylla-cql-core/src/serialize/value.rs
+++ b/scylla-cql-core/src/serialize/value.rs
@@ -1002,7 +1002,19 @@ fn serialize_next_constant_length_elem<'t, T: SerializeValue + 't>(
     Ok(())
 }
 
-fn serialize_next_variable_length_elem<'t, T: SerializeValue + 't>(
+/// Serializes an element with variable-length encoding (vint length prefix + data).
+#[cfg(all(scylla_unstable, feature = "unstable-python-rs"))]
+pub fn serialize_next_variable_length_elem_unstable<'t, T: SerializeValue + 't>(
+    rust_name: &'static str,
+    element_type: &ColumnType,
+    typ: &ColumnType,
+    builder: &mut CellValueBuilder,
+    element: &'t T,
+) -> Result<(), SerializationError> {
+    serialize_next_variable_length_elem(rust_name, element_type, typ, builder, element)
+}
+
+pub(crate) fn serialize_next_variable_length_elem<'t, T: SerializeValue + 't>(
     rust_name: &'static str,
     element_type: &ColumnType,
     typ: &ColumnType,
--- a/scylla-cql/src/frame/response/result.rs
+++ b/scylla-cql/src/frame/response/result.rs
@@ -449,9 +449,7 @@ impl DeserializedMetadataAndRawRows {
         TypedRowIterator::new(raw)
     }
 
-    /// Allows to retrieve raw rows, without the need for deserialization
-    /// Intended to be used only in nodejs-rs driver only.
-    #[cfg(all(scylla_unstable, feature = "unstable-nodejs-rs"))]
+    /// Allows to retrieve raw rows, without the need for deserialization.
     pub fn raw_rows(&self) -> &Bytes {
         &self.raw_rows
     }
--- a/scylla/src/cluster/metadata/mod.rs
+++ b/scylla/src/cluster/metadata/mod.rs
@@ -61,6 +61,7 @@ pub(crate) struct Metadata {
 }
 
 /// Represents a node in the cluster, as fetched from the `system.{peers,local}` tables.
+#[cfg_attr(all(scylla_unstable, feature = "unstable-python-rs"), derive(Clone))] // <- for python-rs HostFilter wrapper to store Peer and reconstruct it from &Peer in accept()
 #[non_exhaustive] // <- so that we can add more fields in a backwards-compatible way
 pub struct Peer {
     /// Unique identifier of the node.
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -53,13 +53,8 @@ pub struct ClusterState {
     pub(crate) cluster_name: Option<String>,
 }
 
-/// Enables printing [ClusterState] struct in a neat way, skipping the clutter involved by
-/// [ClusterState::ring] being large and [Self::keyspaces] debug print being very verbose by default.
-pub(crate) struct ClusterStateNeatDebug<'a>(pub(crate) &'a Arc<ClusterState>);
-impl std::fmt::Debug for ClusterStateNeatDebug<'_> {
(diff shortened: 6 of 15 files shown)


scylla-cql-core/src/deserialize/value.rs
```diff
@@ -1990,7 +1990,8 @@ pub struct FixedLengthBytesSequenceIterator<'frame> {
 }
 
 impl<'frame> FixedLengthBytesSequenceIterator<'frame> {
-    fn new(count: usize, slice: FrameSlice<'frame>) -> Self {
+    /// Creates a new iterator for a fixed-count sequence of `[bytes]` items.
+    pub fn new(count: usize, slice: FrameSlice<'frame>) -> Self {
         Self {
```
Lorak-mmk: why do you need that? I see no usages of `FixedLengthByteSequenceIterator` in python-rs codebase.

scylla-cql-core/src/frame/types.rs
```diff
@@ -278,6 +278,27 @@ pub(crate) fn unsigned_vint_encode(v: u64, buf: &mut Vec<u8>) {
...
+    };
+
+    if extra_bytes != 0 {
+        v += buf.read_uint::<BigEndian>(extra_bytes)?;
+    }
+
+    Ok(v)
+}
+
+#[cfg(not(all(scylla_unstable, feature = "unstable-python-rs")))]
```
Lorak-mmk: I don't think we actually need that. I actually think using that shows a problem with our current python-rs impl. In python-rs we basically re-implement `VectorIterator`, and its `next_constant_length_elem` and `next_variable_length_elem`. To implement them, you do need `unsigned_vint_decode`. 

But why do we implement them? The correct approach imo is to use what already exists (`impl Iterator for VectorIterator`). To do this, we would need to use type `T` that always deserializes, and then in python-rs deserialize such element to our target type.

Two ways to do this:
- Have some wrapper for `Option<FrameSlice<'_>>` in Python-rs, add DeserializeValue impl for it
- Add DeserializeValue impl for `Option<FrameSlice<'_>>` in Rust Driver.

cc @wprzytula

Lorak-mmk: btw even if we decide to publish that, let's not duplicate code - let's have a private function with impl, and then 2 "proxy" functions that just call internal one - one public and one not.

scylla-cql-core/src/frame/types.rs
wprzytula: 🔧 For non-trivial functions, such duplication is not acceptable.

scylla-cql-core/src/serialize/value.rs
wprzytula: 🤔 It's weird we're only exposing the variable-length variant but keep constant-length variant unexposed.

scylla-cql/src/frame/response/result.rs
```diff
@@ -449,9 +449,12 @@ impl DeserializedMetadataAndRawRows {
...
-    #[cfg(all(scylla_unstable, feature = "unstable-nodejs-rs"))]
+    /// Allows to retrieve raw rows, without the need for deserialization.
+    /// Intended to be used only in nodejs-rs and python-rs drivers.
+    #[cfg(all(
+        scylla_unstable,
+        any(feature = "unstable-nodejs-rs", feature = "unstable-python-rs")
+    ))]
     pub fn raw_rows(&self) -> &Bytes {
         &self.raw_rows
     }
```
Lorak-mmk, line 455: Commit: `Make raw_rows() public for python-rs`

This is a type in scylla-cql, not scylla-cql-core, so generally not part of driver public API.

Immediately 2 questions come to my mind:
- How do you even get an object of this type in python-rs? Is it thanks to some API you publish conditionally in some next commit? If so, can you reorder those commits, so that reviewer reading them in order can see the access path?
- Maybe we can just make it pub unconditionally? This is in internal crate after all.

scylla/src/response/query_result.rs
```diff
@@ -127,6 +127,14 @@ impl QueryResult {
         }
     }
 
+    /// Returns the deserialized metadata and raw rows.
+    /// Public only when the `unstable-python-rs` feature is enabled.
+    #[cfg(all(scylla_unstable, feature = "unstable-python-rs"))]
+    pub fn deserialized_metadata_and_rows(&self) -> Option<&DeserializedMetadataAndRawRows> {
+        self.deserialized_metadata_and_rows.as_ref()
+    }
+
```
Lorak-mmk, line 136: Commit: `Make deserialized_metadata_and_rows() public for python-rs`

Ah, here is the access for `DeserializedMetadataAndRawRows`.

Let's definitely reorder this commit with the previous one.

scylla/src/response/query_result.rs
```diff
@@ -415,8 +415,11 @@ impl QueryRowsResult {
...
     /// Allows to obtain reference to `DeserializedMetadataAndRawRows`, without the need for owned object deconstruction.
-    /// Intended to be used only in nodejs-rs driver only.
-    #[cfg(all(scylla_unstable, feature = "unstable-nodejs-rs"))]
+    /// Intended to be used only in nodejs-rs and python-rs drivers.
+    #[cfg(all(
+        scylla_unstable,
+        any(feature = "unstable-nodejs-rs", feature = "unstable-python-rs")
+    ))]
     pub fn raw_rows_with_metadata(&self) -> &DeserializedMetadataAndRawRows {
         &self.raw_rows_with_metadata
```
Lorak-mmk: Why do you need that? I don't see usages in python-rs.

scylla-cql-core/src/serialize/value.rs
```diff
@@ -1002,7 +1002,8 @@ fn serialize_next_constant_length_elem<'t, T: SerializeValue + 't>(
     Ok(())
 }
 
-fn serialize_next_variable_length_elem<'t, T: SerializeValue + 't>(
+/// Serializes an element with variable-length encoding (vint length prefix + data).
+pub fn serialize_next_variable_length_elem<'t, T: SerializeValue + 't>(
     rust_name: &'static str,
     element_type: &ColumnType,
```
Lorak-mmk, line 1031: Let's make this conditionally public, like the rest of the stuff.

When https://github.com/scylladb/scylla-rust-driver/issues/1629 is done we should be able to use that in python-rs and stop using this function.

scylla/src/cluster/metadata/mod.rs
```diff
@@ -61,6 +61,7 @@ pub(crate) struct Metadata {
 }
 
 /// Represents a node in the cluster, as fetched from the `system.{peers,local}` tables.
+#[derive(Clone)]
 #[non_exhaustive] // <- so that we can add more fields in a backwards-compatible way
 pub struct Peer {
     /// Unique identifier of the node.
```
Lorak-mmk, line 67: 1. You can use cfg_attr to only impl this when needed.
2. Why do we need to clone Peer? If we really need to, please explain in the comment.

scylla/src/cluster/state.rs
```diff
@@ -78,6 +78,12 @@ impl std::fmt::Debug for ClusterStateNeatDebug<'_> {
     }
 }
 
+impl std::fmt::Debug for ClusterState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Debug::fmt(&ClusterStateNeatDebug(self), f)
+    }
+}
+
 impl ClusterState {
```
Lorak-mmk, line 76: Commit: "Add Debug for ClusterState"

This commit makes 2 changes: 
- Change ClusterStateNeatDebug to not use Arc
- Add Debug impl for ClusterState

It would be great to split those into separate commits.
The more important thing however, is to explain why this change is correct.
Implementing Debug is the obvious thing - so if in the past `ClusterStateNeatDebug` was introduce instead of just Debug impl, there must have been reasons for that. Please find those reasons, and explain in the commit message why they are wrong or outdated.

Lorak-mmk (changes requested):

scylla-cql-core/src/deserialize/value.rs
patrycja-ziemkiewicz: It is used in sequences deserialization

scylla-cql-core/src/deserialize/frame_slice.rs
```diff
@@ -192,6 +195,23 @@ impl<'frame> FrameSlice<'frame> {
...
+    }
+
+    fn deserialize(
+        _typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        Ok(v)
+    }
+}
+
```
Lorak-mmk, line 214: Don't we need a new type that stores both `Option<FrameSlice<'frame>>` and `&'metadata ColumnType<'metadata>`? cc @wprzytula

scylla-cql/src/frame/response/result.rs
```diff
@@ -449,9 +449,12 @@ impl DeserializedMetadataAndRawRows {
...
-    /// Allows to retrieve raw rows, without the need for deserialization
-    /// Intended to be used only in nodejs-rs driver only.
-    #[cfg(all(scylla_unstable, feature = "unstable-nodejs-rs"))]
+    /// Allows to retrieve raw rows, without the need for deserialization.
+    /// Intended to be used only in nodejs-rs and python-rs drivers.
+    #[cfg(all(
+        scylla_unstable,
+        any(feature = "unstable-nodejs-rs", feature = "unstable-python-rs")
+    ))]
     pub fn raw_rows(&self) -> &Bytes {
```
Lorak-mmk, line 453: This is a private type (in scylla-cql, not scylla-cql-core). Let's make it public unconditionally. 

In general I don't think we need unstable flag in scylla-cql, because we can now make breaking changes there.

scylla-cql-core/src/serialize/value.rs
```diff
@@ -1002,7 +1002,19 @@ fn serialize_next_constant_length_elem<'t, T: SerializeValue + 't>(
     Ok(())
 }
 
-fn serialize_next_variable_length_elem<'t, T: SerializeValue + 't>(
+/// Serializes an element with variable-length encoding (vint length prefix + data).
+#[cfg(all(scylla_unstable, feature = "unstable-python-rs"))]
+pub fn serialize_next_variable_length_elem_<'t, T: SerializeValue + 't>(
+    rust_name: &'static str,
+    element_type: &ColumnType,
```
Lorak-mmk, line 1021: Please change the suffix from `_` to `_unstable` or something like that. The underscore alone is hard to spot and thus confusing.

scylla-cql-core/src/serialize/value.rs
```diff
@@ -980,7 +980,19 @@ fn serialize_sequence<'t, 'b, T: SerializeValue + 't>(
         .map_err(|_| mk_ser_err_named(rust_name, typ, BuiltinSerializationErrorKind::SizeOverflow))
 }
 
-fn serialize_next_constant_length_elem<'t, T: SerializeValue + 't>(
+/// Serializes an element with constant-length encoding
+#[cfg(all(scylla_unstable, feature = "unstable-python-rs"))]
+pub fn serialize_next_constant_length_elem_<'t, T: SerializeValue + 't>(
+    rust_name: &'static str,
```
Lorak-mmk, line 986: ditto

Lorak-mmk (changes requested):

scylla-cql-core/src/deserialize/frame_slice.rs
Lorak-mmk: We already have `RawValue` in scylla-cql-core, so this name is unavailable.

`RowValue`? `RawValueWithMetadata`?

Lorak-mmk: Btw, if you have impl for `Option<FrameSlice<'frame>>`, then you can't deserialize to just `FrameSlice<'frame>`, which would be useful if you don't allow nulls.

If you make impl for `FrameSlice<'frame>` instead, then you can deserialize to both (because there is impl for Option<T>).

So we can:
- keep your impl 
- Switch to impl for `FrameSlice<'frame>`
- Remove it

Not sure which is best.

wprzytula: > Btw, if you have impl for `Option<FrameSlice<'frame>>`, then you can't deserialize to just `FrameSlice<'frame>`, which would be useful if you don't allow nulls.

We discussed it yesterday and my stance was that we don't want to allow `FrameSlice` deserialization from a possibly null value (which we do allow for end types such such as i32, simply failing deserialization if we encounter null), because `FrameSlice` is a special internal representation, and deserialization to it should be an infallible noop. Otherwise, we can introduce bugs, by not handling the null case explicitly. By forcing deserialization to `Option<FrameSlice>`, we harden against misuse. It's trivial for consumers to fail on None anyway if they desire to.

wprzytula: > We already have `RawValue` in scylla-cql-core, so this name is unavailable.
> 
> `RowValue`? `RawValueWithMetadata`?

- Let's not mention `Row` in the single value context.
- `RawValue` is something different from `FrameSlice` and only used in serialization context. It can represent `Unset`, which is specific to serialization too.
- My idea is straightforward: `FrameSliceWithMetadata<'frame, 'metadata>`. I'd allow deserializing `Option<FrameSliceWithMetadata>` in the same way `Option<FrameSlice>` is supported.

Lorak-mmk: > My idea is straightforward: FrameSliceWithMetadata<'frame, 'metadata>. I'd allow deserializing Option<FrameSliceWithMetadata> in the same way Option<FrameSlice> is supported.

Doesn't work: you have no way to get metadata of a null.
We need `FrameSliceWithMetadata<'frame, 'metadata>` which stores `Option<FrameSlice>` and metadata.

Lorak-mmk (approved):

scylla-cql-core/src/deserialize/frame_slice.rs
```diff
@@ -212,6 +212,13 @@ impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for Option<FrameSlic
...
 }
 
+// A type that pairs a frame slice with its column metadata during deserialization.
+// This is useful when you need both the raw serialized data and the type information
+// to properly interpret or further deserialize the value.
+pub struct FrameSliceWithMetadata<'frame, 'metadata> {
+    pub frame_slice: Option<FrameSlice<'frame>>,
+    pub column_type: &'metadata ColumnType<'metadata>,
+}
 #[cfg(test)]
```
wprzytula: 🔧 This should be a docstring, not plain comment.
🔧 Missing newline.

scylla-cql-core/src/deserialize/frame_slice.rs
```diff
@@ -219,6 +219,32 @@ pub struct FrameSliceWithMetadata<'frame, 'metadata> {
...
+        typ: &'metadata ColumnType<'metadata>,
+        v: Option<FrameSlice<'frame>>,
+    ) -> Result<Self, DeserializationError> {
+        Ok(FrameSliceWithMetadata {
+            frame_slice: v,
+            column_type: typ,
+        })
+    }
+}
+
```
wprzytula: 🤔 What's the point of implementing this trait in a separate commit from introducing the struct? IMO the whole point of this struct is to be deserializable, so this constitutes a single logical unit.

wprzytula (changes requested):

Lorak-mmk (approved):

wprzytula (approved):

wprzytula (approved):

Lorak-mmk (approved):

Lorak-mmk (approved):

wprzytula (approved):

PR #1774, opened 2026-07-07
session: Extract execution core

wprzytula:
Refs: #1549

This is another important step on the way to deduplicate request execution paths (currently split between `session.rs` and `pager.rs`).

The PR extracts the execution core from `session.rs` to the newly created `execution.rs`. To ensure that the extraction is correct and absolutely 0 logical changes happen by accident, the changes are put in very small commits, which are carefully crafted as non-controversial no-brainers. 

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -2094,14 +2094,53 @@ impl Session {
         let cluster_state = self.cluster.get_state();
         let request_plan = load_balancing::Plan::new(load_balancer, &routing_info, &cluster_state);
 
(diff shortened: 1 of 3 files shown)


scylla/src/client/session.rs
Lorak-mmk, line 2039: Not outdated, there are 2 Arc: `Arc<ExecutionProfileInner>`, `Arc<Connection>` - both with no clear reason to be there.

Lorak-mmk, line 1: Commit: "session: extract run_request_no_side_effects"

Why no commit message? 😢

Lorak-mmk (approved):

PR #1776, opened 2026-07-09
scylla-cql-core: add CqlTimestamp with now(), MIN/MAX, and arithmetic ops

QuerthDP:
`CqlTimestamp` was a bare newtype with no built-in way to get the current time, no sentinel constants, and no arithmetic - leaving callers to manipulate the raw millisecond field directly.

Add `now()`, `MIN`/`MAX`, and `Add`/`Sub`/`AddAssign`/`SubAssign` against `std::time::Duration` (panicking on overflow, matching `std`/`chrono` conventions) so the type covers the operations most code needs without reaching for an external date/time library.

Also add `checked_duration_since`, mirroring `std::time::Instant`, to compute the duration between two timestamps without the operand-order footgun of a `Sub<CqlTimestamp>` impl.

Refs: https://github.com/scylladb/scylla-cdc-rust/pull/150
Fixes: VECTOR-747

## Pre-review checklist

### DIFF
--- a/scylla-cql-core/src/value.rs
+++ b/scylla-cql-core/src/value.rs
@@ -2,6 +2,7 @@
 //! as well as conversion between them and other types.
 
 use std::net::IpAddr;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use thiserror::Error;
 use uuid::Uuid;
@@ -759,6 +760,61 @@ impl TryInto<chrono_04::NaiveDate> for CqlDate {
 }
 
 impl CqlTimestamp {
+    /// The earliest representable `CqlTimestamp` (milliseconds = `i64::MIN`).
+    pub const MIN: CqlTimestamp = CqlTimestamp(i64::MIN);
+
+    /// The latest representable `CqlTimestamp` (milliseconds = `i64::MAX`).
+    pub const MAX: CqlTimestamp = CqlTimestamp(i64::MAX);
+
+    /// Returns the current wall-clock time as a `CqlTimestamp`.
+    ///
+    /// The returned value holds the number of milliseconds since the Unix epoch
+    /// (1970-01-01 00:00:00 UTC), matching the CQL `timestamp` type semantics.
+    ///
+    /// This is implemented in terms of [`std::time::SystemTime::now`]
+    /// and is therefore **not monotonic**.
+    /// Because of OS/NTP clock adjustments, a later call to `now()` can return
+    /// a smaller value than an earlier call.
+    /// Do not use `now()` to measure elapsed time;
+    /// use [`std::time::Instant`] for that purpose instead.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the system clock is so far removed from the Unix epoch
+    /// (in either direction) that it cannot be represented as `i64` milliseconds.
+    pub fn now() -> Self {
+        match SystemTime::now().duration_since(UNIX_EPOCH) {
+            Ok(d) => Self(
+                i64::try_from(d.as_millis())
+                    .expect("system clock is too far past the Unix epoch to fit in a CqlTimestamp"),
+            ),
+            Err(e) => {
+                let ms = i64::try_from(e.duration().as_millis()).expect(
+                    "system clock is too far before the Unix epoch to fit in a CqlTimestamp",
+                );
+                // `ms` is the result of converting a `u128` (always non-negative) into `i64`,
+                // so it is in the range [0, i64::MAX]. Negating it therefore cannot overflow.
+                Self(-ms)
+            }
+        }
+    }
+
+    /// Returns the amount of time elapsed from `earlier` to `self`.
+    ///
+    /// Mirrors [`std::time::Instant::checked_duration_since`].
+    /// Returns `None` if `earlier` is actually later than `self`,
+    /// instead of silently saturating at zero, since it is easy to swap
+    /// the operands by mistake.
+    pub fn checked_duration_since(self, earlier: CqlTimestamp) -> Option<std::time::Duration> {
+        if self.0 < earlier.0 {
+            return None;
+        }
+        // The ordering check above guarantees `self.0 >= earlier.0`,
+        // so `abs_diff` gives the correct non-negative `u64` difference even
+        // when it wouldn't fit in `i64` (e.g. MAX - MIN).
+        Some(std::time::Duration::from_millis(self.0.abs_diff(earlier.0)))
+    }
+
     fn try_to_chrono_04_datetime_utc(
         &self,
     ) -> Result<chrono_04::DateTime<chrono_04::Utc>, ValueOverflow> {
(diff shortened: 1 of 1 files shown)


scylla-cql-core/src/value.rs
```diff
@@ -759,6 +759,24 @@ impl TryInto<chrono_04::NaiveDate> for CqlDate {
...
+    /// The returned value holds the number of milliseconds since the Unix epoch
+    /// (1970-01-01 00:00:00 UTC), matching the CQL `timestamp` type semantics.
+    pub fn now() -> Self {
+        use std::time::{SystemTime, UNIX_EPOCH};
+        let d = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default();
+        Self(i64::try_from(d.as_millis()).unwrap_or(i64::MAX))
+    }
+
```
Lorak-mmk, line 801: It is imo important to describe in doc comment that this uses SystemTime internally, and thus is not monotonic (later call to `now()` can return smaller value). We already had driver panics related to time monotonicty that I had to do an emergency fix and release for, so let's be careful about this topic.

scylla-cql-core/src/value.rs
```diff
@@ -770,6 +788,33 @@ impl CqlTimestamp {
...
+}
+
+impl std::ops::Sub<CqlTimestamp> for CqlTimestamp {
+    type Output = std::time::Duration;
+
+    fn sub(self, rhs: CqlTimestamp) -> std::time::Duration {
+        let diff_ms = self.0.saturating_sub(rhs.0).max(0) as u64;
+        std::time::Duration::from_millis(diff_ms)
+    }
+}
```
wprzytula, line 873: 💭 I'm afraid it may be unexpected and confusing that we perform saturating arithmetics as the default ops (+, -). To compare, default arithmetic operators perform:
- in debug mode: checked arithmetics with panic,
- in release mode: wrapping arithmetics.

scylla-cql-core/src/value.rs
```diff
@@ -770,6 +788,33 @@ impl CqlTimestamp {
...
+
+impl std::ops::Sub<std::time::Duration> for CqlTimestamp {
+    type Output = CqlTimestamp;
+
+    fn sub(self, rhs: std::time::Duration) -> CqlTimestamp {
+        let ms = i64::try_from(rhs.as_millis()).unwrap_or(i64::MAX);
+        CqlTimestamp(self.0.saturating_sub(ms))
+    }
+}
```
Lorak-mmk, line 864: I don't think `Add` / `Sub` should use saturating semantics. Both std `SystemTime` and chrono `DateTime` use panicking semantics, so it is imo unexpected behavior to do something different. We can of course have methods like `saturating_add` in addition to `Add` trait impl.

Lorak-mmk, line 864: Both chrono and std implement both Add/Sub and AddAssign/SubAssign. Do you know if there is a reason for that? Should we do the same?

scylla-cql-core/src/value.rs
```diff
@@ -770,6 +788,33 @@ impl CqlTimestamp {
...
+
+impl std::ops::Sub<CqlTimestamp> for CqlTimestamp {
+    type Output = std::time::Duration;
+
+    fn sub(self, rhs: CqlTimestamp) -> std::time::Duration {
+        let diff_ms = self.0.saturating_sub(rhs.0).max(0) as u64;
+        std::time::Duration::from_millis(diff_ms)
```
Lorak-mmk: Ditto about saturating.
std uses `duration_since` method, without Sub trait, and this method returns Result.
chrono has `Sub`, but it is neither saturating nor panicking - ther time delta is signed, so the result is always well defined.

With sub this problem is even more important than in add, because it is easy to use wrong order of operands. Imo the std approach (method returning Result) makes most sense here.

Lorak-mmk (changes requested):

QuerthDP:
Changelog:
- add `checked_duration_since` mirroring `std::time::Instant::checked_duration_since`
- add requested docstrings
- change saturating logic to panics
- adjust tests to match the panics logic

scylla-cql-core/src/value.rs
QuerthDP: Added a comment about it

scylla-cql-core/src/value.rs
QuerthDP: Changed. Add/Sub now panics on overflow

scylla-cql-core/src/value.rs
QuerthDP: Switched to panicking semantics as suggested.

I think there's no need for `saturating_add` etc.

QuerthDP: No particular reason I think. I've added them to match their behavior

scylla-cql-core/src/value.rs
QuerthDP: Replaced `Sub<CqlTimestamp>` with `checked_duration_since(earlier)`

scylla-cql-core/src/value.rs
```diff
@@ -759,6 +760,47 @@ impl TryInto<chrono_04::NaiveDate> for CqlDate {
...
+    /// Because of OS/NTP clock adjustments, a later call to `now()` can return
+    /// a smaller value than an earlier call.
+    /// Do not use `now()` to measure elapsed time;
+    /// use [`std::time::Instant`] for that purpose instead.
+    pub fn now() -> Self {
+        let d = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default();
+        Self(i64::try_from(d.as_millis()).unwrap_or(i64::MAX))
+    }
```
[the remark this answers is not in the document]
QuerthDP: I don't know if it's worth doing. It's tradeoff between simplicity / pedantic correctness.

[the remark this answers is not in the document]
Lorak-mmk: Let's use the coderabbit version. One thing I'm not sure about is those calls like `.unwrap_or(i64::MAX)`. If we are in such pathological environment, it may be better to just panic.

scylla-cql-core/src/value.rs
```diff
@@ -759,6 +760,47 @@ impl TryInto<chrono_04::NaiveDate> for CqlDate {
...
+    pub fn checked_duration_since(self, earlier: CqlTimestamp) -> Option<std::time::Duration> {
+        if self.0 < earlier.0 {
+            return None;
+        }
+        // `wrapping_sub` avoids an `i64` overflow panic when self and earlier
+        // are at opposite ends of the range (e.g. MAX - MIN);
+        // the checked ordering above guarantees the true difference fits in a `u64`.
+        let diff_ms = self.0.wrapping_sub(earlier.0);
+        Some(std::time::Duration::from_millis(diff_ms as u64))
+    }
```
wprzytula, line 816: 🤔 Why an explicit check + `wrapping_sub` instead of leveraging the already present standard `checked_sub`?

scylla-cql-core/src/value.rs
```diff
@@ -770,6 +812,58 @@ impl CqlTimestamp {
...
+    fn add(self, rhs: std::time::Duration) -> CqlTimestamp {
+        let ms =
+            i64::try_from(rhs.as_millis()).expect("duration is too large to add to a CqlTimestamp");
+        CqlTimestamp(
+            self.0
+                .checked_add(ms)
+                .expect("overflow when adding Duration to CqlTimestamp"),
+        )
+    }
+}
```
wprzytula, line 840: 💭 The usual approach in Rust wrt overflows is:
- in debug, catch and panic.
- in release, don't pay the cost of checks and wrap.

Here, you panic in both release modes, but also incur the cost of checking for the overflow in both modes.

1. Is it intended?
2. Is it a correct thing to do?

QuerthDP:
Changelog:
- fixed `now()` to correctly return negative timestamps when the system clock is set before epoch
- fixed Add/Sub incorrectly panicking on some in-range results

scylla-cql-core/src/value.rs
Lorak-mmk: > 💭 The usual approach in Rust wrt overflows is:

This is for normal numbers.

Both SystemTime and chrono use panics in both modes.

scylla-cql-core/src/value.rs
QuerthDP: `checked_sub` conflates two different things:
1. `self < earlier` -> genuinely invalid order, should be `None`.
2. `self >= earlier`, but the difference doesn't fit in `i64` -> e.g. `MAX - MIN = u64::MAX` ms, a perfectly valid `Duration`, but it overflows `i64`.

`i64::checked_sub` returns `None` for both cases. Using it directly would make `checked_duration_since(MAX, MIN)` wrongly return `None` instead of `Some(Duration::from_millis(u64::MAX))`.

wprzytula: Maybe it's me having problem with maths, but I don't understand how `wrapping_sub` solves the case you described. `u64::MAX` would rather be an effect of saturation, not wrapping.

scylla-cql-core/src/value.rs
```diff
@@ -759,6 +760,60 @@ impl TryInto<chrono_04::NaiveDate> for CqlDate {
...
+            Ok(d) => Self(
+                i64::try_from(d.as_millis())
+                    .expect("system clock is too far past the Unix epoch to fit in a CqlTimestamp"),
+            ),
+            Err(e) => {
+                let ms = i64::try_from(e.duration().as_millis()).expect(
+                    "system clock is too far before the Unix epoch to fit in a CqlTimestamp",
+                );
+                Self(-ms)
+            }
```
Lorak-mmk, line 798: Worth noting that the innocent looking `-ms` is in general a panicking operation (it will panic for `ms == i64::MIN`). This is not possible here because `e.duration().as_millis()` is a positive number.

scylla-cql-core/src/value.rs
QuerthDP: `MAX - MIN = -1`, when wrapping it becomes `u64::MAX`.

Lorak-mmk: As discussed offline, lets use `if` + `abs_diff`.

Lorak-mmk (changes requested):

QuerthDP:
Changelog:
- changed `wrapping_sub` unclear logic to `abs_diff`

scylla-cql-core/src/value.rs
QuerthDP: Added a comment

QuerthDP:
Changelog:
- add safety comment

scylla-cql-core/src/value.rs
wprzytula: > `MAX - MIN = -1`, when wrapping it becomes `u64::MAX`.

Shouldn't this be `MAX + MIN`?

Lorak-mmk (approved):

scylla-cql-core/src/value.rs
QuerthDP: Both `MAX` and `MIN` are respectively `i64::MAX` and `i64::MIN`, which means both are absolutely about a half of `u64::MAX`.

wprzytula (approved):

PR #1778, opened 2026-07-09
Fix tablets MV bug

### LINKED ISSUE #1742: Tablet-awareness doesn't work with MV
When performing tablet info maintenance, we remove info for tables not present in `tables` map of relevant `Keyspace` object.

Turns out materialized views (so I guess indexes as well, but I didn't check) are not present in this table, so we will prune their info on each metadata refresh.

We need to check the `views` map as well.

Lorak-mmk:
This PR fixes the bug with MV tablet metadata, and adjusts tablets integration test accordingly. I don't think there is much more to say about it.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1742

## Pre-review checklist

### DIFF
--- a/scylla/src/routing/locator/tablets.rs
+++ b/scylla/src/routing/locator/tablets.rs
@@ -612,31 +612,39 @@ impl TabletsInfo {
                 return false;
             }
 
+            // Materialized views are tablet-based too, but they live in a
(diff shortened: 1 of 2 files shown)


wprzytula:
@coderabbitai review

wprzytula (approved):

PR #1780, opened 2026-07-10
Introduce benchmarks to CI

### LINKED ISSUE #1532: Run allocations benchmark in CI
We have a benchmark (disguised as an example) that measures how many allocations per request are typically done. I think we should automatically run it in CI, and post results in a comment.
It is of course not a good way to measure driver performance in general, but it should prevent a whole class of possible regressions.

Action items:
 - Read the benchmark code, see if some improvements should be done to it.
 - Add a CI pipeline that:
     - Runs the benchmark on PR head, and PR base
     - Compares the result
     - Posts the result in a comment, with a visible warning if there is a regression

We already have the semver-checks bot posting a comment. To avoid spamming the PRs, we should figure out a way to post both things in a single comment (and update them independently when PR is updated).

Lorak-mmk:
This PR introduces benchmarks to CI.

The benchmarks use the `gungraun` framework (previously called `iai-callgrind`) to run benchmarks. It is a wrapper for valgrind and its tools.
The benchmarks measure instruction count (and other statistics like cache misses) and allocations (max memory usage, total allocation count etc). They do not measure wall clock time. This is intentional - `gungraun` doesn't support this because the goal is to have reliable consistent results. Allocations and instruction counts are inherently much more stable than execution time.
Benchmarks are implemented in a way that makes framework-specific code minimal: it should not be difficult to later add support to run the same benchmark cases with criterion as well if we wish.

The goal of this PR is NOT introducing a proper set of scenarios - we can do that later. I only introduced a few basic ones that should allow us to catch the most important gains / regressions.
In the future we could add much more cases: migrate existing ones from scylla / scylla-cql, measure tablet feedback handling, session creation, retries etc.
If regression is found, a label will be automatically added to a PR. Benchmark results can be viewed in a job summary. See the example here: https://github.com/Lorak-mmk/scylla-rust-driver/actions/runs/29119057517?pr=15

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1532
Fixes: DRIVER-759

## Pre-review checklist


benchmarks/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+                // Reads (not writes) are used so that the warmup does not insert
+                // rows, which would otherwise inflate the auto-paged `SELECT`
+                // scenario's full-table scan; tablet learning is driven by where
+                // a request is routed, so a read of a (possibly absent) row
+                // teaches the driver the tablet just as a write would.
+                futures::stream::iter(0..TABLET_WARMUP_REQUESTS as i32).for_each_concurrent(64, async |i| {
+                    session
+                        .execute_unpaged(&prepared_select, (i, 2 * i))
+                        .await.unwrap();
+                }).await;
```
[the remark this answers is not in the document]
Lorak-mmk: > and the more widely-supported |i| async { ... }

How tf is it more widely supported in the context of our driver?

> hardcodes 64

This is just a setup, not the benchmark itself, it is fine.

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub fn run_inserts(&self, n: usize) {
+        self.runtime.block_on(async {
+            for i in 0..n as i32 {
+                let result = self.session
+                    .execute_unpaged(&self.prepared_insert, (i, 2 * i))
+                    .await
+                    .unwrap();
+                black_box(result);
+            }
+        })
```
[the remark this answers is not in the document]
Lorak-mmk: No

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub fn run_unpaged_selects(&self, n: usize) {
+        self.runtime.block_on(async {
+            for i in 0..n as i32 {
+                let result = self.session
+                    .execute_unpaged(&self.prepared_select, (i, 2 * i))
+                    .await
+                    .unwrap();
+                black_box(result);
+            }
+        })
```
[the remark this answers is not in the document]
Lorak-mmk: No.

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub fn run_batches(&self, n: usize) {
+        self.runtime.block_on(async {
+            for _ in 0..n as i32 {
+                let result = self.session
+                    .batch(&self.batch, &self.batch_values)
+                    .await
+                    .unwrap();
+                black_box(result);
+            }
+        })
```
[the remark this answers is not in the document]
Lorak-mmk: No

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+                // a request is routed, so a read of a (possibly absent) row
+                // teaches the driver the tablet just as a write would.
+                futures::stream::iter(0..TABLET_WARMUP_REQUESTS as i32)
+                    .for_each_concurrent(64, async |i| {
+                        session
+                            .execute_unpaged(&prepared_select, (i, 2 * i))
+                            .await
+                            .unwrap();
+                    })
+                    .await;
```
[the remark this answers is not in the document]
Lorak-mmk: I don't see how panic makes anything harder to diagnose. On the contrary, it will provide a backtrace.

benchmarks/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Micro-benchmarks that measure, for each request scenario:
+
+- **instructions** executed (via Callgrind),
+- **allocations** — the number of heap blocks allocated (via DHAT), and
+- **peak memory** — the whole-process maximum live heap, "at t-gmax" (via DHAT).
+
+Because the metrics are collected under Valgrind rather than by wall-clock
+timing, they are deterministic enough to compare a pull request against its base
+and reliably surface instruction-count and allocation changes.
```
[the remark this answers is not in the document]
Lorak-mmk: Cluster is not running in the same process, it is not controlled by valgrind here.
Our ClusterWorker can influence the results, but the benchmarks run far too quickly for it to happen.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,116 @@
...
+
+    - name: Benchmark the pull request
+      run: |
+        git checkout "$MERGE_SHA"
+        version=$(cargo metadata --format-version=1 |\
+          jq -r '.packages[] | select(.name == "gungraun").version'
+        )
+        cargo binstall --no-confirm gungraun-runner --version "$version"
+        cargo bench -p benchmarks --bench requests -- \
+          --baseline=base --output-format=json > results.jsonl
```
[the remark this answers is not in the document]
Lorak-mmk: This makes 0 sense, this job is on pull_request_target

benchmarks/benches/requests.rs
```diff
@@ -0,0 +1,84 @@
...
+        // default Callgrind instruction counting.
+        .tool(Dhat::default())
+        // trace-children: Driver doesn't spawn processes, so this won't
+        // influence the benchmarks, but the benchmark may want to spawn
+        // a process (e.g. a proxy with error injection) which we wouldn't
+        // want to trace.
+        // num-callers: Allocations are attributed using call stack. If this number
+        // is slow, then the benchmarking function won't appear there, and the allocation
+        // won't be attributed to the benchmark.
+        .valgrind_args(["--trace-children=no", "--num-callers=500"])
```
wprzytula: 🤔 number is slow?

wprzytula: Also, I don't see the connection between `num-callers` being low and the benchmarking function not appearing (where?).

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,205 @@
+//! Driver micro-benchmark scenarios.
```
wprzytula: 💭 I found such a definition:

> A micro-benchmark is a small test used to measure the speed or memory use of a very tiny piece of code (like a single function or calculation).

Statement execution is big enough to exceed the size of _micro_-benchmarks IMO.

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+impl Default for ScenarioConfig {
+    fn default() -> Self {
+        Self {
+            batch_size: 8,
+            paged_rows: 200,
+            page_size: 20,
+        }
+    }
+}
```
wprzytula, line 67: ♻️ This config contains params only used for scenarios that are not yet introduced in this commit.

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// statements used by the scenarios, pre-populates the data needed by the
+    /// auto-paged `SELECT` scenario and warms up tablet routing.
+    pub fn new(node: &str, config: ScenarioConfig) -> Result<Self> {
+        Self::build(node, config)
+    }
+
+    fn build(node: &str, config: ScenarioConfig) -> Result<Self> {
+        let runtime = tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()?;
```
wprzytula, line 97: 💭 Limiting the benchmarks to current-thread executor seems weird. Why not observe the real-world case of multi-thread executor?

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+                futures::stream::iter(0..TABLET_WARMUP_REQUESTS as i32)
+                    .for_each_concurrent(64, async |i| {
+                        session
+                            .execute_unpaged(&prepared_select, (i, 2 * i))
+                            .await
+                            .unwrap();
+                    })
+                    .await;
+
+                session.refresh_metadata().await?;
```
wprzytula, line 197: 🔧 All other sites use error propagation, but here you resort to `unwrap()`. Why?

benchmarks/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+            for i in 0..n as i32 {
+                let result = self
+                    .session
+                    .execute_unpaged(&self.prepared_insert, (i, 2 * i))
+                    .await
+                    .unwrap();
+                black_box(result);
+            }
+        })
+    }
```
wprzytula, line 232: ❓ Why `unwrap()`?

benchmarks/benches/requests.rs
```diff
@@ -39,6 +39,24 @@ fn setup_default(n: usize) -> State {
...
+            &node_address(),
+            ScenarioConfig {
+                batch_size,
+                ..ScenarioConfig::default()
+            },
+        )
+        .unwrap(),
+        n,
+    )
+}
```
wprzytula, line 58: Again, why `unwrap()`?

Makefile
```diff
@@ -83,6 +83,18 @@ dockerized-test: up
...
+	# Run the driver benchmarks and store the results as the baseline named
+	# "base" to compare against later (e.g. before applying your changes).
+	cargo bench -p benchmarks --bench requests -- --save-baseline=base
+
+.PHONY: bench
+bench: up
+	# Run the driver benchmarks and compare against the "base" baseline saved
+	# by `make bench-baseline` (without overwriting it).
+	cargo bench -p benchmarks --bench requests -- --baseline=base
+
```
wprzytula, line 97: ♻️ This shouldn't come in a later commit than the documentation that mentions this.

benchmarks/benches/requests.rs
Lorak-mmk: I reworded the comment.

benchmarks/src/lib.rs
Lorak-mmk: The main goal of those numbers is to be stable. Multi-threaded executor introduces a lot of instability, and should not matter that much for allocations per request or instructions per request.

benchmarks/src/lib.rs
Lorak-mmk: I don't see much harm from `unwrap()` in benchmarks, and `for_each_concurrent` is a handy tool.

benchmarks/benches/requests.rs
Lorak-mmk: This is a setup function for benchmark. Its return type must match the argument type for benchmark function.

benchmarks/src/lib.rs
Lorak-mmk: I'd like to either unwrap or not in all test functions. In some I unwrap because of `for_each_concurrent`, so I see no reason to behave differently here.

Lorak-mmk:
Addressed review, reworked bench setup a bit to have proper prefilled rows for unpaged select as well.

Lorak-mmk:
Hmm... I think I broke something, the testing PR now reports weird regressions / improvements.

Lorak-mmk:
False alarm, it looks to be a one-time problem.

benchmarks/benches/requests.rs
wprzytula: Will this provide enough context for debugging in case this panics?

wprzytula (approved):

benchmarks/benches/requests.rs
Lorak-mmk: Panic provides backtrace and prints the error, I think it is typically enough - we use it in all the tests.

PR #1783, opened 2026-07-14
pager: Migrate `PagerWorker` to unified execution core

wprzytula:
Refs: #1549

## Summary

This is the most important step on the way to #1549. The unified execution core, extracted from `session.rs` in #1774, is now used as the core of `PagerWorker`, which formerly implemented its own logic on top of `Connection`.

## Commits 

1. Commits up to _pager: `query_first_page` accepts `RequestSpan`_ are IMO a noncontroversial prelude that could be extracted as another PR if reviewers prefer to.
2. Commits from _pager: store Arc<Metrics>_ to _pager: move sending next page out of process_next_page()_ prepare for the main logic replacement step, reducing the clutter of the critical commit.
3. The move: _pager: use unified execution core for `Session::*_iter` APIs_. I tried as much as I could to make it the least visually intrusive way possible. Let me know if it's good enough.
4. Two more commits are refactors: naming, docs, imports.
5. Two commits introduce speculative execution into the pager and update sphinx docs accordingly.

## Worthy Mentions

Opus says:

1. _Two semantic notes worth a mention in the PR:_
    - _Connection is re-resolved via connection_for_shard(shard) on each same-target retry (execution.rs:291), whereas the old pager captured the connection once "to preserve paging cache on the shard." Since paging state is token-based and the shard is fixed, this is fine and matches Session, but it contradicts the old comment's rationale — good to acknowledge._
    - _A valid-but-unexpected non-Rows response on page 2+ now bypasses the retry policy and is counted as a success by metrics + LBP on_request_success, because into_non_error_query_response succeeds and process_next_page returns RequestError after the fiber already completed (pager.rs:463-468). The old code fed it to decide_should_retry and on_request_failure. This is pathological (server never returns such a continuation) and is inherent to the unified success/shape split, so low impact — just flagging the change._
2. _Minor extra allocations vs old: retry_policy.new_session() per page (was reset()), Coordinator::new per attempt (was per node). Both negligible next to a network round-trip._ -> The reason why we can't just keep `reset()` per page is speculative execution: each page can fire arbitrary number of speculative fibers, each of which needs an independent `RetrySession`. We can optimise the non-speculative case though, if reviewers decide so or benchmarks show its importance.

## Further Work

Left for a follow-up: migrating `SingleConnectionPagerWorker` (one used for `Connection::execute_iter()`).

## Pre-review checklist


scylla/src/client/pager.rs
```diff
@@ -160,10 +160,10 @@ enum ShouldFetchMorePages {
 // QueryPager receives them through a channel
 struct PagerWorker {
     load_balancing_policy: Arc<dyn LoadBalancingPolicy>,
-    query_is_idempotent: bool,
-    query_consistency: Consistency,
     retry_session: Box<dyn RetrySession>,
     timeouter: Option<PageQueryTimeouter>,
+    is_idempotent: bool,
+    consistency: Consistency,
     metrics: Metrics,
```
Lorak-mmk: Commit: "pager: drop redundant `query_` prefix from fields"

Is `query_` in `query_consistency` redundant? It suggests that this is the initial consistency, not the mutable possibly downgraded one.

scylla/src/client/pager.rs
```diff
@@ -160,7 +160,7 @@ enum ShouldFetchMorePages {
 // QueryPager receives them through a channel
 struct PagerWorker {
     load_balancing_policy: Arc<dyn LoadBalancingPolicy>,
-    retry_session: Box<dyn RetrySession>,
+    retry_policy: Arc<dyn RetryPolicy>,
     timeouter: Option<PageQueryTimeouter>,
     is_idempotent: bool,
     consistency: Consistency,
```
Lorak-mmk: Commit: "pager: store RetryPolicy, not RetrySession"

I would truly appreciate a bit more explanation in the commit message.
RetrySession seems to me like something that should be enough, due to its resetability. Does the new execution core require RetryPolicy? If so, can we change it to use RetrySession?

Lorak-mmk: Hm..

Now that I think of it, we did discuss the possibility of avoiding RetrySession alloc unless retry is required. To achieve that, execution core does need the policy. 

In any case, commit message explanation would be appreciated. I have no idea if I correctly identified the cause for this change.

scylla/src/client/pager.rs
Lorak-mmk, line 1: Commit: "pager: make paging_state flow as an argument"

Again, its not obvious to me why this is necessary. Please explain in the commit message.

scylla/src/client/pager.rs
```diff
@@ -407,229 +218,134 @@ impl PagerWorker {
...
+
+        // Coordinator stability: try the previous page's connection first.
+        let stability_target = stable_coordinator.map(|coordinator| {
+            // `Coordinator` gets its shard from `Connection`. If `Connection` believes it has `Shard` `None`,
+            // then it means that the connection pool to that connection is not sharded.
+            // In such case, shard is ignored in `connection_for_shard()`.
+            // We must provide a shard to `connection_for_shard()` anyway.
+            // Therefore, let's pass an arbitrarily big shard to catch regressions in this logic earlier.
+            (coordinator.node(), coordinator.shard().unwrap_or(2137))
+        });
```
Lorak-mmk: Note: One scenario I didn't think of before is node getting a quick restart between pages, and enabling driver sharding during this restart. In such case, the coordinator has no shard, but node connection pool is sharded. Is it a problem?

scylla/src/client/pager.rs
```diff
@@ -407,229 +218,134 @@ impl PagerWorker {
...
-        let query_response = match self.timeouter {
-            Some(ref timeouter) => {
-                match tokio::time::timeout_at(timeouter.deadline(), runner).await {
-                    Ok(res) => res,
-                    Err(_) /* tokio::time::error::Elapsed */ => {
-                        self.metrics.inc_request_timeouts();
-                        return Err(RequestTimeoutError(timeouter.timeout_duration()));
-                    }
-                }
-            }
```
Lorak-mmk, line 282: Q: `RequestExecutionParams` stores `load_balancing_policy`. Why? `run_request_no_side_effects` already gets a plan as an argument.

Lorak-mmk (changes requested):

scylla/src/client/pager.rs
wprzytula: I don't see how `query_consistency` suggests that it's initial better than `consistency`. Perhaps you mistake _query_ with _statement_.

wprzytula: In the Pager's glossary, _query_ is the whole execution of all pages.

Lorak-mmk: Ok, then how about naming it `statement_consistency`? Does it make sense?

scylla/src/client/pager.rs
wprzytula: Idk what I should explain in the commit message more than it's already been there, but OK. Now it's expanded.

scylla/src/client/pager.rs
wprzytula: It's not necessarily about the statement. Most of the time it's set on execution profile.

Lorak-mmk: I think the name would be ok anyway, but I won't argue. Resolving.

scylla/src/client/pager.rs
wprzytula: This was actually unnecessary.

scylla/src/client/pager.rs
wprzytula: Shouldn't be, because 2137 is big enough that no node will have that many shards.

Lorak-mmk: We'll surely regret those words in a few years.

scylla/src/client/pager.rs
wprzytula: Because it needs to call `LBP.on_request_success` and `LBP.on_request_failure`.

wprzytula:
Addressed comments.

scylla/src/client/pager.rs
Lorak-mmk: Ah, that makes sense. Thanks.

scylla/src/client/pager.rs
```diff
@@ -198,496 +159,243 @@ impl PagerWorker {
...
-        let runner = async {
-            page_query(connection.clone(), consistency, self.paging_state.clone())
+        // Adapt the paging-aware `page_query` to the two-argument closure
+        // expected by the execution core, injecting the current paging state
+        // and normalizing error responses.
+        let run_request_once = |connection: Arc<Connection>, consistency: Consistency| async move {
+            page_query(connection, consistency, self.paging_state.clone())
                 .await
                 .and_then(QueryResponse::into_non_error_query_response)
         };
```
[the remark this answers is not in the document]
wprzytula: Hmm, @Lorak-mmk Do you think it's serious enough to even bother? It's only about bookkeeping, and only if we get an unexpected response as an 2+ page.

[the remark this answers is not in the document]
Lorak-mmk: The scenario may not be relevant in practice, but perhaps it shows some kind of weakness in our new abstractions? Perhaps execution really should have 2 pub(crate) methods, one for non-iter and first page (handling side effects, trating non-rows as success), and other without side effects (accepts only rows, bookkeeping other responses as errors), both calling one private function underneath?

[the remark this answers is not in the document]
wprzytula: I'm considering your idea.

One doubt: should we only use the new method for iter's first page, or maybe also for `*_single_page` with non-empty `PagingState`?
Conceptually, neither should allow non-Rows responses.
Pragmatically, this is only needed for iter API due to lack of Session available there on 2+ pages to handle side effects.
That said, there really won't ever be a 2+ page in iter APIs that comes with side effects (SchemaChange or SetKeyspace), because the server will never send it as a non-first page.
OTOH, one could put non-zero `PagingState` (crafted or extracted from another request) to call `query_single_page("USE ks")`; then server would likely respond with `SetKeyspace` and ignore the paging state (or error out, but let's keep considering the success case). Then we would error the response out, saying that only `Rows` is accepted as a non-first page, basing the check on the `PagingState` not being `PagingState::start()`.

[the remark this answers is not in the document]
wprzytula: Anyway, I can open a separate issue about improving this. I think it's not a priority. How often does it happen that the 2+ page of the same query is non-Rows? It's always a server bug.

scylla/src/client/pager.rs
wprzytula: I doubt it. It's still a rare combined case:
- **quick** restart of a node,
- the node must have had sharding turned off,
- the node must now have sharding turned on,
- the node must have at least 2138 shards.

Likelihood of such event is extremely low. I don't want to spend too much time on such unlikely cases. We have more important things to do.

scylla/src/client/pager.rs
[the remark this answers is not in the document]
Lorak-mmk: Makes sense.

scylla/src/client/pager.rs
Lorak-mmk: Ok

Lorak-mmk: The consequence is just misrouting anyway, nothing serious.

Lorak-mmk (approved):

scylla/src/client/pager.rs
wprzytula: I could also use `u16::MAX` instead of `2137`. Our protocol implementation guarantees that there is at most `u16::MAX` shards, so `[0, u16::MAX)` - `u16::MAX` shard cannot exist.

PR #1786, opened 2026-07-15
pager: deduplicate `QueryPager`/`PagingExecutor` creation

wprzytula:
Ref: #1549

This deduplicates two `QueryPager`'s constructors: `new_for_query` and `new_for_prepared_statement`. Shared logic, namely resolving various execution config params, is now in `PagingExecutor::new()`.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -548,15 +548,6 @@ where
     }
 }
 
-pub(crate) struct PreparedPagerConfig {
-    pub(crate) prepared: PreparedStatement,
-    pub(crate) values: SerializedValues,
-    pub(crate) execution_profile: Arc<ExecutionProfileInner>,
-    pub(crate) cluster_state: Arc<ClusterState>,
-    pub(crate) metrics: Arc<Metrics>,
-    pub(crate) location_preference: Arc<NodeLocationPreference>,
-}
-
 /// An intermediate object that allows to construct a stream over a query
 /// that is asynchronously paged in the background.
 ///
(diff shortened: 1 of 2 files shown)


scylla/src/client/pager.rs
```diff
@@ -685,46 +725,17 @@ If you are using this API, you are probably doing something wrong."
...
             .access();
 
-        let consistency = statement
-            .config
-            .consistency
-            .unwrap_or(execution_profile.consistency);
         let serial_consistency = statement
             .config
             .serial_consistency
             .unwrap_or(execution_profile.serial_consistency);
```
Lorak-mmk: Commit: "pager: dedup PagingExecutor creation logic"

Weird that consistency is part of PagingExecutor, but serial_consistency is not, and so can't be moved to `new`.

Lorak-mmk: Some for page size I think

Lorak-mmk: I see serial_consistency is addressed. What about page size?

wprzytula: What's the point of moving `page_size` into `PagingExecutor`? It's only accessed via `page_query` closure, which is anyway created out of `PagingExecutor`.

Lorak-mmk: I see now. Perhaps their definition can be moved close to `page_query`, so we'd have the following?
```rust
        let prepared_ref = &prepared;
        let values_ref = &values;
        let consistency = executor.consistency;
        let serial_consistency = executor.serial_consistency;
        let page_size = prepared.get_validated_page_size();
        let page_query = |connection: Arc<Connection>,
                          consistency: Consistency,
                          paging_state: PagingState| async move {
            connection
                .execute_raw_with_consistency(
                    prepared_ref,
                    values_ref,
                    consistency,
                    serial_consistency,
                    Some(page_size),
                    paging_state,
                )
                .await
        };

```

Lorak-mmk (approved):

Issue #1280, opened 2025-03-17
Smarter metadata fetching


Lorak-mmk:
This is an issue about improving the way metadata is fetched. Below I discuss our previous ideas, and the currently proposed approach.

## History

Metadata fetching is expensive (https://github.com/scylladb/scylla-rust-driver/issues/786). Driver fetches metadata every minute (by default), and such fetch requires reading system_schema tables - containing possibly a lot of data.

We did previously consider extracting the metadata from Session (https://github.com/scylladb/scylla-rust-driver/issues/595), so that fetching it is optional.
This however proved difficult - it's really hard to design good API given all the considerations.
Even if we did design some good API, it would still be more difficult to use than the current one, and it does not even address the real issue, it only hides it. The issue is that fetching metadata costs a lot - hiding it does not make it cheaper, it only makes the cost opt-in.

We had ideas to not fetch metadata periodically:
- https://github.com/scylladb/scylla-rust-driver/issues/1008
- https://github.com/scylladb/scylla-rust-driver/issues/546

The question arises: why did we even implement periodic fetching, instead of handling the events?
I did not work in the Rust Driver back then, but the way I understand it it was to avoid request storm when a lot of schema changes are done in a short time. I think this is a very valid reason.

### Assumption

I would still like to get rid of periodically fetching full metadata. I think we can do this, but there is one assumption that must hold: after we receive schema change event X on connection C, we can query system / system_schema tables on connection C, and their contents will already include changes made by X.

I don't think it's unreasonable to expect it from the server, and @avikivity did confirm that Scylla does behave this way.

### First optimization - optional fetch

With the above assumption, we can notice that if we receive an event, fetch the metadata some time after than, and then receive no more events we know that this metadata is consistent. That means we can skip periodic fetching, until we receive another event. 
So the proposed change is:
- When receiving metadata-related event, mark some flag saying metadata is dirty.
- When fetch period (60s by default) finishes, fetch metadata only if the dirty flag is set (and then unset it).

### Second optimization - partial fetch

It is based on the first one. Notice that metadata consists of:
- topology
- schema of each keyspace (they are fully independent of each other)

We can have a flag for each of them, instead of one global flag. Then we only need to fetch the parts that have the flag set, instead of fetching whole metadata. It should not be difficult to implement.

### Third optimization - exponential backoff

Also based on the first optimization. The reason we do periodic fetches instead of reacting to events is to avoid request storm when many schema changes are made.
[...]

Issue #1279, opened 2025-03-17
Implement ControlConnectionPolicy


Lorak-mmk:
As mentioned by @wprzytula in https://github.com/scylladb/scylla-rust-driver/issues/1029 driver tries to open control connection sequentially to the contact points - meaning that it moves to next contact point only after opening the connection to the previous one failed.

This may not always be the desired behavior. 

The other possible improvement is to try opening connections to multiple contact points at the same time, as described in https://github.com/scylladb/scylla-rust-driver/issues/1111

I think that instead of adding such options directly on SessionConfig, we could do it differently, more in line with how similar stuff is implemented in other parts of the driver:

- Introduce ControlConnectionPolicy trait. It would get info about contact points (and other peers after metadata is known), and decide to which of them we should try opening the connection.
- Similarly to LoadBalancingPolicy, we would implement a single default control connection policy. The builder of this policy would expose options described above - randomization and concurrency control.

Is there any reason to not do it this way? If not, I'll close the other 2 issues in favor of this one (and keep them linked here).

roydahan:
@muzarski is his one not needed for cpp-rust-driver 1.0?
If yes, why was it pushed back to 1.x.0 milestone?

muzarski:
> [@muzarski](https://github.com/muzarski) is his one not needed for cpp-rust-driver 1.0? If yes, why was it pushed back to 1.x.0 milestone?

I just moved it out of the `1.1`. I'm not sure if it's going to be prioritized in 1.2 or pushed back to 1.3. Thus, I moved it to `1.x`.

roydahan:
Same as the other one, if required for cpp-rust-driver 1.0 it must be part of 1.1 or 1.2.

CC @dkropachev

muzarski:
With @dkropachev, we decided to drop the priority of this item in cpp-rust driver to p3 - I'm updating the label accordingly.

PR #1792, opened 2026-07-17
Make ClusterWorker the owner of ControlConnection

### LINKED ISSUE #1411: Division of responsibility in metadata code is weird and confusing
Currently we have:
`ControlConnection`: wrapper around `Connection`, created for a short time by `MetadataReader` on each metadata fetch. Has methods like `query_peers`, `query_keyspaces` - so methods that implement fetching metadata from the cluster.
`MetadataReader`: owns control connection. Is one of the 3 places responsible for reconnecting it (!!!!). For actually reading metadata it mostly defers to `ControlConnection`.
`ClusterWorker`: owns `MetadataReader`. Primary owner (= writer) of `ClusterState`. Receives messages of various types and handles them. Second of three places responsible for reconnecting control connection.
Third place responsible for reconnecting control connection is connection pool that the connection is stored in.

Problems:
- It is extremely hard to reason about control connection reconnecting, and handling in general, because there are many places taking part in this, with unclear division of responsibility.
- Responsibility and naming is confusing: `MetadataReader` is definitely a thing that should implement fetching stuff from cluster. `ControlConnection` should take care of e.g. adding server-side timeout and other config, and in the future also preparing and caching statements.

Lorak-mmk:
I'm still not 100% sure about the approach to refactoring ClusterWorker, I have a few local branches with different approaches.

What I am sure about, because every solution I ever considered has that in common, is that we need to move the ownership of CC out of MetadataReader, to be able to more precisely handle when it is broken / recreated. 

This PR does just that. Some intermediate commits are a little clunky, but the split is anyway much better than I initially hoped for.
It is possible that some details (error messages etc) are not perfect yet, but I'll leave it for next PRs. All the important behaviors we had, including handling HostFilter and initial dummy metadata, and how we respond to refresh requests, are preserved.

Ref: https://scylladb.atlassian.net/browse/DRIVER-764
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1411 (IMO)

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/control_connection.rs
+++ b/scylla/src/cluster/control_connection.rs
@@ -7,10 +7,16 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use dashmap::DashMap;
+use tokio::sync::{mpsc, oneshot};
+use tracing::warn;
 
 use crate::client::client_routes::ClientRoutesSubscriber;
 use crate::client::pager::QueryPager;
-use crate::errors::{NextPageError, NextRowError, RequestAttemptError, RequestError};
+use crate::cluster::metadata::UntranslatedEndpoint;
+use crate::errors::{
+    ConnectionError, NextPageError, NextRowError, RequestAttemptError, RequestError,
+};
+use crate::frame::response::event::EventV2 as Event;
 use crate::network::Connection;
 use crate::serialize::row::SerializeRow;
 use crate::statement::Statement;
@@ -20,27 +26,47 @@ const METADATA_QUERY_PAGE_SIZE: i32 = 1024;
 
 pub(crate) type ControlConnectionCache = DashMap<String, PreparedStatement>;
 
+pub(crate) enum ControlConnectionEvent {
+    Broken(ConnectionError),
+    ServerEvent(Event),
+    Shutdown,
+}
+
 /// The single connection used to fetch metadata and receive events from the cluster.
 pub(super) struct ControlConnection {
     conn: Arc<Connection>,
+    endpoint: UntranslatedEndpoint,
     /// The custom server-side timeout set for requests executed on the control connection.
     overridden_serverside_timeout: Option<Duration>,
     cache: Arc<ControlConnectionCache>,
     client_routes_subscriber: Option<Arc<dyn ClientRoutesSubscriber>>,
+
+    error_channel: oneshot::Receiver<ConnectionError>,
+    events_channel: mpsc::Receiver<Event>,
 }
 
 impl ControlConnection {
     pub(super) fn new(
         conn: Arc<Connection>,
+        endpoint: UntranslatedEndpoint,
         cache: Arc<ControlConnectionCache>,
         client_routes_subscriber: Option<Arc<dyn ClientRoutesSubscriber>>,
+        error_channel: oneshot::Receiver<ConnectionError>,
+        events_channel: mpsc::Receiver<Event>,
     ) -> Self {
         Self {
             conn,
+            endpoint,
             overridden_serverside_timeout: None,
             cache,
             client_routes_subscriber,
+            error_channel,
+            events_channel,
         }
+    }
+
+    pub(super) fn endpoint(&self) -> &UntranslatedEndpoint {
+        &self.endpoint
     }
 
     /// Sets the custom server-side timeout set for requests executed on the control connection.
@@ -128,6 +154,41 @@ impl ControlConnection {
             .execute_iter(prepared, serialized_values)
             .await
     }
+
+    pub(crate) async fn wait_for_event(&mut self) -> ControlConnectionEvent {
+        tokio::select! {
+            // Why only `Some`? `None` means that event channel was dropped.
+            // In current implementation (as of writing this comment)
+            // this should not be possible: events sender is stored in HostConnectionConfig,
+            // which is a field of Connection that we own. If we got `None`, then most likely
+            // two things happened:
+            //  - The implementation changed, for example by moving event sender to router.
+            //  - Connection was closed, router shutdown.
+            //  - `tokio::select!` chose this branch instead of error channel.
+            // The best thing we can imo do is ignore this `None`. `error_channel` should receive
+            // info about connection shutdown very soon.
+            Some(cql_event) = self.events_channel.recv() => {
+                ControlConnectionEvent::ServerEvent(cql_event)
+            },
+            maybe_control_connection_failed = &mut self.error_channel => {
+                let err = match maybe_control_connection_failed {
+                    Ok(err) => err,
+                    Err(_recv_error) => {
+                        // If we got here then error channel, in a Connection that we own,
+                        // was dropped without sending anything. This is definitely a bug in the driver!
+                        // We could theoretically recover by dropping a connection and creating new one,
+                        // but we would need to add an error variant to `BrokenConnectionErrorKind` that
+                        // could basically never happen. Let's panic instead.
+                        warn!(concat!("Error sender of control connection unexpectedly dropped. The only case when this ",
+                        "may happen is during runtime shutdown. If you see this and the runtime isn't shutting down, ",
+                        "this is a bug in the driver. Then please open an issue!"));
+                        return ControlConnectionEvent::Shutdown;
+                    },
+                };
+                ControlConnectionEvent::Broken(err)
+            }
+        }
+    }
 }
 
 #[cfg(test)]
@@ -148,6 +209,7 @@ mod tests {
     use crate::cluster::control_connection::ControlConnectionCache;
     use crate::cluster::metadata::UntranslatedEndpoint;
     use crate::cluster::node::ResolvedContactPoint;
+    use crate::network::HostConnectionConfig;
     use crate::network::open_connection;
     use crate::routing::ShardInfo;
     use crate::test_utils::setup_tracing;
@@ -271,12 +333,17 @@ mod tests {
             proxy_addr: SocketAddr,
             feedback_rx: &mut mpsc::UnboundedReceiver<(RequestFrame, Option<u16>)>,
         ) {
-            let (conn, _error_receiver) = open_connection(
-                &UntranslatedEndpoint::ContactPoint(ResolvedContactPoint {
-                    address: proxy_addr,
-                }),
+            let endpoint = UntranslatedEndpoint::ContactPoint(ResolvedContactPoint {
+                address: proxy_addr,
+            });
+            let (events_sender, events_receiver) = mpsc::channel(32);
+            let (conn, error_receiver) = open_connection(
+                &endpoint,
                 None,
-                &Default::default(),
+                &HostConnectionConfig {
+                    event_sender: Some((events_sender, vec![])),
+                    ..Default::default()
+                },
             )
             .await
             .unwrap();
@@ -284,8 +351,11 @@ mod tests {
             let connected_to_scylladb = conn.get_shard_info().is_some();
             let conn_with_default_timeout = ControlConnection::new(
                 Arc::new(conn),
+                endpoint,
                 Arc::new(ControlConnectionCache::new()),
                 None,
+                error_receiver,
+                events_receiver,
             );
 
             // No custom timeout set.
--- a/scylla/src/cluster/metadata/reader.rs
+++ b/scylla/src/cluster/metadata/reader.rs
@@ -109,7 +109,7 @@ impl MetadataReader {
 
         let cc_cache = Arc::new(ControlConnectionCache::new());
 
-        let control_connection_state = Self::make_control_connection(
+        let control_connection_state = Self::make_control_connection_state(
             control_connection_endpoint,
             connection_config.clone(),
             request_serverside_timeout,
@@ -281,7 +281,7 @@ impl MetadataReader {
                 peer.address()
             );
 
-            self.control_connection_state = Self::make_control_connection(
+            self.control_connection_state = Self::make_control_connection_state(
                 peer,
                 self.control_connection_config.clone(),
                 self.request_serverside_timeout,
@@ -393,7 +393,7 @@ impl MetadataReader {
                     .expect("known_peers is empty - should be impossible")
                     .clone();
 
-                self.control_connection_state = Self::make_control_connection(
+                self.control_connection_state = Self::make_control_connection_state(
                     control_connection_endpoint,
                     self.control_connection_config.clone(),
                     self.request_serverside_timeout,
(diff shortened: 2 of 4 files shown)


scylla/src/cluster/metadata/reader.rs
```diff
@@ -540,13 +480,17 @@ impl MetadataReader {
...
         Ok(updated_hosts)
     }
 }
+
+/// Error to report when there was not a single node to even attempt a control
+/// connection on (e.g. all known peers were rejected by the host filter and the
+/// initial contact points failed to resolve to any address).
+fn no_nodes_available_error() -> MetadataError {
+    MetadataError::ConnectionPoolError(ConnectionPoolError::NodeDisabledByHostFilter)
+}
```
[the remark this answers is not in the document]
Lorak-mmk: I'll take care of errors later. I don't want to introduce new variants until I'm sure about them, and this issue is very minor.

scylla/src/cluster/worker.rs
```diff
@@ -344,25 +352,33 @@ impl ClusterWorker {
...
                                 Event::ClientRoutesChange(evt) => {
-                                    let res = self.metadata_reader.fetch_client_route_updates_on_event(&evt).await;
+                                    // We received this event on the control connection, so it must be present.
+                                    let res = match control_connection.as_ref() {
+                                        Some(cc) => self.metadata_reader.fetch_client_route_updates_on_event(cc, &evt).await,
+                                        None => {
+                                            error!("BUG: Received a server event without a control connection.");
+                                            continue;
+                                        }
+                                    };
```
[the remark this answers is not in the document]
Lorak-mmk: Let's be resilient. I hope this will get restructured after further refactors.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -362,10 +362,40 @@ impl MetadataReader {
...
+                    control_connection_endpoint,
+                    self.control_connection_config.clone(),
+                    self.request_serverside_timeout,
+                    Arc::clone(&self.cc_cache),
+                    self.client_routes_subscriber.as_ref().map(Arc::clone),
+                )
+                .await;
+            }
+        }
+    }
```
wprzytula: ♻️ Let's have both conditions in one `if` to reduce indentation.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -362,10 +362,40 @@ impl MetadataReader {
...
+                self.control_connection_state = Self::make_control_connection_state(
+                    control_connection_endpoint,
+                    self.control_connection_config.clone(),
+                    self.request_serverside_timeout,
+                    Arc::clone(&self.cc_cache),
+                    self.client_routes_subscriber.as_ref().map(Arc::clone),
+                )
+                .await;
+            }
+        }
```
wprzytula: 🔧 I don't see why this `clone()` is needed on `endpoint`. It's only accessed by immutable reference, before any mutation is done to `self.control_connection_state`.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -288,21 +288,16 @@ impl MetadataReader {
...
         };
-        let endpoint = working_connection.endpoint().clone();
-
-        let res = working_connection
-            .query_metadata(
-                endpoint.address().port(),
-                &self.keyspaces_to_fetch,
-                self.fetch_schema,
-            )
-            .await;
```
wprzytula: ♻️ I'd prefer the call to `query_metadata_on_cc` to be put below, in a linear code flow, just where `query_metadata` was put.

wprzytula: The more concise `if` branches are, the easier it becomes to read the code.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -341,7 +394,6 @@ impl MetadataReader {
     }
 
     async fn handle_unaccepted_host_in_control_connection(&mut self, metadata: &Metadata) {
-        // We only care about the endpoint of a working control connection here.
         let ControlConnectionState::Working(working_connection) = &self.control_connection_state
```
wprzytula: 🤔 Disappearing comment; looks like LLM's side effects.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -534,3 +586,10 @@ impl MetadataReader {
         Ok(updated_hosts)
     }
 }
+
+/// Error to report when there was not a single node to even attempt a control
+/// connection on (e.g. all known peers were rejected by the host filter and the
+/// initial contact points failed to resolve to any address).
+fn no_nodes_available_error() -> MetadataError {
+    MetadataError::ConnectionPoolError(ConnectionPoolError::NodeDisabledByHostFilter)
+}
```
wprzytula, line 493: ⛏️ I don't see any value in extracting this as a separate function. It negatively impacts code locality, and doesn't shorten anything.

scylla/src/cluster/metadata/reader.rs
wprzytula, line 447: 🤔 Convince me that this is a bug if `control_connection` is `None` here. Couldn't it have changed to `None` in the meantime?
EDIT: I answered myself in another comment.

wprzytula: Can we prevent this scenario at compile time?

scylla/src/cluster/metadata/reader.rs
```diff
@@ -197,12 +187,16 @@ impl MetadataReader {
...
     ///
-    /// On success, updates `known_peers` and returns the working control connection
-    /// together with the fetched metadata.
+    /// On success, updates `known_peers` and returns the fetched metadata together
+    /// with a control connection to use going forward. The returned control
+    /// connection is `None` when metadata was obtained but no usable control
+    /// connection remains: on an `initial` read whose metadata query failed, dummy
+    /// metadata is returned so the session can still start, and the caller is
+    /// expected to re-establish the control connection at the repair cadence.
     async fn establish_cc_and_fetch_metadata(
```
wprzytula: > The returned control is `None` when metadata was obtained but no usable control connection remains

🔧 This misleadingly suggests that the CC was present at some point and metadata was fetched on it, but it no longer exists.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -299,39 +304,59 @@ impl MetadataReader {
...
+                                warn!(
+                                    error = ?err,
+                                    "Initial metadata read failed, proceeding with metadata \
+                                    consisting only of the initial peer list and dummy tokens. \
+                                    This might result in suboptimal performance and schema \
+                                    information not being available."
+                                );
+                                Metadata::new_dummy(&self.known_peers)
+                            }
+                        };
```
wprzytula: ♻️ Let's use `let-else`.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -299,39 +304,59 @@ impl MetadataReader {
...
             }
+
+            return Ok((Some(cc), metadata));
         }
 
-        Err(last_err)
+        match rejected_metadata {
+            Some(metadata) => Ok((None, metadata)),
+            None => Err(last_err),
+        }
```
wprzytula, line 305: ⛏️ This reads more easily for me:
```suggestion
        rejected_metadata
            .ok_or(last_err)
            .map(|metadata| (None, metadata))
```

scylla/src/cluster/worker.rs
```diff
@@ -277,18 +280,23 @@ struct UseKeyspaceRequest {
...
+    pub(crate) async fn work(mut self, initial_cc: Option<ControlConnection>) {
         use tokio::time::Instant;
 
         let control_connection_repair_duration = Duration::from_secs(1); // Attempt control connection repair every second
         let mut last_refresh_time = Instant::now();
 
+        // The control connection is now owned by the worker. `None` means it is
+        // currently broken (or was never established) and needs to be re-established,
+        // which happens during the next metadata refresh.
+        let mut control_connection: Option<ControlConnection> = initial_cc;
```
wprzytula, line 292: 📌  This means that we essentially have a ConstantTimeReconnectionPolicy for CC, with the constant time being 60s (the metadata refresh interval). As a follow-up, I suggest migrating to exponential backoff policy for CC reconnection. Note that this is different from metadata refresh interval: we can wait more time before we update our metadata periodically, but we should not stay without CC for long, because we won't receive any events.

wprzytula: Ah, I've been wrong. We already have a different repair interval (1s). It's still constant time, but at least has more tolerable interval.

Still, introducing an exponential backoff here makes sense, even before implementing a user-facing API for controlling the CC policy (#1279).

scylla/src/cluster/worker.rs
```diff
@@ -344,25 +352,33 @@ impl ClusterWorker {
...
+                            control_connection = None;
                         },
                         ControlConnectionEvent::ServerEvent(event) => {
                             debug!("Received server event: {:?}", event);
                             match event {
                                 Event::TopologyChange(_) => (), // Refresh immediately
                                 Event::ClientRoutesChange(evt) => {
-                                    let res = self.metadata_reader.fetch_client_route_updates_on_event(&evt).await;
+                                    // We received this event on the control connection, so it must be present.
+                                    let res = match control_connection.as_ref() {
```
wprzytula: 💡 Aha, this code convinces me that the `BUG: ...` print in `fetch_client_routes_update_on_event` is actually correct. So my previous request is no longer a thing.

scylla/src/cluster/worker.rs
```diff
@@ -344,25 +352,33 @@ impl ClusterWorker {
...
+                                        Some(cc) => self.metadata_reader.fetch_client_route_updates_on_event(cc, &evt).await,
+                                        None => {
+                                            error!("BUG: Received a server event without a control connection.");
+                                            continue;
+                                        }
```
wprzytula: ♻️ Please use `let-else`.

wprzytula (commented):
In general, looks good. Just nits.

scylla/src/cluster/metadata/reader.rs
Lorak-mmk: This is just temporary code no present in the final version. Do I need to do that?

scylla/src/cluster/metadata/reader.rs
Lorak-mmk: I don't see how to easily do this now. I'll look into it more in the next parts of metadata refactor.

scylla/src/cluster/worker.rs
Lorak-mmk: > Still, introducing an exponential backoff here makes sense, even before implementing a user-facing API for controlling the CC policy (https://github.com/scylladb/scylla-rust-driver/issues/1279).

I don't see much need for it, and definitely not in this PR. CC is different to pool connection:
- It is more criticial. Without it we won't fetch new topology, which may prevent creating pool connections. This means we want fast reconnect.
- There is only one, so impact on the cluster is much less severe than from pool connection. This means we can afford fast reconnect.

scylla/src/cluster/metadata/reader.rs
Lorak-mmk: It is created in two places, and such function is a good place for a comment. Another reason is that this is simply not a good error variant for all situations, and I think improving the error types will be easier later thanks to that.

scylla/src/cluster/metadata/reader.rs
Lorak-mmk: It may not be strictly true in this commit (because in this case metadata is mocked, not fetched), but it becomes true in the very next commit where we also return None CC if CC endpoint is rejected by host filter.

scylla/src/cluster/metadata/reader.rs
Lorak-mmk: Like this?
```rust
                        let Some(metadata) = rejected_metadata.take() else {
                            warn!(
                                error = ?err,
                                "Initial metadata read failed, proceeding with metadata \
                                consisting only of the initial peer list and dummy tokens. \
                                This might result in suboptimal performance and schema \
                                information not being available."
                            );
                            Metadata::new_dummy(&self.known_peers)
                        };
```

This doesn't work, the compiler expects the `else` block to evaluate to `!`.

scylla/src/cluster/metadata/reader.rs
Lorak-mmk: For me not at all. I find it quite difficult to read your version.

Lorak-mmk:
Addressed review comments - some in code, responded to others.

scylla/src/cluster/control_connection.rs
```diff
@@ -128,6 +154,41 @@ impl ControlConnection {
...
+                        // If we got here then error channel, in a Connection that we own,
+                        // was dropped without sending anything. This is definitely a bug in the driver!
+                        // We could theoretically recover by dropping a connection and creating new one,
+                        // but we would need to add an error variant to `BrokenConnectionErrorKind` that
+                        // could basically never happen. Let's panic instead.
+                        warn!(concat!("Error sender of control connection unexpectedly dropped. The only case when this ",
+                        "may happen is during runtime shutdown. If you see this and the runtime isn't shutting down, ",
+                        "this is a bug in the driver. Then please open an issue!"));
+                        return ControlConnectionEvent::Shutdown;
+                    },
```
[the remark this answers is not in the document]
Lorak-mmk: Not introduced in this PR. I'll hold on with fixing, it may go away after further refactors.

scylla/src/cluster/metadata/reader.rs
```diff
@@ -149,226 +107,222 @@ impl MetadataReader {
...
+    /// if this fails.
+    pub(crate) async fn fetch_metadata_on_cc(
+        &mut self,
+        cc: &ControlConnection,
+    ) -> Result<Metadata, MetadataError> {
+        let metadata = self.query_metadata_on_cc(cc).await?;
+        debug!("Fetched new metadata");
+        self.update_known_peers(&metadata);
+        Ok(metadata)
     }
```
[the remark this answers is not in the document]
Lorak-mmk: CC host is checked by host filter when CC is created (`try_establish_on_nodes`). I really don't think that CC being filter or not could possibly change without CC being broken first.

scylla/src/cluster/metadata/reader.rs
wprzytula: Sorry, I should have written `unwrap_or_else()`.

scylla/src/cluster/metadata/reader.rs
wprzytula: In that case, no.

scylla/src/cluster/metadata/reader.rs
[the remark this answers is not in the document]
wprzytula: IMO it's not about whether it _could_ change, but about whether _we should allow_ changing it. The way I understand when `HostFilter` should be asked is whenever a host is refreshed as part of topology refresh. So I think I agree with the coderabbit's comment.

scylla/src/cluster/metadata/reader.rs
wprzytula: Have you fixed this? I don't see a related change here: https://github.com/scylladb/scylla-rust-driver/compare/e20c46577b7928bbe3f7d6a8039fe32c0abb6d6b..6d38c1dbd7a7ecdd91f025c421ffb7ec7c2cfa84

scylla/src/cluster/metadata/reader.rs
wprzytula: Have you fixed this? I don't see a related change here: https://github.com/scylladb/scylla-rust-driver/compare/e20c46577b7928bbe3f7d6a8039fe32c0abb6d6b..6d38c1dbd7a7ecdd91f025c421ffb7ec7c2cfa84

scylla/src/cluster/metadata/reader.rs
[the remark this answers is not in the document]
Lorak-mmk: This would require a significant change in approach (`fetch_metadata_on_cc` must be able to destroy CC), and I don't see any scenario where this could happen, so why waste the effort? The scenarios we care about ( changing IP / DC / Rack) will be handled correctly here.

scylla/src/cluster/metadata/reader.rs
Lorak-mmk: I definitely remember fixing this. I can later check if its really fixed, or got lost in rebases.

scylla/src/cluster/metadata/reader.rs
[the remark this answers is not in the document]
wprzytula: OK.

Lorak-mmk:
Addressed the remaining review comment.

wprzytula (approved):

Issue #879, opened 2023-12-13
Make use of TLS session resumption


Bouncheck:
This issue is half-proposal and half to keep the progress on this somewhere.

It should be possible to optimize establishing subsequent connections by using TLS session resumption mechanisms. This should save a noticeable amount of time and resources, especially when a lot of connections is being used.

It may even be possible to run multiple sessions with the same session information concurrently. While this seems to be improper use of this feature, the tests so far have shown that its technically possible. It would be necessary to check more diligently what would be the security implications and potential problems in more complicated scenarios.

Session resumption requires support from both the server and the client side.

#### Client side

I'm not really familiar with the way scylla rust driver handles making connections, but I can provide an example rust snippet that creates 1 connection and then additional 19 that reuse session information from the first one. Rust openssl allows doing that by using [set_new_session_callback](https://docs.rs/openssl/latest/openssl/ssl/struct.SslConnectorBuilder.html#method.set_new_session_callback) to grab session information.

```rust
use std::net::TcpListener;
use std::thread;
use openssl::ssl::SslContext;
use openssl::ssl::SslMethod;
use std::path::Path;
use openssl::ssl::SslFiletype;
use openssl::ssl::SslSessionCacheMode;
use std::sync::Arc;
use std::sync::Mutex;
use std::net::TcpStream;
use openssl::ssl::Ssl;
use std::io::*;

fn main() {
    let mut ctx = SslContext::builder(SslMethod::tls()).unwrap();
    ctx.set_session_cache_mode(SslSessionCacheMode::CLIENT | SslSessionCacheMode::NO_INTERNAL);
    let session = Arc::new(Mutex::new(None));
    ctx.set_new_session_callback({
        let session = session.clone();
        move |_, sess| *session.lock().unwrap() = Some(sess)
    });
    let ctx = ctx.build();
    let port = 443;
    let ssl = Ssl::new(&ctx).unwrap();
    let stream = TcpStream::connect(("google.com", port)).unwrap();
    let mut stream = ssl.connect(stream).unwrap();
    stream.write_all(b"GET / HTTP/1.1\r\nConnection: Keep-Alive\r\n\r\n").unwrap();
    stream.read_exact(&mut [0]).unwrap();

    let mut streams = Vec::new();
    for i in 1..20 {
        let mut ssl = Ssl::new(&ctx).unwrap();
        unsafe {
            ssl.set_session(session.lock().unwrap().as_ref().unwrap())
                .unwrap();
        }
[...]

PR #1794, opened 2026-07-18
TLS tickets tests

Lorak-mmk:
TLS tickets are a mechanism that allows creating TLS connections quicker
(with a less amount of round-trips, and less CPU time).
Conceptually, it looks like this:
 - We open a full normal TLS connection
 - Server sends us some ticket / tickets
 - We use them when opening more connections to the same host
   in order to skip most of the required handshake.

The mechanism differs between TLS 1.2 and 1.3. From our perspective, the
most important difference is that TLS 1.3 tickets are single-use, while
TLS 1.2 tickets are multi-use.

Also connected to the above and relevant later: Scylla currently always
sends exactly 2 TLS 1.3 tickets. It also doesn't support the TLS
extension that allows the client to request a specific amount of
tickets. It sends 2 unconditionally.

This commit adds tests for TLS tickets support so that we have a proof
of what the current behavior is. The current behavior is that Rustls
supports tickets out of the box, and OpenSSL doesn't (so we'll need to
add support).

There is a caveat with rustls (but we'll hit the same problem with
openssl when we add support). As I mentioned, Scylla sends exactly 2 TLS
1.3 tickets per connection. When we create new connection pool, we open
a first connection, then discover the shard structure, and then open the
rest of the connections concurrently.
First connection will receive 2 tickets. Each will be used for one
subsequent connection. That means if we have a 12-shard node, and we are
missing 11 shards (after the scout connection is opened), then 2 will
use TLS tickets, while the rest will fallback to full handshake.

We could improve this on a server side, by changing the server to send
more tickets, or support RFC 9149 that allows client to request more
tickets. None of those however solve the problem in full, because Rustls
has a hard cap of 8 stored tickets per host. So if we have a node with
more than 9 shards, we'll still fallback to full handshake, no matter
how many tickets we get.

The only resilient solution for rustls is to pace the connection
opening: look at how many tickets we have, and open that many
connection, instead of opening all missing ones concurrently. I think
implementing that properly requires a greater rework of
NodeConnectionPool and is also tied to ReconnectPolicy, so let's address
it some other time.

For that reason, the rustls tests variants only require at least 2
resumptions to happen per node.

Now about the tests. They are located in a new `tls_tickets.rs` file.
This file itself should not be very hard to understand. It has some long
comments that mostly explain what I explained above.
Tests assert both that rustls works, and that openssl doesn't (with
regards to tls tickets).

To function, the tests need to inspect TLS traffic to check what version
was used, if ticket was offered and if it was accepted. For that we need
a TLS proxy, which was implemented in `tls_inspecting_proxy.rs`. This
uses `tls-parser` crate for actual parsing of the wire format, but still
needs a bit of code for everything else.

This file is fully vibe-coded. I think this is ok in this case because:
 - This is a fully self-contained piece of code.
 - It only affects tests.
 - It has perfectly concrete requirements: correctly parse traffic
between driver and Scylla, return 3 parameters (version, offer,
response).
 - If something doesn't work, we'll notice in tests.

I don't understand the TLS protocol, so those parts of the file I didn't
even read. I did verify Tokio interactions: all spawned tasks must be
awaited, so that we have no silent hidden panics.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/879

## Pre-review checklist

### DIFF
--- a/scylla/Cargo.toml
+++ b/scylla/Cargo.toml
@@ -172,3 +172,6 @@ tempfile = "3.19"
 rcgen = "0.14"
 uuid = { version = "1.0", features = ["v1"] }
 env_logger = "0.11"
+# Used by the TLS-inspecting proxy of the session ticket ccm test to classify
+# the handshakes it forwards.
+tls-parser = "0.12.2"
--- a/scylla/tests/integration/ccm/lib/mod.rs
+++ b/scylla/tests/integration/ccm/lib/mod.rs
@@ -17,7 +17,7 @@ use ip_allocator::IpAllocator;
 use tracing::info;
 
 pub(crate) static CLUSTER_VERSION: LazyLock<String> = LazyLock::new(|| {
-    std::env::var("SCYLLA_TEST_CLUSTER").unwrap_or("release:2026.1.0".to_string())
+    std::env::var("SCYLLA_TEST_CLUSTER").unwrap_or("release:2026.2.0".to_string())
 });
 
 static TEST_KEEP_CLUSTER_ON_FAILURE: LazyLock<bool> = LazyLock::new(|| {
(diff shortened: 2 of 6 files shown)


Lorak-mmk:
@coderabbitai full review

wprzytula (approved):

PR #1801, opened 2026-07-27
connection: discard connections after keyspace setup failure

### LINKED ISSUE #1682: Discard new connections when keyspace setup fails
When setting the keyspace on a newly opened connection fails, the pool logs the error but still reports the ready connection as successful with a keyspace attached.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/scylla/src/network/connection_pool.rs#L1233

The ready-connection handler can then accept and publish a connection as if the current keyspace was installed. Later unqualified queries may run with no keyspace, or with the wrong keyspace, instead of retrying or discarding that connection.

Suggested direction: treat keyspace setup failure as a connection setup failure, so the pool retries or drops the connection rather than publishing it.

Guflly:
## Description

A failed `USE` on a newly opened connection is now returned to the pool as a keyspace setup error. The connection is dropped and the pool schedules another refill instead of publishing it as ready.

The regression coverage verifies the failure path preserves the current pool and the success path completes keyspace setup before resharding.

## Validation

- `cargo +1.88.0 test -p scylla --lib network::connection_pool::tests::keyspace_setup_failure -- --nocapture` (2 passed)
- `cargo +1.88.0 check -p scylla --lib --target x86_64-unknown-linux-gnu`
- `git diff --check`

## Pre-review checklist

- [x] The patch is logically coherent.
- [x] The commit messages explain what changes and why.
- [x] The bug fix has regression coverage.
- [x] Affected code compiles and the focused tests pass.
- [x] The description explains the motivation and implementation.
- [x] The new non-exhaustive `ConnectionError` variant is documented.
- [x] User documentation changes are not needed.
- [x] The issue-closing annotation is included below.

Fixes: #1682

### DIFF
--- a/scylla/src/network/connection_pool.rs
+++ b/scylla/src/network/connection_pool.rs
@@ -862,7 +862,7 @@ impl PoolRefiller {
     fn handle_ready_connection(&mut self, evt: OpenedConnectionEvent) {
         let endpoint = self.endpoint_description();
         match evt.result {
-            Err(err) => {
+            Err(ConnectionSetupError::Connection(err)) => {
                 if evt.requested_shard.is_some() {
                     // If we failed to connect to a shard-aware port,
                     // fall back to the non-shard-aware port.
@@ -897,7 +897,37 @@ impl PoolRefiller {
                     }
                 }
             }
+            Err(ConnectionSetupError::Keyspace(err)) => {
+                self.metrics.dec_total_connections();
+                self.had_error_since_last_refill = true;
+                debug!(
+                    "[{}] Failed to set keyspace for new connection: {}",
+                    endpoint, err,
+                );
+
+                if !self.is_filling() && self.is_empty() {
+                    self.update_shared_conns(Some(err.into()));
+                }
+            }
             Ok((connection, error_receiver)) => {
+                // Before the connection can be put to the pool, we need
+                // to make sure that it uses appropriate keyspace
+                if let Some(keyspace) = &self.current_keyspace
+                    && evt.keyspace_name.as_ref() != Some(keyspace)
+                {
+                    // Asynchronously start setting keyspace for this
+                    // connection. It will be received on the ready
+                    // connections channel and will travel through
+                    // this logic again, to be finally put into
+                    // the conns.
+                    self.start_setting_keyspace_for_connection(
+                        connection,
+                        error_receiver,
+                        evt.requested_shard,
+                    );
+                    return;
+                }
+
                 // Update sharding and optionally reshard
                 let shard_info = connection.get_shard_info().as_ref();
                 let sharder = shard_info.map(|s| s.get_sharder());
@@ -927,24 +957,6 @@ impl PoolRefiller {
                         connection.get_shard_aware_port(),
                     );
                     self.shard_aware_port = connection.get_shard_aware_port();
-                }
-
-                // Before the connection can be put to the pool, we need
-                // to make sure that it uses appropriate keyspace
-                if let Some(keyspace) = &self.current_keyspace
-                    && evt.keyspace_name.as_ref() != Some(keyspace)
-                {
-                    // Asynchronously start setting keyspace for this
-                    // connection. It will be received on the ready
-                    // connections channel and will travel through
-                    // this logic again, to be finally put into
-                    // the conns.
-                    self.start_setting_keyspace_for_connection(
-                        connection,
-                        error_receiver,
-                        evt.requested_shard,
-                    );
-                    return;
                 }
 
                 let active_connection_count = self.active_connection_count();
@@ -1060,7 +1072,7 @@ impl PoolRefiller {
                 count_in_metrics(&result);
 
                 OpenedConnectionEvent {
-                    result,
+                    result: result.map_err(ConnectionSetupError::Connection),
                     requested_shard: Some(RequestedShard { shard, sharder }),
                     keyspace_name: None,
                 }
@@ -1073,7 +1085,7 @@ impl PoolRefiller {
                 count_in_metrics(&result);
 
                 OpenedConnectionEvent {
-                    result,
+                    result: result.map_err(ConnectionSetupError::Connection),
                     requested_shard: None,
                     keyspace_name: None,
                 }
@@ -1326,16 +1338,13 @@ impl PoolRefiller {
         let keyspace_name = self.current_keyspace.as_ref().cloned().unwrap();
         self.ready_connections.push(
             async move {
-                let result = connection.use_keyspace(&keyspace_name).await;
-                if let Err(err) = result {
-                    warn!(
-                        "[{}] Failed to set keyspace for new connection: {}",
-                        connection.get_connect_address().ip(),
-                        err,
-                    );
-                }
+                let result = connection
+                    .use_keyspace(&keyspace_name)
+                    .await
+                    .map(|()| (connection, error_receiver))
+                    .map_err(ConnectionSetupError::Keyspace);
                 OpenedConnectionEvent {
-                    result: Ok((connection, error_receiver)),
+                    result,
                     requested_shard,
                     keyspace_name: Some(keyspace_name),
                 }
@@ -1401,9 +1410,14 @@ struct RequestedShard {
 }
 
 struct OpenedConnectionEvent {
-    result: Result<(Connection, ErrorReceiver), ConnectionError>,
+    result: Result<(Connection, ErrorReceiver), ConnectionSetupError>,
     requested_shard: Option<RequestedShard>,
     keyspace_name: Option<VerifiedKeyspaceName>,
+}
+
+enum ConnectionSetupError {
+    Connection(ConnectionError),
+    Keyspace(UseKeyspaceError),
 }
 
 /// Signals that connectivity to a node has changed.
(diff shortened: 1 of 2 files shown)


Guflly:
Fixed both failure paths. Keyspace setup now finishes before sharding state is changed, and dropped connections balance the active connection metric. Added regression coverage for both.

scylla/src/network/connection_pool.rs
```diff
@@ -819,23 +819,15 @@ impl PoolRefiller {
                     }
                 }
             }
+            Err(ConnectionSetupError::Keyspace(err)) => {
+                self.metrics.dec_total_connections();
+                self.had_error_since_last_refill = true;
```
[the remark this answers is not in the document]
Guflly: Fixed in 27ed7d10. The last failed keyspace setup now marks an empty pool as broken and wakes initialization waiters. The original UseKeyspaceError is kept, and the test covers both the state and notification.

scylla/src/network/connection_pool.rs
```diff
@@ -894,7 +894,33 @@ impl PoolRefiller {
...
+                warn!(
+                    "[{}] Failed to set keyspace for new connection: {}",
+                    endpoint, err,
+                );
+            }
```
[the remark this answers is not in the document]
Guflly: Fixed in 27ed7d10. The last failed keyspace setup now marks an empty pool as broken and wakes initialization waiters. The original UseKeyspaceError is kept, and the test covers both the state and notification.

wprzytula (changes requested):
I reviewed this manually and it looks good.

Then I ran Opus review and the important insights are the following:

1. > One design note (not a bug): a transient `UseKeyspaceError::RequestTimeout` now discards an otherwise-healthy TCP connection rather than retrying the `USE` on it. That's safe and simpler, but slightly wasteful. Worth a one-line comment acknowledging the choice.

2. Error handling:
>   - Log level inconsistency: keyspace arm uses `warn!`, the connection arm uses `debug!`. Defensible, but worth aligning intentionally.
>   - The new `ConnectionError::UseKeyspaceError` is a **new public enum variant**, so the checklist item "No public items were introduced" is technically inaccurate. It's non-breaking (`ConnectionError` is `#[non_exhaustive]`), and semantically fits ("connection can no longer be used and should be dropped"). Just flag it as a public-API addition.

3. > Tests — correct, readable, non-flaky, but one coverage gap
>  Gaps / suggestions:
>   - **No unit test for the success path after the reorder** (keyspace set OK → connection added, pool resharded correctly). The happy path is covered by `scylla/tests/integration/session/use_keyspace.rs`, but a focused unit test would round out coverage of the reordered branch and guard against future regressions.
>   - In test 1, `requested_shard: Some(...)` is used specifically to prove no shard-aware fallback fires on keyspace failure — good — but a short comment would make that intent explicit to readers.

<details><summary>Full review here.</summary>
<p>


# Assistant (Plan · Claude Opus 4.8 · 36.5s)

## Summary of the PR

The PR fixes issue #1682: a connection whose `USE keyspace` failed during setup was still being added to the pool (the old `start_setting_keyspace_for_connection` returned `Ok((connection, ...))` even on failure). Such a connection could serve queries against the wrong/default keyspace. The fix threads a new `ConnectionSetupError { Connection, Keyspace }` enum through `OpenedConnectionEvent`, so a keyspace failure is now a distinct error variant that discards the connection and triggers a refill.

Three commits:
1. `discard connections after keyspace setup failure` — core fix
2. `finish setup before updating pool` — moves the keyspace-check to the **top** of the `Ok` arm, before `maybe_reshard`/shard-awareness/shard-port updates
3. `report terminal keyspace setup failures` — reports the error into `shared_conns` and adds `ConnectionError::UseKeyspaceError`

## 1. Connection pool logic correctness

The logic is sound. The key insight of commit 2 (reordering) is correct and important:

- Previously, a connection needing keyspace setup would run `maybe_reshard(sharder)` **before** its setup finished. If the node had resharded, this would **clear the existing healthy pool** (`maybe_reshard` calls `self.conns.clear()` at connection_pool.rs:1110), and then if keyspace setup failed, the pool would be left empty. The reorder defers all sharding side-effects until the connection is fully ready.
- `evt.requested_shard` and `evt.keyspace_name` are correctly threaded through `start_setting_keyspace_for_connection`, so the advanced-shard-awareness mismatch check still runs (just deferred to when the connection comes back ready). No information is lost.
- The single-actor model (all `&mut self` on one task) means the `!self.is_filling() && self.is_empty()` guard and the `is_empty()` re-check inside `update_shared_conns` are race-free.

One design note (not a bug): a transient `UseKeyspaceError::RequestTimeout` now discards an otherwise-healthy TCP connection rather than retrying the `USE` on it. That's safe and simpler, but slightly wasteful. Worth a one-line comment acknowledging the choice.

## 2. Metrics inc/dec correctness — the subtle part, done right

The asymmetry between the two error arms is **correct and non-obvious**:

- `count_in_metrics` only calls `inc_total_connections()` on a **successful** open (connection_pool.rs:1046). A failed *connect* was never counted, so the `Connection` error arm correctly does **not** decrement.
- A keyspace failure happens on a connection that already opened successfully (already inc'd), so the new `dec_total_connections()` at connection_pool.rs:898 correctly rebalances.
- On the `Err(Keyspace)` path, the `connection`/`error_receiver` are dropped inside the async task's `.map(|()| (connection, error_receiver))` closure (never added to `conns`/`excess_connections`), so `remove_connection` won't double-decrement. Balanced.

`dec_total_connections` is a raw `fetch_sub` (would wrap on underflow), but every dec here is paired with a prior inc, so no underflow in practice — consistent with the existing `remove_connection` usage.

## 3. Error handling — uniform where it should be

The keyspace arm mirrors the non-shard-aware connection arm (`had_error_since_last_refill = true`; `if !is_filling && is_empty → update_shared_conns(Some(err.into()))`). The differences are all justified:

- **No shard-aware fallback** in the keyspace arm even when `requested_shard.is_some()` — correct, since the connection itself succeeded (the shard-aware port works); only the keyspace `USE` failed.
- **Sets `had_error_since_last_refill` unconditionally**, unlike the shard-aware *connect* failure which deliberately skips it — correct, a keyspace failure is a genuine error that would recur on the regular port too.

Minor nits:
- Log level inconsistency: keyspace arm uses `warn!`, the connection arm uses `debug!`. Defensible, but worth aligning intentionally.
- The new `ConnectionError::UseKeyspaceError` is a **new public enum variant**, so the checklist item "No public items were introduced" is technically inaccurate. It's non-breaking (`ConnectionError` is `#[non_exhaustive]`), and semantically fits ("connection can no longer be used and should be dropped"). Just flag it as a public-API addition.

## 4. Tests — correct, readable, non-flaky, but one coverage gap

Both tests are correct and non-flaky:

- **`keyspace_setup_failure_triggers_refill`** (unit): verifies `had_error`, empty `ready_connections` (i.e. **no** shard-aware fallback was started), `is_empty`, `need_filling`, `Broken(UseKeyspaceError::RequestTimeout)`, `pool_updated` notified, and metrics decremented to 0. Good, self-contained, no timing.
- **`keyspace_setup_failure_preserves_pool_state`** (proxy): the important regression test for commit 2. `change_request_rules` writes under a `Mutex` synchronously (proxy.rs:413), and the second `open_connection` starts *after* it, so the new connection **deterministically** reports 2 shards — the test genuinely exercises the reshard-deferral. It asserts the pool is *not* resharded/cleared before setup completes and is *preserved* after the failure. No sleeps; `ready_connections.next().await` resolves off the proxy's `server_error` response. Non-flaky. `connection_errors` is never polled, so the kept connection can't be spuriously removed.

Gaps / suggestions:
- **No unit test for the success path after the reorder** (keyspace set OK → connection added, pool resharded correctly). The happy path is covered by `scylla/tests/integration/session/use_keyspace.rs`, but a focused unit test would round out coverage of the reordered branch and guard against future regressions.
- In test 1, `requested_shard: Some(...)` is used specifically to prove no shard-aware fallback fires on keyspace failure — good — but a short comment would make that intent explicit to readers.

## Verdict

The change is correct, the tricky bits (metrics balance and the reorder-before-reshard) are handled properly, and tests are deterministic. I'd recommend only minor follow-ups: align the log level, correct the "no public items" checklist claim (or note the non-breaking variant), and consider adding a success-path unit test for the reordered branch.


</p>
</details>

Guflly:
Updated. I added coverage for successful keyspace setup before resharding, switched the setup-failure log to debug, and corrected the public API checklist item. Both focused keyspace setup tests pass.

Guflly:
Added those two notes. The success-path test, debug log, and public API checklist update are also in place.

Guflly:
Updated this. The success path is now covered and the checklist calls out the new public variant. Discarding on any `USE` failure is intentional, and the keyspace path stays at `warn!` because an opened connection is being dropped. Could you take another look?

scylla/src/network/connection_pool.rs
```diff
@@ -1415,30 +1429,255 @@ impl ConnectivityChangeEvent {
...
+                sharder,
+                connections,
+            }) if sharder.nr_shards.get() == 2
+                && connections.iter().map(Vec::len).sum::<usize>() == 1
+        ));
+
+        match proxy.finish().await {
+            Ok(()) | Err(ProxyError::Worker(WorkerError::DriverDisconnected(_))) => {}
+            Err(err) => panic!("{err}"),
+        }
```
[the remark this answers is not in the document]
wprzytula: Oh, this was a pre-existing bug, right? @Guflly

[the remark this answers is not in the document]
wprzytula: cc @Lorak-mmk

wprzytula (changes requested):
The commit history is tangled, while this PR really consists of two parts:
1. Bug fix for connection metrics imbalance. This solves a pre-existing bug.
2. Behaviour change: discarding connection upon keyspace setup failure.

Could you rewrite the code to consist of these two commits? Right now it's hard to review:
- in "all changes" view, both parts are mixed together;
- per-commit view is even worse with current commits.

Guflly:
Yes, the metrics imbalance was pre-existing. I rewrote the branch into the two requested commits.

wprzytula (commented):
Thank you for the commit rewrite! Now it was very clean to read.

I think we're still missing a (regression) test for the bug fix in the first commit. Could you trigger/mock the scenario that was fixed to make sure the metrics imbalance is gone?

If you decide to write it, I encourage you to first share your test idea, then we'll discuss it, only then let's jump into implementation.

Guflly:
My plan is to add a focused proxy-backed test for the reshard path: start with one non-sharded connection and total_connections at 1, open a second connection with a different sharder, process it, then verify the old connection is removed, the new one is retained, and the metric stays at 1. Does that cover the case you had in mind, or should the test also cover the excess-connection clearing paths?

wprzytula:
> My plan is to add a focused proxy-backed test for the reshard path: start with one non-sharded connection and total_connections at 1, open a second connection with a different sharder, process it, then verify the old connection is removed, the new one is retained, and the metric stays at 1. Does that cover the case you had in mind, or should the test also cover the excess-connection clearing paths?

The direction is IMO good. Excess-connection logic is unfortunately rarely tested, so it will be great if you improve the coverage of it.

Guflly:
Added in f746e4c7. It covers resharding, the excess-limit clear, and the full-pool clear. The focused tests, existing keyspace tests, fmt, and clippy pass.

wprzytula (changes requested):
One last request from me: please put the metrics balance test's commit before the commit that changes the logic wrt discarding connections after keyspace setup failure. This way we will be sure that the test passes with both the old logic and with the new logic.

Guflly:
Reordered. The metric-balance test now comes before the discard change, and the final tree is unchanged. Formatting passes; the focused test hits the existing Unix-only socket code on Windows, so CI will cover the Linux run.

scylla/src/network/connection_pool.rs
```diff
@@ -1610,6 +1611,167 @@ mod tests {
...
+        refiller
+            .ready_connections
+            .push(futures::future::ready(event).boxed());
+        let (use_keyspace_sender, use_keyspace_receiver) = mpsc::channel(1);
+        let worker = tokio::spawn(refiller.run(use_keyspace_receiver));
+
+        tokio::time::timeout(Duration::from_secs(5), async {
+            while metrics.get_total_connections() != 1 {
+                tokio::task::yield_now().await;
+            }
```
wprzytula, line 2004: 🔧 Using `tokio::timeout::sleep(Duration::from_millis(1))` could decrease CPU burn.

wprzytula: Especially that otherwise we keep burning the CPU for 5 seconds.

wprzytula (approved):

scylla/src/network/connection_pool.rs
Guflly: Updated in 1f58e5b4. The loop now sleeps for 1 ms between checks.

scylla/src/network/connection_pool.rs
```diff
@@ -897,7 +897,37 @@ impl PoolRefiller {
...
+                    // this logic again, to be finally put into
+                    // the conns.
+                    self.start_setting_keyspace_for_connection(
+                        connection,
+                        error_receiver,
+                        evt.requested_shard,
+                    );
+                    return;
+                }
+
```
Lorak-mmk, line 930: Commit: connection: discard connections after keyspace setup failure

This is the only thing that is not clear to me. Why is this block moved?
The thing we do between the two locactions:
- Blocking shard awareness if necessary
- Resharding if necessary
- Updating shard-aware port

Neither of those requires keyspace to be set, and they all benefit from being done sooner rather than later.

What am I missing?

Lorak-mmk: If there is some reason for the block to be moved, can you put it in the comment, so noone accidentaly moves it back?

Lorak-mmk: After analysis, I don't think the ordering matters. Either is fine.

Lorak-mmk (approved):

Lorak-mmk (approved):

Issue #1764, opened 2026-07-01
Advanced shard awareness doesn't work in our integration tests


Lorak-mmk:
scylla-proxy supports shard-aware port (aka advanced shard awareness), but in a slightly weird way.

In QueryNode mode, when proxy instance is spawned (look at DoorKeeper::run), it will first query the node and save (only!) shards amount.
Then, when client connection is opened, it will bind a local port that matches the same shard as driver port.

Problem:
Proxy only connects to the specified server address and port. For example, in our commonly used `test_with_3_node_cluster` function, we connect it to all 3 scylla nodes, to port 9042. This means such proxy will always open connections to scylla port 9042, ignoring the shard-aware port sent in SUPPORTED.

The dirtiest fix for this specific issue is to change ports in `test_with_3_node_cluster` to connect to 19042, but it is hardly satisfactory.

Perhaps proxy should take multiple addresses to bind to, like Scylla does? non-shard-aware, non-shard-aware with tls, shard-aware, shard-aware with tls? All of them optional? This would need some handling / changing the SUPPORTED contents probably? It may also require changing our current address translation approach. Hard to tell.

PR #1804, opened 2026-07-29
connection_pool: back off from advanced shard awareness on shard mismatch

### LINKED ISSUE #1447: Disable advanced shard awareness when NAT is detected
When driver discovers that a connection was opened using advanced shard awareness, but it was opened to a different shard than requested, it will start opening normal connection.
It won't do any permanent change in behavior: subsequent connections will still try to use advanced shard awareness, even though it has basically 0 chance of succeeding. This will lead to waste of resources.

Lorak-mmk:
### Problem

 Advanced shard awareness works by choosing a connection's source port so that ScyllaDB assigns the requested shard to it. When something on the path rewrites source ports — a NAT being the usual culprit — the connection lands on an arbitrary shard instead.

 Until now the driver reacted to such a mismatch by merely opening a replacement connection to the regular transport port (handle_ready_connection, the evt.requested_shard.is_some() arm). That repaired the
 single connection but made no lasting change: every subsequent connection attempt still picked a source port, still had virtually no chance of hitting the requested shard, and still burned a connection to
 discover that — for the whole lifetime of the pool. In a NAT'd deployment that is a permanent waste of connections and of pool-filling latency.

 ### Change

 Upon detecting that a connection opened with advanced shard awareness landed on a shard other than the requested one, the pool stops attempting advanced shard awareness towards that node for 5 minutes and fills itself through the regular transport port. One WARN is logged per block, naming the likely cause and the disallow_shard_aware_port escape hatch.

 Design points worth reviewing:

 - Temporary, not permanent. A repaired network setup is picked up without restarting the application. 5 minutes is long enough not to waste connections on attempts that are nearly certain to fail, short enough  to self-heal.
 - Not re-armed while in effect. Otherwise a steady stream of mismatches could postpone the retry indefinitely.
 - A sharder change is not blamed. A shard-count change explains a mismatch just as well (the source port was computed with a stale shard count), so that case is excluded from the diagnosis.

 ### Note for reviewers

We know that shard-aware port doesn't work well with scylla-proxy in our integration tests: https://github.com/scylladb/scylla-rust-driver/issues/1764
This PR will start triggering warnings in this scenario. I don't think this is a big problem.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1447
Refs: https://github.com/scylladb/scylla-rust-driver/issues/1764

 ### Pre-review checklist

### DIFF
--- a/scylla/src/network/connection_pool.rs
+++ b/scylla/src/network/connection_pool.rs
@@ -26,6 +26,7 @@ use std::num::NonZeroUsize;
 use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::pin::Pin;
 use std::sync::{Arc, RwLock, Weak};
+use std::time::Duration;
 use uuid::Uuid;
 
 use tokio::sync::{Notify, mpsc};
@@ -460,6 +461,13 @@ impl NodeConnectionPool {
 
 const EXCESS_CONNECTION_BOUND_PER_SHARD_MULTIPLIER: usize = 10;
 
+/// For how long the driver refrains from using advanced shard awareness towards a node
+/// after discovering that it does not work for that node.
+///
+/// Long enough not to waste connections on an attempt that is nearly certain to fail,
+/// short enough for the driver to recover on its own if the network setup changes.
+const ADVANCED_SHARD_AWARENESS_BLOCK_DURATION: Duration = Duration::from_secs(300);
+
 struct PoolRefiller {
     // Following information identify the pool and do not change
     pool_config: HostPoolConfig,
@@ -473,6 +481,15 @@ struct PoolRefiller {
     // Following fields are updated with information from OPTIONS
     shard_aware_port: Option<u16>,
     sharder: Option<Sharder>,
+
+    // If set, advanced shard awareness (choosing the connection's source port so that ScyllaDB
+    // assigns the desired shard to it) is not attempted until this instant.
+    //
+    // It is set upon discovering that a connection opened with advanced shard awareness landed
+    // on a different shard than the requested one, which means that something in between
+    // (most likely a NAT) rewrites our source ports. Further attempts would then have virtually
+    // no chance of hitting the requested shard, so they would only waste connections.
+    advanced_shard_awareness_blocked_until: Option<tokio::time::Instant>,
 
     // `shared_conns` is updated only after `conns` change
     shared_conns: Arc<ArcSwap<MaybePoolConnections>>,
@@ -553,6 +570,7 @@ impl PoolRefiller {
 
             shard_aware_port: None,
             sharder: None,
+            advanced_shard_awareness_blocked_until: None,
 
             shared_conns,
             conns,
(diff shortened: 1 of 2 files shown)


scylla/src/network/connection_pool.rs
```diff
@@ -717,6 +735,35 @@ impl PoolRefiller {
...
+    }
+
+    /// Stops using advanced shard awareness towards this node for
+    /// [`ADVANCED_SHARD_AWARENESS_BLOCK_DURATION`], after it turned out not to work.
+    fn block_advanced_shard_awareness(&mut self) {
+        // Nothing to log or update if a block is already in effect - the duration is constant,
+        // so re-arming it could only postpone the retry indefinitely under a stream of mismatches.
+        if self.is_advanced_shard_awareness_blocked() {
+            return;
+        }
```
wprzytula, line 753: 🤔 Why would we like to unblock shard awareness when still facing a stream of mismatches?

scylla/src/network/connection_pool.rs
```diff
@@ -717,6 +735,35 @@ impl PoolRefiller {
...
+        warn!(
+            "[{}] A connection opened with advanced shard awareness landed on an unrequested shard, \
+             which suggests that the source port is rewritten (e.g. by a NAT) on the way to the node. \
+             Not using advanced shard awareness for this node for the next {} s. \
+             Requests will still be routed to the correct shards, but the connection pool \
+             will be filled less efficiently. If you intentionally want to not use shard aware port, \
+             use `disallow_shard_aware_port` config option.",
+            self.endpoint_description(),
+            ADVANCED_SHARD_AWARENESS_BLOCK_DURATION.as_secs(),
+        );
```
wprzytula, line 765: I'd add:
_Consider if it's possible to repair your network configuration to prevent source port rewriting._

wprzytula (approved):
Looks good! 🚀

scylla/src/network/connection_pool.rs
Lorak-mmk: Hm.. I don't think this matters that much. 
When the shard-aware port is blocked, we won't use it and thus won't get new mismatches from new attempts.
What is possible is a big `connect_timeout`, and an old attempt getting a mismatch long after we blocked the shard-aware port. Semantics where the 5 minutes are counted from first mismatch just make more sense to me intuitively. 

But again, this is not an important edge case so I can change that if you want.

Lorak-mmk:
- Added suggested text to the warning
- Changed comment on session builder from "5 minutes" to "few minutes" in order to not suggest to users that this value is stable.

scylla/src/network/connection_pool.rs
wprzytula: I also feel this is not important. Let's leave it as-is, I was just curious why you followed this reasoning and even bothered. I guess it was AI.

wprzytula (approved):

scylla/src/network/connection_pool.rs
Lorak-mmk: AI wrote all this of course, but I did look at it and thought the semantics it invented are reasonable. With that, the interval without a warning is actually 5 minutes. If we removed this check, the function would need to be rewritten some other way to not spam the warning.

PR #1810, opened 2026-07-31
Extract `scylla-ccm-bridge` crate for reuse

wprzytula:
## What

This PR extracts ccm bridge lib as a separate crate. 

## Why

I want to reuse the CCM bridge in CPP RS integration testing, especially for TLS and client routes there.

## How

- missing docstrings are added;
- items are made `pub`.
- the new crate has two features: one for client routes (needs `scylla`, `scylla-cql`, and `scylla-proxy`) and one for exposing a facility method to create a `SessionBuilder` from CCM `Cluster` (needs `scylla`). When the features are disabled, there's no dependency on any other `scylla*` crate;
- I fixed doc tests so that now they compile correctly.

## Pre-review checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "scylla-cql",
     "scylla-cql-core",
     "scylla-proxy",
+    "scylla-ccm-bridge",
     "utils/analyze_errors_size",
     "utils/find_stale_test_keyspaces",
     "utils/generate_book_tests",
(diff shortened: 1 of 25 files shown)


Lorak-mmk (approved):
Commit structure is not the greatest - the last commit that could just move files and add Cargo.toml also publishes functions and adds docstrings - but I don't care about this much in a PR like this.

Great, thanks! Now we'll be also able to run examples in CI.

PR #1811, opened 2026-07-31
Bump syn to v3 and darling to 0.24

Lorak-mmk:
v3 is a major version, but we were only slightly affected by breaking changes. Changelog says:
> Some enums no longer provide From impls. Construct the variant by name
instead, such as Expr::Array(e) in place of e.into().

In our case, ImplItemFn no longer implements Into<ImplItem>, so in 2 places `.into()` calls had to be changed to manual variant construction.

## Pre-review checklist

### DIFF
--- a/scylla-macros/Cargo.toml
+++ b/scylla-macros/Cargo.toml
@@ -16,8 +16,8 @@ proc-macro = true
 all-features = true
 
 [dependencies]
-darling = "0.23.0"
-syn = "2.0"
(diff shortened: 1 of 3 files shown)


Lorak-mmk:
@coderabbitai You added "semver-checks-breaking" label. This change is, to the best of my knowledge, not semver breaking and doesn't force us to bump our major version. Don't do that/

wprzytula (approved):

PR #1813, opened 2026-07-31
Run all examples in CI

### LINKED ISSUE #839: Run examples in the CI

### LINKED ISSUE #1675: Fix UDT example query to select the UDT column
The user-defined type example creates a table with columns `k` and `my`, but later queries different column names.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/examples/user-defined-type.rs#L56

As a result, running the example fails with an invalid-column query instead of demonstrating UDT deserialization.

Suggested direction: select the UDT column from the example table, for example selecting `my` from the table created earlier in the example.

### LINKED ISSUE #1676: Fix CQL time types example reads to select column t
The time and timestamp example tables define their value column as `t`, but the read queries select another column name.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/examples/cql-time-types.rs#L125

Scylla rejects the first such read as an unknown column, so the example aborts before demonstrating time and timestamp deserialization.

Suggested direction: update those reads to select the `t` column created by the example.

### LINKED ISSUE #1677: Fix tracing example row type for paged query
The paged tracing example queries one text column, but the stream is typed as a three-column tuple.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/examples/tracing.rs#L89

The row stream performs row type checking, so the example fails with a column-count mismatch before it can print tracing IDs.

Suggested direction: use a one-column tuple for this query, or select columns matching the tuple type.

### LINKED ISSUE #1688: Make schema agreement example exercise manual waiting
The schema-agreement example leaves automatic schema-agreement waiting enabled while also demonstrating a manual wait.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/examples/schema_agreement.rs#L16

Schema-changing queries already await schema agreement by default, so if the create statement times out the example exits before the manual wait, and if it succeeds the manual wait checks an already-agreed schema. That makes the timeout branch misleading for the behavior the example is trying to show.

Suggested direction: disable automatic schema-agreement waiting for this example, or remove the manual timeout handling.

Lorak-mmk:
This PR fixes bugs in the examples, improves their structure, and makes
all of them runnable. CI job that runs them all is added.

## Why

We had 26 examples and no way of knowing whether any of them still worked. Nothing
ran them, so the only thing enforced was that they compiled — and compiling turns
out to say very little. What that missed:

- `tls_openssl.rs` told the reader to hand-edit `scylla.yaml`, mount certificates
  into a docker container and connect on port 9142, carried a
  `FIXME(wprzytula): Adjust to rustls.`, and loaded its CA from `./test/tls/ca.crt`
  — **a path that does not exist in this repository**.
- `client_routes.rs` dialled a hardcoded `127.0.0.1:12137`. Nothing in this
  repository has ever listened there.
- `custom_load_balancing_policy.rs` built a session and returned without issuing a
  single request. Its favourite datacenter was `"PL"`, which exists on no cluster we
  run, so the policy silently took its fallback branch every time. The half worth
  demonstrating had never executed.
- `auth.rs` connected with `.user("cassandra", "cassandra")` to a cluster running
  `AllowAllAuthenticator`, which ignores credentials. It would have passed with the
  wrong password, or with no `.user()` call at all.
- Two examples queried columns their own `CREATE TABLE` never defined.

Each of these type-checked for years. So this PR does two things: it fixes the
examples, and it removes the mechanism that let them rot — from now on CI runs
every one of them.

## What you get

- `make run-examples` runs all the examples.
- A CI job does both on every change.
- `examples/README.md` says what each example teaches, so you can pick one without
  opening files.

## Structure

`examples/` now contains exactly 26 example files, `Cargo.toml` and `README.md`.
Cluster scaffolding too big to sit inside an example lives in `examples/ci/`, which
holds no examples at all.

Four examples need a cluster the compose file cannot provide, so they build one with
`scylla-ccm-bridge`, the same way the integration tests do. In those files the setup
is fenced between `--- CI setup ---` banner comments — it is scaffolding, not the
lesson, and a reader can skip straight past it.

## The commits

Sixteen commits, in four groups.

**Hygiene** — small, independently correct fixes.

| # | Commit |
| --- | --- |
| 1 | `examples: fix queries selecting nonexistent columns` |
| 2 | `examples: drop dependencies no example uses` |
| 3 | `examples: declare the chrono features the examples rely on` |

**Structure** — one example, one lesson.

| # | Commit |
| --- | --- |
| 4 | `examples: move driver metrics out of the basic example` |
| 5 | `examples: merge parallel and parallel-prepared` |
| 6 | `examples: cut the boilerplate copied into the schema agreement example` |
| 7 | `examples: use snake_case for every example name` |
| 8 | `examples: rename two examples whose names no longer describe them` |
| 9 | `examples: add an example for batch statements` |

Commit 7 is a pure rename, so the semantically empty change stays separate from the
ones that alter behaviour. Commit 9 fills a real gap: `Session::batch` had a book
chapter and no example, and it is the technique most often confused with firing many
independent requests.

**Cluster setup** — the four examples a plain cluster cannot serve.

| # | Commit |
| --- | --- |
| 10 | `examples: bring up a TLS cluster with CCM so the TLS examples can run` |
| 11 | `examples: authenticate against a cluster that actually enforces it` |
| 12 | `examples: exercise the custom load balancing policy on two datacenters` |
| 13 | `examples: stand up a client-routes topology so the example can run` |

Each is self-contained and each ends with a check that would fail if the feature
silently did not engage — the TLS examples ask `system.clients` whether the server
considers the connection encrypted, and `auth.rs` follows a successful connection
with a deliberately wrong password to prove the credentials are load-bearing.

**Running them.**

| # | Commit |
| --- | --- |
| 14 | `examples: add script to run all and README` |
| 15 | `ci: run the examples` |

**Bonus doc fix**

Commit 16: `docs: stop teaching deprecated metrics accessors` |

## Caveats

**Is `cqlsh_rs` really runnable?** It is interactive, but with stdin on `/dev/null` it
sees EOF and exits 0. The runner redirects stdin for every example, which also keeps
any future example from blocking CI on a prompt.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/839

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1675
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1676
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1677
Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1688

## Pre-review checklist


wprzytula:
> **Running them.**
> # 	Commit
> 14 	`examples: make running the whole directory one command`
> 15 	`ci: run the examples`
> 16 	`docs: stop teaching deprecated metrics accessors`

:question: How is commit 16 related to running the examples?

Lorak-mmk:
> > **Running them.**
> > # Commit
> > 14 	`examples: make running the whole directory one command`
> > 15 	`ci: run the examples`
> > 16 	`docs: stop teaching deprecated metrics accessors`
> 
> ❓ How is commit 16 related to running the examples?

Fixed. 

@wprzytula Ready for review now.
From the previous version there were some simplifications:
- Script to run examples is shortened, no longer keeps a list of ccm vs non-ccm example, which also simplifies CI
- Some code deduplication related to client routes

This second part is one thing that I did not carefully review - you know this code much better so could you see if the changes are reasonable?

scylla-ccm-bridge/src/client_routes.rs
```diff
@@ -1348,43 +1338,68 @@ fn build_client_routes_change_body(connection_ids: &[String], host_ids: &[String
...
-fn serde_json_routes(routes: &[RouteEntry]) -> String {
+/// Serialize routes into the REST API's JSON payload without pulling in serde_json.
+fn serde_json_routes(routes: &[ClientRoute]) -> String {
     let entries: Vec<String> = routes
         .iter()
         .map(|r| {
             format!(
                 r#"{{"connection_id":"{}","host_id":"{}","address":"{}","port":{},"tls_port":{}}}"#,
-                r.connection_id, r.host_id, r.address, r.port, r.tls_port
+                r.connection_id,
```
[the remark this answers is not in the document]
Lorak-mmk: Not introduced by this PR

.github/workflows/rust.yml
```diff
@@ -164,6 +164,39 @@ jobs:
...
+        if: ${{ always() }}
+        run: docker compose -f test/cluster/docker-compose.yml logs
+      - name: Remove cluster
+        run: docker compose -f test/cluster/docker-compose.yml down
+
+      # The remaining examples need a cluster the compose file cannot provide -
+      # TLS, password authentication, two datacenters, published client routes -
+      # so each brings up its own with scylla-ccm, as the ccm tests do.
+      - name: Install scylla-ccm
+        run: pip3 install https://github.com/scylladb/scylla-ccm/archive/master.zip
```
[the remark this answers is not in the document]
Lorak-mmk: We want to use master

.github/workflows/rust.yml
```diff
@@ -164,6 +164,39 @@ jobs:
...
 
+  # Runs every example in `examples/`. Compiling them is already covered by the
+  # static checks; this job exists because compiling proves nothing about whether
+  # an example still describes reality - the stale ones we found had type-checked
+  # for years while pointing at cluster setups nobody had.
+  examples:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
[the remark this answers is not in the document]
Lorak-mmk: The token has only read permissions, it is harmless

scripts/run-examples.sh
```diff
@@ -0,0 +1,86 @@
...
+# driver with an opaque spawn failure and nothing names the real cause.
+if ! command -v ccm > /dev/null; then
+    echo "ERROR: ccm is not installed, and some examples start a cluster of their own." >&2
+    echo "See examples/README.md; install it with:" >&2
+    echo "  uv tool install \"git+https://github.com/scylladb/scylla-ccm.git\"" >&2
+    exit 1
+fi
+
+# Ask cargo for the full set of example targets instead of maintaining a list,
+# so that a newly added example is run without any further wiring.
```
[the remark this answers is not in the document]
Lorak-mmk: This is intentional

scylla-ccm-bridge/src/client_routes.rs
```diff
@@ -1402,26 +1417,21 @@ async fn post_client_routes_raw(node_ip: IpAddr, json_body: &str) -> Result<(),
...
-         Host: {}:{}\r\n\
+         Host: {}\r\n\
          Content-Type: application/json\r\n\
          Content-Length: {}\r\n\
          Connection: close\r\n\
          \r\n\
          {}",
-        node_ip,
-        10000,
+        api_addr,
```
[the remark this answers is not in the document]
Lorak-mmk: Not introduced by this PR, and not really important here.

examples/Cargo.toml
```diff
@@ -27,9 +24,8 @@ tracing = { version = "0.1.25", features = ["log"] }
 tracing-subscriber = { version = "0.3.14", features = ["env-filter"] }
 chrono = { version = "0.4", default-features = false }
 time = { version = "0.3.22" }
-uuid = { version = "1.0", features = ["v1"] }
+uuid = "1.0"
```
wprzytula, line 27: 📝 The `"v1"` feature was added to support doctests executed in `Timeuuid` section of the book.

examples/schema_agreement.rs
```diff
@@ -8,79 +17,59 @@ use std::time::Duration;
...
     let uri = env::var("SCYLLA_URI").unwrap_or_else(|_| "172.42.0.2:9042".to_string());
 
     println!("Connecting to {uri} ...");
 
     let session: Session = SessionBuilder::new()
         .known_node(uri)
-        .schema_agreement_interval(Duration::from_secs(1)) // check every second for schema agreement if not agreed first check
+        // Important: by default schema agreement is automatically awaited after each DDL,
+        // making our manual awaits no-op in this case.
+        .auto_await_schema_agreement(false)
```
wprzytula, line 29: 💭 Do you think _DDL_ is a widely known notion? Won't readers be confused?
At least, I'd say _DDL statement_.

examples/schema_agreement.rs
```diff
@@ -8,79 +17,59 @@ use std::time::Duration;
...
+        // How long to sleep between two checks while the nodes disagree.
```
wprzytula: ⛏️ two -> consecutive

examples/schema_agreement.rs
```diff
@@ -8,79 +17,59 @@ use std::time::Duration;
...
-            "CREATE TABLE IF NOT EXISTS examples_ks.schema_agreement (a int, b int, c text, primary key (a, b))",
-            &[],
-        )
-        .await?;
-
+    // This wait is not optional. Automatic waiting is off, and the statement
+    // below is sent to whichever node the load balancing policy picks - very
+    // likely not the one that just created the keyspace. Without waiting for
+    // the nodes to agree first, creating the table fails, sooner or later, with
+    // "Can't find a keyspace examples_ks".
```
wprzytula, line 44: 🤔 Is this a genuine error string?

examples/Cargo.toml
```diff
@@ -74,14 +78,14 @@ path = "execution_profile.rs"
...
-path = "logging.rs"
-
 [[example]]
 name = "logging_log"
 path = "logging_log.rs"
 
+[[example]]
+name = "logging_tracing"
+path = "logging_tracing.rs"
+
```
wprzytula, line 103: 💡 Maybe: _logging_via_log_ and _logging_via_tracing_?

docs/source/statements/batch.md
```diff
@@ -146,6 +149,13 @@ For more information about sending values in a statement see [Statement values](
 
 
 ### Performance
-Batch statements do not use token/shard aware load balancing, batches are sent to a random node.
-
-Use [prepared statements](prepared.md) for best performance
+A batch is sent to a single coordinator, which then has to fan its statements
+out to the replicas of each partition. That coordinator is chosen token- and
+shard-aware when the batch's first statement is a prepared statement whose
```
wprzytula: > That coordinator is chosen token- and shard-aware

⛏️ This reads weird.

examples/auth.rs
```diff
@@ -1,26 +1,111 @@
...
+        .user(USER, PASSWORD)
+        .build()
+        .await?;
 
-    session.query_unpaged("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await.unwrap();
-    session
-        .query_unpaged("DROP TABLE IF EXISTS examples_ks.auth;", &[])
+    session.query_unpaged("CREATE KEYSPACE IF NOT EXISTS examples_ks WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor' : 1}", &[]).await?;
+    println!("Connected, and the keyspace was created.");
+
```
wprzytula, line 65: 💭 By deleting `DROP TABLE`, you make the example non-idempotent. This is probably true for other examples too, but is it desired?

examples/custom_load_balancing_policy.rs
```diff
@@ -56,23 +67,146 @@ impl LoadBalancingPolicy for CustomLoadBalancingPolicy {
...
+    // The policy above falls back to *every* node if it does not recognise the
+    // datacenter name, so a typo would make the rest of this example pass for
+    // the wrong reason. Compare the name against the topology the driver
+    // actually discovered before trusting anything below.
+    let cluster_state = session.get_cluster_state();
+    let datacenters: BTreeSet<&str> = cluster_state
+        .get_nodes_info()
+        .iter()
+        .filter_map(|node| node.datacenter.as_deref())
+        .collect();
```
wprzytula, line 123: ❓ Any reason why use `BTreeSet` instead of `HashSet`?

examples/ci/client_routes_cluster.rs
wprzytula, line 1: ❓ Why not reuse the existing NLB impl from scylla-proxy?

examples/README.md
```diff
@@ -0,0 +1,99 @@
...
+  examples ignore it, since they create their own cluster.
+
+## What each example teaches
+
+| Example | Teaches |
+| --- | --- |
+| `auth` | Connecting to a cluster that requires password authentication - and what a wrong password looks like. |
+| `basic` | The end-to-end basics: connect, create a keyspace and a table, insert, and read rows as tuples, as derived structs, or untyped. |
+| `batch` | Grouping `INSERT`/`UPDATE`/`DELETE` statements into one batch, what atomicity it does and does not buy, and why a batch is not a throughput tool. |
+| `client_routes` | Reaching nodes through the private endpoints they publish in `system.client_routes` instead of the addresses they gossip. |
```
wprzytula: 🔧 ScyllaDB doesn't gossip. It uses Raft for topology.

examples/ci/client_routes_cluster.rs
wprzytula, line 1: ❓ Why not reuse existing NLB impl from scylla-proxy?

examples/README.md
```diff
@@ -0,0 +1,99 @@
...
+| `compare_tokens` | Computing a partition key's token in the driver, checking it against the token the cluster reports, and listing the replicas that own it. |
+| `cql_time_types` | Reading and writing `date`, `time` and `timestamp` using `chrono`, `time`, and the raw `CqlDate`/`CqlTime`/`CqlTimestamp` types. |
+| `cqlsh_rs` | A miniature `cqlsh`: a REPL with CQL keyword completion that executes whatever you type. |
+| `custom_deserialization` | Implementing `DeserializeValue` by hand so that a column can be read directly into your own type. |
+| `custom_load_balancing_policy` | Writing a load balancing policy of your own - here one restricted to a single datacenter - and checking which node really coordinated each request. |
+| `custom_serialization` | Deriving `SerializeRow` for your own structs, including generic ones, and passing them as statement values. |
```
wprzytula, line 45: 🤔 Are these examples asymmetrical?
- `custom_deserialization` teaches `DeserializeValue`,
- `custom_serialization` teaches `SerializeRow`.

What about `SerializeValue` and `DeserializeRow`?
How is `user_defined_type` example related to this?

examples/README.md
```diff
@@ -0,0 +1,99 @@
...
+| `parallel` | Saturating the driver with bounded concurrent requests, using a prepared statement so every request is routed straight to a replica. |
+| `query_history` | Collecting the history of a request's execution: its attempts, retries and speculative executions. |
+| `schema_agreement` | Taking control of waiting for schema agreement instead of paying for the driver's automatic wait after every schema change. |
+| `select_paging` | Iterating over a large result with the automatic row stream, and driving paging by hand with `PagingState`. |
+| `speculative_execution` | Configuring a speculative execution policy so that one slow replica does not hold up an idempotent request. |
+| `tls_openssl` | Connecting to a cluster over TLS with the `openssl` backend. |
+| `tls_rustls` | Connecting to a cluster over TLS with the `rustls` backend. |
+| `tower` | Wrapping a `Session` in a `tower::Service`. |
+| `tracing` | Enabling CQL tracing for queries, prepares, executions, paged reads and batches, and fetching the resulting tracing information. |
+| `user_defined_type` | Mapping a CQL user-defined type onto a Rust struct with `SerializeValue`/`DeserializeValue`. |
```
wprzytula, line 61: 📌 I think we should group our examples into directories by topic.
This will also make it more discoverable what examples we're still missing.

Example topics:
- performance (parallel)
- marshalling/deser (custom deser, UDT)
- observability (tracing, logging, metrics, query_history)
- security (TLS, auth)
- configuration (exec profiles)
- load balancing (enforce coordinator, custom LBP, speculative execution?)

I don't yet know how to categorise batches, select paging, get by name, tower, schema agreement, cqlsh-rs, basic, client_routes, compare_tokens.

wprzytula: WDYT @Lorak-mmk? I can open an issue if you believe this is worth pursuing.

wprzytula (changes requested):
Good job! Let's have our examples finally runnable.

examples/README.md
Lorak-mmk: I don't think we have that many examples that it would help. It would also decrease ease of navigation imo, forcing users to navigate subdirectories.

wprzytula: > I don't think we have that many examples that it would help.

Really? Categories always help me when getting to know something.

examples/Cargo.toml
Lorak-mmk: Ok, but now that the book tests are executed as doctests, we no longer need it here, right?

examples/custom_load_balancing_policy.rs
Lorak-mmk: No reason. Changed to HashSet.

examples/README.md
Lorak-mmk: Please open an issue about improving them.

examples/auth.rs
Lorak-mmk: Not sure what you mean.
Each example creates the same KS (examples_ks). They use `CREATE KEYSPACE IF NOT EXISTS ` for that so it is idempotent.

Some examples also create a table, with a name unique to the example - and they also use `IF NOT EXIST` for that. 

After my changes, `auth` example no longer creates any table. I don't see the problem.

examples/schema_agreement.rs
Lorak-mmk: Yes.

Lorak-mmk:
```
cassandra@cqlsh> CREATE TABLE IF NOT EXISTS examples_ks3.schema_agreement (a int, b int, c text, primary key (a, b));
InvalidRequest: Error from server: code=2200 [Invalid query] message="Can't find a keyspace examples_ks3"
```

examples/Cargo.toml
wprzytula: Yes. Just wanted you to know. Also maybe you want to update the commit message.

examples/auth.rs
wprzytula: I wrongly claimed non-idempotence. What I meant was _cleaning up after oneself_. If an example creates a table and then does not drop it, it does not clean up.
What I also omitted was that the `auth` example no longer creates any table. This resolves my question.

examples/README.md
wprzytula: #1828

examples/README.md
wprzytula: #1829

examples/ci/client_routes_cluster.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// The endpoint registered for the node in `system.client_routes`.
+    pub fn routed_endpoint(&self) -> SocketAddr {
+        self.routed.listen_addr()
+    }
+
+    /// The address the node gossips, which no client here can reach.
+    pub fn gossiped_node_address(&self) -> &str {
+        &self.gossiped_node_address
+    }
```
wprzytula: 🔧 Again, nodes of ScyllaDB don't gossip.

wprzytula (approved):

examples/tls_rustls.rs
```diff
@@ -0,0 +1,74 @@
...
+            &[],
+        )
+        .await?
+        .into_rows_result()?;
+    for row in rows_result.rows::<(Option<bool>, Option<String>, Option<String>)>()? {
+        let (enabled, protocol, cipher) = row?;
+        println!(
+            "server sees a connection: ssl_enabled={enabled:?}, protocol={protocol:?}, cipher={cipher:?}"
+        );
+    }
```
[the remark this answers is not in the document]
wprzytula: Is this resolved?

examples/Cargo.toml
Lorak-mmk: Adjusted.

Lorak-mmk:
- Eradicated mentions of gossip
- Mentioned the historical reason for the `v1` feature in the commit message.

Lorak-mmk:
Rebased on main.

Lorak-mmk:
Rebased on main

PR #1815, opened 2026-08-03
Minimal metadata fetching

### LINKED ISSUE #1741: `fetch_schema_metadata` disabled doesn't work well with tablet-awareness (and token-awareness in general)
Schema metadata consists of 2 abstract parts:
- Metadata necessary for token-awareness to function fully correctly: paritioner of each table, replication strategy of each keyspace (for vnodes), list of tablet-based keyspaces (for tablets).
- Optional metadata: all the rest (for example column list of tables).

In the past I was under the impression that `fetch_schema_metadata` disables the optional part, but still fetches the mandatory part.
This turns out to not be the case. When `fetch_schema_metadata` is disabled, keyspace list and replication strategies are fetched, but table list is not (so partitioners are also not).

Note: there is also `keyspaces_to_fetch` which causes us to not download ANY info about other keyspaces - a nice tool for user if their app only accesses a subset of keyspaces.

For vnodes it will causes token-awareness to not work for tables not using Murmur3 (like CDC tables).
For tablets it is worse: we remove tablet info for tables not present in metadata, so with `fetch_schema_metadata == false`, tablet-awareness won't function at all.

I find it quite weird that with `fetch_schema_metadata == false` we fetch keyspaces, but not fetch tables list. Both of those are part of necessary metadata, so I see no reason to fetch one but not the other.

Proposal: the typical case I see for disabling `fetch_schema_metadata` is when user has no need to inspect the metadata in `ClusterState` - but they still expect driver to function well. For that reason, let's fetch table list (+ partitioners) when `fetch_schema_metadata` is disabled - just don't fetch column list, pk, ck, which is the heaviest part of metadata. Users can still disable this for unused keyspaces with `keyspaces_to_fetch`. If in the future someone reports a use case for which they need to disable everything, we can introduce such option.

patrycja-ziemkiewicz:
## Summary
This PR introduces the fetch_full_schema_metadata option to both SessionBuilder and SessionConfig. The option defaults to true and only takes effect when fetch_schema_metadata is enabled.

When full schema metadata fetching is disabled, the driver retrieves only the lightweight metadata required for query routing, such as table and view names, partitioners, and tablet information. Detailed schema data, including column definitions, partition keys, clustering keys, and user-defined types, is not fetched.

The behavior of fetch_schema_metadata(false) has also been updated. It now disables all schema metadata fetching, including keyspaces, tables, views, partitioners, and other schema information. As a result, token-awareness and tablet-awareness are unavailable in this mode. 

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1741

## Pre-review checklist

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -8,6 +8,7 @@ use super::{Compression, PoolSize, SelfIdentity, WriteCoalescingDelay};
 use crate::authentication::AuthenticatorProvider;
 use crate::client::client_routes::ClientRoutesConfig;
 use crate::client::execution::{RequestExecutionParams, RunRequestResult};
+use crate::cluster::metadata::{SchemaMetadataFetchLevel, SchemaMetadataFetchMode};
 use crate::cluster::node::KnownNode;
 use crate::cluster::{Cluster, ClusterNeatDebug, ClusterState};
 use crate::errors::DbError;
@@ -281,6 +282,11 @@ pub struct SessionConfig {
     /// If true, full schema is fetched with every metadata refresh.
     pub fetch_schema_metadata: bool,
 
+    /// If true, full schema details (columns, partition keys, clustering keys, UDTs) are fetched.
+    /// Only has effect when fetch_schema_metadata is true.
+    /// If false, only lightweight metadata needed for routing (table names, partitioners, tablet info) is fetched.
+    pub fetch_full_schema_metadata: bool,
+
     /// Custom timeout for requests that query metadata.
     pub metadata_request_serverside_timeout: Option<Duration>,
 
@@ -422,6 +428,7 @@ impl SessionConfig {
             timestamp_generator: None,
             keyspaces_to_fetch: Vec::new(),
             fetch_schema_metadata: true,
+            fetch_full_schema_metadata: true,
             metadata_request_serverside_timeout: Some(Duration::from_secs(2)),
             keepalive_interval: Some(Duration::from_secs(30)),
             keepalive_timeout: Some(Duration::from_secs(30)),
@@ -1157,11 +1164,20 @@ impl Session {
             }
         };
 
+        let schema_metadata_fetch_mode = match (
+            config.fetch_schema_metadata,
+            config.fetch_full_schema_metadata,
+        ) {
+            (false, _) => SchemaMetadataFetchMode::Disabled,
+            (true, false) => SchemaMetadataFetchMode::Enabled(SchemaMetadataFetchLevel::Minimal),
+            (true, true) => SchemaMetadataFetchMode::Enabled(SchemaMetadataFetchLevel::Full),
+        };
+
         let cluster = Cluster::new(
             known_nodes,
             pool_config,
             config.keyspaces_to_fetch,
-            config.fetch_schema_metadata,
+            schema_metadata_fetch_mode,
             config.metadata_request_serverside_timeout,
             config.hostname_resolution_timeout,
             config.host_filter,
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -60,7 +60,7 @@ use std::sync::{Arc, OnceLock};
 use std::time::Duration;
 use thiserror::Error;
 use tokio::time::timeout;
-use tracing::{Instrument, debug, error};
+use tracing::{Instrument, debug, error, warn};
 use uuid::Uuid;
 
 pub(crate) const TABLET_CHANNEL_SIZE: usize = 8192;
@@ -535,6 +535,15 @@ impl SessionConfig {
             return Err(NewSessionError::IllegalConfig(
                 "Keepalive interval must be non-zero".into(),
             ));
+        }
+
+        if !self.fetch_schema_metadata && self.fetch_full_schema_metadata {
+            warn!(
+                "Inconsistent schema metadata configuration: `fetch_schema_metadata` is `false`, \
+                         but `fetch_full_schema_metadata` is `true`. Because overall schema metadata fetching \
+                         is turned off, ALL schema metadata fetching is now disabled, and `fetch_full_schema_metadata` \
+                         will have no effect. Token-aware and tablet-aware routing will NOT function."
+            );
         }
 
         // Ensure no illegal configuration with Client Routes
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -279,7 +279,8 @@ pub struct SessionConfig {
     /// If empty, fetch all keyspaces
     pub keyspaces_to_fetch: Vec<String>,
 
-    /// If true, full schema is fetched with every metadata refresh.
+    /// If true, schema metadata is fetched with every metadata refresh.
+    /// If false, NO schema metadata is fetched at all (no keyspaces, no tables, nothing).
     pub fetch_schema_metadata: bool,
 
     /// If true, full schema details (columns, partition keys, clustering keys, UDTs) are fetched.
--- a/scylla/src/client/session_builder.rs
+++ b/scylla/src/client/session_builder.rs
@@ -986,6 +986,15 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
     /// Set the fetch schema metadata flag.
     /// The default is true.
     ///
+    /// When false, NO schema metadata is fetched - the driver will not
+    /// fetch keyspaces, tables, partitioners, or any other schema information.
+    /// Token-awareness and tablet-awareness will NOT function.
+    ///
+    /// This should ONLY be used in very specific scenarios where you want to
+    /// completely eliminate schema metadata queries and can accept degraded
+    /// driver performance. For most use cases, you should use
+    /// `fetch_full_schema_metadata(false)` instead.
+    ///
     /// # Example
     /// ```
     /// # use scylla::client::session::Session;
--- a/scylla/src/client/session_builder.rs
+++ b/scylla/src/client/session_builder.rs
@@ -1004,6 +1004,42 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
         self
     }
 
+    /// Set the fetch full schema metadata flag.
+    /// This only has effect when `fetch_schema_metadata` is true.
+    /// The default is true.
+    ///
+    /// When false, the driver fetches lightweight metadata necessary for proper
+    /// routing (table names, partitioners, tablet info) but skips heavy metadata
+    /// like column details, partition/clustering key information, and UDTs.
+    /// This significantly reduces memory usage and metadata fetch time while
+    /// maintaining full driver routing functionality.
+    ///
+    /// When true (default), the driver fetches complete schema information
+    /// including all columns, keys, and UDTs, which can be inspected via
+    /// ClusterState API.
+    ///
+    /// Note: When set to false, ClusterState::compute_token
+    /// and ClusterState::get_endpoints will not work.
+    ///
+    /// # Example
+    /// ```
+    /// # use scylla::client::session::Session;
+    /// # use scylla::client::session_builder::SessionBuilder;
+    /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let session: Session = SessionBuilder::new()
+    ///     .known_node("127.0.0.1:9042")
+    ///     .fetch_schema_metadata(true)
+    ///     .fetch_full_schema_metadata(false)  // Fetch only minimal metadata
+    ///     .build()
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn fetch_full_schema_metadata(mut self, fetch: bool) -> Self {
+        self.config.fetch_full_schema_metadata = fetch;
+        self
+    }
+
     /// Set the server-side timeout for metadata queries.
     /// The default is `Some(Duration::from_secs(2))`. It means that
     /// the all metadata queries will be set the 2 seconds timeout
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -32,8 +32,8 @@ use uuid::Uuid;
 
 use super::{
     CollectionType, Column, ColumnKind, ColumnType, Keyspace, MaterializedView, Metadata,
-    MissingUserDefinedType, NativeType, Peer, SingleKeyspaceMetadataError, Strategy, Table,
-    UserDefinedType,
+    MissingUserDefinedType, NativeType, Peer, SchemaMetadataFetchLevel, SchemaMetadataFetchMode,
+    SingleKeyspaceMetadataError, Strategy, Table, UserDefinedType,
 };
 
 use crate::DeserializeRow;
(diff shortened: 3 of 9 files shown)


scylla/src/client/session_builder.rs
```diff
@@ -986,6 +981,15 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
     ///
+    /// When false, NO schema metadata is fetched - the driver will not
+    /// fetch keyspaces, tables, partitioners, or any other schema information.
+    /// Token-awareness and tablet-awareness will NOT function.
+    ///
+    /// This should ONLY be used in very specific scenarios where you want to
+    /// completely eliminate schema metadata queries and can accept degraded
+    /// driver performance. For most use cases, you should use
+    /// `fetch_full_schema_metadata(false)` instead.
+    ///
```
Lorak-mmk, line 997: Commit: "update: docstring for fetch_schema_metadata"

This commit mentions `fetch_full_schema_metadata` but this option doesn't yet exist. The comment should be edited to mention this config when the config as actually added.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -1405,6 +1405,57 @@ impl ControlConnection {
...
+    #[expect(clippy::result_large_err)]
+    async fn query_tables_minimal(
+        &self,
+        keyspaces_to_fetch: &[String],
+        partitioners: &mut PerKsTable<Option<String>>,
+    ) -> Result<PerKeyspaceResult<PerTable<Table>, SingleKeyspaceMetadataError>, MetadataError>
+    {
+        let rows = self
+            .query_filter_keyspace_name::<(String, String)>(
+                "SELECT keyspace_name, table_name FROM system_schema.tables",
```
Lorak-mmk: Commit: "add: query_table_minimal helper method"

This commit adds a function that is never called. I would prefer it to be added in the same commit that starts using it to avoid having dead code.

If you insist on having it in separate commit, you should put `#[expect(dead_code)]` on it, otherwise clippy / check will complain, breaking your assertion that `All commits compile, pass static checks and pass test.`

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -1405,6 +1405,57 @@ impl ControlConnection {
...
+
+    /// Query table names only (without column details) and merge with partitioners.
+    /// This is used when fetch_full_schema_metadata is false.
+    #[expect(clippy::result_large_err)]
+    async fn query_tables_minimal(
+        &self,
+        keyspaces_to_fetch: &[String],
+        partitioners: &mut PerKsTable<Option<String>>,
+    ) -> Result<PerKeyspaceResult<PerTable<Table>, SingleKeyspaceMetadataError>, MetadataError>
+    {
```
Lorak-mmk: I wonder, does it have to be a separate function? Maybe its cleaner to use the original one and give it a parameter telling if it should fetch full metadata? It could result in less code duplication, but I did not really look into it to see if its feasible.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -1456,6 +1456,64 @@ impl ControlConnection {
...
+    async fn query_views_minimal(
+        &self,
+        keyspaces_to_fetch: &[String],
+        partitioners: &mut PerKsTable<Option<String>>,
+    ) -> Result<
+        PerKeyspaceResult<PerTable<MaterializedView>, SingleKeyspaceMetadataError>,
+        MetadataError,
+    > {
+        let rows = self
+            .query_filter_keyspace_name::<(String, String, String)>(
```
Lorak-mmk: Commit: add: query_views_minal helper method

My previous 2 comments apply here as well.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -1456,6 +1456,64 @@ impl ControlConnection {
...
+    /// This is used when fetch_full_schema_metadata is false.
+    #[expect(clippy::result_large_err)]
+    async fn query_views_minimal(
+        &self,
+        keyspaces_to_fetch: &[String],
+        partitioners: &mut PerKsTable<Option<String>>,
+    ) -> Result<
+        PerKeyspaceResult<PerTable<MaterializedView>, SingleKeyspaceMetadataError>,
+        MetadataError,
+    > {
```
Lorak-mmk: Commit: "add: query_views_minal helper method"

Typo. Minal -> minimal

scylla/src/client/session.rs
```diff
@@ -1165,6 +1165,7 @@ impl Session {
             pool_config,
             config.keyspaces_to_fetch,
             config.fetch_schema_metadata,
+            config.fetch_full_schema_metadata,
```
Lorak-mmk: For user-facing config we'll use 2 bools because of backwards compatibility. Internally, we can use a 3-variant enum to statically handle the one combination that doesn't make much sense.

scylla/src/client/session_builder.rs
```diff
@@ -1008,6 +1008,39 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
+    ///     .build()
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn fetch_full_schema_metadata(mut self, fetch: bool) -> Self {
+        self.config.fetch_full_schema_metadata = fetch;
+        self
+    }
+
```
Lorak-mmk, line 1051: Commit: add: fetch_full_schema_metadata method to SessionBuilder

Let's have this commit just after the one that adds the config value

Lorak-mmk (changes requested):

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -625,7 +634,16 @@ impl ControlConnection {
                     udts,
                 ))
             } else {
-                Ok((HashMap::new(), HashMap::new(), HashMap::new()))
+                // Fetch minimal schema: just table/view names and partitioners
+                // This is needed for token-awareness and tablet-awareness to work
+                let mut all_partitioners = self.query_table_partitioners().await?;
```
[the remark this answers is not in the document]
patrycja-ziemkiewicz: @Lorak-mmk , @wprzytula I used an existing function to fetch partitioners. I noticed that even in the existing code, when fetch_full_schema_metadata=true and fetch_schema_metadata=true, it already fetches all partitioners regardless of keyspaces_to_fetch. I wanted to check if fetching all partitioners is done on purpose here.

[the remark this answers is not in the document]
Lorak-mmk: I didn't look into this, but it sounds like a bug. You can add a commit that fixes it.

scylla/src/client/session_builder.rs
```diff
@@ -981,6 +981,15 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
     ///
+    /// When false, NO schema metadata is fetched - the driver will not
+    /// fetch keyspaces, tables, partitioners, or any other schema information.
+    /// Token-awareness and tablet-awareness will NOT function.
+    ///
+    /// This should ONLY be used in very specific scenarios where you want to
+    /// completely eliminate schema metadata queries and can accept degraded
+    /// driver performance. For most use cases, you should use
+    /// `fetch_full_schema_metadata(false)` instead.
+    ///
```
Lorak-mmk, line 997: I'm still a bit afraid there might be (vnodes) users who turn `fetch_schema_metadata` off to improve performance, and we will introduce regression for them.

Maybe we should print a warning if `fetch_schema_metadata` is disabled? Or at least if it is disabled while `fetch_full_schema_metadata` is enabled?

scylla/tests/integration/metadata/configuration.rs
```diff
@@ -303,8 +327,46 @@ async fn test_turning_off_schema_fetching() {
...
+        .ddl(
+            "CREATE TABLE table_a (
+                a int PRIMARY KEY,
+                b text
+            )",
+        )
+        .await
+        .unwrap();
+
+    session.refresh_metadata().await.unwrap();
```
Lorak-mmk: You didn't disable auto-refresh, so no need to call it manually - it is done automatically after each DDL.
If you want to refresh just once as an optimization, you need to disable metadata fetch on schema change in session builder.

Lorak-mmk (changes requested):

scylla/src/client/session_builder.rs
patrycja-ziemkiewicz: I believe warning when fetch_schema_metadata=false sounds reasonable, as token awareness and tablet awareness will not function then.

scylla/src/client/session_builder.rs
```diff
@@ -1009,6 +1009,14 @@ impl<K: SessionBuilderKind> GenericSessionBuilder<K> {
...
+            warn!(
+                "Disabling schema metadata fetching (`fetch_schema_metadata(false)`). \
+                No keyspace, table, or partitioner info will be retrieved. \
+                Token-awareness and tablet-awareness will NOT function, which may lead to degraded routing performance."
+            );
+        }
+
         self.config.fetch_schema_metadata = fetch;
         self
     }
```
Lorak-mmk, line 1014: This is a wrong place for such warning. Put it in `SessionConfig::validate`, and only warn on the "weird" case that suggests a user might encounter a regression: `fetch_schema_metadata` false and `fetch_full_schema_metadata` true.

Lorak-mmk (changes requested):

Lorak-mmk (approved):
Let's wait for @wprzytula before merging.

wprzytula:
@patrycja-ziemkiewicz Please rebase, then re-ask for my review.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -625,7 +634,16 @@ impl ControlConnection {
...
+                Ok::<_, MetadataError>((
+                    self.query_tables_minimal(keyspaces_to_fetch, &mut all_partitioners)
+                        .await?,
+                    self.query_views_minimal(keyspaces_to_fetch, &mut all_partitioners)
+                        .await?,
+                    HashMap::new(),
+                ))
             }
```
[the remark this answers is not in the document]
wprzytula: @patrycja-ziemkiewicz Is this true? What are you going to do about this?

scylla/src/client/session_builder.rs
wprzytula: > Maybe we should print a warning if `fetch_schema_metadata` is disabled? Or at least if it is disabled while `fetch_full_schema_metadata` is enabled?

The second idea is actually optimal: migrating users that have `fetch_schema_metadata` disabled will then get a notice that something's wrong with their configuration. At the same time, users who consciously disable `fetch_full_schema_metadata` too won't get the warning.

scylla/src/client/session.rs
```diff
@@ -1157,11 +1164,20 @@ impl Session {
             }
         };
 
+        let schema_metadata_fetch_mode = match (
+            config.fetch_schema_metadata,
+            config.fetch_full_schema_metadata,
+        ) {
+            (false, _) => SchemaMetadataFetchMode::Disabled,
```
wprzytula, line 1181: 🔧 Here I'd emit a warning if `(false, true)` is provided. This will inform migrating vnode users that they're going to lose token awareness.

scylla/src/cluster/metadata/mod.rs
```diff
@@ -32,6 +32,18 @@ use uuid::Uuid;
...
+pub(crate) enum SchemaMetadataFetchMode {
+    Disabled,
+    Enabled(SchemaMetadataFetchLevel),
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum SchemaMetadataFetchLevel {
+    Minimal,
+    Full,
+}
```
wprzytula, line 45: 💭 Wouldn't one 3-variant enum do?

scylla/src/client/session.rs
```diff
@@ -537,6 +537,15 @@ impl SessionConfig {
...
 
+        if !self.fetch_schema_metadata && self.fetch_full_schema_metadata {
+            warn!(
+                "Inconsistent schema metadata configuration: `fetch_schema_metadata` is `false`, \
+                         but `fetch_full_schema_metadata` is `true`. Because overall schema metadata fetching \
+                         is turned off, ALL schema metadata fetching is now disabled, and `fetch_full_schema_metadata` \
+                         will have no effect. Token-aware and tablet-aware routing will NOT function."
+            );
+        }
+
```
wprzytula, line 548: 💭 Aha, so you added a warning in `validate()` instead of the match where I suggested it. I guess it's OK.

scylla/tests/integration/metadata/configuration.rs
```diff
@@ -38,6 +39,29 @@ fn map_fedback_message<'a, T, F: Fn(RequestFrame) -> T + 'a>(
...
 }
 
+async fn setup_minimal_schema_metadata_session() -> (Session, String) {
+    let session = create_new_session_builder().build().await.unwrap();
+    let supports_tablets = scylla_supports_tablets(&session).await;
+    let session = create_new_session_builder()
+        .fetch_full_schema_metadata(false)
+        .build()
+        .await
+        .unwrap();
```
wprzytula, line 53: 🔧 Let's put a comment why those two Sessions are needed. Also, let's help future devs avoid costly debugging we had to do. For that, I'd add a warning in `scylla_supports_tablets`'s docstring that the function needs full schema metadata to function correctly.

scylla/src/cluster/metadata/fetching.rs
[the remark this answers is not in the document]
Lorak-mmk: Document that those APIs don't work in minimal mode.

scylla/tests/integration/metadata/configuration.rs
wprzytula: ❓ @Lorak-mmk Aren't there any driver-internal checks that require full schema metadata?

scylla/tests/integration/metadata/configuration.rs
```diff
@@ -318,8 +378,73 @@ async fn test_turning_off_schema_fetching() {
...
+    assert!(table.clustering_key.is_empty());
+    assert!(keyspace.user_defined_types.is_empty());
+
+    session.ddl(format!("DROP KEYSPACE {ks}")).await.unwrap();
+}
+
+#[tokio::test]
+async fn test_token_awareness_with_minimal_schema_metadata() {
+    setup_tracing();
+    let (session, ks) = setup_minimal_schema_metadata_session().await;
```
wprzytula, line 398: 💭 This tests vnodes-based token awareness. Is tablet-based token awareness also verified when full schema metadata is not fetched?

scylla/tests/integration/metadata/configuration.rs
Lorak-mmk: WDYM? You mean like tablet maintenance? I am not aware of any driver-internal process that needs full metadata.

scylla/tests/integration/metadata/configuration.rs
patrycja-ziemkiewicz: Yes it is also verified https://github.com/scylladb/scylla-rust-driver/pull/1815/changes#diff-4d95cb51059316ccce41b95ae97aaf12fcf4ffd65df692780af6483345ae5eb3 here.

scylla/src/cluster/metadata/mod.rs
patrycja-ziemkiewicz: Separating them allows us to match on Enabled(_) for operations shared between Minimal and Full, while completely bypassing schema metadata logic when Disabled. This avoids duplicating match arms and handling unreachable branches.

wprzytula (approved):

PR #1821, opened 2026-08-05
Improve handling Scylla versions for tests

Lorak-mmk:
Recently @wprzytula discovered an issue: when version provided to CCM to start a cluster is not full (like `release:2026.1` instead of `release:2026.1.5`), then EACH ccm invocation will contact AWS to resolve the full version, slowing down tests.

We have other problems with version handling:
- docker compose uses `:latest` tag, so we have no control when the version is bumped in CI
- docker compose uses `:latest` tag, so contributors stay on old versions unless they `docker pull` manually.
- Version used by ccm is not kept in sync at all, and so it drifts away with time.

This PR fixes those problems by introducing `scylla_version.env` file, which is used as a default both by docker and ccm.
Version in this file will be kept up to date by Renovate.

## Pre-review checklist


renovate.jsonc
```diff
@@ -10,6 +10,21 @@
...
+                "/^scylla_version\\.env$/"
+            ],
+            "matchStrings": [
+                "(?m)^SCYLLA_VERSION=(?<currentValue>\\d+\\.\\d+\\.\\d+)$"
+            ],
+            "depNameTemplate": "scylladb/scylla",
+            "datasourceTemplate": "docker",
+            "versioningTemplate": "semver"
+        }
+    ],
```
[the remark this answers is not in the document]
Lorak-mmk: I have no idea what you mean, it is inside the root. I don't see anything wrong with it.

CONTRIBUTING.md
```diff
@@ -39,11 +39,15 @@ Most of them live in `scylla/tests/integration`, but there are still a few lefto
...
 Fortunately there is no need to invoke `docker compose` manually, everything can be handled by our `Makefile`.
 
-To run a cargo test suite, use the command below (note that you must have Docker, Docker Compose V2, and [cargo-nextest](https://nexte.st/) installed):
+To run a cargo test suite, use the command below (note that you must have Docker, Docker Compose V2 (at least v2.20), and [cargo-nextest](https://nexte.st/) installed):
 ```bash
 make test
 ```
 
+The ScyllaDB version used by the `docker compose` cluster is pinned in `scylla_version.env` in the repository root, and Renovate keeps it up to date.
+It must be a full, three-component version (e.g. `2026.2.2`) rather than a truncated one (e.g. `2026.2`).
```
wprzytula: 🔧 Let's provide rationale for this guideline.

wprzytula: I see it's written in `scylla_version.env`, but I believe it should be present in `CONTRIBUTING.md` - also there or only there, as you prefer.

scylla-ccm-bridge/src/lib.rs
```diff
@@ -21,12 +21,71 @@ use futures::FutureExt;
...
+            })
+        });
+    assert!(
+        valid,
+        "SCYLLA_VERSION in scylla_version.env must be a full three-component numeric version \
+         (e.g. `2026.2.2`), got `{version}`. A truncated version such as `2026.2` is accepted \
+         by ccm, but it makes every single ccm invocation - even against an already-running \
+         cluster - query AWS to resolve the latest patch number, which slows the tests down \
+         badly.",
+    );
```
wprzytula, line 84: 💭 This won't accept release candidate versions (e.g., `rc2`). Is this intentional?

CONTRIBUTING.md
```diff
@@ -44,9 +44,10 @@ To run a cargo test suite, use the command below (note that you must have Docker
 make test
 ```
 
-The ScyllaDB version used by the `docker compose` cluster is pinned in `scylla_version.env` in the repository root, and Renovate keeps it up to date.
+The ScyllaDB version used for testing is pinned in `scylla_version.env` in the repository root, and Renovate keeps it up to date.
+That single file pins the version for both the `docker compose` cluster and the ccm tests.
 It must be a full, three-component version (e.g. `2026.2.2`) rather than a truncated one (e.g. `2026.2`).
-To use a different version ad hoc, set `SCYLLA_VERSION` in your environment - it takes precedence over the value from the file.
+To run the `docker compose` cluster on a different version ad hoc, set `SCYLLA_VERSION` in your environment - it takes precedence over the value from the file.
```
wprzytula: ❓ Why does this change belong to commit `test(ccm): take the default cluster version from scylla_version.env`?

renovate.json
```diff
@@ -10,6 +10,21 @@
...
+                "/^scylla_version\\.env$/"
+            ],
+            "matchStrings": [
+                "(?m)^SCYLLA_VERSION=(?<currentValue>\\d+\\.\\d+\\.\\d+)$"
+            ],
+            "depNameTemplate": "scylladb/scylla",
+            "datasourceTemplate": "docker",
+            "versioningTemplate": "semver"
+        }
+    ],
```
wprzytula, line 28: 🔧 I'd appreciate comments explaining what each setting do.
❓ Does `"semver"` `"versioningTemplate"` restrict to semver-compatible tags, or is it the regex in `"matchStrings"` that achieves that goal?
❓ Can we actually put comments in this file? I see you previously put one with `//`, but this is .json, not .jsonc. Will Renovate handle this correctly? The comment is coloured red in GitHub, which suggests it's bad. Comments in other files get coloured grey.

scylla-ccm-bridge/src/lib.rs
Lorak-mmk: Yes. I think we should only use stable releases in CI. For one-time manual testing of some rc we can run tests manually.

renovate.json
Lorak-mmk: > ❓ Can we actually put comments in this file? I see you previously put one with //, but this is .json, not .jsonc. Will Renovate handle this correctly?

We can, and it will. I renamed to .jsonc to prevent confusion.

Lorak-mmk: > ❓ Does "semver" "versioningTemplate" restrict to semver-compatible tags, or is it the regex in "matchStrings" that achieves that goal?

AFAIU "versioningTemplate": "semver" prevents bumping to tags other than full versions. matchStrings informs renovate which part of the file contains the version number to be replaced.

wprzytula (approved):
🚀 Great job! Let's leverage Renovate wherever possible.

Issue #1670, opened 2026-06-06
ClusterWorker and the control connection can deadlock on high volume of CQL events


piodul:
The two key facts are:

1. When `Connection::reader` receives a CQL event, it immediately pushes it to the `event_sender` channel if it has one. This channel is set for control connections and it has limited capacity of 32 entries.
2. The `ClusterWorker::work` is implemented as a loop that waits for one for some events - including the CQL events from the control connection, but also explicit requests to fetch metadata. After receiving some event it may decide to refresh the cluster's metadata before starting to wait for the next event.

The following deadlock can occur if the cluster is hammered with schema changes:

- A user request to refresh metadata arrives,
- The cluster worker picks the request from the channel user requests channel,
- 33 more schema change events arrive, causing the `Connection::reader` for the control connection to block on sending the event to the channel,
- The cluster worker proceeds to refresh metadata. Refreshing metadata works by sending queries on the control connection. However, because the `Connection::reader` future is stuck on the control connection, it cannot read the responses to the queries from the socket and thus cannot forward them to the cluster worker that is waiting for them.

As a result:

- The control connection deadlocks because the event channel cannot be drained by the cluster worker,
- The cluster worker cannot drain the event channel because it is stuck waiting for responses to the metadata queries, which themselves are stuck on the control connection.

The bug reproduces on my machine if I run the whole test suite, but I believe I wrote a reliable reproducer (which requires placing a strategic sleep):

```diff
diff --git a/scylla/src/cluster/worker.rs b/scylla/src/cluster/worker.rs
index 65a092af..b0067465 100644
--- a/scylla/src/cluster/worker.rs
+++ b/scylla/src/cluster/worker.rs
@@ -453,6 +453,9 @@ impl ClusterWorker {
             // Perform the refresh
             debug!("Requesting metadata refresh");
             last_refresh_time = Instant::now();
+
+            tokio::time::sleep(Duration::from_secs(5)).await;
+
             let refresh_res = self.perform_refresh().await;
 
             // Send refresh result if there was a request
diff --git a/scylla/tests/integration/session/internal_requests.rs b/scylla/tests/integration/session/internal_requests.rs
index de557e98..23697e34 100644
--- a/scylla/tests/integration/session/internal_requests.rs
+++ b/scylla/tests/integration/session/internal_requests.rs
@@ -11,7 +11,10 @@ use scylla_proxy::{
 use tokio::sync::mpsc;
 use tracing::info;
 
[...]

Issue #1410, opened 2025-07-30
ClusterWorker: Tablets updating can be delayed by metadata fetching


Lorak-mmk:
ClusterWorker process messages from several sources, one at a time.
One of them is that a new tablet was received in a response custom payload. Processing this is quite simple: we clone ClusterState, edit its tablet info, and replace the old state. It should be quick, and does not perform IO.
Another message type is use_keyspace request. Processing it means setting one field, and spawning a task - again quick and easy.
Then there are messages related to metadata.  Processing them can mean fetching metadata (possibly long), or even trying to reconnect CC. During this time, tablets / use_keyspace cannot be processed.

One solution that came to my mind is to have 2 workers. One will be MetadataWorker. It will own Control Connection, and take care of fetching new metadata, which it will then send to ClusterWorker.
ClusterWorker will be responsible for:
- Receiving Tablet info and updating ClusterState with it
- Processing use_keyspace requests
- Transform Metadata into ClusterState, and update it

This unfortunately doesn't completely solve the issue - transforming Metadata into ClusterState is also a potentially long operation (it starts connection pools and waits for them to connect), so it could still prevent tablet / use_keyspace from being executed.
We also can perform this transformation on MetadataWorker easily - resulting ClusterState would get out of sync in regards to current keyspace / tablet info.,

Issue #1793, opened 2026-07-17
Add `ControlConnectionQueryFallback` option


patrycja-ziemkiewicz:
### Summary

Add an option to allow routing application queries through the control connection when node connection pools are unavailable, similar to the Python driver's `ControlConnectionQueryFallback`.

### Background

The Python driver provides a `ControlConnectionQueryFallback` setting on the `Cluster` object with three modes:

- **`Disabled`** (default) - requires a usable node pool for application queries. If none is available, the driver raises `NoHostAvailable`.
- **`Fallback`** - still attempts to create node pools, but allows application queries to fall back to the control connection when no usable pool is available. Session startup proceeds even if all initial pool attempts fail.
- **`SkipPoolCreation`** - disables node-pool creation entirely and routes all application queries through the control connection.

### Current state in the Rust driver

The Rust driver's `ControlConnection` is used exclusively for metadata fetching and event listening. There is currently no way to route user queries through it when connection pools are unavailable, the driver will return a connection error instead.

### References

- Python driver implementation: `cassandra.cluster.ControlConnectionQueryFallback` ([source](https://github.com/scylladb/python-driver/blob/master/cassandra/cluster.py))
- Related UDS discussion: being able to skip pool creation is a prerequisite for single-connection usage over Unix domain sockets : https://github.com/scylladb/scylla-rust-driver/issues/1616

Lorak-mmk:
In Rust Driver I would take a different approach than in Python, specifically lets skip the "Fallback" mode. I think using it is invalid most of the time, or even always.

A single-node session (where only CC address works) is only feasible for workloads where performance doesn't matter, like CQLSH, some heartbeats, DDLs, fetching schema.
If you need performance, you need normal session, and CC fallback is useless.
If you don't need performance, then using the simplified session makes more sense (it should connect faster, use less resources, have less problems).

Another observation: Unix proxy only makes sense for single-node session.

Because of all that, one idea I have for public API is simply a new field in SessionConfig, `maintenance_mode`, which is either `None`, or `Some` storing a single contact point (in which case normal contact points must be empty). This single contact point may be a unix socket (in addition to IP address / DNS name).
For implementation, we can either share a single CC (more difficult) or have a CC + a NodeConnectionPool opened to the same address as CC (imo simpler and good enough for now).

You'll still need to figure out many details, I'm sure @wprzytula will help.

patrycja-ziemkiewicz:
## Notes on possible approaches

[...]

PR #1823, opened 2026-08-06
Split metadata worker

Lorak-mmk:
This PR splits existing ClusterWorker into 2 workers. The new one is called MetadataWorker.
This is a prerequisite (and I think the last prerequisite!)  for fixing our issues with metadata fetching and other internal processes.

# Why?

I spent A LOT of time thinking about refactoring the ClusterWorker to make the internal tasks non-blocking ( ref: https://github.com/scylladb/scylla-rust-driver/issues/1670 ), enabling smarter metadata fetching ( ref: https://github.com/scylladb/scylla-rust-driver/issues/1280 ) and removing unnecessary delays (ref: https://github.com/scylladb/scylla-rust-driver/issues/1410 ). I thought about this a lot (both myself and with various AIs), made some PoCs (again, both myself and with AIs).

My conclusion is that it all can be done, but writing code this way is inherently complicated, and makes ClusterWorker mostly incomprehensible. I certainly wouldn't be sure of its correctness.

How to improve on this? We can notice that there are 2 categories of tasks that ClusterWorker performs. Those that require a working control connection (waiting on events, waiting on broken CC notification, fetching metadata, handling metadata refresh requests) and those that don't (handling connectivity events, handling tablets feedback, creating ClusterState, handling use_keyspace).
We can split those tasks into 2 workers, where the CC worker passes its results to ClusterWorker using a channel.

After such split, we can work on improving each worker independently, and each of them is easier to comprehend (because the combinatorial explosion of task interactions is limited).

The independence mentioned above is quite important: we can now introduce smart metadata fetching and make MetadataWorker non-blocking without even touching ClusterWorker and thinking about interactions between its tasks.

# Implementation

MetadataWorker communicates with ClusterWorker in exactly one way: by providing MetadataUpdate on a newly introduced merge channel. Merge channel is SPSC channel holding at most one value, introduced in this PR.
The idea is that we don't control the time when ClusterWorker consumes produced MetadataUpdates, and we don't want to `await` on `send`, so we do something else: merge the metadata updates. This turns out to be quite easy, and solves the problem nicely.

The only slightly complicated part is handling refresh requests: MetadataWorker is the one receiving them, then after fetching new metadata it passes them to ClusterWorker in the MetadataUpdate, and ClusterWorker responds to them after publishing new ClusterState.

Initially, MetadataWorker is created in a way that minimizes code changes. Later it is refactor to a form that I always imagined for it: with a `work_on_cc` method that takes a CC, and performs work on it (handling events etc), until it breaks. 

# Commits

First 3 commits introduce some new code, that is initially mostly dead.
I am not completely happy with doing it that way, but I don't really see any alternatives, and the pieces of code (ClientRoutesUpdate, merge channel, MetadataUpdate) are self-contained enough to be reviewed independently.

Then we have the main commit ("Split off MetadataWorker") that splits the worker. It is written in a way that minimizes code changes: new worker looks very similar to the old one.

Then there are 2 last commits that refactor the new worker to get it into the shape that I want it in. The main goal of this is having separate loops for the case where we have working CC and the one without. That way in the main loop we always have working CC, getting rid of the Option.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/1670
Ref: https://github.com/scylladb/scylla-rust-driver/issues/1410 (partially solved by this PR!)

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/metadata/merge_channel.rs
+++ b/scylla/src/cluster/metadata/merge_channel.rs
@@ -1,0 +1,292 @@
+#![allow(dead_code)] // `expect` would need to be put in a lot of places.
+//! A single-producer single-consumer, capacity-one, merge-on-send channel.
+//!
+//! This is the primitive over which the metadata worker (producer) talks to the
+//! cluster worker (consumer). The consumer can be busy for a long time - it
+//! builds a new `ClusterState` and awaits connection pools - while the producer
+//! keeps generating updates: freshly fetched metadata, partial client-routes
+//! snapshots, node UP/DOWN hints, replies owed to explicit refresh requests.
+//!
+//! Queueing all of those would make the consumer apply stale intermediate
+//! states one after another; dropping them would lose information. Instead,
+//! this channel holds *at most one* in-flight value and hands the producer
+//! mutable access to it, so that a new update can be **merged** into the
+//! pending one ([`Sender::modify`]).
+//!
+//! The single-producer/single-consumer discipline is enforced statically: both
+//! [`Sender::modify`] and [`Receiver::recv`] take `&mut self`, and neither
+//! endpoint is `Clone`.
+//!
+//! # Implementation notes
+//!
+//! - The slot is guarded by a [`std::sync::Mutex`], not a `tokio` one: the
+//!   critical section is purely synchronous (it never awaits). Lock poisoning
+//!   is handled the same way as elsewhere in this crate - with `unwrap()`. The
+//!   only code inside the critical section that can panic is the
+//!   caller-provided merge closure; the producer of this channel is a driver
+//!   worker task, so a panicking closure means the worker is already dead and
+//!   the session is unusable. Propagating the poison as a panic to the consumer
+//!   is then strictly better than silently continuing with a possibly
+//!   half-merged update.
+//! - Wakeups go through [`tokio::sync::Notify`].
+//! - Endpoint liveness is tracked with explicit `AtomicBool` flags rather than
+//!   `Arc::strong_count`. `Arc`'s refcount is only decremented *after* our
+//!   `Drop` impl returns, so a receiver woken from `Drop for Sender` would
+//!   still observe a strong count of 2 and park again, losing the wakeup
+//!   forever. An explicit flag set *before* notifying has no such race.
+
+use std::sync::Arc;
+use std::sync::Mutex;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use thiserror::Error;
+use tokio::sync::Notify;
+
+/// State shared by the two endpoints of the channel.
+struct Shared<T> {
+    /// The single in-flight value, if any.
+    slot: Mutex<Option<T>>,
+    /// Signals "the slot may have changed" or "the sender was dropped".
+    notify: Notify,
+    /// Set by `Drop for Sender`, before notifying the receiver.
+    sender_dropped: AtomicBool,
+    /// Set by `Drop for Receiver`.
+    receiver_dropped: AtomicBool,
+}
+
+/// The producing endpoint of a [`merge_channel`].
+pub(crate) struct Sender<T> {
+    shared: Arc<Shared<T>>,
+}
+
+/// The consuming endpoint of a [`merge_channel`].
+pub(crate) struct Receiver<T> {
+    shared: Arc<Shared<T>>,
+}
+
+/// Returned by [`Sender::modify`] when the [`Receiver`] has been dropped,
+/// so that no update can ever be observed again.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Error)]
+#[error("the receiving end of the channel has been dropped")]
+pub(crate) struct SendError;
+
+/// Creates a capacity-one, merge-on-send channel.
+pub(crate) fn merge_channel<T>() -> (Sender<T>, Receiver<T>) {
+    let shared = Arc::new(Shared {
+        slot: Mutex::new(None),
+        notify: Notify::new(),
+        sender_dropped: AtomicBool::new(false),
+        receiver_dropped: AtomicBool::new(false),
+    });
+    (
+        Sender {
+            shared: Arc::clone(&shared),
+        },
+        Receiver { shared },
+    )
+}
+
+impl<T> Sender<T> {
+    /// Merges an update into the pending value.
+    ///
+    /// `f` is applied to the currently pending value, which is `None` if the
+    /// receiver already took the previous one (or if nothing was sent yet).
+    ///
+    /// If, after `f` returns, the slot holds a value, the receiver is notified.
+    /// If `f` leaves the slot as `None` - i.e. it decided there is nothing to
+    /// send, or it cancelled a previously pending update - no notification is
+    /// emitted: there would be nothing for the receiver to take, and a spurious
+    /// wakeup would only make it re-check the slot and park again.
+    ///
+    /// Returns [`SendError`] if the receiver has been dropped. In that case `f`
+    /// is not applied at all.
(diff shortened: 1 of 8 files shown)


Lorak-mmk:
@patrycja-ziemkiewicz I requested your review as well because:
 - In my opinion reviewing code is really good way of learning, and you don't really have much opportunity for that when working on Python-rs
 - This PR significantly changes the code that touches your task ( https://github.com/scylladb/scylla-rust-driver/issues/1793 ). I guess your task is a bit less urgent now that we aim to start migration with test.py instead of SCT, but its worth completing anyway for the sake of cqlsh-rs. I think my change should only make yours easier, because ClusterWorker will no longer need any modifications, because it will no longer operate on control connection.

If you don't have time to review, its fine - we can merge without it. Don't sacrifice other tasks for it, treat it as lower priority.

scylla/src/cluster/metadata/worker.rs
```diff
@@ -0,0 +1,358 @@
...
+                        .take()
+                        .map(|request| request.response_chan);
+                    if self
+                        .send_update(|slot| {
+                            MetadataUpdate::merge_metadata(slot, metadata, response_chan)
+                        })
+                        .is_err()
+                    {
+                        return ControlFlow::Break(());
+                    }
```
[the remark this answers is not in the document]
Lorak-mmk: This may be true, but I'm not sure how to fix it. If send_update fails, then ClusterWorker is already finished:
- Either because user dropped the Session, in which case dropping the response channel is fine because no-one waits for it.
- Or because Tokio runtime is shutting down, and ClusterWorker was dropped before us. In such case it is possible that `Session` still exists, and there is someone waiting for response, so the panic would really happen.

I don't see how to get the ownership of response_chan back, nor do I see an appropriate error. WDYT @wprzytula ?

scylla/src/cluster/metadata/update.rs
```diff
@@ -0,0 +1,458 @@
...
+    /// `WHERE connection_id IN ? AND host_id IN ?`, so it returns the full cross product,
+    /// while the event's semantics is a zip of both lists.
+    pub(crate) fn from_event(
+        event: &ClientRoutesChangeEvent,
+        relevant_connection_ids: &[String],
+        fetched: &ClientRoutes,
+    ) -> Self {
+        #[deny(clippy::wildcard_enum_match_arm)]
+        let (connection_ids, host_ids) = match event {
+            ClientRoutesChangeEvent::UpdateNodes {
```
wprzytula, line 168: 📌 WDYT about enabling this lint for the whole codebase? Then using wildcard enum match will require explicit `allow`.

wprzytula: cc @patrycja-ziemkiewicz @hmarszalek - we can do the same in the wrapper drivers.

scylla/src/cluster/metadata/merge_channel.rs
```diff
@@ -0,0 +1,299 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct SendError;
+
+impl Display for SendError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str("the receiving end of the channel has been dropped")
+    }
+}
+
+impl std::error::Error for SendError {}
```
wprzytula: ⛏️ Why not derive `thiserror::Error`, just like we do everywhere else?

scylla/src/cluster/metadata/merge_channel.rs
wprzytula, line 1: 💯 Brilliant work!

scylla/src/cluster/metadata/update.rs
```diff
@@ -1,10 +1,143 @@
...
+use crate::cluster::metadata::{ClientRoute, ClientRoutes, Metadata};
+use crate::errors::MetadataError;
 use crate::frame::response::event::ClientRoutesChangeEvent;
 
+/// An explicit request to refresh cluster metadata, sent by
+/// [`Cluster::refresh_metadata`](crate::cluster::worker::Cluster::refresh_metadata).
+#[derive(Debug)]
+pub(in super::super) struct RefreshRequest {
+    pub(in super::super) response_chan: tokio::sync::oneshot::Sender<Result<(), MetadataError>>,
+}
```
wprzytula, line 17: ⛏️ Let's import `tokio::sync::oneshot` to reduce verbosity.

scylla/src/cluster/metadata/worker.rs
```diff
@@ -0,0 +1,271 @@
...
+use super::merge_channel;
+use super::reader::MetadataReader;
+
+/// Works in the background to keep learning about the cluster: fetches metadata
+/// periodically and on request, and drains CQL server events from the control
+/// connection.
+///
+/// See the [module-level documentation](self) for the division of
+/// responsibilities between this worker and the cluster worker.
+pub(in super::super) struct MetadataWorker {
```
wprzytula: 🔧 I don't see any module-level documentation in this commit (`Split off MetadataWorker`). Perhaps it's added later.

scylla/src/cluster/worker.rs
```diff
@@ -518,72 +401,95 @@ impl ClusterWorker {
...
-        let client_routes_updated_hosts = match (
-            self.client_routes_subscriber.as_ref(),
-            metadata.client_routes.take(),
-        ) {
-            (Some(subscriber), Some(routes)) => subscriber.replace_client_routes(routes),
-            _ => Default::default(),
-        };
-
+        metadata: Metadata,
+        process_up_hints: impl FnOnce(&ClusterState),
```
wprzytula: ❓ What are `hosts_to_refill` and `up_hints` about? I don't see such names. Is this comment a leftover?

wprzytula (commented):
🚀 AWESOME WORK!

scylla/src/cluster/worker.rs
Lorak-mmk: Oh yes, sorry about that. I modifed this code quite a bit since the comment was written.

scylla/src/cluster/metadata/worker.rs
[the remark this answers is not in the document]
wprzytula: Why not respond with `Ok(())` in this case? This seems benign in both situations.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: I executed `cargo clippy --all-targets --workspace -- -Wclippy::wildcard_enum_match_arm` and there are quite a lot of those warnings. We can consider it for production code imo, not tests.

scylla/src/cluster/metadata/worker.rs
Lorak-mmk: Heh, I removed the AI-generated comment because it was trash, and then forgot to write one myself.

scylla/src/cluster/metadata/worker.rs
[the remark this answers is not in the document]
Lorak-mmk: This is a pre-existing bug, opened https://github.com/scylladb/scylla-rust-driver/issues/1824

scylla/src/cluster/metadata/worker.rs
Lorak-mmk: I don't think more comment is needed. Edited this section accordingly.

wprzytula: > See the [module-level documentation](self) for the division of responsibilities between this worker and the cluster worker.

This is still present, and there's still no documentation of this module.

wprzytula (changes requested):

Lorak-mmk:
Adjusted the doc comment.

wprzytula (approved):

PR #1830, opened 2026-08-08
Migrate single connection pager to the unified execution core

### LINKED ISSUE #1549: Deduplicate request execution paths

wprzytula:
Fixes: #1549

## Summary

This PR concludes the unification of execution paths by migrating the `SingleConnectionPagingExecutor` to the unified execution core. There are no behavioral changes intended, other than arising from switching the logic to the unified core.

The key notion of this PR is _genericisation_.

The APIs exposed by the `execution` module had the following constraints:
- Metrics must be passed to REP (for use in speculative execution).
- LBP must be passed to REP (`RequestExecutionParams`) and available for providing feedback (`on_request_success` and `on_request_failure`).
- The _query plan_ must be an iterator over `(NodeRef, Shard)` targets.
- `Coordinator` must be returned from a successful execution.

This PR lifts those limitations by:
- making `Metrics` optional and only (statically) required if SEP is passed.
- making both `Target` and `Coordinator` generic.

Extensive testing for `Connection::execute_iter()` API is added. 

## Details

<details><summary>Detailed description by Opus 5</summary>
<p>

# Genericising the unified execution core

## Background

Request execution — load balancing, retries, speculative execution, history,
metrics, client-side timeout — used to be implemented independently in several
places: `Session`'s non-paged methods, the `QueryPager` worker, and the
control-connection paging path (`SingleConnectionPagingExecutor`). Earlier work
collapsed the first two into a single layered core in
`scylla/src/client/execution.rs`:

- `RequestExecutionParams::run_request_no_side_effects` — no `Session` access,
  no side-effect handling; applies the client-side timeout and dispatches
  speculative-execution fibers.
- `RequestExecutionParams::run_request_speculative_fiber` — a single fiber:
  iterates the execution plan, picks connections, runs the per-attempt closure,
  applies the retry policy.
- the outermost `run_request` layer, which handles `USE <keyspace>` and
  schema-agreement side effects, stays on `Session` / `PagingExecutor` because
  it needs session state.

`SingleConnectionPagingExecutor` could not join that core, because the core was
hardwired to load balancing: the plan was an `Iterator<Item = (NodeRef, Shard)>`,
connections came from `node.connection_for_shard(shard)`, every attempt fed
success/failure back into an LBP, every attempt produced a `Coordinator`, and
metrics were mandatory. The control connection has none of those: it has one
fixed connection, no `Node`, no LBP, and no metrics sink. So it carried its own
duplicate loop — its own timeout handling, its own error mapping, its own
worker.

This PR removes each of those hardwired assumptions, one per commit, and then
migrates the executor.

## The central abstraction: `AttemptTarget`

The key generalisation is that "a thing an attempt can be sent to" is now a
trait rather than a `(NodeRef, Shard)` tuple:

```rust
pub(crate) trait AttemptTarget {
    type Coordinator;
    async fn get_connection(&self) -> Result<Arc<Connection>, ConnectionPoolError>;
    fn coordinator(&self, connection: &Arc<Connection>) -> Self::Coordinator;
    fn on_attempt_success(&self, elapsed: Duration);
    fn on_attempt_failure(&self, elapsed: Duration, error: &RequestAttemptError);
}
```

An execution plan is now simply an `Iterator<Item = Target>`. The fiber loop is
unchanged in structure — `'targets_in_plan` / `'same_target_retries`, connection
acquisition failure skips to the next target, retry decisions map onto the two
loop labels — but every LBP-specific operation is delegated to the target.

Two implementations:

- `NodeAttemptTarget<'a>` — a `(node, shard)` pair plus the `&dyn
  LoadBalancingPolicy` and `&RoutingInfo` needed for feedback. Acquires a
  connection via `node.connection_for_shard(shard)`, produces a real
  `Coordinator`, and forwards success/failure to the policy. Call sites build it
  by mapping over the load balancing plan:
  `plan.map(|(node, shard)| NodeAttemptTarget::new(node, shard, lbp, routing_info))`.
- `SingleConnectionTarget` — wraps one `Arc<Connection>`. `get_connection`
  always succeeds, `Coordinator = ()`, and both feedback hooks are no-ops. The
  plan is `std::iter::once(...)`.

Because `Coordinator` is an associated type, "no coordinator" is expressed
statically as `()` rather than as a runtime `Option` the caller must handle.
`run_request_no_side_effects` now returns `RequestExecutionOutcome<Target::Coordinator>`
— a named struct replacing the previous `(RunRequestResult<_>, Coordinator)`
tuple, which had become awkward once the second field's type varied.

## Supporting decouplings

Each of these existed only to let the core stop assuming a load-balanced,
metric-ed, session-backed request.

**LBP out of `RequestExecutionParams`.** The load balancing policy and
`RoutingInfo` are no longer fields of the params struct; they moved into
`NodeAttemptTarget`, where they belong. `RequestExecutionParams::new` therefore
returns `(Self, &dyn LoadBalancingPolicy)` — callers that do load balancing take
the policy and build targets with it; the control-connection path simply never
sees one.

**Metrics made optional, tied to speculative execution.** The two fields became
one:

```rust
metrics_and_speculative_policy: Option<(&Arc<Metrics>, Option<&dyn SpeculativeExecutionPolicy>)>,
```

This is not just optionality — it statically encodes the invariant that
speculative execution cannot be requested without a metrics sink (the
speculative-execution machinery requires one). The control connection passes
`None` and all the `inc_*` / `log_query_latency` helpers become no-ops.

**`RetrySession` moved onto `ExecuteRequestContext`.** The retry session used to
be created inside the fiber. It is now constructed per fiber at the call site
and threaded through the context, which keeps the fiber body free of policy
construction and makes the per-fiber ownership explicit.

**Statically optional `Coordinator`.** Groundwork for `Target::Coordinator`:
the coordinator became a type-level rather than value-level option across the
execution result types.

## Migrating `SingleConnectionPagingExecutor`

With the above in place, the executor drops its bespoke loop and becomes a thin
layer over the core.

Structurally, it stops storing a boxed `fetcher` closure and instead stores the
request's constituents (`connection`, `prepared`, `values`, `page_size`,
`consistency`, `serial_consistency`, `request_timeout`), so each page can build
its own `run_request_once` closure.

`fetch_one_page` constructs `RequestExecutionParams` with:
- `retry_policy: &FallthroughRetryPolicy` — the control-connection path has
  never retried; a fallthrough policy makes every attempt terminal, preserving
  behaviour while going through the shared retry machinery;
- `metrics_and_speculative_policy: None`;
- `history_listener: None`;
- `request_timeout` — now applied by the core's timeout layer rather than by a
  hand-rolled `tokio::time::timeout`.

and a one-element `SingleConnectionTarget` plan.

Consequences:

- `RequestTimeoutError` and its ad-hoc mapping are deleted; timeouts now surface
  as `RequestError::RequestTimeout` from the shared layer, and the double-nested
  `Result<Result<_, _>, _>` return shape collapses into a single `RequestError`.
- Response interpretation is factored into `page_from_outcome`, which handles
  `IgnoredWriteError` (unreachable in practice, mapped to an empty final page)
  and rejects non-Rows responses.
- The path gains proper `RequestSpan` tracing (`make_span`), which it previously
  lacked, and the spawned worker task now inherits the caller's span.
- The first page is still fetched on the caller's task; only pages 2+ run in the
  spawned `fetch_remaining_pages` worker.

-----------------------------------------

</p>
</details> 

## Net effect

There is now exactly one implementation of the request-attempt loop. Whether a request is dispatched across a load-balancing plan with retries, speculative fibers and metrics, or sent once down a fixed control connection, it goes
through the same code; the differences are expressed as an `AttemptTarget` implementation plus `None`s in the params struct, and are checked by the type system where possible rather than by runtime branching.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/execution.rs
+++ b/scylla/src/client/execution.rs
@@ -333,9 +333,9 @@ impl<'a> RequestExecutionParams<'a> {
         let mut current_consistency: Consistency = self.consistency;
 
         'targets_in_plan: for (node, shard) in request_plan {
-            let span = trace_span!("Executing request", node = %node.address, shard = %shard);
+            let span = trace_span!("Executing request on chosen target", node = %node.address, shard = %shard);
             'same_target_retries: loop {
-                trace!(parent: &span, "Execution started");
+                trace!(parent: &span, "Execution attempt started");
                 let connection = match node.connection_for_shard(shard).await {
                     Ok(connection) => connection,
                     Err(e) => {
--- a/scylla/src/client/execution.rs
+++ b/scylla/src/client/execution.rs
@@ -221,6 +221,46 @@ impl AttemptTarget for NodeAttemptTarget<'_> {
     }
 }
 
+/// A target that always uses one specific connection. Used by the control
+/// connection, which has no [`Node`](crate::cluster::Node) and no load
+/// balancing.
+pub(crate) struct SingleConnectionTarget {
+    connection: Arc<Connection>,
+}
+
+impl SingleConnectionTarget {
+    pub(crate) fn new(connection: Arc<Connection>) -> Self {
+        Self { connection }
+    }
+}
+
+impl AttemptTarget for SingleConnectionTarget {
+    type Coordinator = ();
+
+    async fn get_connection(&self) -> Result<Arc<Connection>, ConnectionPoolError> {
+        Ok(Arc::clone(&self.connection))
+    }
+
+    fn coordinator(&self, _connection: &Arc<Connection>) {}
+
+    fn on_attempt_success(
+        &self,
+        _load_balancing_policy: &dyn LoadBalancingPolicy,
+        _routing_info: &RoutingInfo<'_>,
+        _elapsed: Duration,
+    ) {
+    }
+
+    fn on_attempt_failure(
+        &self,
+        _load_balancing_policy: &dyn LoadBalancingPolicy,
+        _routing_info: &RoutingInfo<'_>,
+        _elapsed: Duration,
+        _error: &RequestAttemptError,
+    ) {
+    }
+}
+
 /// Outcome of [`run_request_no_side_effects`]/[`run_request_speculative_fiber`].
 pub(crate) struct RequestExecutionOutcome<C> {
     /// The successful (or ignored-write) result.
--- a/scylla/src/client/execution.rs
+++ b/scylla/src/client/execution.rs
@@ -13,7 +13,7 @@ use crate::observability::driver_tracing::RequestSpan;
 use crate::observability::history::{self, HistoryListener};
 use crate::observability::metrics::Metrics;
 use crate::policies::load_balancing::{self, LoadBalancingPolicy, RoutingInfo};
-use crate::policies::retry::{RequestInfo, RetryDecision, RetryPolicy};
+use crate::policies::retry::{RequestInfo, RetryDecision, RetryPolicy, RetrySession};
 use crate::policies::speculative_execution::{self, SpeculativeExecutionPolicy};
 use crate::response::{Coordinator, NonErrorQueryResponse};
 use crate::statement::StatementConfig;
@@ -136,6 +136,7 @@ struct HistoryData<'a> {
 
 /// Per-fiber execution context.
 struct ExecuteRequestContext<'a> {
+    retry_session: Box<dyn RetrySession>,
     history_data: Option<HistoryData<'a>>,
     routing_info: &'a load_balancing::RoutingInfo<'a>,
     request_span: &'a RequestSpan,
@@ -264,6 +265,7 @@ impl<'a> RequestExecutionParams<'a> {
                             &shared_request_plan,
                             &run_request_once,
                             ExecuteRequestContext {
+                                retry_session: self.retry_policy.new_session(),
                                 history_data,
                                 routing_info,
                                 request_span,
@@ -290,6 +292,7 @@ impl<'a> RequestExecutionParams<'a> {
                         request_plan,
                         &run_request_once,
                         ExecuteRequestContext {
+                            retry_session: self.retry_policy.new_session(),
                             history_data,
                             routing_info,
                             request_span,
@@ -338,12 +341,11 @@ impl<'a> RequestExecutionParams<'a> {
         &self,
         request_plan: impl Iterator<Item = (NodeRef<'a>, Shard)>,
         run_request_once: impl Fn(Arc<Connection>, Consistency) -> QueryFut,
-        context: ExecuteRequestContext<'a>,
+        mut context: ExecuteRequestContext<'a>,
     ) -> Option<Result<(RunRequestResult<NonErrorQueryResponse>, Coordinator), RequestError>>
     where
         QueryFut: Future<Output = Result<NonErrorQueryResponse, RequestAttemptError>>,
     {
-        let mut retry_session = self.retry_policy.new_session();
         let mut last_error: Option<RequestError> = None;
         let mut current_consistency: Consistency = self.consistency;
 
@@ -421,7 +423,7 @@ impl<'a> RequestExecutionParams<'a> {
                     consistency: current_consistency,
                 };
 
-                let retry_decision = retry_session.decide_should_retry(request_info);
+                let retry_decision = context.retry_session.decide_should_retry(request_info);
                 trace!(
                     parent: &span,
                     retry_decision = ?retry_decision
--- a/scylla/src/client/execution.rs
+++ b/scylla/src/client/execution.rs
@@ -1,3 +1,28 @@
+//! Unified request-execution core shared by all request paths.
+//!
+//! Historically, request execution (load balancing, retries, speculative
+//! execution, history, metrics) was implemented independently in several
+//! places: [`Session`](crate::client::session::Session) non-paged methods, the
+//! [`QueryPager`](crate::client::pager::QueryPager) worker and the control
+//! connection worker. This module collapses all of that into a single set
+//! of layered functions:
+//!
+//! - [`RequestExecutionParams::run_request_no_side_effects`] - takes no `Session`
+//!   and does not handle side effects. It applies the client-side timeout and dispatches
+//!   speculative-execution fibers. It is generic over the *source* of
+//!   connections (see [`AttemptTarget`]) so that it works both with a
+//!   load-balancing plan of nodes and with a single fixed connection (as used
+//!   by the control connection).
+//! - [`RequestExecutionParams::run_request_speculative_fiber`] - a single speculative fiber:
+//!   iterates the execution plan, picks connections, runs the per-attempt
+//!   closure and applies the retry policy.
+//!
+//! The outermost layer (`run_request`), which additionally handles
+//! side effects coming from `USE <keyspace>` and schema-changing statements,
+//! lives on [`Session`](crate::client::session::Session) because it needs
+//! access to session state. An analogous side-effects-handling layer lives
+//! [`PagingExecutor`](crate::client::pager::PagingExecutor).
+
 use std::future::Future;
 use std::net::SocketAddr;
 use std::sync::Arc;
--- a/scylla/src/client/execution.rs
+++ b/scylla/src/client/execution.rs
@@ -10,6 +10,7 @@ use tracing::trace_span;
 
 use crate::client::execution_profile::ExecutionProfileInner;
 use crate::cluster::NodeRef;
+use crate::errors::ConnectionPoolError;
 use crate::errors::RequestAttemptError;
 use crate::errors::RequestError;
 use crate::frame::types::{Consistency, SerialConsistency};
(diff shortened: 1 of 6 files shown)


scylla/src/client/execution.rs
```diff
@@ -319,24 +478,24 @@ impl<'a> RequestExecutionParams<'a> {
...
         let mut last_error: Option<RequestError> = None;
         let mut current_consistency: Consistency = self.consistency;
 
-        'nodes_in_plan: for (node, shard) in request_plan {
-            let span = trace_span!("Executing request", node = %node.address, shard = %shard);
-            'same_node_retries: loop {
-                trace!(parent: &span, "Execution started");
-                let connection = match node.connection_for_shard(shard).await {
+        'targets_in_plan: for target in request_plan {
+            let span = trace_span!("Executing request on chosen target");
```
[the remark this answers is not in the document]
wprzytula: I leave it for a different PR which will focus on improving the structure and quality of our emitted traces.

scylla/src/client/execution.rs
```diff
@@ -69,10 +69,13 @@ pub(crate) struct RequestExecutionParams<'a> {
...
+    /// The following two fields are grouped to statically enforce that speculative execution
+    /// is only used when metrics sink is provided.
+    ///
+    /// Metrics sink, if any. The control connection has no metrics.
+    /// Speculative execution policy, if any. Only fires for idempotent requests.
+    pub(crate) metrics_and_speculative_policy:
+        Option<(&'a Arc<Metrics>, Option<&'a dyn SpeculativeExecutionPolicy>)>,
     /// Client-side request timeout, if any.
     pub(crate) request_timeout: Option<Duration>,
     /// History listener, if any.
```
Lorak-mmk, line 114: Commit: "execution: require metrics for speculative execution"

I started to write a comment about why we even need that (only one SEP needs metrics, the rest work fine without). Then I understood it is about `Context`: it has `Arc<Metrics>` which we can't produce artificially.
But now I think: why can't we can't keep the metrics clone in CC, keeping them non-optional?

Sure, it may not make that much sense to use SEP on SingleConnection (or maybe we just don't see use case for now?), but imo it seems better to be consistent.

scylla/src/client/execution.rs
```diff
@@ -98,7 +96,7 @@ impl<'a> RequestExecutionParams<'a> {
         execution_profile: &'a ExecutionProfileInner,
         metrics: &'a Arc<Metrics>,
         request_kind: RequestPaging,
-    ) -> Self {
+    ) -> (Self, &'a dyn LoadBalancingPolicy) {
         let is_idempotent = statement_config.is_idempotent;
```
Lorak-mmk, line 128: Commit: execution: take LBP out of REP

The previous commits introduced some complexity - as expected - but it was not too bad, and I thought the benefit of the unification was more important here. 

This commit is the first place where I'm starting to wonder if its even worth it, because now we break the nice property of having all exec params in a single struct.

Lorak-mmk: I'm trying to think of some alternative approach that doesn't involve making RequestExecutionParams over some trait implemented for LBP and for something else.


Maybe the problem is that we want `RequestExecutionParams` to cover all the cases?
There was already ownership problem with the pager, requiring to duplicate some logic there.
Perhaps we should have RequestExecutionParams, and SingleConnectionRequestExecutionParams (and maybe OwnedRequestExecutionParams). They all would have a method to run the request, and underneath would call a common function?

wprzytula: > Commit: execution: take LBP out of REP

This commit is not strictly necessary. I can implement a `NoOpLBP` and pass it to the REP in `SingleConnectionPagingExecutor`.

wprzytula: Done.

scylla/src/client/execution.rs
wprzytula: I'm not sure we need to work on this. I can consider further cleanup in the future. For now, this ugliness does not propagate to `session.rs` - it's contained in `execution.rs`.

scylla/src/client/pager.rs
```diff
@@ -495,111 +496,189 @@ impl PagingExecutor {
...
+    /// Fetches exactly one page from the fixed connection by delegating to
+    /// [`run_request_no_side_effects`] with a single-connection plan.
+    async fn fetch_one_page(
+        &self,
+        paging_state: &PagingState,
+        request_span: &RequestSpan,
+    ) -> Result<RequestExecutionOutcome<()>, RequestError> {
+        // The control-connection-style path never retries: a fallthrough retry
+        // policy makes every attempt terminal.
+        let retry_policy = FallthroughRetryPolicy::new();
```
Lorak-mmk, line 557: We have an issue for client-side timeouts in CC, but not for retries in CC. Let's open one.

scylla/src/client/pager.rs
```diff
@@ -495,111 +496,189 @@ impl PagingExecutor {
...
+        let exec_params = RequestExecutionParams {
+            is_idempotent: true, // I assume that requests sent on CC are idempotent SELECTs.
```
Lorak-mmk: Why is that something you need to hardcode, and not take from the statement?

I remember setting this flag on CC statements.

scylla/src/client/pager.rs
```diff
@@ -495,111 +496,189 @@ impl PagingExecutor {
...
-                        return Err(RequestTimeoutError(timeout));
-                    }
-                }
+
+        let routing_info = RoutingInfo {
+            consistency: self.consistency,
+            serial_consistency: self.serial_consistency,
+            // We're anyway executing on the Control Connection against locally replicated system tables,
+            // so the token is not useful for anything.
+            token: None,
```
Lorak-mmk: I wonder if it may have implications for tablet routing v2 (no token = no tablet = tablet feedback). Can we calculate the token just to be safe?

Lorak-mmk (changes requested):

scylla/src/client/pager.rs
wprzytula: https://github.com/scylladb/scylla-rust-driver/issues/1836

wprzytula:
`affects-performance` label was added by the CI job, so I ran the benchmarks locally to see the culprit.

Observations:
- **there definitely are more CPU instructions**.
- there's no change in allocations.
- L1, LL, RAM hits are unstable, but tend negatively.

Conclusions:
- **This PR does worsen the performance on the hot path by a minor factor.**
- The win is that we get fully unified execution paths now.
- It's up to the reviewers to decide whether this regression is acceptable or I should think harder on optimising.

<details><summary>This is how it looks:</summary>
<p>


```
requests::requests::insert counts_0:(setup_default(100))                                                                                                                                                
 ======= CALLGRIND ====================================================================                                                                                                                
 Baselines:                               |base                                                                                                                                                        
 Instructions:                     2483462|2469356              (+0.57124%) [+1.00571x]                                                                                                                
 L1 Hits:                          3557918|3541909              (+0.45199%) [+1.00452x]                                                                                                                
 LL Hits:                           209489|202041               (+3.68638%) [+1.03686x]                                                                                                                
 RAM Hits:                             106|103                  (+2.91262%) [+1.02913x]                                                                                                                
 Total read+write:                 3767513|3744053              (+0.62659%) [+1.00627x]                                                                                                                
 Estimated Cycles:                 4609073|4555719              (+1.17114%) [+1.01171x]                                                                                                                
 ======= DHAT =========================================================================                                                                                                                
 Total bytes:                        29464|29464                (No change)                                                                                                                            
 Total blocks:                         901|901                  (No change)                                                                                                                            
 At t-gmax bytes:                        0|0                    (No change)                                                                                                                            
 At t-gmax blocks:                       0|0                    (No change)                                                                                                                            
 At t-end bytes:                         0|0                    (No change)                                                                                                                            
 At t-end blocks:                        0|0                    (No change)                                                                                                                            
 Reads bytes:                        40700|40700                (No change)                                                                                                                            
 Writes bytes:                       43400|43400                (No change)                                                                                                                            
requests::requests::unpaged_select counts_0:(setup_default(100))                                                                                                                                        
 ======= CALLGRIND ====================================================================                                                                                                                
 Baselines:                               |base                                                                                                                                                        
 Instructions:                     2542896|2537309              (+0.22019%) [+1.00220x]                                                                                                                
 L1 Hits:                          3646281|3641232              (+0.13866%) [+1.00139x]                                                                                                                
 LL Hits:                           221311|215990               (+2.46354%) [+1.02464x]                                                                                                                
 RAM Hits:                             105|111                  (-5.40541%) [-1.05714x]                                                                                                                
 Total read+write:                 3867697|3857333              (+0.26868%) [+1.00269x]                                                                                                                
 Estimated Cycles:                 4756511|4725067              (+0.66547%) [+1.00665x]                                                                                                                
 ======= DHAT =========================================================================                                                                                                                
 Total bytes:                        54464|54464                (No change)                                                                                                                            
 Total blocks:                         901|901                  (No change)                                                                                                                            
 At t-gmax bytes:                        0|0                    (No change)                                                                                                                            
 At t-gmax blocks:                       0|0                    (No change)                                                                                                                            
 At t-end bytes:                         0|0                    (No change)                                                                                                                            
 At t-end blocks:                        0|0                    (No change)                                                                                                                            
 Reads bytes:                        53100|53100                (No change)                                                                                                                            
 Writes bytes:                       74000|74000                (No change)                                                                                                                            
requests::requests::paged_select counts_0:(setup_default(100))                                                                                                                                          
 ======= CALLGRIND ====================================================================                                                                                                                
 Baselines:                               |base                                                                                                                                                        
 Instructions:                   223977341|223106801            (+0.39019%) [+1.00390x]                                                                                                                
 L1 Hits:                        332926173|331809462            (+0.33655%) [+1.00337x]                                                                                                                
 LL Hits:                         11152175|10802079             (+3.24101%) [+1.03241x]                                                                                                                
 RAM Hits:                             643|641                  (+0.31201%) [+1.00312x]                                                                                                                
 Total read+write:               344078991|342612182            (+0.42813%) [+1.00428x]                                                                                                                
 Estimated Cycles:               388709553|385842292            (+0.74312%) [+1.00743x]                                                                                                                
 ======= DHAT =========================================================================                                                                                                                
 Total bytes:                     13481464|13481464             (No change)                                                                                                                            
 Total blocks:                      156801|156801               (No change)                                                                                                                            
 At t-gmax bytes:                        0|0                    (No change)                                                                                                                            
 At t-gmax blocks:                       0|0                    (No change)                                                                                                                            
 At t-end bytes:                         0|0                    (No change)                                                                                                                            
 At t-end blocks:                        0|0                    (No change)                                                                                                                            
 Reads bytes:                     32966700|32841700             (+0.38061%) [+1.00381x]                                                                                                                
 Writes bytes:                    66213200|64513600             (+2.63448%) [+1.02634x]                                                                                                                
requests::requests::batch sizes_0:(setup_batch(64, 100))                                                                                                                                                
 ======= CALLGRIND ====================================================================                                                                                                                
 Baselines:                               |base                                                                                                                                                        
 Instructions:                     4460821|4355908              (+2.40852%) [+1.02409x]                                                                                                                
 L1 Hits:                          6364761|6236290              (+2.06005%) [+1.02060x]                                                                                                                
 LL Hits:                           244217|236664               (+3.19144%) [+1.03191x]                                                                                                                
 RAM Hits:                             398|382                  (+4.18848%) [+1.04188x]                                                                                                                
 Total read+write:                 6609376|6473336              (+2.10154%) [+1.02102x]                                                                                                                
 Estimated Cycles:                 7599776|7432980              (+2.24400%) [+1.02244x]                                                                                                                
 ======= DHAT =========================================================================                                                                                                                
 Total bytes:                       936664|936664               (No change)                                                                                                                            
 Total blocks:                        1501|1501                 (No change)                                                                                                                            
 At t-gmax bytes:                        0|0                    (No change)                                                                                                                            
 At t-gmax blocks:                       0|0                    (No change)                                                                                                                            
 At t-end bytes:                         0|0                    (No change)                                                                                                                            
 At t-end blocks:                        0|0                    (No change)                                                                                                                            
 Reads bytes:                       727600|727600               (No change)                                                                                                                            
 Writes bytes:                      743100|743100               (No change)          

```

</p>
</details>

Lorak-mmk:
I think this regression is acceptable. What is interesting is why did this even happen. Can you ask AI to try to investigate the cause of the regression?

Lorak-mmk (approved):

wprzytula:
I checked that the dominant part of the regression is introduced in `60968fb6d pager: migrate SingleConnectionPagingExecutor to unified core`:
```
requests::requests::batch sizes_0:(setup_batch(64, 100))
  ======= CALLGRIND ====================================================================
  Baselines:                               |base
  Instructions:                     4434776|4355908              (+1.81060%) [+1.01811x]
  L1 Hits:                          6332080|6236290              (+1.53601%) [+1.01536x]
  LL Hits:                           242189|236664               (+2.33453%) [+1.02335x]
  RAM Hits:                             398|382                  (+4.18848%) [+1.04188x]
  Total read+write:                 6574667|6473336              (+1.56536%) [+1.01565x]
  Estimated Cycles:                 7556955|7432980              (+1.66790%) [+1.01668x]
```

After a longer discussion with Opus, these are the important take-aways:

_At CGU=16 the same commit measured +1.81%. So roughly ⅔–¾ of the reported regression is codegen-unit partitioning shuffling unrelated code._

_And the noise floor is larger than the signal_

_Re-running the same unmodified binary five times:_

- _batch: −0.027%, +0.085%, +0.139%, +0.370%, +0.309% — a 0.40 pp spread_
- _paged_select: −0.147%, +0.272%, +0.140%, −0.039%, −0.023% — a 0.42 pp spread_

_Same binary, same input, different instruction counts. Callgrind is deterministic per-execution, so this is the benchmark itself: it drives a real three-node cluster, and network timing, connection-pool state, background tasks and tokio scheduling change how much code actually runs each time._

This is its conclusion:

_The reported batch regression decomposes roughly as: ~1.0 pp CGU-partitioning wobble in unrelated functions, ~0.4 pp run-to-run noise, ~0.1–0.3 pp real cost in the execution core. The genericisation and the second instantiation are essentially free. There is no meaningful performance problem in 60968fb6d to fix._

_What is worth fixing is the harness, since it currently can't resolve changes below ~0.5%: pin codegen-units = 1 in the root Cargo.toml, and either run each scenario N times taking the minimum, or point the benchmarks at a fixed single-node/mock target so the network isn't in the measurement loop._

Lorak-mmk:
Ok. Feel free to merge.

PR #1831, opened 2026-08-08
examples: complete custom serde traits

### LINKED ISSUE #1828: Custom ser/de examples are asymmetrical/missing
## What I spotted

🤔 Are these examples asymmetrical?
- `custom_deserialization` teaches `DeserializeValue`,
- `custom_serialization` teaches `SerializeRow`.

What about `SerializeValue` and `DeserializeRow`?
How is `user_defined_type` example related to this?

_Originally posted by @wprzytula in https://github.com/scylladb/scylla-rust-driver/pull/1813#discussion_r3720780299_

## Suggested solution

1. Inspect examples: `custom_serialization`, `custom_deserialization`, `user_defined_type`.
2. See what's missing and add it.
3. Think about better structuring of those examples.

ColumbusLabs:
Complete the custom serialization/deserialization examples so both value- and row-level traits are demonstrated symmetrically.

- implement `SerializeValue` for a custom text wrapper and use it from a derived `SerializeRow`
- derive `DeserializeRow` around the existing zero-copy custom `DeserializeValue`
- clarify the example index descriptions

Validation:
- `cargo check -p examples --examples`
- `make static` (all local checks passed through Clippy and book generation; the final rustdoc leak check could not run because the nightly toolchain is not installed)

Fixes: #1828

### DIFF
--- a/examples/custom_deserialization.rs
+++ b/examples/custom_deserialization.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use scylla::DeserializeRow;
 use scylla::client::session::Session;
 use scylla::client::session_builder::SessionBuilder;
 use scylla::deserialize::value::DeserializeValue;
@@ -32,7 +33,11 @@ async fn main() -> Result<()> {
     #[derive(PartialEq, Eq, Debug)]
     struct MyType<'a>(&'a str);
 
-    impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for MyType<'frame> {
+    // The value may borrow for any lifetime shorter than the response frame.
+    impl<'frame, 'metadata, 'a> DeserializeValue<'frame, 'metadata> for MyType<'a>
+    where
+        'frame: 'a,
+    {
         fn type_check(
             typ: &scylla::frame::response::result::ColumnType,
         ) -> std::result::Result<(), scylla::deserialize::TypeCheckError> {
@@ -57,8 +62,15 @@ async fn main() -> Result<()> {
         .await?
         .into_rows_result()?;
 
-    let (v,) = rows_result.single_row::<(MyType,)>()?;
-    assert_eq!(v, MyType("asdf"));
+    // DeserializeRow can be derived for a struct representing a whole row.
+    #[derive(Debug, DeserializeRow)]
+    struct MyRow<'a> {
+        // Field name must correspond to the CQL column's name.
+        v: MyType<'a>,
+    }
+
+    let row = rows_result.single_row::<MyRow>()?;
+    assert_eq!(row.v, MyType("asdf"));
 
     println!("Ok.");
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -40,9 +40,9 @@ cargo run -p examples --example basic
 | `compare_tokens` | Computing a partition key's token in the driver, checking it against the token the cluster reports, and listing the replicas that own it. |
 | `cql_time_types` | Reading and writing `date`, `time` and `timestamp` using `chrono`, `time`, and the raw `CqlDate`/`CqlTime`/`CqlTimestamp` types. |
 | `cqlsh_rs` | A miniature `cqlsh`: a REPL with CQL keyword completion that executes whatever you type. |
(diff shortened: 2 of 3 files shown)


examples/custom_deserialization.rs
```diff
@@ -32,7 +33,10 @@ async fn main() -> Result<()> {
     #[derive(PartialEq, Eq, Debug)]
     struct MyType<'a>(&'a str);
 
-    impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for MyType<'frame> {
+    impl<'frame, 'metadata, 'a> DeserializeValue<'frame, 'metadata> for MyType<'a>
+    where
+        'frame: 'a,
+    {
```
wprzytula, line 40: ❓ Why this lifetime change? Why do you need to introduce `'a`?

examples/custom_deserialization.rs
```diff
@@ -57,8 +61,14 @@ async fn main() -> Result<()> {
         .await?
         .into_rows_result()?;
 
-    let (v,) = rows_result.single_row::<(MyType,)>()?;
-    assert_eq!(v, MyType("asdf"));
+    // DeserializeRow can be derived for a struct representing a whole row.
+    #[derive(Debug, DeserializeRow)]
+    struct MyRow<'a> {
+        v: MyType<'a>,
+    }
```
wprzytula, line 70: 🔧 Let's add a comment above the `v` field:
> Field name must correspond to the CQL column's name.

wprzytula (changes requested):

wprzytula:
Also: how is `user_defined_type` example related to the modified two? Doesn't it duplicate their contents?

examples/custom_deserialization.rs
ColumbusLabs: The separate lifetime is required by the derived DeserializeRow implementation: the row may borrow for a lifetime shorter than the response frame, so MyType must implement DeserializeValue whenever the frame outlives that borrow. Reverting to MyType<'frame> fails cargo check at the derive with 'lifetime may not live long enough'. I added a short code comment to make that relationship explicit.

examples/custom_deserialization.rs
ColumbusLabs: Added the requested comment directly above the v field.

ColumbusLabs:
The user_defined_type example is complementary rather than a duplicate: it derives SerializeValue and DeserializeValue for a Rust struct mapped to a CQL UDT. The custom serialization/deserialization examples now show manually implementing the value traits for scalar column values, alongside the derived row traits. The README descriptions call out that distinction, so I left user_defined_type unchanged.

examples/custom_deserialization.rs
```diff
@@ -32,7 +33,11 @@ async fn main() -> Result<()> {
     #[derive(PartialEq, Eq, Debug)]
     struct MyType<'a>(&'a str);
 
-    impl<'frame, 'metadata> DeserializeValue<'frame, 'metadata> for MyType<'frame> {
+    // The value may borrow for any lifetime shorter than the response frame.
+    impl<'frame, 'metadata, 'a> DeserializeValue<'frame, 'metadata> for MyType<'a>
+    where
+        'frame: 'a,
+    {
```
wprzytula, line 40: 💭  Wouldn't this be possible even if left just `'frame` and `'metadata` lifetimes?
Lifetimes variance allows shorter lifetimes to satisfy this `impl trait` block, meaning that the value doesn't have to live as long as the response frame. Am I mistaken? cc @Lorak-mmk

ColumbusLabs: I tested both two-lifetime forms against the current head. Using MyType<'frame> makes the DeserializeRow derive fail with 'lifetime may not live long enough' because the derived row lifetime may be shorter than 'frame. Using MyType<'_> fails inside deserialize because 'frame is not known to outlive the anonymous borrow. The three-lifetime form with 'frame: 'a is what connects the response buffer lifetime to the shorter borrow required by the derived row. With that bound restored, cargo check -p examples --example custom_deserialization passes, so I believe the current bound is required.

wprzytula (approved):

Lorak-mmk (approved):

Issue #1628, opened 2026-04-29
Allow serialization of some iterators


Lorak-mmk:
Consider a user that wants to serialize their own custom types, and of course do that in ergonomic and performant way.
That means:
- No manual byte operations, the serialization impl should reuse our impls and other helper tools
- No unnecessary cloning

I encountered this when trying to improve latte tool.
When such a user has their own custom Vec<T>, and wants to serialize it, there is a problem: T may not implement SerializeValue!
User would like to do `.iter().map(|elem| SomeWrapper(elem))` (where `SomeWrapper` implements SerializeValue), but there is a problem: we do not implement SerializeValue for iterators. So the user has to do serialization manually, or allocate a Vec for no reason.

PR #1835, opened 2026-08-11
Make metadata fetching non-blocking

### LINKED ISSUE #1670: ClusterWorker and the control connection can deadlock on high volume of CQL events

Lorak-mmk:
This PR makes metadata fetching done in MetadataWorker non-blocking.

What does it even mean?

Until now we had a bug, specifically a deadlock.
Events from CC are handled in a `select!` in a worker loop. However,
metadata fetch is an async work done outside this `select!`, so during the metadata fetch
we did not handle events. If the event channel became full during the fetch, the CC could no longer receive responses from the server because its reader would hand on trying to send an event to the channel.

The solution (implemented here): only perform async work in a branch of `select!`. That way we never starve any work we need to do.
Two most important structures here:
- `FetchPlan` records what fetches we need to do (but didn't start yet)
- `PendingFetches` kepp the futures with currently running fetches.

Before each `select!` we look at both and start fetches that can and must be started.

This approach has one more advantage: when implementing smarter metadata fetching ( https://github.com/scylladb/scylla-rust-driver/issues/1280 ) it will make it easy to perform multiple partial fetches (e.g. topology and client routes) concurrently.

## Problem with initial fetch

The solution above applies to `work_on_cc`, so when the CC is already created and its initial metadata fetch.
Unfortunately it is possible for this problem to also happen during the initial metadata fetch!
The idea for a solution is ofc the same: we need to drain the events when reading metadata.

Some details were problematic: FetchPlan is private to MetadataWorker, and I wanted to keep it that way.
So the solution here is to make MetadataReader accept a callback (in the form of trait because of rustc). Impl of this trait is provided by the worker, and what it does is create an FetchPlan from the receive events (as expected).

Commits up to "Add regression test for event flood during metadata fetch" address the first path (`work_on_cc`).
Later commits address the second path: metadata fetch done on connection creation.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1670

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/metadata/cc_establisher.rs
+++ b/scylla/src/cluster/metadata/cc_establisher.rs
@@ -1,4 +1,4 @@
-//! This module contains the [`MetadataReader`] struct, which is responsible for
+//! This module contains the [`ControlConnectionEstablisher`] struct, which is responsible for
 //! creating control connections and fetching cluster metadata through them.
 //!
 //! The control connection is a dedicated connection to one of the cluster nodes
@@ -6,13 +6,13 @@
 //! - Fetch cluster metadata (topology, schema, token ring information)
 //! - Receive server-side events (topology changes, schema changes, status changes)
 //!
-//! [`MetadataReader`] establishes control connections, including:
+//! [`ControlConnectionEstablisher`] establishes control connections, including:
 //! - Connection establishment to contact points or known peers
 //! - Iterating over known peers and initial contact points on connection failure
 //! - Host filtering to ensure the control connection is established to an accepted node
 //!
-//! Ownership of the established control connection lives outside the reader (in the
-//! metadata worker); the reader only knows how to create one. The metadata queries
+//! Ownership of the established control connection lives outside the establisher (in the
+//! metadata worker); the establisher only knows how to create one. The metadata queries
 //! themselves are methods on [`ControlConnection`], configured at its creation.
 
 use std::sync::Arc;
@@ -40,13 +40,13 @@ use crate::utils::safe_format::IteratorSafeFormatExt;
 
 /// Maintains the persistent state needed to create control connections and
 /// fetch cluster metadata. The established control connection itself is owned by
-/// the caller (the metadata worker), not by the reader.
-pub(crate) struct MetadataReader {
+/// the caller (the metadata worker), not by the establisher.
+pub(crate) struct ControlConnectionEstablisher {
     // =======================================================================================
-    // Configuration values - they will stay the same during whole lifetime of MetadataReader.
+    // Configuration values - they will stay the same during whole lifetime of ControlConnectionEstablisher.
     // =======================================================================================
     control_connection_config: ConnectionConfig,
-    /// Configuration stamped onto every control connection this reader creates;
+    /// Configuration stamped onto every control connection this establisher creates;
     /// governs what the control connection's metadata queries fetch.
     cc_config: ControlConnectionConfig,
     hostname_resolution_timeout: Option<Duration>,
@@ -56,18 +56,18 @@ pub(crate) struct MetadataReader {
     initial_known_nodes: Vec<KnownNode>,
 
     // ====================================================================
-    // Mutable state of MetadataReader. It will change during its lifetime.
+    // Mutable state of ControlConnectionEstablisher. It will change during its lifetime.
     // ====================================================================
-    // when a control connection fails, MetadataReader tries to connect to one of known_peers
+    // when a control connection fails, ControlConnectionEstablisher tries to connect to one of known_peers
     known_peers: Vec<UntranslatedEndpoint>,
     cc_cache: Arc<ControlConnectionCache>,
 }
 
 /// The per-candidate metadata fetch that
-/// [`MetadataReader::establish_cc_and_fetch_metadata`] runs on each candidate
+/// [`ControlConnectionEstablisher::establish_cc_and_fetch_metadata`] runs on each candidate
 /// connection - implemented by the metadata worker.
 ///
-/// When `MetadataReader` open a new `Connection`, it needs to fetch initial
+/// When `ControlConnectionEstablisher` open a new `Connection`, it needs to fetch initial
 /// `Metadata` on it before returning it as a new CC. A candidate is such
 /// potential CC until the `Metadata` is done fetching.
 ///
@@ -91,8 +91,8 @@ pub(super) trait FetchOnCandidate {
     ) -> Result<(TopologyUpdateGuard, Self::Payload), MetadataError>;
 }
 
-impl MetadataReader {
-    /// Creates a new MetadataReader.
+impl ControlConnectionEstablisher {
+    /// Creates a new ControlConnectionEstablisher.
     ///
     /// Resolves the initial contact points and populates the initial known peers
     /// list. Does **not** establish a control connection — use
@@ -120,7 +120,7 @@ impl MetadataReader {
 
         let cc_cache = Arc::new(ControlConnectionCache::new());
 
-        Ok(MetadataReader {
+        Ok(ControlConnectionEstablisher {
             control_connection_config: connection_config,
             cc_config: ControlConnectionConfig {
                 keyspaces_to_fetch,
@@ -449,13 +449,13 @@ impl MetadataReader {
     }
 }
 
-/// Freshly fetched [`Metadata`] that the [`MetadataReader`] has not yet absorbed.
+/// Freshly fetched [`Metadata`] that the [`ControlConnectionEstablisher`] has not yet absorbed.
 ///
 /// [`ControlConnection::query_metadata`] returns its result wrapped in this
-/// guard, so that the fetched metadata cannot be used without the reader
+/// guard, so that the fetched metadata cannot be used without the establisher
 /// updating its known peers from it first: the only way to extract the
 /// [`Metadata`] is [`apply`](Self::apply).
-#[must_use = "the fetched metadata must be applied to the MetadataReader"]
+#[must_use = "the fetched metadata must be applied to the ControlConnectionEstablisher"]
 pub(super) struct TopologyUpdateGuard {
     metadata: Metadata,
 }
@@ -465,10 +465,10 @@ impl TopologyUpdateGuard {
         Self { metadata }
     }
 
-    /// Updates the reader's known peers from the fetched metadata and releases
+    /// Updates the establisher's known peers from the fetched metadata and releases
     /// the metadata itself.
-    pub(super) fn apply(self, reader: &mut MetadataReader) -> Metadata {
-        reader.update_known_peers(&self.metadata);
+    pub(super) fn apply(self, establisher: &mut ControlConnectionEstablisher) -> Metadata {
+        establisher.update_known_peers(&self.metadata);
         self.metadata
     }
 }
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -47,7 +47,6 @@ use crate::errors::{
     MetadataError, MetadataFetchError, MetadataFetchErrorKind, NextPageError, NextRowError,
     PeersMetadataError, RequestAttemptError, RequestError, TablesMetadataError, UdtMetadataError,
 };
-use crate::frame::response::event::ClientRoutesChangeEvent;
 use crate::routing::Token;
 
 use super::reader::TopologyUpdateGuard;
@@ -552,18 +551,20 @@ impl ControlConnection {
         Ok(routes)
     }
 
-    /// Performs a partial fetch of `system.client_routes`. Partial means that filtering is done
-    /// not only by connection ids known to the driver (which is always the case), but also
-    /// by host ids - only for the hosts whose ids are present in the event payload.
+    /// Performs a partial fetch of `system.client_routes`.
+    ///
+    /// Such partial fetch is restricted to the given (connection id, host id) pairs, as accumulated from
+    /// CLIENT_ROUTES_CHANGE:UPDATE_NODES events. Those pairs are additionally filtered
+    /// to only query connection ids that the driver is interested in.
     ///
     /// Returns the resulting [`ClientRoutesUpdate`], or `None` if there is no subscriber
-    /// configured or the event contained no connection ids relevant to this driver. Merging
+    /// configured or no pair has a connection id relevant to this driver. Merging
     /// the update into the [`ClientRoutesSubscriber`]'s knowledge is the caller's job.
     ///
     /// [`ClientRoutesSubscriber`]: crate::client::client_routes::ClientRoutesSubscriber
-    pub(super) async fn fetch_client_routes_update_on_event(
+    pub(super) async fn fetch_client_routes_update(
         &self,
-        evt: &ClientRoutesChangeEvent,
+        pairs: &HashSet<(String, Uuid)>,
     ) -> Result<Option<ClientRoutesUpdate>, MetadataError> {
         let Some(subscriber) = self.config.client_routes_subscriber.as_ref() else {
             // No subscriber, but received an event? Strange enough, but nothing to be done here.
@@ -571,41 +572,38 @@ impl ControlConnection {
             return Ok(None);
         };
 
-        #[deny(clippy::wildcard_enum_match_arm)]
-        let (connection_ids, host_ids) = match evt {
-            ClientRoutesChangeEvent::UpdateNodes {
-                connection_ids,
-                host_ids,
-            } => (connection_ids, host_ids),
-            _ => unreachable!("clippy testifies that the match is exhaustive"),
-        };
+        let monitored_pairs = pairs
+            .iter()
+            .filter(|(conn_id, _host_id)| subscriber.get_connection_ids().contains(conn_id));
 
-        // TODO: this is wasteful - it allocates both strings and a vec.
+        // The pairs cannot be queried as such: the query is
+        // `WHERE connection id IN ? AND host id IN ?`, which fetches the whole
+        // cross product of the listed ids - possibly more routes than the
+        // pairs name, for example `(A, Y)` for pairs `[(A, X), (B, Y)]`. This
+        // is a tradeoff - the only alternative is issuing multiple queries,
+        // one per connection id. The surplus routes are ignored when the
+        // update is built from the pairs - see
+        // [`ClientRoutesUpdate::from_pairs`].
+        //
+        // TODO: this is wasteful - it allocates strings and vecs.
         // This won't be a performance problem, because UPDATE_NODES events are not frequent.
         // As an optimization, we can implement ser/de for some special new iterator type,
         // to avoid the need to allocate when serializing collections.
-        let connection_ids: Vec<String> = connection_ids
-            .iter()
-            .filter(|&conn_id| subscriber.get_connection_ids().contains(conn_id))
-            .cloned()
-            .collect();
+        let (connection_ids, host_ids): (HashSet<String>, HashSet<Uuid>) =
+            monitored_pairs.cloned().unzip();
 
         if connection_ids.is_empty() {
-            // The event contained no relevant connection IDs.
+            // The events contained no relevant connection IDs.
             // Nothing to be done.
             return Ok(None);
         }
 
-        // Although this is vaguely documented, the semantics of an event with connection ids [A, B, C] and host ids [X, Y, Z]
-        // is that the following entries were added/updated/removed: `[(A, X), (B, Y), (C, Z)]`.
-        // Unfortunately, we can't really query Scylla this way. Therefore, we do the query: `WHERE connection id IN ? AND host id IN ?`,
-        // which fetches possibly more routes than necessary, for example `(A, Z)` or `(C, Y)`.
-        // This is a tradeoff - the only alternative is issuing multiple queries, one per connection id.
-        // I believe the tradeoff here is correct.
-        let client_routes = self.query_client_routes(&connection_ids, host_ids).await?;
+        let client_routes = self
+            .query_client_routes(&Vec::from_iter(connection_ids), &Vec::from_iter(host_ids))
+            .await?;
 
-        Ok(Some(ClientRoutesUpdate::from_event(
-            evt,
+        Ok(Some(ClientRoutesUpdate::from_pairs(
+            pairs,
             subscriber.get_connection_ids(),
             &client_routes,
         )))
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -40,13 +40,14 @@ use crate::DeserializeRow;
 use crate::client::pager::QueryPager;
 use crate::cluster::NodeAddr;
 use crate::cluster::control_connection::ControlConnection;
-use crate::cluster::metadata::ClientRoutes;
+use crate::cluster::metadata::{ClientRoutes, ClientRoutesUpdate};
 use crate::deserialize::DeserializeOwnedRow;
 use crate::errors::{
     ClientRoutesMetadataError, DbError, KeyspaceStrategyError, KeyspacesMetadataError,
     MetadataError, MetadataFetchError, MetadataFetchErrorKind, NextPageError, NextRowError,
     PeersMetadataError, RequestAttemptError, RequestError, TablesMetadataError, UdtMetadataError,
 };
+use crate::frame::response::event::ClientRoutesChangeEvent;
 use crate::routing::Token;
 
 type PerKeyspace<T> = HashMap<String, T>;
@@ -156,20 +157,22 @@ impl fmt::Display for InvalidCqlType {
 }
 
 impl ControlConnection {
-    pub(crate) async fn query_metadata(
-        &self,
-        connect_port: u16,
-        keyspace_to_fetch: &[String],
-        schema_metadata_fetch_mode: SchemaMetadataFetchMode,
-        client_routes_connection_ids: Option<&[String]>,
-    ) -> Result<Metadata, MetadataError> {
+    /// Queries the cluster metadata (peers, schema, client routes) as configured
+    /// by this control connection's
+    /// [`ControlConnectionConfig`](crate::cluster::control_connection::ControlConnectionConfig).
+    pub(super) async fn query_metadata(&self) -> Result<Metadata, MetadataError> {
+        let connect_port = self.endpoint().address().port();
+
         let peers_query = self.query_peers(connect_port);
-        let keyspaces_query = self.query_keyspaces(keyspace_to_fetch, schema_metadata_fetch_mode);
+        let keyspaces_query = self.query_keyspaces(
+            &self.config.keyspaces_to_fetch,
+            self.config.schema_metadata_fetch_mode,
+        );
         let client_routes_query = async {
-            let Some(connection_ids) = client_routes_connection_ids else {
+            let Some(subscriber) = self.config.client_routes_subscriber.as_ref() else {
                 return Ok(None);
             };
-            self.query_client_routes(connection_ids, &[])
+            self.query_client_routes(subscriber.get_connection_ids(), &[])
                 .await
                 .map(Some)
         };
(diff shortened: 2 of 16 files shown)


scylla/src/cluster/metadata/reader.rs
```diff
@@ -267,8 +250,8 @@ impl MetadataReader {
                 }
             };
 
-            let metadata = match self.query_metadata_on_cc(&cc).await {
-                Ok(metadata) => metadata,
+            let metadata = match cc.query_metadata().await {
+                Ok(topology_update) => topology_update.apply(self),
```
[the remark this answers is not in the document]
Lorak-mmk: Huh, this is true. And I don't see an easy way to solve it, need to think about it.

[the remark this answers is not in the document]
Lorak-mmk: Deferring after the fetch of course doesn't work because we would lose events.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -40,13 +40,14 @@ use crate::DeserializeRow;
 use crate::client::pager::QueryPager;
```
wprzytula, line 40: ⛏️ commit `Make ControlConnection self-sufficient for metadata queries`:
typo: `and next the next commit`

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -156,20 +157,22 @@ impl fmt::Display for InvalidCqlType {
...
+    /// [`ControlConnectionConfig`](crate::cluster::control_connection::ControlConnectionConfig).
+    pub(super) async fn query_metadata(&self) -> Result<Metadata, MetadataError> {
+        let connect_port = self.endpoint().address().port();
+
         let peers_query = self.query_peers(connect_port);
-        let keyspaces_query = self.query_keyspaces(keyspace_to_fetch, schema_metadata_fetch_mode);
+        let keyspaces_query = self.query_keyspaces(
+            &self.config.keyspaces_to_fetch,
+            self.config.schema_metadata_fetch_mode,
+        );
```
wprzytula, line 176: commit `Make ControlConnection self-sufficient for metadata queries`:
♻️ This commit would be IMO more readable if split into two:
1. Group config items into `ControlConnectionConfig`.
2. The remaining part of this refactor.

wprzytula: Actually, extracting the `ControlConnectionConfig` is not even mentioned in the commit message. It really shouldn't be part of this commit. I don't see how it's connected to moving ownership/borrowing between `MetadataReader` and `ControlConnection`.

wprzytula (commented):
commit `Introduce TopologyUpdateGuard`:
💯 Nice! Such type-level guards help maintain invariants a lot, and are zero-cost.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -552,60 +551,67 @@ impl ControlConnection {
         Ok(routes)
     }
 
-    /// Performs a partial fetch of `system.client_routes`. Partial means that filtering is done
-    /// not only by connection ids known to the driver (which is always the case), but also
-    /// by host ids - only for the hosts whose ids are present in the event payload.
+    /// Performs a partial fetch of `system.client_routes`, restricted to the given
+    /// (connection id, host id) pairs, as accumulated from
+    /// CLIENT_ROUTES_CHANGE:UPDATE_NODES events.
```
wprzytula: ❓ Why delete the definition of _partialness_ of the fetch?

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -552,60 +551,67 @@ impl ControlConnection {
...
-        }
-
-        // Although this is vaguely documented, the semantics of an event with connection ids [A, B, C] and host ids [X, Y, Z]
-        // is that the following entries were added/updated/removed: `[(A, X), (B, Y), (C, Z)]`.
-        // Unfortunately, we can't really query Scylla this way. Therefore, we do the query: `WHERE connection id IN ? AND host id IN ?`,
-        // which fetches possibly more routes than necessary, for example `(A, Z)` or `(C, Y)`.
-        // This is a tradeoff - the only alternative is issuing multiple queries, one per connection id.
-        // I believe the tradeoff here is correct.
-        let client_routes = self.query_client_routes(&connection_ids, host_ids).await?;
+        let client_routes = self.query_client_routes(&connection_ids, &host_ids).await?;
```
wprzytula: ♻️ This is arguably cleaner and avoids intermediate `Vec` allocation:
```rust
        let monitored_pairs = pairs
            .iter()
            .filter(|(conn_id, _host_id)| subscriber.get_connection_ids().contains(conn_id));

        // The pairs cannot be queried as such: the query is
        // `WHERE connection id IN ? AND host id IN ?`, which fetches the whole
        // cross product of the listed ids - possibly more routes than the
        // pairs name, for example `(A, Y)` for pairs `[(A, X), (B, Y)]`. This
        // is a tradeoff - the only alternative is issuing multiple queries,
        // one per connection id. The surplus routes are ignored when the
        // update is built from the pairs - see
        // [`ClientRoutesUpdate::from_pairs`].
        //
        // TODO: this is wasteful - it allocates strings and vecs.
        // This won't be a performance problem, because UPDATE_NODES events are not frequent.
        // As an optimization, we can implement ser/de for some special new iterator type,
        // to avoid the need to allocate when serializing collections.
        let (connection_ids, host_ids): (HashSet<String>, HashSet<Uuid>) =
            monitored_pairs.cloned().unzip();

        if connection_ids.is_empty() {
            // The events contained no relevant connection IDs.
            // Nothing to be done.
            return Ok(None);
        }

        let client_routes = self
            .query_client_routes(&Vec::from_iter(connection_ids), &Vec::from_iter(host_ids))
            .await?;
```

wprzytula: 📝 Also, conversion from `HashSet` into `Vec` will no longer be needed when I implement #1628.

scylla/src/cluster/metadata/update.rs
```diff
@@ -538,7 +547,7 @@ mod tests {
         let host_x = Uuid::new_v4();
         let host_y = Uuid::new_v4();
 
-        // The event lists (conn-1, host_x) — relevant — and (conn-other, host_y),
+        // The event lists (conn-1, host_x) - relevant - and (conn-other, host_y),
```
wprzytula: ❓ Out of curiosity: did you teach your LLM agent to replace all long hyphens with the ordinary hyphens?

scylla/src/cluster/metadata/update.rs
```diff
@@ -562,10 +571,10 @@ mod tests {
         );
     }
 
-    // `ClientRoutesUpdate::from_event`: an event pair with no matching route in the
-    // re-fetch yields an explicit `None` entry, i.e. a deletion.
+    // `ClientRoutesUpdate::from_pairs`: an event pair with no matching route in
+    // the re-fetch yields an explicit `None` entry, i.e. a removal.
```
wprzytula: ❓ Why change `deletion` into `removal`? Is there a meaningful difference between them?

scylla/src/cluster/metadata/worker.rs
```diff
@@ -60,6 +62,19 @@ pub(in super::super) struct MetadataWorker {
...
+enum EventAction {
+    /// Nothing to fetch - the event was fully handled synchronously.
+    None,
+    /// Fetch full metadata: the event describes a change (e.g. in topology)
+    /// that only a full fetch picks up.
+    FetchFull,
+    /// Fetch the client routes for these (connection id, host id) pairs, as
+    /// listed in a CLIENT_ROUTES_CHANGE:UPDATE_NODES event.
+    FetchClientRoutes { pairs: HashSet<(String, Uuid)> },
+}
```
wprzytula, line 116: commit `Extract synchronous event handling in MetadataWorker`:
🤔 Introducing a new enum is definitely not a _pure code motion_ declared by the commit message.
Also, the commit message is very hard for me to parse.

scylla/src/cluster/metadata/worker.rs
```diff
@@ -34,9 +37,8 @@ const CONTROL_CONNECTION_REPAIR_INTERVAL: Duration = Duration::from_secs(1);
 /// The control connection is therefore not a field, but a value owned by the
 /// currently running loop.
 pub(in super::super) struct MetadataWorker {
-    /// Fetches metadata and (re-)establishes control connections. Also keeps the
-    /// known peers and the client-routes subscriber, which it needs for
-    /// connection ids and event registration.
+    /// Establishes control connections (fetching metadata in the process) and
+    /// keeps the known peers to establish them to.
     metadata_reader: MetadataReader,
```
wprzytula: 💭 Isn't this now more a _ControlConnectionEstablisher_ than _MetadataReader_? Perhaps we should rename it accordingly?
- `ControlConnection` now reads metadata itself.
- The current `MetadataReader` just _establishes control connections and keeps the known peers to establish them to_, which is essentially managing opening control connections.

scylla/src/cluster/metadata/worker.rs
```diff
@@ -75,6 +82,214 @@ enum EventAction {
...
+                fetch: Box::pin(cc.query_metadata()),
+            };
+        }
+
+        // Partial client-routes work starts only when nothing at all is in
+        // flight: a running full fetch could not subsume it (it may have read
+        // the routes table already), and a running client-routes fetch
+        // serializes with it per type, so that newer results never lose to
+        // older ones.
+        if matches!(self, PendingFetches::Idle)
```
wprzytula: ⛏️ Consider adding an explicit `return` at the end of the first `if` to simplify reasoning about the code.

scylla/src/cluster/metadata/worker.rs
```diff
@@ -75,6 +82,214 @@ enum EventAction {
...
+            // `plan` cannot be `Full` here: were it, the full fetch would have
+            // just been started above. Taking the plan thus never discards
+            // owed work.
+            && let FetchPlan::Partial(request) = std::mem::take(plan)
+        {
+            *self = PendingFetches::Partial {
+                client_routes_fetch: Box::pin(request.fetch_on(cc)),
+            };
+        }
```
wprzytula: ♻️ For future safety, let's use exhaustive `match` instead of `if let` on the returned value of `std::mem::take()`. Write `unreachable!()` with explanations where appropriate.

wprzytula (commented):
Posting review for the first part (up until `Add regression test for event flood during metadata fetch`).

scylla/src/cluster/metadata/worker.rs
Lorak-mmk: Definitely, I was thinking about it myself and was planning to do it later. I can do it now if you want.

scylla/src/cluster/metadata/cc_establisher.rs
```diff
@@ -62,6 +62,26 @@ pub(crate) struct MetadataReader {
     cc_cache: Arc<ControlConnectionCache>,
 }
 
+/// The per-candidate metadata fetch that
+/// [`MetadataReader::establish_cc_and_fetch_metadata`] runs on each candidate
+/// connection - implemented by the metadata worker.
```
wprzytula, line 68: ❓ What the _candidate_ that this naming suggests? A candidate for what?

scylla/src/cluster/metadata/worker.rs
```diff
@@ -351,21 +355,83 @@ impl MetadataWorker {
...
+    ///
+    /// The rule of [`work_on_cc`](Self::work_on_cc) applies here too: the
+    /// connection's reader task delivers server events through a bounded
+    /// channel and blocks once it fills, no longer processing query responses
+    /// either - and events flow from the moment the connection is opened (it
+    /// REGISTERs during setup). Awaiting `query_metadata` without draining
+    /// events would thus let a burst of events deadlock the establishment.
+    /// Hence the fetch is awaited only in `select!`, with event draining as
+    /// the other branch.
+    async fn fetch_on_candidate(
```
wprzytula: > /// The per-candidate fetch of [`establish`](Self::establish):

🤔 How is `fetch_on_candidate` _the per-candidate fetch of `establish`_? I can't parse this.

scylla/src/cluster/metadata/worker.rs
```diff
@@ -351,21 +355,83 @@ impl MetadataWorker {
...
+            tokio::select! {
+                fetch_result = &mut fetch => {
+                    return fetch_result.map(|topology_update| (topology_update, plan));
+                }
+
+                cc_event = cc_events.wait_for_event() => match cc_event {
+                    ControlConnectionEvent::ServerEvent(event) => {
+                        // `Break` (the cluster worker is gone) is deliberately
+                        // ignored - see the doc comment.
+                        let _ = Self::absorb_server_event(updates, event, &mut plan);
```
wprzytula, line 427: 🤔  I can't see the explanation in the doc comment:

```rust
    /// The per-candidate fetch of [`establish`](Self::establish): runs the
    /// full metadata query while draining the candidate's server events.
    ///
    /// # Establishment never blocks either
    ///
    /// The rule of [`work_on_cc`](Self::work_on_cc) applies here too: the
    /// connection's reader task delivers server events through a bounded
    /// channel and blocks once it fills, no longer processing query responses
    /// either - and events flow from the moment the connection is opened (it
    /// REGISTERs during setup). Awaiting `query_metadata` without draining
    /// events would thus let a burst of events deadlock the establishment.
    /// Hence the fetch is awaited only in `select!`, with event draining as
    /// the other branch.
```

Lorak-mmk:
Rebased on main, adressed comments from the first part of the review.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: Nope, it did it all by itself. Reverted, sorry for that.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: Reverted

scylla/src/cluster/metadata/worker.rs
Lorak-mmk: I've split the commit into two. First extracts `publish_metadata`, the other introduces `EventAction` and `handle_server_event`. Messages were rewritten from scratch by me.

scylla/src/cluster/metadata/worker.rs
Lorak-mmk: Fixed.

Lorak-mmk:
Addressed comments for the first part of the PR.

scylla/src/cluster/metadata/cc_establisher.rs
Lorak-mmk: Rewrote the commit message, hopefully it explains that well.

Lorak-mmk:
Addressed most of the other review comments. Rewrote many commit messages to be more understandable. Some commits were split into 2 to make it easier to review.
One comment remains: https://github.com/scylladb/scylla-rust-driver/pull/1835#discussion_r3767404484 because I forgot why this error is ignored. I'll fix it shortly.

Lorak-mmk:
Addressed the last review comment.

wprzytula (approved):

patrycja-ziemkiewicz (approved):

Lorak-mmk:
I noticed that in the commit where I rename MetadataReader I forgot about the module and variables names. I updated it, and will push the new version shortly. This is a simple mechanical change only in internals, so I'll assume that approves still apply because I don't want to postpone the release any longer.

Diff of the change:
```patch
diff --git a/scylla/src/cluster/metadata/reader.rs b/scylla/src/cluster/metadata/cc_establisher.rs
similarity index 97%
rename from scylla/src/cluster/metadata/reader.rs
rename to scylla/src/cluster/metadata/cc_establisher.rs
index 829b85fc5..724d799c3 100644
--- a/scylla/src/cluster/metadata/reader.rs
+++ b/scylla/src/cluster/metadata/cc_establisher.rs
@@ -11,8 +11,8 @@
 //! - Iterating over known peers and initial contact points on connection failure
 //! - Host filtering to ensure the control connection is established to an accepted node
 //!
-//! Ownership of the established control connection lives outside the reader (in the
-//! metadata worker); the reader only knows how to create one. The metadata queries
+//! Ownership of the established control connection lives outside the establisher (in the
+//! metadata worker); the establisher only knows how to create one. The metadata queries
 //! themselves are methods on [`ControlConnection`], configured at its creation.
 
 use std::sync::Arc;
@@ -40,13 +40,13 @@ use crate::utils::safe_format::IteratorSafeFormatExt;
 
 /// Maintains the persistent state needed to create control connections and
 /// fetch cluster metadata. The established control connection itself is owned by
-/// the caller (the metadata worker), not by the reader.
+/// the caller (the metadata worker), not by the establisher.
 pub(crate) struct ControlConnectionEstablisher {
     // =======================================================================================
     // Configuration values - they will stay the same during whole lifetime of ControlConnectionEstablisher.
     // =======================================================================================
     control_connection_config: ConnectionConfig,
-    /// Configuration stamped onto every control connection this reader creates;
+    /// Configuration stamped onto every control connection this establisher creates;
     /// governs what the control connection's metadata queries fetch.
     cc_config: ControlConnectionConfig,
     hostname_resolution_timeout: Option<Duration>,
@@ -452,7 +452,7 @@ impl ControlConnectionEstablisher {
 /// Freshly fetched [`Metadata`] that the [`ControlConnectionEstablisher`] has not yet absorbed.
 ///
 /// [`ControlConnection::query_metadata`] returns its result wrapped in this
-/// guard, so that the fetched metadata cannot be used without the reader
+/// guard, so that the fetched metadata cannot be used without the establisher
 /// updating its known peers from it first: the only way to extract the
 /// [`Metadata`] is [`apply`](Self::apply).
 #[must_use = "the fetched metadata must be applied to the ControlConnectionEstablisher"]
@@ -465,10 +465,10 @@ impl TopologyUpdateGuard {
         Self { metadata }
     }
 
-    /// Updates the reader's known peers from the fetched metadata and releases
+    /// Updates the establisher's known peers from the fetched metadata and releases
     /// the metadata itself.
-    pub(super) fn apply(self, reader: &mut ControlConnectionEstablisher) -> Metadata {
-        reader.update_known_peers(&self.metadata);
+    pub(super) fn apply(self, establisher: &mut ControlConnectionEstablisher) -> Metadata {
+        establisher.update_known_peers(&self.metadata);
         self.metadata
     }
 }
diff --git a/scylla/src/cluster/metadata/fetching.rs b/scylla/src/cluster/metadata/fetching.rs
index 228ed346d..e08fa4968 100644
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -49,7 +49,7 @@ use crate::errors::{
 };
 use crate::routing::Token;
 
-use super::reader::TopologyUpdateGuard;
+use super::cc_establisher::TopologyUpdateGuard;
 
 type PerKeyspace<T> = HashMap<String, T>;
 type PerKeyspaceResult<T, E> = PerKeyspace<Result<T, E>>;
@@ -163,8 +163,8 @@ impl ControlConnection {
     /// [`ControlConnectionConfig`](crate::cluster::control_connection::ControlConnectionConfig).
     ///
     /// The result comes wrapped in a [`TopologyUpdateGuard`]: the fetched metadata
-    /// affects which peers the [`ControlConnectionEstablisher`](super::reader::ControlConnectionEstablisher)
-    /// establishes control connections to, so the reader must absorb it before
+    /// affects which peers the [`ControlConnectionEstablisher`](super::cc_establisher::ControlConnectionEstablisher)
+    /// establishes control connections to, so the establisher must absorb it before
     /// anything else can use it.
     pub(super) async fn query_metadata(&self) -> Result<TopologyUpdateGuard, MetadataError> {
         let connect_port = self.endpoint().address().port();
diff --git a/scylla/src/cluster/metadata/mod.rs b/scylla/src/cluster/metadata/mod.rs
index 26e48a2d8..18176bb6d 100644
--- a/scylla/src/cluster/metadata/mod.rs
+++ b/scylla/src/cluster/metadata/mod.rs
@@ -17,9 +17,9 @@
 //  - client routes:
 //    - [ClientRoute]
 
+pub(super) mod cc_establisher;
 mod fetching;
 pub(crate) mod merge_channel;
-pub(super) mod reader;
 pub(crate) mod update;
 pub(super) mod worker;
 
diff --git a/scylla/src/cluster/metadata/worker.rs b/scylla/src/cluster/metadata/worker.rs
index f811e4fc8..07f1ba2fc 100644
--- a/scylla/src/cluster/metadata/worker.rs
+++ b/scylla/src/cluster/metadata/worker.rs
@@ -20,8 +20,8 @@ use crate::frame::response::event::ClientRoutesChangeEvent;
 use crate::frame::response::event::EventV2 as Event;
 use crate::frame::response::event::StatusChangeEvent;
 
+use super::cc_establisher::{ControlConnectionEstablisher, FetchOnCandidate, TopologyUpdateGuard};
 use super::merge_channel;
-use super::reader::{ControlConnectionEstablisher, FetchOnCandidate, TopologyUpdateGuard};
 
 /// How often the worker attempts to establish a control connection while it has none.
 const CONTROL_CONNECTION_REPAIR_INTERVAL: Duration = Duration::from_secs(1);
@@ -41,7 +41,7 @@ const CONTROL_CONNECTION_REPAIR_INTERVAL: Duration = Duration::from_secs(1);
 pub(in super::super) struct MetadataWorker {
     /// Establishes control connections (fetching metadata in the process) and
     /// keeps the known peers to establish them to.
-    metadata_reader: ControlConnectionEstablisher,
+    cc_establisher: ControlConnectionEstablisher,
 
     // This value determines how frequently the metadata
     // worker will refresh the cluster metadata
@@ -324,13 +324,13 @@ impl Future for PendingFetches<'_> {
 
 impl MetadataWorker {
     pub(in super::super) fn new(
-        metadata_reader: ControlConnectionEstablisher,
+        cc_establisher: ControlConnectionEstablisher,
         cluster_metadata_refresh_interval: Duration,
         refresh_channel: tokio::sync::mpsc::Receiver<RefreshRequest>,
         updates: merge_channel::Sender<MetadataUpdate>,
     ) -> Self {
         Self {
-            metadata_reader,
+            cc_establisher,
             cluster_metadata_refresh_interval,
             refresh_channel,
             updates,
@@ -368,7 +368,7 @@ impl MetadataWorker {
         // `Session::connect`, which downstream crates compute the layout of.
         // One allocation per establishment attempt.
         let fetch: Pin<Box<dyn Future<Output = _> + Send>> = Box::pin(
-            self.metadata_reader
+            self.cc_establisher
                 .establish_cc_and_fetch_metadata(initial, &mut fetcher),
         );
         let (kept, metadata) = fetch.await?;
@@ -577,7 +577,7 @@ impl MetadataWorker {
 
         // The in-flight fetches. Their futures borrow `cc` and nothing else,
         // so the loop stays free to use `self` (in particular
-        // `self.metadata_reader`) while fetches are in flight.
+        // `self.cc_establisher`) while fetches are in flight.
         let mut pending_fetches = PendingFetches::Idle;
 
         loop {
@@ -597,7 +597,7 @@ impl MetadataWorker {
                     match fetch_outcome {
                         FetchOutcome::Full(Ok(topology_update)) => {
                             debug!("Fetched new metadata");
-                            let metadata = topology_update.apply(&mut self.metadata_reader);
+                            let metadata = topology_update.apply(&mut self.cc_establisher);
                             self.publish_metadata(metadata)?;
                         }
                         FetchOutcome::Full(Err(err)) => {
diff --git a/scylla/src/cluster/worker.rs b/scylla/src/cluster/worker.rs
index a70b4f761..235e80c3a 100644
--- a/scylla/src/cluster/worker.rs
+++ b/scylla/src/cluster/worker.rs
@@ -28,8 +28,8 @@ use tracing::{debug, error, info, trace};
 use uuid::Uuid;
 
 use super::metadata::Metadata;
+use super::metadata::cc_establisher::ControlConnectionEstablisher;
 use super::metadata::merge_channel::{self, merge_channel};
-use super::metadata::reader::ControlConnectionEstablisher;
 use super::metadata::worker::MetadataWorker;
 use super::state::ClusterState;
 
@@ -100,7 +100,7 @@ impl Cluster {
         let client_routes_subscriber = client_routes_address_translator
             .map(|translator| translator as Arc<dyn ClientRoutesSubscriber>);
 
-        let metadata_reader = ControlConnectionEstablisher::new(
+        let cc_establisher = ControlConnectionEstablisher::new(
             known_nodes,
             hostname_resolution_timeout,
             pool_config.connection_config.clone(),
@@ -117,7 +117,7 @@ impl Cluster {
         let (metadata_updates_sender, metadata_updates_receiver) = merge_channel();
 
         let mut metadata_worker = MetadataWorker::new(
-            metadata_reader,
+            cc_establisher,
             cluster_metadata_refresh_interval,
             refresh_receiver,
             metadata_updates_sender,
```

PR #1837, opened 2026-08-12
Control Connection client side timeouts

### LINKED ISSUE #1419: Client-side timeouts in CC
Control Connection does not use any execution profiles or client-side request timeouts. This makes it possible for the request to hang indefinitely.
We introduced support for Scylla's server-side timeout overriding. it would be great to also introduce client-side timeouts to 100% avoid indefinite hangs.
There is one danger in that: client-side timeout should not be shorted than server-side timeout. If it is, we risk a situation where we timeout a perfectly good request that the DB would soon respond to.

Two possibilities I see:
1. Set client-side timeout automatically based on server-side one (e.g. +0.5s).
2. Expose another API for client-side. Add a warning if it is lower than server-side.

First one is less configurable, and makes the name `metadata_request_serverside_timeout` a bit counterintuitive (because it also sets client-side timeout), but is fully resilient to the footgun I mentioned.
Second one is more configurable, and avoids the naming weirdness, but it is possible to encounter the issue if user ignores the warning.

Open questions:
How does this interact with Control Connection Policy that we want to implement at some point?
Will we want to expose more configuration of CC? Retries, execution profile, etc? The API may then become quite weird.

Lorak-mmk:
This PR introduces client-side timeouts to control connection queries.
The timeout is by default derived from server-side timeout:
- If server-side timeout is present, then 1s is added to it
- If it isn't present, a default of 30s is used

The timeout can be overwritten by the user.

Additionally, I changed the default server-side timeout from 2s to 30s.
Our default timeout is 30s, and we don't want to waste work (by dropping requests that db would respond to at some point). I don't think that metadata timeouts should be lower than the default one.
This may become more of a problem with sl:driver now released on server side, which is why I made the change now.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1419

Based on https://github.com/scylladb/scylla-rust-driver/pull/1830

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/control_connection.rs
+++ b/scylla/src/cluster/control_connection.rs
@@ -23,6 +23,23 @@ use crate::statement::prepared::PreparedStatement;
 
 const METADATA_QUERY_PAGE_SIZE: i32 = 1024;
 
+/// Timeouts applied to requests executed on the control connection.
+///
+/// The server-side timeout is only an override of the server's own limit, appended
+/// to the statement as a `USING TIMEOUT` clause; it is a ScyllaDB-only feature, so
+/// it is silently inapplicable to other targets (e.g. Cassandra).
+#[derive(Clone, Copy, Debug, Default)]
+pub(crate) struct MetadataRequestTimeouts {
+    pub(crate) serverside_override: Option<Duration>,
+}
+
+impl MetadataRequestTimeouts {
+    /// The server-side timeout override actually in effect.
+    fn serverside(&self, target_is_scylladb: bool) -> Option<Duration> {
+        self.serverside_override.filter(|_| target_is_scylladb)
+    }
+}
+
 pub(crate) type ControlConnectionCache = DashMap<String, PreparedStatement>;
 
 pub(crate) enum ControlConnectionEvent {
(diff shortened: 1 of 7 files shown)


scylla/src/cluster/control_connection.rs
```diff
@@ -142,12 +183,16 @@ impl ControlConnection {
...
+        let mut prepared: PreparedStatement = self
+            .get_or_prepare_statement(statement)
+            .await
+            .map_err(|attempt_err| {
+                NextRowError::NextPageError(NextPageError::RequestFailure(attempt_err.into()))
+            })?;
+
+        // Set on this per-use clone rather than in `get_or_prepare_statement`, so that
+        // the shared cache never bakes a timeout in.
+        prepared.set_request_timeout(Some(self.request_timeouts.clientside()));
```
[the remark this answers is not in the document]
Lorak-mmk: This is intentional. Client-side timeouts for PREPARE will be added some other time.

wprzytula:
Please rebase on main.

scylla/tests/integration/metadata/configuration.rs
```diff
@@ -24,7 +24,7 @@ use tokio::sync::mpsc::error::TryRecvError;
 use tracing::info;
 
 // By default, custom metadata request timeout is set to 2 seconds.
-const DEFAULT_TIMEOUT: Duration = Duration::from_secs(2);
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(30);
```
wprzytula, line 27: @mykaul @andrzej-jackowski-scylladb Do you think such default makes sense for metadata queries timeout?

wprzytula (approved):

scylla/tests/integration/metadata/configuration.rs
andrzej-jackowski-scylladb: @nuivall, do you have any opinion?

I think that `Our default timeout is 30s, and [...] I don't think that metadata timeouts should be lower than the default one.` is a reasonable point.

Lorak-mmk:
Rebased on main

Lorak-mmk:
Rebased on main

wprzytula (approved):

PR #1849, opened 2026-08-18
Preparations for partial metadata fetching

Lorak-mmk:
This PR performs various refactors and simplifications in preparations for partial metadata fetching https://github.com/scylladb/scylla-rust-driver/issues/1280

There should be no semantic change - definitely not any significant ones. Just refactors.
This is one of the rare PRs where AI was not used. Why? Recent changes in those ares (MetadataWorker, ClusterWorker etc) were quite rapid, and I feel that code quality and my understanding of it were a bit hurt as a result of that. This PR is en effort to improve this situation.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/1280

## Pre-review checklist

- [x] I have split my patch into logically separate commits.
- [x] All commit messages clearly explain what they change and why.
- [ ] ~~I added relevant tests for new features and bug fixes.~~
- [x] All commits compile, pass static checks and pass test.
- [x] PR description sums up the changes and reasons why they should be introduced.
- [x] I have provided docstrings for the public items that I want to introduce.
- [ ] ~~I have adjusted the documentation in `./docs/source/`.~~
- [x] I added appropriate `Fixes:` annotations to PR description.

PR #1850, opened 2026-08-19
Smarter metadata fetching: topology

Lorak-mmk:
Based on https://github.com/scylladb/scylla-rust-driver/pull/1849

This PR changes the behavior on TOPOLOGY_CHANGE events. Previously, full metadata was fetched on each such event.
This PR now introduces a partial fetch, so only the topology (`system.local` and `system.peers`) is fetched.

Additionally, I added the exact same fetch on STATUS_CHANGE. This is because nodes `rpc_address` can change with just a restart. If we get rid of a periodic topology fetch, we would miss such changes.

Last commit introduces a small optimization: if the new topology is identical, there is no reason to publish new ClusterState.

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -202,6 +202,27 @@ impl ControlConnection {
         }))
     }
 
+    /// Queries only the cluster topology: the peer list (`system.peers` and
+    /// `system.local`).
+    ///
+    /// The `cluster_name` that comes along the peers is dropped: it is fixed for
+    /// the lifetime of a cluster, so the one read by the last full fetch stands.
+    ///
+    /// The result comes wrapped in a [`TopologyUpdateGuard`] for the same reason
+    /// as in `query_metadata`: the establisher must absorb the new peer list
+    /// before anything else uses it.
+    pub(super) async fn query_topology(
+        &self,
+    ) -> Result<TopologyUpdateGuard<Vec<Peer>>, MetadataError> {
+        let connect_port = self.endpoint().address().port();
+
+        let (peers, _cluster_name) = self.query_peers(connect_port).await?;
+
+        Self::validate_peers(&peers)?;
+
+        Ok(TopologyUpdateGuard::new(peers))
+    }
+
     /// Rejects a peer list that could not describe a working cluster, so that
     /// the caller treats the fetch as failed instead of publishing it.
     fn validate_peers(peers: &[Peer]) -> Result<(), PeersMetadataError> {
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -192,15 +192,7 @@ impl ControlConnection {
 
         let (peers, cluster_name) = peers_and_cluster_name;
 
-        // There must be at least one peer
-        if peers.is_empty() {
-            return Err(MetadataError::Peers(PeersMetadataError::EmptyPeers));
-        }
-
-        // At least one peer has to have some tokens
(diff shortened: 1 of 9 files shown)


Lorak-mmk:
@coderabbitai review

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -192,22 +192,51 @@ impl ControlConnection {
...
         }
 
-        Ok(TopologyUpdateGuard::new(Metadata {
-            peers,
-            keyspaces,
-            cluster_name,
-            client_routes,
-        }))
+        Ok(())
     }
```
wprzytula, line 240: ♻️ Extraction of `validate_peers` from `query_metadata` could come as a pre-commit.

scylla/src/cluster/metadata/update.rs
```diff
@@ -61,6 +60,12 @@ pub(in super::super) struct PartialMetadataChanges {
     /// Partial client-routes snapshots fetched in response to
     /// CLIENT_ROUTES_CHANGE events.
     pub(in super::super) client_routes_updates: Option<ClientRoutesUpdate>,
+    /// The peer list fetched in response to TOPOLOGY_CHANGE events.
+    ///
+    /// Not partial within its own aspect: a topology fetch always reads the
+    /// whole peer list, so this replaces the topology of the state it is
+    /// applied to.
+    pub(in super::super) peers: Option<Vec<Peer>>,
```
wprzytula, line 68: ❓ But why? If we only fetch topology in reaction to NODE_ADDED / NODE_REMOVED events, then wouldn't it be enough to fetch topology for that given node?
Perhaps the problem lies in identification: the events carry an IP address, while we identify nodes by host ID?

scylla/tests/integration/utils.rs
```diff
@@ -727,3 +726,27 @@ pub(crate) fn inject_status_change_down(running_proxy: &RunningProxy, addr: Sock
...
+    );
+}
+
+/// Injects a forged `TOPOLOGY_CHANGE NEW_NODE` event for `addr`.
+pub(crate) fn inject_topology_change_new_node(running_proxy: &RunningProxy, addr: SocketAddr) {
+    inject_event_to_ccs(
+        running_proxy,
+        cluster_change_event_body("TOPOLOGY_CHANGE", "NEW_NODE", addr),
+    );
+}
```
wprzytula, line 752: ❓ What about `REMOVED_NODE`?

scylla/src/cluster/worker.rs
wprzytula: 🏕️ `wheter` -> `whether`

Lorak-mmk:
Rebased on main.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: Yes. I felt like fetching partial topology would open a can of edge-cases, for a small benefit.

scylla/tests/integration/utils.rs
Lorak-mmk: I can add it if you want. I just thought the value of such test is low (we handle all topology change events the same way). For status change it had more sense because we do different things on up and down.

Lorak-mmk:
- Addressed review comments
- Dropped the optimization that checked duplicate topology. I think we can do such things later, right now I'm aiming for a simple working solution.

Lorak-mmk:
@coderabbitai full review

wprzytula (approved):

Lorak-mmk:
Rebased on main

wprzytula (approved):

Issue #1433, opened 2025-09-08
Test "test_schema_await_with_unreachable_node" is flaky


Lorak-mmk:
It sometimes fails on dropping keyspace.
I did not investigate it a lot, but I see some possible issues with it:
- It assumes that dropping must work, but it is intended that even creating keyspace won't always work (see callsites of run_some_ddl_with_unreachable_node). Perhaps we should have a separate session for dropping?
- Test case 4 assumes that control connection is still opened to node 0. It could be broken by case 3.

Lorak-mmk:
Happened again: https://github.com/scylladb/scylla-rust-driver/actions/runs/18713576618/job/53367519182?pr=1445

Lorak-mmk:
I had an idea that maybe a connection is dropped in one test case, and session is not refilling it before next test case. 
However, run sent by @wprzytula is failing on a first test case, which is really surprising. 
Posting the log archive of this run below so that it doesn't get lost.

[logs_49215947399.zip](https://github.com/user-attachments/files/23544162/logs_49215947399.zip)

I looked at it, and afaict this is what is happening:
1. Session is correctly established, with connections to all 3 nodes through proxt (let's call them nodes 0, 1, 2).
2. We enter the first test case, and change proxy rules accordingly
3. CREATE KEYSPACE is sent to node 1 (as expected), and reaches the server
4. Non-error result is sent back from the server, and reaches the driver, so the keyspace was successfully created
5. Driver tries to await schema agreement, so it send a query to `system.local` to all 3 nodes
6. Query to node 1 is intercepted by the proxy, and connection dropped
7. Other queries complete successfully
8. Driver does not attempt another agreement check
9. Driver sends DROP KEYSPACE
10. Connection that was broken in step 6 is re-opened. I don't think it matters, but adding it for completeness
11. Driver receives error from DROP KEYSPACE
12. Driver receives two EVENTs related to schema changes in other tests
13. Driver receives EVENT about keyspace that we created in step 3 (!)
14. Driver reports error about keyspace we tried to drop not existing.

So, I see potentially two issues:
- Issue with a test: DROP KEYSPACE needs schema with keyspace to be propagated to the node we send it to. We can achieve this by either performing schema agreement after clearing proxy rules, or sending DROP KEYSPACE to the same node that we sent CREATE KEYSPACE to
[...]

PR #1853, opened 2026-08-20
Partial schema fetch

### LINKED ISSUE #1280: Smarter metadata fetching

Lorak-mmk:
This PR introduces partial schema fetching: driver can now fetch only the keyspaces affected by SCHEMA_CHANGE events, instead of all keyspaces.
This is the new behavior on metadata refresh intervals. Previously, all metadata was fetched then. Now only the affected schema is.
This finalizes the effort of smarter metadata fetching. We could try to make it more granular, but it would be much more difficult, and most likely fragile - I don't want to do that unless a strong reason comes up.

A nice property of the new system: on a stable cluster with no schema changes no metadata queries will be issued on metadata refresh intervals.

A new config is introduced: `PeriodicFetchMode`. By default, it is set to the new behavior above, but the user can set it to `FullMetadata` to get back the old behavior in case the CQL events turn out to not be reliable enough.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1280
Fixes: https://scylladb.atlassian.net/browse/DRIVER-764

## Pre-review checklist

### DIFF
--- a/docs/source/connecting/connecting.md
+++ b/docs/source/connecting/connecting.md
@@ -53,8 +53,14 @@ If you need to share `Session` with different threads / Tokio tasks etc. use `Ar
 
 ## Metadata
 
-The driver refreshes the cluster metadata periodically, which contains information about cluster topology as well as the cluster schema. By default, the driver refreshes the cluster metadata every 60 seconds.
-However, you can set the `cluster_metadata_refresh_interval` to a non-negative value to periodically refresh the cluster metadata. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
+The driver keeps an up-to-date view of the cluster topology (and client routes, if used) and of the cluster schema.
+Topology and client routes are refreshed in reaction to the server events announcing a change.
+Schema changes are different: the keyspaces that `SCHEMA_CHANGE` events name are accumulated and re-read by a periodic refresh, which batches the burst of events that a single DDL statement produces.
+
+That refresh runs every 60 seconds by default; `cluster_metadata_refresh_interval` sets the interval. A shorter one picks schema changes up sooner, at the cost of more frequent metadata queries.
+
+Only the affected keyspaces are re-read, not the whole schema.
+If your cluster's events cannot be relied upon, `periodic_metadata_fetch_mode(PeriodicFetchMode::FullMetadata)` makes each refresh re-read the whole metadata instead.
 
 
 ```{eval-rst}
--- a/scylla/src/cluster/metadata/fetching.rs
+++ b/scylla/src/cluster/metadata/fetching.rs
@@ -26,6 +26,7 @@ use std::time::Instant;
 use crate::frame::response::result::{ColumnSpec, TableSpec};
 use crate::parse_utils::{ParseErrorCause, ParseResult, ParserState};
 use futures::{FutureExt, Stream, StreamExt, TryStreamExt, future, stream};
+use itertools::{Either, Itertools};
 use rand::Rng;
 use tracing::{debug, trace, warn};
 use uuid::Uuid;
@@ -40,6 +41,7 @@ use crate::DeserializeRow;
 use crate::client::pager::QueryPager;
 use crate::cluster::NodeAddr;
 use crate::cluster::control_connection::ControlConnection;
+use crate::cluster::metadata::update::{FetchedKeyspace, SchemaUpdate};
 use crate::cluster::metadata::{ClientRoutes, ClientRoutesUpdate};
 use crate::deserialize::DeserializeOwnedRow;
 use crate::errors::{
@@ -513,6 +515,16 @@ impl TryFrom<ClientRoutesEntry> for super::ClientRoute {
     }
 }
 
+/// What the last SCHEMA_CHANGE event seen for a keyspace did to it.
+pub(crate) enum LastKeyspaceChange {
+    /// The keyspace itself was dropped, so it is to be removed from the schema
+    /// metadata rather than re-read.
+    Dropped,
+    /// Anything else - the keyspace was created or altered, or an object in it
+    /// was created, altered or dropped - so its metadata is to be re-read.
+    Altered,
+}
+
 impl ControlConnection {
     pub(super) async fn query_client_routes(
         &self,
@@ -635,6 +647,74 @@ impl ControlConnection {
             subscriber.get_connection_ids(),
             &client_routes,
         )))
+    }
+
+    /// Performs a partial fetch of the schema metadata: re-reads the schema
+    /// parts described by `request`.
+    ///
+    /// Each keyspace in `request` is marked as either dropped, or altered.
+    /// The altered keyspaces are fetched from the cluster.
+    /// The dropped ones will be reported as `FetchedKeyspace::Absent`,
+    /// without asking cluster for them.
+    /// If a keyspace marked as `altered` is not present in the cluster,
+    /// it will also be reported as `FetchedKeyspace::Absent`.
+    ///
+    /// Keyspaces in `request` are filtered by this control connection's configuration.
+    ///
+    /// Returns `None` when the configuration leaves nothing to do at all -
+    /// there is then no update to publish.
+    pub(super) async fn fetch_schema_update(
+        &self,
+        request: HashMap<String, LastKeyspaceChange>,
+    ) -> Result<Option<SchemaUpdate>, MetadataError> {
+        if matches!(
+            self.config.schema_metadata_fetch_mode,
+            SchemaMetadataFetchMode::Disabled
+        ) {
+            // No schema metadata is kept at all, so there is nothing to update.
+            return Ok(None);
+        }
+
+        // An empty `keyspaces_to_fetch` means "no restriction".
+        let is_fetched = |keyspace: &String| {
+            self.config.keyspaces_to_fetch.is_empty()
+                || self.config.keyspaces_to_fetch.contains(keyspace)
+        };
+
+        // Split the request into a list of keyspaces to fetch,
+        // and a map containing keyspaces that were dropped.
+        // Later this map will be extended with a fetch result.
+        let (to_fetch, mut keyspaces): (Vec<_>, HashMap<_, _>) = request
+            .into_iter()
+            .filter(|(name, _)| is_fetched(name))
+            .partition_map(|(name, state)| match state {
+                LastKeyspaceChange::Altered => Either::Left(name),
+                LastKeyspaceChange::Dropped => Either::Right((name, FetchedKeyspace::Absent)),
+            });
+
+        if to_fetch.is_empty() {
+            // Careful: an empty keyspace list means "every keyspace" to
+            // `query_keyspaces`, so it must not be called with one.
+            return Ok((!keyspaces.is_empty()).then_some(SchemaUpdate { keyspaces }));
+        }
+
+        let mut fetched = self
+            .query_keyspaces(&to_fetch, self.config.schema_metadata_fetch_mode)
+            .await?;
+
+        keyspaces.extend(to_fetch.into_iter().map(|keyspace| {
+            // A keyspace the fetch found no row for is gone: either the event
+            // announcing its drop is still on its way, or it was dropped
+            // without one. Either way it is dropped from the metadata rather
+            // than left stale.
+            let entry = match fetched.remove(&keyspace) {
+                Some(metadata) => FetchedKeyspace::Present(metadata),
+                None => FetchedKeyspace::Absent,
+            };
+            (keyspace, entry)
+        }));
+
+        Ok(Some(SchemaUpdate { keyspaces }))
     }
 
     fn query_filter_keyspace_name<'a, R>(
--- a/scylla/src/cluster/metadata/mod.rs
+++ b/scylla/src/cluster/metadata/mod.rs
@@ -48,6 +48,32 @@ pub(crate) enum SchemaMetadataFetchLevel {
     Full,
 }
 
+/// What the periodic cluster metadata refresh re-reads.
+///
+/// Set with
+/// [`SessionBuilder::periodic_metadata_fetch_mode`](crate::client::session_builder::SessionBuilder::periodic_metadata_fetch_mode);
+/// how often the refresh happens is a separate setting,
+/// [`SessionBuilder::cluster_metadata_refresh_interval`](crate::client::session_builder::SessionBuilder::cluster_metadata_refresh_interval).
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum PeriodicFetchMode {
+    /// The schema of the keyspaces that `SCHEMA_CHANGE` events named since the
+    /// previous refresh, and nothing else. The default.
+    ///
+    /// Every other aspect of the metadata is re-read in reaction to the server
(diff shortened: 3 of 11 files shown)


scylla/src/cluster/metadata/worker.rs
```diff
@@ -284,6 +357,17 @@ impl<'cc> PendingFetches<'cc> {
...
 
+        // The tick advances whether or not schema work is owed.
+        // We could only restart it on schema fetch. The advantage is that
+        // we would react quicker (immediately) to a lone change. The disadvantage
+        // is that if there is a lot of connected clients, they would all send the request to
+        // the server at the same moment. The main reason to have partial fetches is to
+        // make load on the server lighter in such scenario, so the disadvantage is more
+        // important.
+        if periodic_tick_due {
+            *next_refresh_deadline = deadline_after(Instant::now(), refresh_interval);
```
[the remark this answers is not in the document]
Lorak-mmk: As far as I can tell this was always the case. I don't think we need to fix it now.

scylla/tests/integration/metadata/partial_fetch.rs
```diff
@@ -137,3 +136,249 @@ async fn assert_event_triggers_only_a_topology_fetch(
...
+    session: &Session,
+    is_expected: impl Fn(&Arc<ClusterState>) -> bool,
+) -> Arc<ClusterState> {
+    loop {
+        let state = session.get_cluster_state();
+        if is_expected(&state) {
+            return state;
+        }
+        tokio::time::sleep(Duration::from_millis(50)).await;
+    }
```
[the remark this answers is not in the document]
Lorak-mmk: We have timeouts for the whole tests, so it is fine.

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -637,6 +649,74 @@ impl ControlConnection {
...
+        // Split the request into a list of keyspaces to fetch,
+        // and a map containing keyspaces that were dropped.
+        // Later this map will be extended with a fetch result.
+        let (to_fetch, mut keyspaces): (Vec<_>, HashMap<_, _>) = request
+            .into_iter()
+            .filter(|(name, _)| is_fetched(name))
+            .partition_map(|(name, state)| match state {
+                LastKeyspaceChange::Altered => Either::Left(name),
+                LastKeyspaceChange::Dropped => Either::Right((name, FetchedKeyspace::Absent)),
+            });
```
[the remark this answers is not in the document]
Lorak-mmk: What is reordered is drop events (table drop vs keyspace drop) which doesn't affect our correctness. It will only make the drop optimization useless in such case.

I don't know of a bug causing Scylla to reorder CREATE vs DROP.

Lorak-mmk:
@coderabbitai full-review

scylla/src/cluster/metadata/update.rs
wprzytula, line 1: commit `MetadataUpdate: Handle per-keyspace error`:
❓ IIUC, this is an improvement that was accidentally missed in previous PRs? And the enhancement is unrelated to the remaining part of the PR?

scylla/src/cluster/metadata/update.rs
```diff
@@ -279,6 +284,41 @@ impl ClientRoutesUpdate {
...
+    /// [`merge`](Self::merge).
+    pub(crate) keyspaces: HashMap<String, FetchedKeyspace>,
+}
+
+/// What a partial schema fetch established about one keyspace.
+// The `Present` payload is large (a `Keyspace` is a few hundred bytes) and
+// `Absent` is empty, but boxing it would trade an allocation per fetched
+// keyspace for a saving of a few hundred bytes per absent one - in a map that
+// holds the handful of keyspaces one refresh interval's events named, and lives
+// only until the update is applied.
```
wprzytula: > in a map that holds the handful of keyspaces one refresh interval's events named

❓ I fail to parse this grammar. Can you help me?

wprzytula: I also don't understand the explanation of the tradeoff chosen.

scylla/src/cluster/metadata/update.rs
wprzytula, line 168: ♻️ Can we deduplicate this per-keyspace result merging?

scylla/src/cluster/metadata/update.rs
```diff
@@ -195,6 +229,41 @@ impl MetadataUpdate {
...
+                update.metadata_changes = Some(MetadataChanges::Partial(partial));
+            }
+            Some(MetadataChanges::Partial(partial)) => partial.merge_schema_update(schema),
+            Some(MetadataChanges::Full {
+                metadata,
+                refresh_responses: _,
+            }) => {
+                // The partial fetch is necessarily newer, by the argument in
+                // `merge_topology_update`.
+                for (name, keyspace) in schema.keyspaces {
```
wprzytula, line 248: > // The partial fetch is necessarily newer, by the argument in `merge_topology_update`.

♻️ Can we write this argument instead of referencing it?

scylla/src/cluster/metadata/update.rs
```diff
@@ -319,6 +388,28 @@ pub(crate) enum FetchedKeyspace {
...
+                    if matches!(
+                        self.keyspaces.get(&name),
+                        Some(FetchedKeyspace::Present(Ok(_)))
+                    ) => {}
+                newer_keyspace => {
+                    self.keyspaces.insert(name, newer_keyspace);
+                }
+            }
+        }
+    }
```
wprzytula, line 411: ❓ Why does this mention `HashMap::extend`?

scylla/src/cluster/metadata/fetching.rs
```diff
@@ -637,6 +649,74 @@ impl ControlConnection {
...
+            // than left stale.
+            let entry = match fetched.remove(&keyspace) {
+                Some(metadata) => FetchedKeyspace::Present(metadata),
+                None => FetchedKeyspace::Absent,
+            };
+            (keyspace, entry)
+        }));
+
+        Ok(Some(SchemaUpdate { keyspaces }))
+    }
```
wprzytula: 🔧 I don't see `dropped` anywhere here (it's mentioned in the docstring).

scylla/src/cluster/metadata/worker.rs
```diff
@@ -284,6 +347,17 @@ impl<'cc> PendingFetches<'cc> {
             return;
         }
```
wprzytula, line 381: 💭 Shall we rename `next_refresh_deadline` to `next_schema_refresh_deadline`?

scylla/src/cluster/metadata/worker.rs
```diff
@@ -284,6 +347,17 @@ impl<'cc> PendingFetches<'cc> {
...
+        // The tick advances whether or not schema work is owed.
+        // We could only restart it on schema fetch. The advantage is that
+        // we would react quicker (immediately) to a lone change. The disadvantage
+        // is that if there is a lot of connected clients, they would all send the request to
+        // the server at the same moment. The main reason to have partial fetches is to
+        // make load on the server lighter in such scenario, so the disadvantage is more
+        // important.
+        if periodic_tick_due {
+            *next_refresh_deadline = deadline_after(Instant::now(), refresh_interval);
+        }
```
wprzytula, line 392: 📝 I believe the trade-off is correct here.

docs/source/connecting/connecting.md
```diff
@@ -53,8 +53,11 @@ If you need to share `Session` with different threads / Tokio tasks etc. use `Ar
 
 ## Metadata
 
-The driver refreshes the cluster metadata periodically, which contains information about cluster topology as well as the cluster schema. By default, the driver refreshes the cluster metadata every 60 seconds.
-However, you can set the `cluster_metadata_refresh_interval` to a non-negative value to periodically refresh the cluster metadata. This is useful when you do not have unexpected amount of traffic or when you have an extra traffic causing topology to change frequently.
+The driver keeps an up-to-date view of the cluster topology and of the cluster schema. Most of it is refreshed in reaction to the server events announcing a change, so no polling is involved. Schema changes are the exception: the keyspaces that `SCHEMA_CHANGE` events name are accumulated and re-read by a periodic refresh, which batches the burst of events that a single DDL statement produces.
```
wprzytula: 🤔 This is confusing:

> Most of it is refreshed in reaction to the server events announcing a change, so no polling is involved. Schema changes are the exception:

1. Schema changes are refreshed in reaction to server events;
2. Schema changes do not involve polling.

So they are not an exception to the description above.

scylla/src/cluster/metadata/mod.rs
```diff
@@ -48,6 +48,32 @@ pub(crate) enum SchemaMetadataFetchLevel {
...
+    AffectedKeyspaces,
+    /// All metadata selected by the session's schema-fetch configuration,
+    /// ignoring which keyspaces the `SCHEMA_CHANGE` events named.
+    ///
+    /// This is the behaviour of driver versions that did not handle
+    /// `SCHEMA_CHANGE` events, kept as an escape hatch for clusters whose
+    /// events cannot be relied upon. It costs a query per schema table per
+    /// refresh interval, hence it is not the default.
+    FullMetadata,
+}
```
wprzytula, line 75: I'm not sure we should derive `Copy`. It may bite us later.

scylla/tests/integration/metadata/partial_fetch.rs
```diff
@@ -9,9 +9,10 @@ use std::time::Duration;
 use scylla::client::session::Session;
```
wprzytula, line 9: commit `Add IT for PeriodicFetchMode::FullMetadata`
> Without it we would have to test for this code path.

🔧 I think it should be "no test".

wprzytula (changes requested):

scylla/src/cluster/metadata/worker.rs
Lorak-mmk: Not really - I added the config to perform the full refresh instead if user wants.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: Yes, but more of a bugfix than an enhancement. But the whole "per keyspace errors" is such an exotic edge case that I don't think its worth it to open an issue, mention in changelog separately etc.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: Rewritten the comment myself.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: Notice that the code is a bit different: `Some(FetchedKeyspace::Present(Ok(old_ks)))` vs `Some(Ok(old_ks))`. Introducing some abstraction for deduplicating this would imo do more bad than good.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: I don't think there is an argument needed, the fetched schema update is obviously newer than the previously fetched metadata. I'll adjust the comment.

scylla/src/cluster/metadata/update.rs
Lorak-mmk: Because it was used before I fixed per-keyspace errors handling, and I forgot to adjust the comment.

scylla/src/cluster/metadata/fetching.rs
Lorak-mmk: This was a stale doc after I changed the argument from 2 maps to a single map. Fixed.

docs/source/connecting/connecting.md
Lorak-mmk: Fixed.

scylla/src/cluster/metadata/mod.rs
Lorak-mmk: Thanks for being diligent about that!

Lorak-mmk:
Addressed most review comments, responded to others.

scylla/src/cluster/metadata/update.rs
wprzytula: OK.

wprzytula: OK.

wprzytula:
About yet another CI failure caused by #1433: Can't we just disable the test for Cassandra until we fix it? It's obviously flaky, and having it in the CI is more harm than benefit.

wprzytula (approved):
Great to see this optimisation finally in the driver! 💯

PR #1856, opened 2026-08-24
PM-344: point call_jira_sync.yml at github-automation main

dani-tweig:
`.github/workflows/call_jira_sync.yml` pinned the reusable Jira-sync workflow to commit `99e0314`, so fixes merged to `scylladb/github-automation` main never reached this repository.

This points it at `@main` instead:

```diff
-    uses: scylladb/github-automation/.github/workflows/main_pr_events_jira_sync.yml@99e0314ffd5efbebc137945c051f676628f340de # main
+    uses: scylladb/github-automation/.github/workflows/main_pr_events_jira_sync.yml@main
```

Only that one line changes — the `on:` triggers, `permissions:`, the `caller_action` input and the `caller_jira_auth` secret are all untouched, so no repository configuration is needed.

Fixes: PM-344

Part of PM-336.

### DIFF
--- a/.github/workflows/call_jira_sync.yml
+++ b/.github/workflows/call_jira_sync.yml
@@ -11,7 +11,7 @@ permissions:
 
 jobs:
   jira-sync:
-    uses: scylladb/github-automation/.github/workflows/main_pr_events_jira_sync.yml@99e0314ffd5efbebc137945c051f676628f340de # main
+    uses: scylladb/github-automation/.github/workflows/main_pr_events_jira_sync.yml@main
     with:
       caller_action: ${{ github.event.action }}
     secrets:


Lorak-mmk:
All our repos have "Require actions to be pinned to a full-length commit SHA" marked in actions settings, so this will fail (it didn't on this PR, because the action uses pull_request_target, so the definition from main branch is used, not the definition from this PR).

The correct solution is imo to setup Renovate to update the workflow regularly.

Lorak-mmk:
Btw Renovate should already work here. If it isn't, we need to investigate why.

roydahan (changes requested):
@dani-tweig leave it pinned but pointing to newer commit.

Lorak-mmk:
Please note that generally there should be no need to update it manually. We do have Renovate configured, and it works. We had a bump of this workflow very recently: https://github.com/scylladb/scylla-rust-driver/pull/1864 to f56f3669cce08717ab616b4f6f0f8a62990766ff , a commit from 24.08. Renovate will send updates for this action each Saturday.

Lorak-mmk:
Pre-emptively closing - if there is actually anything to do here please let me know.
Using Renovate for this is something we should do in all the other repos as well imo.

PR #1857, opened 2026-08-26
Add an example iterating the token ring and its replicas

### LINKED ISSUE #651: Add documentation / example about iterating over token ranges
I want to add efficient full table scans in Catalytic. According to https://www.scylladb.com/2017/02/13/efficient-full-table-scans-with-scylla-1-6/, an efficient way of doing this is, is by querying with `token(id)`.

There are some tips mentioned in that article how the ranges could be calculated, this involves the amount of cores and cpu’s. I was hoping this information can be queryed or is already available, so creating a method somewhere which returns the efficient token ranges would be easy.

ChrisJr404:
Adds a runnable example that walks the cluster's token ring and lists the replicas that own each token range, using `ReplicaLocator::ring` and `ClusterState::get_token_endpoints`. It follows the approach @wprzytula described on the issue.

Fixes: #651

## Pre-review checklist

### DIFF
--- a/examples/token_ring.rs
+++ b/examples/token_ring.rs
@@ -5,7 +5,7 @@
 //! at a given token starts just after the previous token in the ring and is owned
 //! by the replicas the cluster returns for any token inside it. Which nodes those
 //! are depends on the keyspace's replication strategy, so replica sets are looked
-//! up per keyspace and table.
+//! up per keyspace.
 
 use anyhow::Result;
 use scylla::client::session::Session;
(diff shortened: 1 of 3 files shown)


Lorak-mmk:
Why did you decide to remove our PR checklist?

ChrisJr404:
Sorry, that was unintentional. I've restored the pre-review checklist in the description.

ChrisJr404:
Sorry about that, I trimmed the checklist when I rewrote the description. It is restored in the PR body now with the pre-review items filled in.

On the bot's two notes about the example: both are fair. I will initialize `previous_token` from the ring's final entry so the wrapping range prints its real lower bound, and use an empty table name in `get_token_endpoints` so the example stays on the global vnode ring rather than picking up tablet routing on tablet-enabled keyspaces. Will push that update.

ChrisJr404:
Pushed the example update:

- `previous_token` is now seeded from the ring's last token, so the first (wrap-around) range prints its real lower bound instead of `(highest token, ...]`. An empty ring just skips the loop.
- `get_token_endpoints` is now called with an empty table name so the lookup stays on the keyspace-wide vnode ring rather than picking up per-table tablet routing on tablet-enabled keyspaces.

Compiles and passes fmt/clippy locally.

examples/token_ring.rs
```diff
@@ -0,0 +1,80 @@
+//! Walking the cluster's token ring and, for each token range, listing the
+//! replicas that own it.
+//!
+//! The token ring is a sorted, circular sequence of tokens. The range that ends
+//! at a given token starts just after the previous token in the ring and is owned
+//! by the replicas the cluster returns for any token inside it. Which nodes those
```
wprzytula, line 6: 🔧 This merits an example. Please write a sample token ring and describe which node owns which range.

examples/token_ring.rs
```diff
@@ -0,0 +1,80 @@
...
+        // Replicas that own the range ending at this token. `get_token_endpoints`
+        // resolves them through the keyspace's replication strategy, so a token
+        // range can have different replicas in different keyspaces. An empty table
+        // name keeps the lookup on the keyspace-wide vnode ring instead of any
+        // per-table tablet routing.
+        let replicas: Vec<NodeAddr> = cluster_state
+            .get_token_endpoints("examples_ks", "", *token)
+            .into_iter()
+            .map(|(node, _shard)| node.address)
+            .collect();
```
wprzytula, line 61: 🔧 Tablets are only mentioned here, as a side note. The distinction between tablet-based and vnodes-based routing, as well as interactions between both of them and the ring, should be central to this example.

examples/token_ring.rs
```diff
@@ -0,0 +1,80 @@
...
+            println!(
+                "Range ({}, {}] is owned by {:?}",
+                previous,
+                token.value(),
+                replicas
+            );
+        }
+
+        previous_token = Some(token.value());
+    }
```
wprzytula, line 75: 🔧 This **quietly** assumes vnodes, which is nowhere clearly stated, and not at all in the module's docstring.

wprzytula (changes requested):

PR #1863, opened 2026-08-28
connection: add tests for orphaning + transport factory

piodul:
There are currently no tests for the orphaning logic. Add some tests that use the proxy to simulate requests getting stuck on the DB side and verify that the orphaning logic reacts accordingly.

Additionally, introduce the TransportFactory object which allows connections created in the tests to communicate over DuplexStreams instead of TcpStreams. The DuplexStream is an in-memory bidirectional stream of bytes similar in function to a TCP stream but bypasses the system entirely. Thanks to this bypass, it is now possible to write tests that use tokio's paused timer and auto-advance functionality effectively - with TCP streams, even if the process communicates with itself, the task waiting for the data on the other side might not get woken up by the time tokio decides to auto-advance the timer - thinking that there is currently nothing to do. This functionality is actively used in the new tests but will definitely be useful in future tests.

## Pre-review checklist

### DIFF
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -7,6 +7,7 @@ use crate::frame::{
 use crate::{RequestOpcode, TargetShard};
 use bytes::Bytes;
 use compression::no_compression;
+use itertools::Either;
 use scylla_cql::frame::types::read_string_multimap;
 use std::collections::HashMap;
 use std::env::VarError;
@@ -15,9 +16,10 @@ use std::future::Future;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::sync::atomic::{AtomicBool, AtomicU8, AtomicU32, Ordering};
 use std::sync::{Arc, Mutex};
-use tokio::io::{AsyncRead, AsyncWrite};
+use tokio::io::{AsyncRead, AsyncWrite, DuplexStream};
 use tokio::net::{TcpListener, TcpSocket, TcpStream};
 use tokio::sync::mpsc::error::TryRecvError;
+use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tokio::sync::{broadcast, mpsc};
 use tracing::{debug, error, info, trace, warn};
 
@@ -343,7 +345,7 @@ impl Proxy {
         let (finish_guard, finish_waiter) = mpsc::channel(1);
 
         let (error_propagator, error_sink) = mpsc::unbounded_channel();
-        let (doorkeepers, running_nodes): (Vec<_>, Vec<RunningNode>) = self
+        let (addresses_and_doorkeepers, running_nodes): (Vec<_>, Vec<RunningNode>) = self
             .nodes
             .into_iter()
             .map(|node| {
@@ -365,29 +367,86 @@ impl Proxy {
                         cc_event_sender: cc_event_sender.clone(),
                     }
                 };
-                (
+                let proxy_addr = node.proxy_addr();
+                let doorkeeper = {
                     Doorkeeper::spawn(
                         node,
                         terminate_signaler.clone(),
                         finish_guard.clone(),
                         error_propagator.clone(),
                         cc_event_sender,
-                    ),
-                    running,
-                )
+                    )
+                };
+                let address_and_doorkeeper = async move { Ok((proxy_addr, doorkeeper.await?)) };
+                (address_and_doorkeeper, running)
             })
             .unzip();
 
-        for doorkeeper in doorkeepers {
-            doorkeeper.await?; // await doorkeeper creation, including binding to a socket
+        let mut duplex_submitter_map = HashMap::new();
+        for address_and_doorkeeper in addresses_and_doorkeepers {
+            let (proxy_addr, duplex_submitter) = address_and_doorkeeper.await?; // await doorkeeper creation, including binding to a socket
+            duplex_submitter_map.insert(proxy_addr, duplex_submitter);
         }
+
+        let transport_factory = Arc::new(TransportFactory {
+            duplex_submitter_map,
+        });
 
         Ok(RunningProxy {
             terminate_signaler,
             finish_waiter,
             running_nodes,
             error_sink,
+            transport_factory,
         })
+    }
+}
+
+/// Creates [`DuplexStream`]s that directly connect to the associated proxy's
+/// nodes.
+///
+/// This is an alternative method of establishing a transport layer connection
+/// to the proxy nodes. The purpose of this alternative method is to allow
+/// writing tests that use tokio's paused time feature. The advantage of using
+/// a channel is that, when some data is sent on the channel, the waiting
+/// task will be immediately notified and scheduled for execution. For
+/// system sockets this is not guaranteed and the receiving task may be
+/// notified after a real-time delay, which can trick tokio's timer auto-advance
+/// and make it move the time forward, even if the task on the receiving
+/// side would have been notified earlier in a real scenario.
+pub struct TransportFactory {
+    duplex_submitter_map: HashMap<SocketAddr, UnboundedSender<(DuplexStream, SocketAddr)>>,
+}
+
+impl TransportFactory {
+    pub fn connect(
+        &self,
+        connect_address: SocketAddr,
+        source_ip: Option<IpAddr>,
+        source_port: Option<u16>,
+    ) -> Result<tokio::io::DuplexStream, std::io::Error> {
+        let sender = self
+            .duplex_submitter_map
+            .get(&connect_address)
+            .ok_or_else(|| {
+                std::io::Error::new(
+                    std::io::ErrorKind::HostUnreachable,
+                    format!("No host under address {connect_address}"),
+                )
+            })?;
+
+        let source_ip = source_ip.unwrap_or(IpAddr::V4(Ipv4Addr::LOCALHOST));
+        let source_port = source_port.unwrap_or_else(|| rand::random_range(49152..=65535));
+        let source_address = SocketAddr::from((source_ip, source_port));
+
+        let (duplex1, duplex2) = tokio::io::duplex(8 * 1024);
+        sender.send((duplex2, source_address)).map_err(|_| {
+            std::io::Error::new(
+                std::io::ErrorKind::ConnectionRefused,
+                format!("The proxy node {connect_address} no longer accepts connections"),
+            )
+        })?;
+        Ok(duplex1)
     }
 }
 
@@ -502,6 +561,7 @@ pub struct RunningProxy {
     finish_waiter: FinishWaiter,
     pub running_nodes: Vec<RunningNode>,
     error_sink: ErrorSink,
+    transport_factory: Arc<TransportFactory>,
 }
 
 impl RunningProxy {
@@ -535,6 +595,11 @@ impl RunningProxy {
     /// Waits until an error occurs in proxy. If proxy finishes with no errors occurred, returns Err(()).
     pub async fn wait_for_error(&mut self) -> Option<ProxyError> {
         self.error_sink.recv().await
+    }
+
+    /// Returns the [`TransportFactory`] object associated with the proxy.
+    pub fn transport_factory(&self) -> Arc<TransportFactory> {
+        Arc::clone(&self.transport_factory)
     }
 
     /// Requests termination of all proxy workers and awaits its completion.
@@ -578,6 +643,11 @@ struct Doorkeeper {
     shards_count: Option<u16>,
     error_propagator: ErrorPropagator,
     cc_event_sender: Arc<Mutex<HashMap<usize, mpsc::UnboundedSender<ResponseFrame>>>>,
+    duplex_receiver: UnboundedReceiver<(DuplexStream, SocketAddr)>,
+    // The other side of the `duplex_receiver`. Not used directly by the doorkeeper
+    // but kept as a field in order to keep `duplex_receiver` open which allows
+    // to simplify some logic and not care about it getting closed.
+    _duplex_submitter: UnboundedSender<(DuplexStream, SocketAddr)>,
 }
 
 impl Doorkeeper {
@@ -587,7 +657,7 @@ impl Doorkeeper {
         finish_guard: FinishGuard,
         error_propagator: ErrorPropagator,
         cc_event_sender: Arc<Mutex<HashMap<usize, mpsc::UnboundedSender<ResponseFrame>>>>,
-    ) -> Result<(), DoorkeeperError> {
+    ) -> Result<UnboundedSender<(DuplexStream, SocketAddr)>, DoorkeeperError> {
         let listener = TcpListener::bind(node.proxy_addr())
             .await
             .map_err(|err| DoorkeeperError::DriverConnectionAttempt(node.proxy_addr(), err))?;
@@ -615,6 +685,7 @@ impl Doorkeeper {
             )
         };
 
+        let (duplex_submitter, duplex_receiver) = mpsc::unbounded_channel();
         let doorkeeper = Doorkeeper {
             shards_count: None, // temporarily, until Doorkeeper examines its ShardAwareness
             node,
@@ -623,9 +694,11 @@ impl Doorkeeper {
             finish_guard,
             error_propagator,
             cc_event_sender,
+            duplex_receiver,
+            _duplex_submitter: duplex_submitter.clone(),
         };
         tokio::task::spawn(doorkeeper.run());
-        Ok(())
+        Ok(duplex_submitter)
     }
 
     async fn run(mut self) {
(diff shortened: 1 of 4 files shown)


piodul:
Hmm, I used these tests to successfully do some refactors to orphaning on my local branch (and found some bugs with them), but indeed the tests are not entirely correct at the moment - some requests are not orphaned but instead they time out. I started debugging debugging them, changed the proxy to forge all requests, but now it looks like responses to the requests don't come quickly enough. I suspect it's because the test operates in paused time mode, responses are sent on a socket that has to go through docker networking and the responses arrive after a short delay, whereas the paused time mode decides to advance the clock by a couple of seconds because there seems to be no work to do.

Not sure how to fix this, maybe the communication between the connection and the proxy should go through an in-process channel instead of a socket. This would require some refactoring to the connection establishment code.

Lorak-mmk:
> I suspect it's because the test operates in paused time mode, responses are sent on a socket that has to go through docker networking and the responses arrive after a short delay, whereas the paused time mode decides to advance the clock by a couple of seconds because there seems to be no work to do.

Yep, this is why I've also never been able to write paused_time tests operating on a full Session. It would be great to solve this.

> Not sure how to fix this, maybe the communication between the connection and the proxy should go through an in-process channel instead of a socket. This would require some refactoring to the connection establishment code.

Yeah, it would require a refactor to:
-  be able to give connection something we could call a `connection_factory` that gets an IP and returns `impl AsyncRead + AsyncWrite`.
- Adjust scylla_proxy to also operate this way.

If we do that for connection, then there would be very little additional work required to pass it trough the session (behind an unstable feature) and be able to have paused_time tests with a full Session and proxy dry mode.

piodul:
> > Not sure how to fix this, maybe the communication between the connection and the proxy should go through an in-process channel instead of a socket. This would require some refactoring to the connection establishment code.
> 
> Yeah, it would require a refactor to:
> 
> * be able to give connection something we could call a `connection_factory` that gets an IP and returns `impl AsyncRead + AsyncWrite`.
> * Adjust scylla_proxy to also operate this way.
> 
> If we do that for connection, then there would be very little additional work required to pass it trough the session (behind an unstable feature) and be able to have paused_time tests with a full Session and proxy dry mode.

Yeah, I think I can do something like that. Looks like tokio has [DuplexStream](https://docs.rs/tokio/latest/tokio/io/struct.DuplexStream.html) which sounds like a valid replacement for a TcpStream in this case.

piodul:
v2:
- Added the `TransportFactory`
- Updated the tests
  - They now use the new `TransportFactory`
  - Updated the code to actually orphan some of the requests as they should have been (I did not apply `tokio::timeout` on some of them as I should have)
  - Removed `test_orphaning_wakes_up_again_correctly` as it tests behavior specific to a reimplementation of the orphaner that I have developed locally (spoilers!), I believe it's not as interesting for the current implementation
  - Reworked `test_orphaning_tolerates_many_short_lived_orphaned_requests`, added `test_orphaning_closes_connection_after_gradual_increase_of_orphaned_requests` which is a "negative" version of the other test case

piodul:
Automatic benchmarks reported a regression in the number of instructions executed in the "batch" benchmark (about 6%). I did not modify the request code path, only the connection establishment path. How reliable is the benchmark job, @Lorak-mmk ? Should I look into the regression?

piodul:
> How reliable is the benchmark job, @Lorak-mmk ?

Probably not very much, if PRs that merely _add examples_ get flagged... https://github.com/scylladb/scylla-rust-driver/pull/1857

piodul:
@coderabbitai full review

Lorak-mmk:
> Automatic benchmarks reported a regression in the number of instructions executed in the "batch" benchmark (about 6%). I did not modify the request code path, only the connection establishment path. How reliable is the benchmark job, @Lorak-mmk ? Should I look into the regression?

The benchmarks are quite new, and the stability is not great - probably for the similar reasons you initial tests failed, actually. So ignore them for now.

scylla-proxy/src/proxy.rs
```diff
@@ -874,11 +955,22 @@ impl Doorkeeper {
...
+                v = self.duplex_receiver.recv(), if !self.duplex_receiver.is_closed() => {
+                    // If the channel got closed then retry, in the next iteration
+                    // this branch of the `select!` statement will be disabled
+                    // by the `is_closed` condition.
+                    let Some((duplex, addr)) = v else { continue };
+                    (Either::Right(duplex), addr)
+                },
+            };
+        };
         info!(
```
wprzytula, line 976: ❓ Naive question: why is this logic so complicated?
I'd think that we either use TCP sockets or Duplex channels. Why allow both in a single select, racing them? Why not restrict this function to one specific kind of a channel, specified as a parameter?

scylla/src/network/connection.rs
```diff
@@ -3432,4 +3435,306 @@ mod tests {
...
+            mk_rule("DROP", RequestReaction::drop_frame()),
+            mk_rule(
+                "DELAY_HALF_THRESHOLD",
+                make_delayed_response(super::OLD_AGE_ORPHAN_THRESHOLD / 2),
+            ),
+            mk_rule(
+                "DELAY_DOUBLE_THRESHOLD",
+                make_delayed_response(2 * super::OLD_AGE_ORPHAN_THRESHOLD),
+            ),
+        ];
```
wprzytula, line 3482: ♻️ Let's extract those strings as constants.

scylla/src/network/connection.rs
```diff
@@ -3432,4 +3435,306 @@ mod tests {
...
+
+        // All queries should fail because of a timeout
+        for task in query_tasks {
+            let result = task.await.unwrap();
+            let _ = result.expect_err("One of the requests did not time out as expected");
+        }
+
+        // The orphaner should shut down the connection within one second
+        let err = error_receiver.await.unwrap();
+        assert_too_many_orphaned_streams_error(err);
```
wprzytula: 🔧 Let's mention `super::OLD_AGE_ORPHAN_THRESHOLD` instead of _one second_.

scylla/src/network/connection.rs
```diff
@@ -3432,4 +3435,306 @@ mod tests {
...
+
+    /// Verifies that a number of persistently orphaned requests that is large
+    /// but still below the threshold will not cause the connection to be closed.
+    #[tokio::test(start_paused = true)]
+    async fn test_orphaning_tolerates_long_running_orphaned_requests_under_count_threshold() {
+        setup_tracing();
+        let (_proxy, conn, mut error_receiver) = orphaning_test_setup().await;
+
+        let mut query_tasks = Vec::new();
+        for _ in 0..super::OLD_ORPHAN_COUNT_THRESHOLD / 2 {
```
wprzytula: 💭 Can we use `OLD_ORPHAN_COUNT_THRESHOLD` instead of `OLD_ORPHAN_COUNT_THRESHOLD / 2`? To check that the threshold is enforced exactly.

wprzytula (commented):
The tested scenarios are really well-thought-out, thanks!

scylla-proxy/src/proxy.rs
piodul: The duplex receiver is a mpsc channel receiver, so if all corresponding senders are closed then `recv()` will return `None`. I need to account for that and fall back to listening on the TCP socket only, if it happens. The loop is a bit clunky, but I think it works (maybe it would look nicer with tail recursion, if it gets stabilized and as long as it works in async).

I can change the proxy to listen on either TCP socket or on the Duplex-receiving channel, but:

- Users will have now to choose the Duplex mode explicitly when the proxy is created,
- I still need to handle the case when `duplex_receiver` gets closed, so the error handling logic will spill to higher layer so that the Doorkeeper gets gracefully closed without complaining in the logs - so I'll have to move the complexity somewhere else.

scylla/src/network/connection.rs
piodul: Yes, good point, will change that.

scylla-proxy/src/proxy.rs
wprzytula: I presume that the proxy will only use one transport kind during its lifetime. If this is wrong, correct me.

> The duplex receiver is a mpsc channel receiver, so if all corresponding senders are closed then `recv()` will return `None`. I need to account for that and fall back to listening on the TCP socket only, if it happens.

> The loop is a bit clunky, but I think it works (maybe it would look nicer with tail recursion, if it gets stabilized and as long as it works in async).
> 
> I can change the proxy to listen on either TCP socket or on the Duplex-receiving channel, but:
> 
>   * Users will have now to choose the Duplex mode explicitly when the proxy is created,

Seems OK. This is a distinct proxy's mode, similarly to the Dry Mode.

> 
>   * I still need to handle the case when `duplex_receiver` gets closed, so the error handling logic will spill to higher layer so that the Doorkeeper gets gracefully closed without complaining in the logs - so I'll have to move the complexity somewhere else.

I'm perfectly OK with the loop; I'm not OK with mixing TCP and Duplex receivers in one select, if only one of those is expected to ever deliver anything.

wprzytula: BTW, Opus says:

_**The complexity is avoidable without restructuring anything.** The whole loop/`is_closed`/`continue` dance exists only because `recv()` can return `None` on a closed channel and then keeps returning it immediately (busy-loop). But the sender side is a `HashMap` of `UnboundedSender`s owned by `TransportFactory`, which is owned by `RunningProxy` — so closure only happens at proxy teardown. If the `Doorkeeper` simply keeps its own clone of the corresponding `UnboundedSender` as a field, the channel can *never* be closed while the doorkeeper is alive, `recv()` never yields `None`, and `make_driver_stream` collapses to a plain two-branch `select!` with no loop, no guard, no `continue`:_

```rust
let (driver_stream, driver_addr) = tokio::select! {
    v = self.listener.accept() => {
        let (stream, addr) = v.map_err(...)?;
        (Either::Left(stream), addr)
    }
    v = self.duplex_receiver.recv() => {
        // unreachable: we hold a sender ourselves
        let (duplex, addr) = v.expect("duplex channel cannot close: Doorkeeper holds a sender");
        (Either::Right(duplex), addr)
    }
};
```

_That removes the author's *entire* stated justification for the complexity, and it does so without moving error handling up a layer or adding a mode flag — i.e. it dodges both of the objections they raised against your suggestion._

Is it right?

piodul: > I presume that the proxy will only use one transport kind during its lifetime. If this is wrong, correct me.

I guess it depends on what the future test writers will decide to do, you can create two connections where one will use a real socket and the other will work over DuplexStreams - I think that this should work perfectly fine. Not sure what would be the use case, though.

> Is it right?

Yes, I guess the doorkeeper keeps an `mpsc::Sender` as a field of the `Doorkeeper` then `recv()` is guaranteed not to return a `None` here. Seems slightly hacky to me, but maybe if I manage to restructure the code so that the sender is kept somewhere close to `make_driver_stream` in the code then this will reduce the distance of this spooky-action-at-a-distance and will be better.

Alternatively, if I change the code to listen only on one kind of receiver then I can change this async function to just hang indefinitely in case of a `None`. This will have the same effect as if the sender were never dropped but also nothing were sent through it anymore.

wprzytula: > > I presume that the proxy will only use one transport kind during its lifetime. If this is wrong, correct me.
> 
> I guess it depends on what the future test writers will decide to do, you can create two connections where one will use a real socket and the other will work over DuplexStreams - I think that this should work perfectly fine. Not sure what would be the use case, though.

OK, let's not restrict this. The complexity arising from this possible (although unlikely) use case is acceptable. 

> > Is it right?
> 
> Yes, I guess the doorkeeper keeps an `mpsc::Sender` as a field of the `Doorkeeper` then `recv()` is guaranteed not to return a `None` here. Seems slightly hacky to me, but maybe if I manage to restructure the code so that the sender is kept somewhere close to `make_driver_stream` in the code then this will reduce the distance of this spooky-action-at-a-distance and will be better.

Let's explore this direction, to still support the mixed-transport proxies.

> 
> Alternatively, if I change the code to listen only on one kind of receiver then I can change this async function to just hang indefinitely in case of a `None`. This will have the same effect as if the sender were never dropped but also nothing were sent through it anymore.

piodul: Done, I added a `_duplex_submitter` field to the `Doorkeeper`, in the end.

scylla/src/network/connection.rs
piodul: Done

scylla/src/network/connection.rs
piodul: Done, changed the comment to mention the named constant.

scylla/src/network/connection.rs
piodul: Done

piodul:
v3:
- Simplified the logic around `select!`
  - Now `Doorkeeper::spawn` creates the mpsc channel instead of `Proxy::run`, I thought it looks cleaner this way as otherwise I would have to pass two halves of the mpsc channel through `Doorkeeper::spawn`
- Extracted fake statement strings recognized by the proxy into constants
- Adjusted one comment to mention the orphan age threshold constant instead of its current literal value
- Adjusted one test to fill orphaned requests up to the count threshold instead only half

piodul:
v3.1: appeased clippy

piodul:
Indeed I missed the `config.local_ip_address` thing. I wonder why it wasn't pointed out by the bot when I first submitted the PR.

piodul:
v3.2: now passing `config.local_ip_address` instead of `None` to `TransportFactory::connect`

piodul:
> At all five sites, start_paused = true pauses Tokio time before orphaning_test_setup() resolves the database host, starts the proxy, and opens the connection. This real I/O can race with virtual-time advancement. Remove start_paused and call stop_the_clock() after setup.

There is some validity to this concern, but the tests aren't fully running in dry mode and are using a real scylla cluster to perform the handshake etc. Perhaps I should change the tests to run in dry mode but it's not nice that one has to manually forge the handshake in each of those tests (I can see that the dry tests do it). Perhaps the proxy should offer some helpers when in dry mode to simulate a proper handshake.

scylla-proxy/src/proxy.rs
```diff
@@ -365,32 +367,89 @@ impl Proxy {
...
+            .ok_or_else(|| {
+                std::io::Error::new(
+                    std::io::ErrorKind::HostUnreachable,
+                    format!("No host under address {connect_address}"),
+                )
+            })?;
+
+        let source_ip = source_ip.unwrap_or(IpAddr::V4(Ipv4Addr::LOCALHOST));
+        let source_port = source_port.unwrap_or_else(|| rand::random_range(49152..=65535));
+        let source_address = SocketAddr::from((source_ip, source_port));
```
wprzytula, line 440: 💭  While improbable, port collisions due to same number sampled are possible. Are we going to ignore them?

scylla-proxy/src/proxy.rs
```diff
@@ -537,6 +597,11 @@ impl RunningProxy {
         self.error_sink.recv().await
     }
 
+    /// Returns the [`TransportFactory`] object associated with the proxy.
+    pub fn transport_factory(&self) -> Arc<TransportFactory> {
+        Arc::clone(&self.transport_factory)
+    }
+
```
wprzytula, line 604: ⛏️ Just out of curiosity, why do you clone eagerly here instead of returning a reference to the Arc? It's not how the things are done in Rust I think.

piodul: I just wanted to make this API slightly easier to use. I anticipate that practically every time this method is used you will need an owned `Arc`, so I wanted to save a `.clone()` for the consumers of the API. I can change it to return a reference to an `Arc` if you think it will be better that way.

scylla-proxy/src/proxy.rs
piodul: Not sure, can a duplicate IP+source port combination confuse the proxy somehow?

piodul: If we want to be super thorough with this and prevent duplicate IP+port combinations then it will be difficult, especially if we want to allow TCP + duplex at the same time. If you have any TCP connections and you establish a duplex connection you need to make sure that you don't reuse a tuple already used by TCP, and vice versa. If the source port is unspecified then the operating system chooses the port, so you would have to somehow reserve the ports that are being fake-used by duplex streams.

If you disallow simultaneous TCP + duplex use then it sounds more doable, but still kind of a PITA - you would have to introduce a per-proxy-node allocator of IP+port tuples.

I'm not sure if this complexity is justified, but maybe I'm missing something and it is necessary.

scylla-proxy/src/proxy.rs
wprzytula: It's not important, it can stay this way.

scylla-proxy/src/proxy.rs
wprzytula: As you say, it's too much effort to really bother.

wprzytula (approved):

Lorak-mmk (approved):
Nice tests, thanks!

wprzytula (approved):

piodul:
I've sent a follow-up to make the tests run the proxy in dry mode: https://github.com/scylladb/scylla-rust-driver/pull/1873

PR #1866, opened 2026-08-31
Introduce minimal `BoundStatement`

### LINKED ISSUE #941: Introduce `BoundStatement`

wprzytula:
Fixes: #941
Based on: #1878

----------------------------

### Credits

This PR was inspired by #1321 by @nrxus. The original PR did two more things:
1. It replaced internal usage of `PreparedStatement` in the driver with `BoundStatement`.
    @Lorak-mmk didn't like this, so I dropped this part. Justification: in the future we might want to optimise serialization of values with `PreparedStatement` to elide the `SerializedValues` intermediate buffer and save allocation. This is not feasible with `BoundStatement`.
2. It introduced `BoundBatch`. This feature is TODO. 

----------------------------

# What

A _minimal_ implementation of `BoundStatement` is added. `BoundStatement` represents a prepared statement that has already been bound with values to be executed with. It is purposely a lightweight pair of both, with no separate config held.

Included:
- tests
- an example
- docs at `docs/source`.

### `BoundStatement` owns `PreparedStatement`, not borrows

The primary use cases for `BoundStatement` include:
1. keeping multiple statements with different bind parameter types inside a homogenous collection, typically a non-lifetime-parametrised one;
2. passing a full statement ready for execution across thread boundaries (which requires `'static` - references are forbidden).

So a borrowing `BoundStatement` would not help much.

### `BoundStatement` doesn't keep its own `StatementConfig`

This is too keep it lightweight. As it owns its `PreparedStatement`, it can configure it directly.

## TODOs

What I mean by _minimal_ is, among others, the following limitations/TODOs:

1. For now, there's no way to configure the `BoundStatement` after its creation. It only inherits the configuration from the `PreparedStatement` it's built from. #1876
2. We have an idea of introducing _Binders_ - an owned builder-like pattern for binding values one-by-one - either by appending, by arbitrary index, or by name. They are not yet done here. #1877
3. `BoundStatement` cannot be put into a `Batch`. `BoundBatch` is not a thing yet. #468

--------------------------

## Pre-review checklist


wprzytula:
Rebased on main.

scylla/src/statement/bound.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// Returns the token that would be computed for executing the provided bound statement.
+    pub fn calculate_token(&self) -> Result<Option<Token>, PartitionKeyError> {
+        self.extract_partition_key_and_calculate_token()
+            .map(|p| p.map(|(_, t)| t))
+    }
+
+    /// Returns the prepared statement behind the `BoundStatement`.
+    pub fn prepared(&self) -> &PreparedStatement {
+        &self.prepared
+    }
```
Lorak-mmk, line 46: My initial thought was that we could give access to `&mut PreparedStatement` so that users could configure it, bit it doesn't work of course: user could replace the statement, and the values would no longer be type-safe.

scylla/src/client/session.rs
```diff
@@ -865,6 +875,58 @@ impl Session {
...
+    /// As all results come in one response (no paging is done!), the memory footprint and latency may be huge
+    /// for statements returning rows (i.e. SELECTs)! Prefer this method for non-SELECTs, and for SELECTs
+    /// it is best to use paged requests:
+    /// - to receive multiple pages and transparently iterate through them, use [execute_iter](Session::execute_iter).
+    /// - to manually receive multiple pages and iterate through them, use [execute_single_page](Session::execute_single_page).
+    ///
+    /// Prepared statements are much faster than unprepared statements:
+    /// * Database doesn't need to parse the statement string upon each execution (only once)
+    /// * They are properly load balanced using token aware routing
+    ///
```
Lorak-mmk: User of this API is already using a prepare statement, do we need this note?

scylla/src/client/session.rs
```diff
@@ -990,6 +1128,56 @@ impl Session {
             .await
     }
 
+    /// Execute a bound statement with paging.\
+    /// This method will query all pages of the result.\
+    /// Requires a [BoundStatement] generated by binding values to a [PreparedStatement].
+    ///
+    /// Returns an async iterator (stream) over all received rows.\
+    /// Page size can be specified in the [PreparedStatement] passed to the function.
```
Lorak-mmk: Why did you add backslashes here?

docs/source/statements/bound.md
```diff
@@ -0,0 +1,143 @@
...
+# }
+```
+
+### Configuration
+
+The first iteration of the API deliberately exposes no options on
+`BoundStatement` itself. Configure the `PreparedStatement` (consistency, page
+size, execution profile, ...) *before* binding - the bound statement inherits
+all of its settings, and you can inspect them through
+`BoundStatement::prepared`.
```
Lorak-mmk: > The first iteration of the API deliberately

I don't think user-facing docs should mention such iterations. Just say the we currently don't expose those options on bound statement.

Lorak-mmk (changes requested):
> It introduced BoundBatch. This feature is TODO.

For BoundBatch we have https://github.com/scylladb/scylla-rust-driver/issues/468 so its fine to postpone it.

> For now, there's no way to configure the BoundStatement after its creation. It only inherits the configuration from the PreparedStatement it's built from.

Please open an issue for that if you don't plan to introduce that in this PR.

> We have an idea of introducing Binders - an owned builder-like pattern for binding values one-by-one - either by appending, by arbitrary index, or by name. They are not yet done here.

Ditto

scylla/src/client/session.rs
wprzytula: Agreed, no need for this.

scylla/src/client/session.rs
wprzytula: It's a leftover from old docstrings. When the PR was originally created a looong time ago, docstrings of other execution APIs contained backslashes. Now I forgot to remove them when rebasing on main.

docs/source/statements/bound.md
wprzytula: I'll change the first sentence into:
> `BoundStatement` does not expose configuration modifiers.

Others look appropriate to me.

wprzytula:
Rebased on main.

scylla/src/client/session.rs
wprzytula: Actually, `execute_iter` has them, too.

wprzytula:
> > It introduced BoundBatch. This feature is TODO.
> 
> For BoundBatch we have #468 so its fine to postpone it.
> 
> > For now, there's no way to configure the BoundStatement after its creation. It only inherits the configuration from the PreparedStatement it's built from.
> 
> Please open an issue for that if you don't plan to introduce that in this PR.

Opened #1876.
> 
> > We have an idea of introducing Binders - an owned builder-like pattern for binding values one-by-one - either by appending, by arbitrary index, or by name. They are not yet done here.
> 
> Ditto

Opened #1877.

scylla/src/client/session.rs
Lorak-mmk: I don't think there is a need for those backslashes. You can verify in rendered cargo doc. My guess is they can be safely removed.

wprzytula: https://github.com/scylladb/scylla-rust-driver/pull/1878

wprzytula:
Addressed comments.

scylla/src/client/session.rs
```diff
@@ -810,7 +815,8 @@ impl Session {
     /// * Database doesn't need to parse the statement string upon each execution (only once)
     /// * They are properly load balanced using token aware routing
     ///
-    /// > ***Warning***\
+    /// > ***Warning***
+    /// >
     /// > For token/shard aware load balancing to work properly, all partition key values
```
Lorak-mmk, line 821: Out of scope" we should be using warning blocks for that, as we do in e.g. `prepare`:
```
    /// <div class="warning">
    ///
    /// **Warning!**
    ///
    /// For token/shard aware load balancing to work properly, all partition key values
    /// must be sent as bound values
    /// (see [performance section](https://rust-driver.docs.scylladb.com/stable/statements/prepared.html#performance))
    ///
    /// </div>
    ///
```

Lorak-mmk (dismissed):

Lorak-mmk (approved):

PR #1870, opened 2026-08-31
treewide: Avoid relying on names of unnamed markers

### LINKED ISSUE #1869: Avoid reliance on synthesized bind-variable names
When unnamed bind marker is used in a statement (`?`), server must generate a name for that. It generally uses a column name for that, or something else that is reasonable. However, users should avoid relying on that, because those names are not really specified anywhere, so there is a risk of them changing between server releases.

We should make sure to not use such markers for name-binding in our examples / documentation / most tests.

JIRA: https://scylladb.atlassian.net/browse/DRIVER-899

Lorak-mmk:
In the binary CQL protocol, each bind marker in PREPARED response must have a name. If a user didn't provide one (because they used unnamed `?` marker), then server must generate one.
In general we should advise users to not rely on those names, in case they unexpectedely change between server versions. To do that, we need to avoid relying on those generated names in our examples / docs / most tests. 

This PR does just that. I intentionally left a test that specifically asserts simple unnamed markers are named after the column they bind to - `scylla/tests/integration/statements/prepared.rs:436 — test_prepared_statement_col_specs`. I think such test is useful to tell us if even such basic behavior breaks.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1869
Fixes: https://scylladb.atlassian.net/browse/DRIVER-899

## Pre-review checklist

### DIFF
--- a/docs/source/statements/values.md
+++ b/docs/source/statements/values.md
@@ -35,7 +35,7 @@ session
     .await?;
 
 // Sending an integer and a string using a named struct.
-// Names of fields must match names of columns in request,
+// Names of fields must match names of bind markers in the request,
 // but having them in the same order is not required.
 // If the fields are in the same order, you can use attribute:
 // `#[scylla(flavor = "enforce_order")]`
@@ -54,10 +54,10 @@ let int_string = IntString {
 };
 
 session
-    .query_unpaged("INSERT INTO ks.tab (a, b) VALUES(?, ?)", int_string)
+    .query_unpaged("INSERT INTO ks.tab (a, b) VALUES(:a, :b)", int_string)
     .await?;
 
-// You can use named bind markers in statement if you want
+// You can name bind markers differently than the columns if you want
 // your names in struct to be different than column names.
 #[derive(SerializeRow)]
 struct IntStringCustom {
--- a/examples/custom_serialization.rs
+++ b/examples/custom_serialization.rs
@@ -52,7 +52,7 @@ async fn main() -> Result<()> {
 
     session
         .query_unpaged(
-            "INSERT INTO examples_ks.custom_serialization (k, my) VALUES (?, ?)",
+            "INSERT INTO examples_ks.custom_serialization (k, my) VALUES (:k, :my)",
             to_insert,
         )
         .await?;
@@ -71,7 +71,7 @@ async fn main() -> Result<()> {
 
     session
         .query_unpaged(
-            "INSERT INTO examples_ks.custom_serialization (k, my) VALUES (?, ?)",
+            "INSERT INTO examples_ks.custom_serialization (k, my) VALUES (:k, :my)",
             to_insert_2,
         )
         .await?;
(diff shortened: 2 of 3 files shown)


docs/source/statements/values.md
```diff
@@ -54,7 +54,7 @@ let int_string = IntString {
 };
 
 session
-    .query_unpaged("INSERT INTO ks.tab (a, b) VALUES(?, ?)", int_string)
+    .query_unpaged("INSERT INTO ks.tab (a, b) VALUES(:a, :b)", int_string)
     .await?;
```
wprzytula, line 58: 😕  I don't understand the rationale at all.

> In the binary CQL protocol, each bind marker in PREPARED response must have a name. If a user didn't provide one (because they used unnamed `?` marker), then server must generate one.
> In general we should advise users to not rely on those names, in case
they unexpectedely change between server versions.

1. You're only modifying `query_unpaged()` invocations, which has nothing to do with `PREPARED` response.
2. You're saying that _users should not rely on those names_. I don't see how the code you're changing relied on those names. It only used `?`, so it accepted any (arbitrary) name. If the name change between server versions, this won't have any impact (or at least I'm unable to see any).

Lorak-mmk: The struct we pass here (`int_string`) is the following:
```rust
#[derive(SerializeRow)]
struct IntString {
    a: i32,
    b: String,
}
```

By default, `SerializeRow` uses by-name binding. This means it looks at each column in RowSerializationContext, takes its name, and looks for its field with the same name.

But what are those names in `RowSerializationContext`? Those are of course names of bind markers. But we did not provide names for bind markers - we are using `?`. So where do thosse names come from? They come from the server, and in this case the server uses the relevant column name for the bind marker name.

> You're only modifying query_unpaged() invocations, which has nothing to do with PREPARED response.

Of course they do. Those are unprepared queries with values, so driver has to prepare them transparently, and then it gets the PREPARED response it uses to create the RowSerializationContext.

> You're saying that users should not rely on those names. I don't see how the code you're changing relied on those names. It only used ?, so it accepted any (arbitrary) name. If the name change between server versions, this won't have any impact (or at least I'm unable to see any).

We are binding struct fields BY NAME, so we are relying on the fact that the server decided to name first `?` `a`, and the second `?` `b`.

Lorak-mmk:
Rebased on main

Lorak-mmk:
Changed the comments in docs because they were a bit wrong after the change.

docs/source/statements/values.md
```diff
@@ -54,10 +54,10 @@ let int_string = IntString {
...
 
-// You can use named bind markers in statement if you want
+// You can name bind markers differently than the coluymns if you want
```
wprzytula: 🔧 typo: columns

wprzytula (changes requested):

wprzytula (approved):

Lorak-mmk:
Rebased on main

wprzytula (approved):

PR #1874, opened 2026-09-01
Add a new OpenSSL variant of TlsContext

Lorak-mmk:
Our `TlsContext::OpenSsl010` variant accepts a ready `SslContext`. It is also a tuple variant, so there is no possibility to accept additional args other than the context.

This is bad for several reasons:
- We want to support TLS tickets. Configuring them properly requires using a `SslContextBuilder::set_new_session_callback`.
- We have no sound way to transform `SslContext` into `SslContextBuilder` (both use refcounted `SSL_CTX` underneath, builder guarantees refcount == 1, we have no control over refcount of user-given `SslContext`)
- Apart from the context, we may (and with high probability will) want to offer further customizations, for example using a hardcoded name for hostname verification. There is no space for such settings.

This PR adds a new `OpenSsl 0.10`-based `TlsContext` variant. This variant stores a new `OpenSsl010Context` struct.
This struct has private fields, and is constructed from either `SslContextBuilder` or `SslConnectorBuilder`.
In the future we can add builder-style methods to this struct (and I will do so for TLS tickets).

Ref: https://github.com/scylladb/scylla-rust-driver/issues/879
Ref: https://scylladb.atlassian.net/browse/DRIVER-170

## Pre-review checklist

### DIFF
--- a/docs/source/connecting/tls.md
+++ b/docs/source/connecting/tls.md
@@ -69,6 +69,10 @@ address in its subject alternative name. Replace that trust with `set_cert_store
 Beware that `set_ca_file` *adds* to the trusted set rather than replacing it, so it does not
 narrow anything down on its own.
 
+**_NOTE:_** Passing an already-built `SslContext` (`TlsContext::OpenSsl010`) still works, but is
+deprecated since 1.9.0: the driver cannot configure a context it did not build, so such a context
+cannot support TLS session tickets.
+
 For example, if database certificate is in the file `ca.crt`:
 ```rust
 # extern crate scylla;
--- a/docs/source/connecting/tls.md
+++ b/docs/source/connecting/tls.md
@@ -23,7 +23,7 @@ For rustls, we use `ServerName::IpAddress`, which is passed to `ClientConnection
 To enable use of TLS using `openssl`, add in `Cargo.toml`:
 
 ```toml
-scylla = { version = "0.4", features = ["openssl-010"] }
+scylla = { version = "1.8", features = ["openssl-010"] }
 openssl = "0.10.70"
 ```
 
--- a/docs/source/connecting/tls.md
+++ b/docs/source/connecting/tls.md
@@ -51,35 +51,48 @@ Then install the package with `openssl`:
     ```
 
 ### Using TLS
-To use TLS you will have to a `TlsContext`. For convenience, both an
-openssl
-[`SslContext`](https://docs.rs/openssl/0.10.33/openssl/ssl/struct.SslContext.html)
-and a rustls
+To use TLS you will have to create a `TlsContext`. For the openssl backend, build an
+[`SslConnector`](https://docs.rs/openssl/0.10/openssl/ssl/struct.SslConnector.html)
+builder and wrap it in an `OpenSsl010Config`; for rustls, an `Arc` of a
 [`ClientConfig`](https://docs.rs/rustls/latest/rustls/client/struct.ClientConfig.html)
-can be automatically converted to a `TlsContext` when passing to
-`SessionBuilder`.
+is automatically converted to a `TlsContext` when passing it to `SessionBuilder`.
 
-**_NOTE:_** Recommended API in `openssl` crate is `SslConnector`, because it has safer defaults. Please use it, and then call `into_context()` to
-get `SslContext` instance you can pass to the driver.
+**_NOTE:_** `SslConnector` is the recommended openssl API, because it has safe defaults
+(most notably, it verifies the peer's certificate). If you really need full control, you
+can build the context from a raw `SslContextBuilder` with
+`OpenSsl010Config::from_dangerous_builder`, which has no such defaults.
+
+**_NOTE:_** `SslConnector` also trusts the system's CA store, because it calls
+`set_default_verify_paths`. If your cluster has its own CA, that may be wider than you want: a
+certificate chaining to any public root would be accepted as long as it carries the node's IP
+address in its subject alternative name. Replace that trust with `set_cert_store`, as below.
+Beware that `set_ca_file` *adds* to the trusted set rather than replacing it, so it does not
+narrow anything down on its own.
 
 For example, if database certificate is in the file `ca.crt`:
 ```rust
 # extern crate scylla;
 # extern crate openssl;
-use scylla::client::session::Session;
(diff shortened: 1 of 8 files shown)


scylla/src/client/session.rs
```diff
@@ -154,16 +154,36 @@ impl std::fmt::Debug for Session {
...
     #[cfg(feature = "openssl-010")]
+    #[deprecated(
+        since = "1.9.0",
+        note = "the driver cannot configure a context it did not build, so this variant \
+                cannot support TLS session tickets; use `OpenSsl010Context` instead"
+    )]
     OpenSsl010(openssl::ssl::SslContext),
+    /// TLS context backed by OpenSSL 0.10, created from a builder.
+    #[cfg(feature = "openssl-010")]
+    OpenSsl010Builder(OpenSsl010Context),
```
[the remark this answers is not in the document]
Lorak-mmk: I'll wait for @wprzytula for an opinion about name. I am open to changing the name, but the argumentation above doesn't convince me: user doesn't see what is inside it. User only sees that this is constructed from a builder.

scylla/src/client/session.rs
```diff
@@ -154,16 +154,36 @@ impl std::fmt::Debug for Session {
 #[non_exhaustive]
 pub enum TlsContext {
     /// TLS context backed by OpenSSL 0.10.
+    ///
+    /// Deprecated in favour of [`TlsContext::OpenSsl010Builder`] / [`OpenSsl010Context`].
     #[cfg(feature = "openssl-010")]
+    #[deprecated(
+        since = "1.9.0",
+        note = "the driver cannot configure a context it did not build, so this variant \
+                cannot support TLS session tickets; use `OpenSsl010Context` instead"
```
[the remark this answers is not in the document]
Lorak-mmk: No need imo, it will be done in very next PR.

docs/source/connecting/tls.md
```diff
@@ -51,35 +51,48 @@ Then install the package with `openssl`:
     ```
```
wprzytula, line 51: commit's `tls: accept an openssl context builder, not a built context` message:

> Recommending `SslConnector` has one sharp edge worth the documentation
> it gets here: it pre-trusts the system CA store, and `set_ca_file` adds
> to that rather than replacing it. Every recipe therefore installs the
> cluster's CA with `set_cert_store`, which replaces the default store.

❓ What _recipe_ do you mean? Where does this happen? Is this user's responsibility or is it something that you hardened against in the driver?

wprzytula: The documentation below answers the last question, but the first two remain unanswered.

scylla/src/network/tls/openssl.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub fn from_dangerous_builder(mut builder: SslContextBuilder) -> Self {
+        Self::configure(&mut builder);
+        Self {
+            context: builder.build(),
+        }
+    }
+
+    /// The single place where the driver configures a context builder,
+    /// shared by both constructors.
+    fn configure(_builder: &mut SslContextBuilder) {}
```
wprzytula: 💭 I guess this `empty` configure is a stub for future configuration, right?
A comment clarifying this would help. Also, mentioning that `SslConnectorBuilder` derefs to `SslContextBuilder` is a crucial note to understand what's going on here.

scylla/src/network/tls/openssl.rs
wprzytula, line 1: ♻️ I dislike that this single commit both refactors (creates `openssl.rs` mod and extracts `new_ssl` function, among others) and introduces a new variant. Please split.

scylla/src/client/session.rs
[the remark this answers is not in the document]
wprzytula: User sees that (opaque) `OpenSsl10Context` is inside, and a _context_ is not a _builder_. This is what Copilot means. I think this is slightly inelegant.

docs/source/connecting/tls.md
```diff
@@ -23,7 +23,7 @@ For rustls, we use `ServerName::IpAddress`, which is passed to `ClientConnection
 To enable use of TLS using `openssl`, add in `Cargo.toml`:
 
 ```toml
-scylla = { version = "0.4", features = ["openssl-010"] }
+scylla = { version = "1.9", features = ["openssl-010"] }
```
wprzytula: 🔧 1.9 hasn't been released yet.
I see you're already forward-looking 😄

wprzytula (changes requested):
Looks correct overall but requires some fixes.

scylla/src/client/session.rs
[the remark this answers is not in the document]
Lorak-mmk: Do you have an idea for better name (/names)?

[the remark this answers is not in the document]
wprzytula: I'd rename both the variant and the held struct into `OpenSsl10Config`.

scylla/src/network/tls/openssl.rs
Lorak-mmk: There already was a comment mentioning dereferencing:
        // `SslConnectorBuilder` cannot be unwrapped into its `SslContextBuilder`,
        // but it dereferences to it, which is enough to configure it.

docs/source/connecting/tls.md
Lorak-mmk: Reworded this part of the message.

Lorak-mmk:
Addressed review comments, rebased on main

Lorak-mmk:
Added back a commit that went missing by accident

docs/source/connecting/tls.md
wprzytula: This one is still unresolved.

wprzytula (approved):
Apart from not-yet-released `scylla` version recommended, LGTM.

docs/source/connecting/tls.md
Lorak-mmk: It was intentional, I didn't think you want me to change that. But ok, I'll change it. I guess it won't fall back behind again if we grep the driver when bumping all the versions.

Lorak-mmk:
Fixed the last comment (Version)

wprzytula (approved):

PR #1875, opened 2026-09-01
Tls tickets support for OpenSsl

### LINKED ISSUE #879: Make use of TLS session resumption

Lorak-mmk:
This PR adds (very basic) TLS tickets support for our openssl backend.
The implementation here is aimed to match what Rustls does, so:
 - Ticket store is keyed by (translated) IP
 - Storing max 8 TLS 1.3 tickets per IP (+ 1 TLS 1.2 ticket)
 
 Store limits node count to 256, removing the least recently updated one when inserting a new one over the limit.

There are many deficiencies with those implementations, but its a good start - we'll at least support the mechanism itself.
We can improve it later, and I'll open an issue for that when we merge this PR, describing problems and potential solutions.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/879
Fixes: https://scylladb.atlassian.net/browse/DRIVER-170

## Pre-review checklist

### DIFF
--- a/docs/source/connecting/tls.md
+++ b/docs/source/connecting/tls.md
@@ -104,4 +104,31 @@ let session: Session = SessionBuilder::new()
 # }
 ```
 
+### TLS session tickets
+
+A context built through `OpenSsl010Config` resumes TLS sessions by default: the driver banks the
+session tickets the cluster issues (per node, respecting the TLS 1.3 single-use rule) and offers
+them on later connections. A TLS 1.2 ticket is reusable, so every connection after the first
+resumes; a TLS 1.3 ticket is spent on one connection, so only as many resume as the server issued
+tickets. The rustls backend does this natively, with no configuration.
+
+To turn it off:
+
+```rust
+# extern crate scylla;
+# extern crate openssl;
+use scylla::client::session::OpenSsl010Config;
+use openssl::ssl::{SslConnector, SslMethod};
+
+# use std::error::Error;
+# fn check_only_compiles() -> Result<(), Box<dyn Error>> {
+let builder = SslConnector::builder(SslMethod::tls())?;
+let context = OpenSsl010Config::new(builder).set_use_tls_tickets(false);
+# Ok(())
+# }
+```
+
+The deprecated `TlsContext::OpenSsl010` never resumes: the driver cannot install a session
+callback on a context it did not build.
+
 See the full [openssl example](https://github.com/scylladb/scylla-rust-driver/blob/main/examples/tls_openssl.rs) and [rustls example](https://github.com/scylladb/scylla-rust-driver/blob/main/examples/tls_rustls.rs) for more details.
--- a/docs/source/connecting/tls.md
+++ b/docs/source/connecting/tls.md
@@ -132,3 +132,19 @@ The deprecated `TlsContext::OpenSsl010` never resumes: the driver cannot install
 callback on a context it did not build.
 
 See the full [openssl example](https://github.com/scylladb/scylla-rust-driver/blob/main/examples/tls_openssl.rs) and [rustls example](https://github.com/scylladb/scylla-rust-driver/blob/main/examples/tls_rustls.rs) for more details.
+
+### TLS tickets limitations
+
+TLS 1.2 tickets should work perfectly fine.
+TLS 1.3 tickets are single-use, and:
+ - At the time of writing this, Scylla server offerc exactly 2 tickets on each connection
+ - Driver has no mechanism to take into consideration the amount of stored tickets when deciding when and how many connections to open.
+
+Because of that, TLS 1.3 tickets may often be not offered by the driver. Consider the following scenario:
+ - Node has 20 shards
+ - Driver opens first connection, gets 2 TLS 1.3 tickets.
+ - Driver opens remaining 19 connections.
+ - Only 2 will use tickets, 17 will perform full handshake.
+
+ This problem happens both with Rustls and Openssl backends.
+ Use TLS 1.2 for now if you care about tickets being offered whenever possible.
--- a/scylla/src/network/tls/openssl.rs
+++ b/scylla/src/network/tls/openssl.rs
@@ -1,5 +1,7 @@
 //! The OpenSSL 0.10 TLS backend of the driver.
 
+use std::net::SocketAddr;
+
 use openssl::error::ErrorStack;
 use openssl::ssl::{Ssl, SslConnectorBuilder, SslContext, SslContextBuilder};
 
@@ -48,15 +50,18 @@ impl OpenSsl010Config {
     fn configure(_builder: &mut SslContextBuilder) {}
 
     /// Creates a fresh per-connection [`Ssl`] object out of this context.
-    pub(crate) fn new_ssl(&self) -> Result<Ssl, ErrorStack> {
-        new_ssl(&self.context)
+    pub(crate) fn new_ssl(&self, node_address: SocketAddr) -> Result<Ssl, ErrorStack> {
+        new_ssl(&self.context, node_address)
     }
 }
 
 /// Creates a fresh per-connection [`Ssl`] object out of an [`SslContext`].
-pub(crate) fn new_ssl(context: &SslContext) -> Result<Ssl, ErrorStack> {
+pub(crate) fn new_ssl(context: &SslContext, node_address: SocketAddr) -> Result<Ssl, ErrorStack> {
     let mut ssl = Ssl::new(context)?;
     ssl.set_connect_state();
+    // Makes OpenSSL verify the node's certificate against its IP address.
+    // Corresponds to `X509_VERIFY_PARAM_set1_ip`.
+    ssl.param_mut().set_ip(node_address.ip())?;
     Ok(ssl)
 }
 
--- a/scylla/src/network/tls/openssl.rs
+++ b/scylla/src/network/tls/openssl.rs
@@ -1,11 +1,180 @@
 //! The OpenSSL 0.10 TLS backend of the driver.
 
-use std::net::SocketAddr;
+use std::collections::hash_map::Entry;
+use std::collections::{HashMap, VecDeque};
+use std::net::{IpAddr, SocketAddr};
+use std::sync::{Arc, LazyLock, Mutex};
 
 use openssl::error::ErrorStack;
-use openssl::ssl::{Ssl, SslConnectorBuilder, SslContext, SslContextBuilder};
+use openssl::ex_data::Index;
+use openssl::ssl::{
+    Ssl, SslConnectorBuilder, SslContext, SslContextBuilder, SslSession, SslSessionCacheMode,
+    SslSessionRef, SslVersion,
+};
 
 use crate::client::session::TlsContext;
+
+/// How many TLS 1.3 tickets we are willing to bank for a single node.
+///
+/// TLS 1.3 tickets are single use, so unused ones accumulate; without a cap a long-lived
+/// session talking to a chatty server would grow the store without bound. Eight matches
+/// rustls' own per-host limit, which makes the two backends behave alike under load.
+const MAX_TLS13_TICKETS_PER_HOST: usize = 8;
+
+/// How many nodes we are willing to bank sessions for.
+///
+/// Far above any real cluster's node count, so it only ever trips on churn: rolling
+/// restarts and elastic scale-in/out retire node IPs the driver will never reconnect to,
+/// whose tickets would otherwise be held (a DER `SSL_SESSION` embeds the peer's
+/// certificate, so ~10 kB each) for the process lifetime.
+const MAX_NODES: usize = 256;
+
+/// A single TLS session banked for a node, DER-encoded.
+#[derive(Clone)]
+struct NodeTicket(Vec<u8>);
+
+impl NodeTicket {
+    fn der(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+/// Sessions banked for a single node.
+#[derive(Default)]
+struct TicketBank {
+    /// The latest TLS 1.2 session, if any.
+    ///
+    /// RFC 5077 tickets are reusable - nothing forbids offering the same one again - so
+    /// we only keep the newest and hand out copies of it indefinitely.
+    tls12: Option<NodeTicket>,
+
+    /// TLS 1.3 tickets, oldest first.
+    ///
+    /// A TLS 1.3 ticket is single use: RFC 8446 appendix C.4 ("Client Tracking
+    /// Prevention") says clients SHOULD NOT reuse one, because a resumed ClientHello
+    /// carries the ticket as its `PskIdentity` and reuse makes connections linkable by a
+    /// passive observer. Hence a FIFO queue that a ticket leaves for good when used.
+    tls13: VecDeque<NodeTicket>,
+
+    /// Value of [`Nodes::clock`] at the last [`TicketStore::store`] for this node,
+    /// which orders nodes by staleness when the [`MAX_NODES`] cap is reached.
+    ///
+    /// A counter rather than an `Instant`: no clock reads, and it cannot tie.
+    stored_at: u64,
+}
+
+/// The [`TicketStore`]'s contents, all of it behind one lock.
+#[derive(Default)]
+struct Nodes {
+    tickets: HashMap<IpAddr, TicketBank>,
+    /// Ticks once per [`TicketStore::store`].
+    clock: u64,
+}
+
+/// TLS sessions harvested from completed handshakes, to be offered by later connections.
+///
+/// Keyed per node address, because a ticket is only resumable against the server that
+/// issued it. Offering a TLS 1.3 ticket to the wrong node would additionally spend it for
+/// nothing, as it is single use.
+/// We could key by host id, which would help in some cases where multiple translated IPs
+/// connect to a single node (e.g. some client routes deployments) but:
+/// - It would deviate from rustls behavior.
+/// - It would be more complicated (for example when handling connections for which
+///   host id is unknown, like control connection).
+///
+/// Sessions are kept as DER rather than as live `SslSession` objects. That is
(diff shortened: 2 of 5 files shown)


docs/source/connecting/tls.md
```diff
@@ -104,4 +104,31 @@ let session: Session = SessionBuilder::new()
 # }
 ```
```
wprzytula, line 105: commit message:

> Tickets are kept as DER and decoded per connection. OpenSSL marks a live
SSL_SESSION not-resumable when its connection is torn down without a clean
TLS shutdown, which is how every driver connection ends; a DER copy is
immune to that.

❓ I don't get the conclusion. I don't see what marking an SSL session as non-resumable has to do with a DER ticket, or rather why the ticket is immune to the problem of marking the session as non-resumable.

scylla/src/network/tls/openssl.rs
```diff
@@ -1,19 +1,182 @@
...
+    /// we only keep the newest and hand out copies of it indefinitely.
+    tls12: Option<Vec<u8>>,
+
+    /// TLS 1.3 tickets, oldest first.
+    ///
+    /// A TLS 1.3 ticket is single use: RFC 8446 appendix C.4 ("Client Tracking
+    /// Prevention") says clients SHOULD NOT reuse one, because a resumed ClientHello
+    /// carries the ticket as its `PskIdentity` and reuse makes connections linkable by a
+    /// passive observer. Hence a FIFO queue that a ticket leaves for good when used.
+    tls13: VecDeque<Vec<u8>>,
```
wprzytula: ⛏️ Can we have a newtype `NodeTicket` for `Vec<u8>`? I'm an enthusiast of strong typing raw bytes.

scylla/src/network/tls/openssl.rs
```diff
@@ -1,19 +1,182 @@
...
+        if nodes.tickets.len() >= MAX_NODES && !nodes.tickets.contains_key(&node_ip) {
+            let stalest = nodes
+                .tickets
+                .iter()
+                .min_by_key(|(_, tickets)| tickets.stored_at)
+                .map(|(&ip, _)| ip);
+            if let Some(ip) = stalest {
+                nodes.tickets.remove(&ip);
+            }
+        }
```
wprzytula, line 128: 💭 Since we need a minimal element from `tickets`, shouldn't we rather have `tickets` as `BTreeMap`, which would [support cheaply obtaining the lowest value](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.first_key_value)?

wprzytula: Otherwise churn scenarios will pessimistically involve iterating over 256 elements on each connection being opened. Likely perfectly acceptable, but since we are already using BTreeMap in the driver, why not use it here too?

scylla/src/network/tls/openssl.rs
```diff
@@ -37,19 +203,102 @@ impl OpenSsl010Context {
...
     /// shared by both constructors.
-    fn configure(_builder: &mut SslContextBuilder) {}
+    fn configure(builder: &mut SslContextBuilder, tickets: &Arc<TicketStore>) {
+        // `CLIENT`, because client-side session caching is off by default and without it
+        // the new-session callback would never be called at all. `NO_INTERNAL_STORE`,
+        // because our own store replaces OpenSSL's internal one - unlike OpenSSL's, it is
+        // keyed per node and it respects the TLS 1.3 single-use rule.
+        builder.set_session_cache_mode(
+            SslSessionCacheMode::CLIENT | SslSessionCacheMode::NO_INTERNAL_STORE,
+        );
```
wprzytula, line 232: ❓ Do you say that OpenSSL's store is incompatible with TLS 1.3?! It's shocking.

scylla/tests/integration/ccm/tls_tickets.rs
```diff
@@ -62,20 +66,29 @@ const CQL_PORT: u16 = 9042;
...
         match self {
+            // Resumes natively.
             Backend::Rustls023 => true,
-            // For either openssl flavour the driver builds a fresh `Ssl` from the
-            // `SslContext` per connection and never calls `SSL_set_session`.
-            Backend::OpenSsl010 | Backend::OpenSsl010Builder => false,
+            // The driver banks the sessions the server issues (via the new-session
+            // callback it installs on the context it builds) and, unless tickets are
+            // switched off, calls `SSL_set_session` with one on every new connection.
+            // With them off it still banks them, but offers nothing.
```
wprzytula: ❓ Why bank the tickets if they're explicitly turned off by user? It seems a waste of resources.

scylla/src/network/tls/openssl.rs
Lorak-mmk: Your wording could suggest that it works for other TLS versions. It does not. It is a server-only thing, useless for a client. There isn't even any function in openssl to access it.
<details>

<summary>From Opus</summary>

### 1. There is no client read path in the library at all                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 lookup_sess_in_cache() (ssl/ssl_sess.c:500) is the only function that reads ctx->sessions. It has exactly one caller — ssl_get_prev_session() at ssl_sess.c:633 — whose first line is:                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 ```c                                                                                                                                                                                                                                                                                                                                                                                                                                     
   int ssl_get_prev_session(SSL_CONNECTION *s, CLIENTHELLO_MSG *hello)                                                                                                                                                                                                                                                                                                                                                                    
   {                                                                                                                                                                                                                                                                                                                                                                                                                                      
       /* This is used only by servers. */                                                                                                                                                                                                                                                                                                                                                                                                
 ```                                                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 The external-cache escape hatch is server-only too — SSL_CTX_sess_set_get_cb(3): "(TLS server only.)" … "The get_session_cb() is only called on SSL/TLS servers". And SSL_CTX_set_session_cache_mode(3) states the policy outright:                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 │ SSL_SESS_CACHE_CLIENT: Client sessions are added to the session cache. As there is no reliable way for the OpenSSL library to know whether a session should be reused or which session to choose … the application must select the session to be reused by using the SSL_set_session(3) function.                                                                                                                                      
 │                                                                                                                                                                                                                                                                                                                                                                                                                                        
 │ SSL_SESS_CACHE_NO_INTERNAL_LOOKUP: … As automatic lookup only applies for SSL/TLS servers, the flag has no effect on clients.                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 ### 2. Even with a read path, the key is unusable                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 The cache is an LHASH keyed on (ssl_version, session_id) only. For a TLS 1.3 client session OpenSSL derives that id itself (statem_clnt.c:3335-3356):                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 ```c                                                                                                                                                                                                                                                                                                                                                                                                                                     
    * We choose the former approach because this fits in with assumptions                                                                                                                                                                                                                                                                                                                                                                 
    * elsewhere in OpenSSL. The session ID is set to the SHA256 hash of the ticket.                                                                                                                                                                                                                                                                                                                                                       
   if (!EVP_Digest(s->session->ext.tick, ticklen, s->session->session_id, &sess_len, sctx->sha256, NULL))                                                                                                                                                                                                                                                                                                                                 
 ```                                                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 Key = SHA256(ticket). To look a session up you must already possess the ticket. There is no host / IP / SNI dimension in the cache. SSL_CTX_sessions(3) also warns the LHASH "must not be modified directly".                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 ### 3. rust-openssl bolts the door anyway                                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 Present: add_session, remove_session (both unsafe), session_cache_size, the three callbacks. Absent from both openssl 0.10.81 and openssl-sys 0.9.117: SSL_CTX_sessions, all lh_* iteration, SSL_CTX_sess_number, SSL_CTX_flush_sessions, SSL_SESSION_is_resumable, SSL_SESSION_dup, SSL_SESSION_get0_hostname. Even the unsupported walk-the-LHASH-yourself route needs new FFI declarations. rust-openssl's own doc on                 
 SslSessionCacheMode::CLIENT repeats the upstream line, and its two client-side tests use precisely CLIENT | NO_INTERNAL — the config you shipped.                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 ### 4. The decisive one: the store cannot retain your tickets                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 ssl_sess.c:1321:                                                                                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 ```c                                                                                                                                                                                                                                                                                                                                                                                                                                     
   int ssl_clear_bad_session(SSL_CONNECTION *s)                                                                                                                                                                                                                                                                                                                                                                                           
   {                                                                                                                                                                                                                                                                                                                                                                                                                                      
       if ((s->session != NULL) && !(s->shutdown & SSL_SENT_SHUTDOWN) && !(SSL_in_init(...) || SSL_in_before(...))) {                                                                                                                                                                                                                                                                                                                     
           SSL_CTX_remove_session(s->session_ctx, s->session);                                                                                                                                                                                                                                                                                                                                                                            
 ```                                                                                                                                                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                          
 Called from ossl_ssl_connection_free (ssl_lib.c:1608) and ossl_ssl_connection_reset (ssl_lib.c:582). No close_notify → the session is yanked from the cache and stamped not_resumable = 1 — and remove_session_lock sets that flag outside the "was it in the cache" branch, identically in 1.1.1, 3.0, 3.5 and master. So NO_INTERNAL_STORE does not shield the live object either.

</details>

docs/source/connecting/tls.md
Lorak-mmk: Openssl tracks which ticket was issued on which connection. When a connection is uncleany closed (which we always do), tickets issued by it are invalidated.
This is a legacy mechanism, deemed unnecessary in TLS 1.1 standard. Openssl has it for legacy reasons.

The way around it is laundering the tickets, by transforming them to DER and back - that way the connection between ticket and connection can't be tracked by openssl.

scylla/src/network/tls/openssl.rs
wprzytula: 1. So now you say that _an internal OpenSSL's store_ is a server-only thing?
2. Why do they even turn it on by default on client then?
3. Shouldn't we turn the internal store off always then?

scylla/src/network/tls/openssl.rs
Lorak-mmk: We need to query by IP to get a ticket, but query by `stored_at` to get a minimum.

We can't replace the map with BTreeMap keyed by clock, because then we won't be able to query by IP.
We can have a BTreeMap in addition to HashMap, BTreeMap would map last store time to an ip, but imo it makes this code more complicated, and I doubt this will ever become a performance issue.

Can I leave it as-is?

scylla/src/network/tls/openssl.rs
Lorak-mmk: TBH I didn't investigate it that deeply. I did check, both myself and with AI, that we have no way to use it for our purposes. Anything else seems unrelated.

Lorak-mmk:
Addressed review comments.
Added a new commit that adds to `tls.md` description of TLS 1.3 tickets limitation.

Lorak-mmk:
Rebased on main

wprzytula (approved):

Issue #1859, opened 2026-08-27
Driver config reporting: Initial (downcasting-based) implementation


Lorak-mmk:
Scylla exposes all values received in a STARTUP frame in a `client_options` column of `system.clients` table.
We want to have drivers report their configuration there for easier debugging of various issues.

The current idea is to have each connection report a `SESSION_ID` uuid, to correlate various connections of a single session, and control connection to report a full JSON config under `DRIVER_CONFIG` key.

Schema for config reporting is present in the repo of each driver, e.g. https://github.com/scylladb/gocql/blob/master/docs/driver-config-schema.json
It allows for custom keys to be sent, to let drivers report their config that doesn't fit the common schema.

There is basically one place with non-obvious implementation: policies. 
I see 2 approaches:
 - Extend relevant traits with a new (optional) method that produces some Config object. Pro: custom policies can report properly, which is important for us because of wrappers. Con: backwards-compatibility risk if schema changes, more difficult designs.
 - Downcasting: check if the policy is one of our built-in policies, and report its config accordingly. Basically no backwards-compat risk, and easier impl, but custom policies can't produce their own report.
 
Other drivers (C#, gocql, java) went with a downcasting approach. I propose to do the same for now, and at some later time introduce a proper interface to improve reporting in the wrappers.

PR #1882, opened 2026-09-02
Send SESSION_ID in STARTUP

Lorak-mmk:
ScyllaDB stores every STARTUP option in system.clients.client_options. Until now nothing there told the connections of one session apart from those of another client, so a row could only be attributed by address and port - which reveals nothing once a client opens dozens of connections.

Generate a UUID per Session and report it on every connection, so that rows can be grouped by session, and expose it as Session::session_id() so the same value can be logged client side and the two correlated.

This will be necessary for Driver Config Reporting: report will only be sent on CC, and we need a way to connect the connection of a single Session.
This is a nice feature on its own, so it deserves a separate PR.

Ref: https://github.com/scylladb/scylla-rust-driver/issues/1859
Ref: https://scylladb.atlassian.net/browse/DRIVER-378

## Pre-review checklist

### DIFF
--- a/docs/source/connecting/connecting.md
+++ b/docs/source/connecting/connecting.md
@@ -51,6 +51,20 @@ Creating many `Session`'s in one application (e.g. `Session` per thread / per To
 
 If you need to share `Session` with different threads / Tokio tasks etc. use `Arc<Session>` - all methods of `Session` take `&self`, so it doesn't hinder the functionality in any way.
 
+## Session identification
+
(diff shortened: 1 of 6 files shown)


wprzytula (approved):

Lorak-mmk:
Rebased on main. Please re-approve.

Lorak-mmk:
Why did it close, wtf.

Lorak-mmk:
Huh, approve didn't get stale somehow, so I'm merging.

PR #1883, opened 2026-09-02
statement: make BoundStatement configurable

### LINKED ISSUE #1876: Allow configuring `BoundStatement` after it's constructed
## Problem Statement

#1866 introduced `BoundStatement`, but provided no way to change its configuration after binding. This means that all settings must be done on `PreparedStatement` before binding. In some cases this might be inconvenient (e.g. because the settings can only be known right before execution, which happens elsewhere than the binding).

## Naive Solution

Expose `prepared_mut(&mut self)` method on `BoundStatement`. While this would allow changing all the settings as desired, this would also create a hole in our type-level guarantees: a `BoundStatement` could be fed with a wrong `PreparedStatement` by calling `std::mem::swap()` on `prepared_mut()`. Thus this is invalid.

## Suggested Solution

Introduce setters to `BoundStatement`, essentially duplicating what `PreparedStatement` exposes. While verbose and boilerplatey, it's the best we can offer.

w3lld1:
## Summary

- I added the full set of public `PreparedStatement` configuration mutators to `BoundStatement`.
- Each method delegates to the owned prepared statement, so configuration can change after binding without exposing mutable access to it.
- I added regression coverage for setting and resetting every forwarded option after binding.
- I updated the bound-statement book chapter to document and demonstrate post-binding configuration.

## Validation

- `cargo check -p scylla --lib`
- `cargo test -p scylla --test integration --no-run`
- `cargo clippy -p scylla --lib -- -D warnings -A unfulfilled-lint-expectations`
- `cargo test -j 2 -p scylla --doc book_tests::statements::bound` (4 passed)
- `git diff --check upstream/main...HEAD`

The integration test binary compiles successfully. I could not run the database-backed integration test locally because this environment cannot access the Docker socket, so I am leaving the runtime integration suite to CI.

## Pre-review checklist

- [ ] All commits compile, pass static checks and pass test. (Compilation and static checks pass; the database-backed runtime test is deferred to CI as noted above.)

Fixes #1876

### DIFF
--- a/docs/source/statements/bound.md
+++ b/docs/source/statements/bound.md
@@ -110,9 +110,9 @@ session.execute_unpaged(&prepared, (12345,)).await?; // serializes them again
 
 ### Configuration
 
-`BoundStatement` does not expose configuration modifiers. Configure the `PreparedStatement`
-(consistency, page size, execution profile, ...) *before* binding - the bound statement
-inherits all of its settings, and you can inspect them through `BoundStatement::prepared`.
+`BoundStatement` inherits the prepared statement's configuration when it is created. You can
+also update options such as consistency, page size, and execution profile directly on the bound
+statement before execution.
 
 ```rust
 # extern crate scylla;
@@ -121,15 +121,12 @@ inherits all of its settings, and you can inspect them through `BoundStatement::
 # async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
 use scylla::statement::Consistency;
 
-let mut prepared = session
+let prepared = session
     .prepare("INSERT INTO ks.tab (a) VALUES(?)")
     .await?;
 
-// Set the options first...
-prepared.set_consistency(Consistency::One);
-
-// ...then bind. The bound statement will be executed with Consistency::One.
-let bound = prepared.bind(&(12345,))?;
+let mut bound = prepared.bind(&(12345,))?;
+bound.set_consistency(Consistency::One);
 assert_eq!(bound.prepared().get_consistency(), Some(Consistency::One));
 
 session.execute_bound_unpaged(&bound).await?;
--- a/docs/source/statements/bound.md
+++ b/docs/source/statements/bound.md
@@ -121,13 +121,16 @@ statement before execution.
 # async fn check_only_compiles(session: &Session) -> Result<(), Box<dyn Error>> {
 use scylla::statement::Consistency;
 
-let prepared = session
+let mut prepared = session
     .prepare("INSERT INTO ks.tab (a) VALUES(?)")
     .await?;
+prepared.set_consistency(Consistency::Quorum);
+prepared.set_page_size(100);
 
 let mut bound = prepared.bind(&(12345,))?;
 bound.set_consistency(Consistency::One);
 assert_eq!(bound.prepared().get_consistency(), Some(Consistency::One));
+assert_eq!(bound.prepared().get_page_size(), 100);
 
 session.execute_bound_unpaged(&bound).await?;
 # Ok(())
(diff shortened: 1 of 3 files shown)


wprzytula:
>     * [ ]  I have adjusted the documentation in `./docs/source/`. (No book changes are needed; the new public API is documented directly in its Rust docstrings.)

False. The book mentions that `BoundStatement` doesn't allow configuration after it's been created.

w3lld1:
Thanks — updated the book in `134019ba`. The configuration section now documents that a bound statement can be configured after binding and demonstrates `BoundStatement::set_consistency()` directly.

Validation:
- `cargo test -j 2 -p scylla --doc book_tests::statements::bound` — 4 passed
- `git diff --check`

scylla/tests/integration/statements/bound.rs
```diff
@@ -210,6 +211,77 @@ async fn bound_statement_inherits_prepared_statement_config(session: &Session) {
...
+    bound.set_load_balancing_policy(None);
+    bound.set_execution_profile_handle(None);
+
+    let reset = bound.prepared();
+    assert_eq!(reset.get_consistency(), None);
+    assert_eq!(reset.get_serial_consistency(), None);
+    assert!(reset.get_retry_policy().is_none());
+    assert!(reset.get_load_balancing_policy().is_none());
+    assert!(reset.get_execution_profile_handle().is_none());
+}
```
wprzytula, line 296: 🔧 Let's make this test ensure that the new configuration specified on `BoundStatement` overrides what was explicitly set before on the `PreparedStatement`, and that those config parts that are not overridden remain unchanged.

docs/source/statements/bound.md
```diff
@@ -121,15 +121,12 @@ inherits all of its settings, and you can inspect them through `BoundStatement::
...
-// Set the options first...
-prepared.set_consistency(Consistency::One);
-
-// ...then bind. The bound statement will be executed with Consistency::One.
-let bound = prepared.bind(&(12345,))?;
+let mut bound = prepared.bind(&(12345,))?;
+bound.set_consistency(Consistency::One);
 assert_eq!(bound.prepared().get_consistency(), Some(Consistency::One));
 
 session.execute_bound_unpaged(&bound).await?;
```
wprzytula, line 135: 🔧 This example, for completeness, should show both configuring `PreparedStatement` before binding and `BoundStatement` after binding.

wprzytula (changes requested):

w3lld1:
Thanks — addressed both follow-up points in `09500198`:

- the integration test now starts with explicit prepared-statement configuration, verifies that bound-statement overrides win, and verifies that a non-overridden option remains inherited;
- the book example now demonstrates the same before/after-binding flow.

Validation: `cargo fmt --all -- --check` and `git diff --check`. The database-backed integration test remains deferred to CI.

wprzytula:
Please rebase on main.

w3lld1:
Rebased the branch onto the current `main` in `a32add37`.

Fresh validation on the rebased head:
- `cargo check -j 2 -p scylla`
- `git diff --check origin/main...HEAD`

wprzytula (approved):

Lorak-mmk (approved):

Lorak-mmk:
> Validation

This section is so fascinating to me. It is completely and utterly useless, yet it appears in all automated PRs. Why is AI always producing it?

Lorak-mmk:
Rebased on main

wprzytula:
> > Validation
> 
> This section is so fascinating to me. It is completely and utterly useless, yet it appears in all automated PRs. Why is AI always producing it?

AI agents rely on the section, trusting it, not to duplicate the checks themselves. This is what I observed. They tend to complain when the `Validation`/`Verification` section is missing from the cover letter.

Lorak-mmk:
PR cover letter is not a good place for it. For working locally, they should use their own notes. For assertion of tests passing on a PR, the CI status is the answer, not the cover letter.

PR #1885, opened 2026-09-02
Use one ClusterState per request

### LINKED ISSUE #1818: Use one ClusterState per request
When executing request, we should fetch `ClusterState` at one point and use it for whole request. If we use one state for some things, and other state for other things, our model of statement execution and its invariants are gone.

I noticed that in `Session::execute`:
- We fetch state for tracing purposes:
```rust
        if !span.span().is_disabled()
            && let (Some(table_spec), Some(token)) = (routing_info.table, token)
        {
            let cluster_state = self.get_cluster_state();
            let replicas = cluster_state.get_token_endpoints_iter(table_spec, token);
            span.record_replicas(replicas)
        }
```
- But also fetch it in `Session::run_request` 

This specific case is very minor: other state is only used for tracing. We should however check if this is the only case - there may be others, more serious.

Lorak-mmk:
This PR moves ClusterState load on execution path out of run_request, and applies relevant necessary changes:
1. Tablet hint is calculated in `execute`, and no longer passed as an argument to `run_request_once` lambda.
2. Trace log that printed replicas of prepared statement now uses the same ClusterState as the whole request.

Additionally, I made a small fix to recent tablets v2 PR: Now the driver tries to compute a tablet hint for transparently-prepared QUERY request. If it fails, we just fallback to 0.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1818

## Pre-review checklist

- [ ] ~~I have adjusted the documentation in `./docs/source/`.`~~

### DIFF
--- a/scylla/src/client/session.rs
+++ b/scylla/src/client/session.rs
@@ -1273,6 +1273,8 @@ impl Session {
             ));
         }
 
+        let cluster_state = self.get_cluster_state();
+
         let execution_profile = batch
             .get_execution_profile_handle()
             .unwrap_or_else(|| self.get_default_execution_profile_handle())
@@ -1315,10 +1317,10 @@ impl Session {
             Coordinator,
         ) = self
             .run_request(
+                &cluster_state,
                 routing_info,
                 exec_params,
-                // A BATCH frame carries no tablet-version block, so the hint is unused here.
-                |connection: Arc<Connection>, consistency: Consistency, _: u8| async move {
+                |connection: Arc<Connection>, consistency: Consistency| async move {
                     connection
                         .batch_with_consistency(batch, values_ref, consistency, serial_consistency)
                         .await
@@ -1612,6 +1614,8 @@ impl Session {
             .unwrap_or_else(|| self.get_default_execution_profile_handle())
             .access();
 
+        let cluster_state = self.get_cluster_state();
+
         let exec_params = RequestExecutionParams::new_for_session_apis(
             &statement.config,
             &execution_profile,
@@ -1637,10 +1641,10 @@ impl Session {
             Coordinator,
         ) = self
             .run_request(
+                &cluster_state,
                 routing_info,
                 exec_params,
-                // A QUERY frame carries no tablet-version block, so the hint is unused here.
-                |connection: Arc<Connection>, consistency: Consistency, _: u8| {
+                |connection: Arc<Connection>, consistency: Consistency| {
                     // Needed to avoid moving query and values into async move block
                     let values_ref = &values;
                     let paging_state_ref = &paging_state;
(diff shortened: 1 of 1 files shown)


scylla/src/client/session.rs
```diff
@@ -1315,10 +1317,11 @@ impl Session {
             Coordinator,
         ) = self
             .run_request(
+                &cluster_state,
                 routing_info,
                 exec_params,
                 // A BATCH frame carries no tablet-version block, so the hint is unused here.
-                |connection: Arc<Connection>, consistency: Consistency, _: u8| async move {
+                |connection: Arc<Connection>, consistency: Consistency| async move {
```
wprzytula: 🔧 The comment is no longer relevant.

scylla/src/client/session.rs
```diff
@@ -1675,8 +1675,13 @@ impl Session {
...
+                                    // We could calculate the token here, and then
+                                    // a proper tablet hint. Without it, we'll receive
+                                    // unnecessary tablet feedback. However:
+                                    //  - It would require adding `PartitionKeyError` to `RequestAttemptError`. Do we want it?
+                                    //  - Unprepared statements arent performance-critical, maybe the overhead of
+                                    //    getting a tablet info on each such request is acceptable?
+                                    // TODO: Decide on the above.
                                     0,
                                 )
                                 .await
```
wprzytula, line 1695: 💭 I'm in favour of computing this properly.
1. Server might have (now or in the future) some metrics for misrouted prepared executions, and this will confusingly increase them.
2. Having another variant of `RequestAttemptError` is really benign IMO. The error type itself is not new.

scylla/src/client/session.rs
```diff
@@ -1637,10 +1642,11 @@ impl Session {
             Coordinator,
         ) = self
             .run_request(
+                &cluster_state,
                 routing_info,
                 exec_params,
                 // A QUERY frame carries no tablet-version block, so the hint is unused here.
-                |connection: Arc<Connection>, consistency: Consistency, _: u8| {
+                |connection: Arc<Connection>, consistency: Consistency| {
```
dawmd, line 1648: This comment is also redundant (at least for) now.

scylla/src/client/session.rs
dawmd: If QUERY requests are rare enough not to affect performance, I think either approach is OK. At the end of the day, it's a matter of who should pay the price: the server by generating and sending back tablet (+ the cost of network), or the driver by computing the token here. But if it's a rare event, then the cost should be negligible.

The point @wprzytula makes about the metrics is interesting. AFAIK, we don't track this in the server right now, but that's a good idea. Maybe we should consider adding them.

tl;dr I think it's okay to send the actual tablet version block here, as long as it's convenient for the driver. If it's problematic for whatever reason, I'd probably lean more towards keeping the code as-is, at least until (/unless) we introduce those metrics.

Lorak-mmk:
I added the computation of tablet hint for transparently prepared QUERY.

Turns out I didn't even need to extend RequestAttemptError: fail of this computation doesn't have to error the request, we can just fallback to 0.

dawmd (approved):

wprzytula (approved):

PR #1886, opened 2026-09-04
Expose compute_token_preserialized for nodejs-rs

hmarszalek:
Allow access to the `compute_token_preserialized` method with the unstable-node-js feature enabled.

This is needed for https://github.com/scylladb/nodejs-rs-driver/pull/529.

## Pre-review checklist

### DIFF
--- a/scylla/src/cluster/state.rs
+++ b/scylla/src/cluster/state.rs
@@ -735,7 +735,10 @@ impl ClusterState {
 }
 
 /// Additional API for interop-based code.
-#[cfg(all(scylla_unstable, feature = "unstable-csharp-rs"))]
+#[cfg(all(
+    scylla_unstable,
+    any(feature = "unstable-csharp-rs", feature = "unstable-nodejs-rs")
+))]
 impl ClusterState {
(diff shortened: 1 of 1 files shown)


Lorak-mmk:
I looked at the code in your PR that uses this:
```rust
   pub fn compute_token<'env>(
        &self,
        env: &'env Env,
        components: Vec<EncodedValuesWrapper>,
        keyspace: String,
        table: String,
    ) -> JsResult<JsInstance<'env, js_constructible_class::Token>> {
        with_custom_error_sync(|| {
            self.with_cluster_snapshot(env, |snapshot| {
                let mut values = SerializedValues::new();
                for component in &components {
                    values
                        .add_value(component, &PREENCODED_COLUMN_TYPE)
                        .map_err(make_js_error)?;
                }
                let token = snapshot
```


I don't think this need any new unstable API. You can have a newtype `PreSerializedValues(Vec<EncodedValuesWrapper>)` and implement `SerializeRow` for it. The implementation would simply ignore `RowSerializationContext`, and for each of your values call `make_cell_writer` on the row writer, and then `set_value` on the cell writer.

PR #1887, opened 2026-09-04
scylla: Hide ConsistencyMode behind unstable feature

### LINKED ISSUE #1884: Expose ConsistencyMode behind an unstable feature
The user is not able to write their own `DefaultPolicy` with the type being crate-private.

dawmd:
ConsistencyMode is needed to implement DefaultPolicy. As long as
it's crate-private, the user cannot implement it on their own.
Expose it conditionally so it's accessible if needed.

Also, drop ConsistencyMode::Local. Local strong consistency is not
implemented yet, so there's no point in exposing it -- it's dead code.

Fixes: scylladb/scylla-rust-driver#1884

Backport: no. The feature is experimental + the relevant code is only present on `main`.

## Pre-review checklist

- [ ] (N/A) I have adjusted the documentation in `./docs/source/`.

### DIFF
--- a/scylla/src/cluster/metadata/mod.rs
+++ b/scylla/src/cluster/metadata/mod.rs
@@ -184,13 +184,31 @@ impl Peer {
 /// The consistency mode of a keyspace, as reported by the `consistency` column of
 /// `system_schema.scylla_keyspaces`.
 ///
-/// Deliberately crate-private. Strong consistency is an experimental server-side feature and
-/// these mode names come straight from it, so making the type public would freeze names we may
-/// yet have to change - and an enum's variants cannot be renamed without a major release. Should
-/// this ever become public, it needs `#[non_exhaustive]`, which is pointless until then.
+/// Strong consistency is an experimental feature, and so this type is still unstable.
+/// That's why it's gated behind an unstable feature.
 //
 // Note: Local strong consistency isn't implemented yet.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+#[cfg(all(scylla_unstable, feature = "unstable-strong-consistency"))]
+pub enum ConsistencyMode {
+    /// Eventual consistency. Covers every non-tablet keyspace and every keyspace
+    /// on a server that does not report a consistency mode.
+    Eventual,
+    /// Global strong consistency (`consistency = 'global'`): the keyspace uses
+    /// strongly-consistent (Raft-based) tablets.
+    Global,
+}
+
+/// The consistency mode of a keyspace, as reported by the `consistency` column of
+/// `system_schema.scylla_keyspaces`.
+///
+/// Strong consistency is an experimental feature, and so this type is still unstable.
+/// Hence crate-private.
+//
+// Note: Local strong consistency isn't implemented yet.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg(not(all(scylla_unstable, feature = "unstable-strong-consistency")))]
 pub(crate) enum ConsistencyMode {
     /// Eventual consistency. Covers every non-tablet keyspace and every keyspace
     /// on a server that does not report a consistency mode.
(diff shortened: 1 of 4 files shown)


Lorak-mmk (changes requested):
As a rule, we require `cfg(scylla_unstable)` for all unstable features, so that users can't enable them accidentally. This is the same thing that Tokio does: https://docs.rs/tokio/latest/tokio/#unstable-features.

You need to guard the items with `#[cfg(all(scylla_unstable, feature = "unstable-strong-consistency"))]`

dawmd:
> As a rule, we require `cfg(scylla_unstable)` for all unstable features, so that users can't enable them accidentally. This is the same thing that Tokio does: https://docs.rs/tokio/latest/tokio/#unstable-features.
> 
> You need to guard the items with `#[cfg(all(scylla_unstable, feature = "unstable-strong-consistency"))]`

Oh, I see. That makes sense.

v2:
* Extended the guards with `scylla_unstable`.

Lorak-mmk (approved):
I'll wait for @wprzytula with merging.

scylla/src/cluster/metadata/mod.rs
```diff
@@ -188,13 +188,13 @@ impl Peer {
...
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub(crate) enum ConsistencyMode {
     /// Eventual consistency. Covers every non-tablet keyspace and every keyspace
     /// on a server that does not report a consistency mode.
     Eventual,
-    /// Reserved for a future per-datacenter strong-consistency mode (`consistency = 'local'`).
-    Local,
     /// Global strong consistency (`consistency = 'global'`): the keyspace uses
     /// strongly-consistent (Raft-based) tablets.
     Global,
```
wprzytula, line 200: 💭 Now that I think of it, shouldn't the variants be rather `StrongLocal` and `StrongGlobal`, in contrast to `Eventual`?

scylla/src/cluster/metadata/mod.rs
```diff
@@ -216,7 +234,14 @@ pub struct Keyspace {
...
-    /// Crate-private along with [`ConsistencyMode`] itself; see that type for why.
+    /// Public only behind the `unstable-strong-consistency` feature because
+    /// strong consistency is still experimental.
+    #[cfg(all(scylla_unstable, feature = "unstable-strong-consistency"))]
+    pub consistency_mode: ConsistencyMode,
+    /// The consistency mode of the keyspace.
+    ///
+    /// Strong consistency is still experimental; hence crate-private.
+    #[cfg(not(all(scylla_unstable, feature = "unstable-strong-consistency")))]
     pub(crate) consistency_mode: ConsistencyMode,
```
wprzytula, line 245: 💭 This works, but is a bit inelegant - a switch based on a flag is always distracting.

Wouldn't it be cleaner to leave the field always `pub(crate)`, but conditionally expose a `pub` getter?

scylla/src/cluster/metadata/mod.rs
dawmd: Scylla uses those names:
* when creating a keyspace
* internally
* in the system table\*

I think we should keep them as-is for consistency.

\*: https://github.com/scylladb/scylladb/blob/46f68969253f5c57c5ae1099a94c9401713eed7f/db/schema_tables.cc#L1199-L1202 and https://github.com/scylladb/scylladb/blob/46f68969253f5c57c5ae1099a94c9401713eed7f/data_dictionary/data_dictionary.cc#L618-L639.

scylla/src/cluster/metadata/mod.rs
dawmd: I can do that. We'll just need to remember to get rid of the getter once the field is exposed unconditionally.

scylla/src/cluster/metadata/mod.rs
wprzytula: OK, fair.

scylla/src/cluster/metadata/mod.rs
wprzytula: I leave the decision up to you and @Lorak-mmk.

dawmd: I'm fine with either, so I guess I lean more towards keeping the code as-is (just to avoid pushing a new version :smile:).

dawmd:
FYI @aleksbykov this PR should expose `ConsistencyMode` again.

wprzytula (approved):

Lorak-mmk (approved):

Lorak-mmk:
Rebased on main.

aleksbykov:
> FYI @aleksbykov this PR should expose `ConsistencyMode` again.

@dawmd , thanks, i have rebuilt the cql-stress to latest master of rustdriver while no new release

PR #1890, opened 2026-09-04
scylla/tests: Move tablets_routing_v2 tests to CCM suite

### LINKED ISSUE #1888: TABLETS_ROUTING_V2 integration tests are always silently skipped
Three out of the four tests in the file rely on the nodes using the experimental feature `strongly-consistent-tables`. However, the test suite doesn't allow us to enable it -- the nodes are reused. Setting the feature globally (in `test/cluster/scylla-cluster.yml`) isn't an option either because the tests must also work against versions of Scylla that don't have that feature.

Move the tests to the CCM suite and explicitly enable the experimental feature for the tests.

Affected version: cf24aa057a27066ef51c4b4ee231dbd05e2a08d0.

dawmd:
We move the tests to the CCM test suite because they depend on the nodes
running with the experimental feature strongly-consistent-tables enabled.
That's not feasible outside CCM -- the nodes are reused by integration
tests, and we cannot enable the feature globally because the tests must
be able to be run against multiple versions of Scylla, some of which may
not support it (e.g., 2025.1.10).

After moving the tests to the new suite, we manually configure the
cluster. The tests are skipped when run against Scylla that doesn't
support the experimental feature or TABLETS_ROUTING_V2.

Fixes: scylladb/scylla-rust-driver#1888

Backport: no. The affected code is only present on the main branch.

## Pre-review checklist

- [ ] (N/A -- tests) I have provided docstrings for the public items that I want to introduce.
- [ ] (N/A) I have adjusted the documentation in `./docs/source/`.


wprzytula (approved):

Lorak-mmk:
> Backport: no. The affected code is only present on the main branch.

Generally, there is no need for such line here. In drivers we live in a beautiful world in which backports basically never happen.

Lorak-mmk (approved):

Lorak-mmk:
Rebased on main,

PR #1896, opened 2026-09-05
tests: introduce new test for compression

### LINKED ISSUE #1424: Have a frame compression integration tests
Have tests for available compression methods:
1. lz4
2. snappy

Test should create frames of different frame sizes (compressed):
1. `size < 1000`
2. `1000 < size > 2000` 
3. 1mb or around it

Migrated to Jira: [DRIVER-142](https://scylladb.atlassian.net/browse/DRIVER-142)

[DRIVER-142]: https://scylladb.atlassian.net/browse/DRIVER-142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

wprzytula:
A revival of #1218 authored by @fruch.

### Original cover letter

test are trying both algorithms (snappy, lz4), while counting the size of the frames coming into scylla

for reference we also have one test doing the same without compression

test main structure was borrowed from java-driver, the usage of proxy to count the packets is unique to this change.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1424

## Pre-review checklist

### DIFF
--- a/scylla-proxy/src/frame.rs
+++ b/scylla-proxy/src/frame.rs
@@ -43,7 +43,8 @@ impl FrameParams {
     /// Tells whether the frame carried a compressed body on the wire.
     ///
     /// Note that frames handed out by the proxy always expose a *decompressed*
-    /// body, so this flag is the only way to tell that compression was in play.
+    /// body, so this flag (together with [`RequestFrame::wire_body_len`]) is the
+    /// only way to tell that compression was in play.
     pub const fn is_compressed(&self) -> bool {
         self.flags & flag::COMPRESSION != 0
     }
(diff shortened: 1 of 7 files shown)


scylla/tests/integration/session/compression.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+#[tokio::test]
+async fn should_execute_queries_without_compression_1mb() {
+    test_compression(None, 1_000_000, 1_000_000..1_010_000).await;
+}
+
+#[tokio::test]
+async fn should_execute_queries_with_snappy_compression() {
+    test_compression(Some(Compression::Snappy), 1_000, 1_000..1_500).await;
```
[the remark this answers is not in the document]
wprzytula: What's the point of such tests?

scylla-proxy/src/frame.rs
```diff
@@ -61,19 +62,37 @@ pub(crate) enum FrameOpcode {
...
+    /// The frame body, decompressed if the frame arrived compressed.
     pub body: Bytes,
+    /// Number of body bytes as they appeared on the wire, i.e. *before* decompression.
+    /// For frames that were not compressed this is equal to `body.len()`.
+    ///
+    /// This is what makes it possible to observe how effective compression is;
+    /// [`Self::body`] is always decompressed, so its length says nothing about
+    /// how much data actually travelled over the network.
+    pub wire_body_len: usize,
+}
```
Lorak-mmk, line 78: Note: as new pub field in exhaustive struct. Need major bump in scylla-proxy, but because scylla-proxy is at "0.0.7" every bump is major.

scylla-proxy/src/proxy.rs
```diff
@@ -1818,6 +1818,7 @@ mod tests {
             params: recvd_params,
             opcode: recvd_opcode,
             body: recvd_body,
+            ..
         } = read_request_frame(conn, compression).await.unwrap();
         assert_eq!(recvd_params, HARDCODED_OPTIONS_PARAMS);
```
Lorak-mmk, line 1823: Commit: scylla-proxy: record wire-level body length on frames

This struct doesn't have a lot of fields. When adding a new field and updating matches, I'd prefer to explicitly exclude it instead of using `..`. That way, when we add another new field someday, we won't miss some match that should be updated.

Lorak-mmk (changes requested):

wprzytula:
Rebased on main.

scylla-proxy/src/proxy.rs
wprzytula: Strong agree. I missed this Claude's decision.

wprzytula:
Addressed the review comment.

Lorak-mmk (approved):

PR #1899, opened 2026-09-05
lib.rs: expose VectorSerializationErrorKind

wprzytula:
Mistakenly, we didn't expose `VectorSerializationErrorKind` in the `scylla`'s API. This made it impossible to match on the enum's variants. As similar error kinds for other types are exposed, this one gets exposed now, too.

## Pre-review checklist


Lorak-mmk (approved):
Interesting, I think I was fixing the same thing for deser in the past and that I check there are no other places like that...

Lorak-mmk:
Rebased on main.

PR #1902, opened 2026-09-06
proxy: select socket family from real node address

### LINKED ISSUE #1691: Select proxy shard-aware socket family from the real node address
The proxy chooses the shard-aware outgoing socket family from the proxy listener address instead of the real node address.

Code permalink: https://github.com/scylladb/scylla-rust-driver/blob/e04a12a9950cb3d051288541d5439835098c5cc2/scylla-proxy/src/proxy.rs#L897

With mixed IPv4/IPv6 proxy and node addresses, the cluster-side connect can fail with an address-family error. IPv6 shard-aware setups can also fail before connecting because the later bind path uses an IPv4 wildcard.

Suggested direction: select the socket and bind address family from the real node address used for the cluster-side connection.

be-student:
The shard-aware proxy now creates its outgoing socket and wildcard bind address from the real node's IP family. This allows mixed-family proxy/node configurations and IPv6 shard-aware connections.

Validation: focused IPv4/IPv6 regression test, `cargo fmt --all -- --check`, and `cargo clippy -p scylla-proxy --all-targets -- -D warnings`.

Fixes: #1691

### DIFF
--- a/scylla-proxy/src/proxy.rs
+++ b/scylla-proxy/src/proxy.rs
@@ -988,15 +988,10 @@ impl Doorkeeper {
         real_addr: SocketAddr,
     ) -> Result<(TcpStream, Option<TargetShard>), DoorkeeperError> {
         let mut cluster_stream = if let Some(shards) = self.shards_count {
-            let socket = match self.node.proxy_addr().ip() {
-                std::net::IpAddr::V4(_) => TcpSocket::new_v4(),
-                std::net::IpAddr::V6(_) => TcpSocket::new_v6(),
-            }
-            .map_err(DoorkeeperError::SocketCreate)?;
+            let (socket, mut desired_addr) = shard_aware_socket(real_addr, driver_addr.port())
+                .map_err(DoorkeeperError::SocketCreate)?;
 
             let shard_preserving_addr = {
-                let mut desired_addr =
-                    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), driver_addr.port());
                 while socket.bind(desired_addr).is_err() {
                     // in search for a port that translates to the desired shard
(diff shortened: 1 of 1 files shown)


wprzytula (changes requested):
🔧 The unit test of the mere `shard_aware_socket` is not enough. Have an integration test that really starts a proxy, to check the whole execution path.

be-student:
Addressed the review in 842f11cb. The unit test now allocates a source port dynamically, and a new integration test starts the proxy with an IPv4 listener and an IPv6 real node to exercise the complete mixed-family connection path. Both focused tests pass.

scylla-proxy/src/proxy.rs
```diff
@@ -1723,6 +1718,24 @@ impl ProxyWorker {
     }
 }
 
+fn shard_aware_socket(
```
wprzytula: 🔧 
```suggestion
fn open_shard_aware_socket_to_real_node(
```

wprzytula: Let's be explicit.

wprzytula (changes requested):
One last naming nit from my side.

be-student:
Addressed the latest review in 0dcaafa5. Renamed the helper as requested and made both dynamic-port tests retry on AddrInUse to close the probe/rebind race. Formatting and both targeted proxy tests pass.

wprzytula (approved):

Lorak-mmk (approved):
In general looks fine, but the commit structure is a mess. Third commits performs a rename because first commit used wrong name (first commit should be edited instead), and fixes tests introduced before.

@wprzytula I don't think it makes much sense to improve it for such small change that is almost surely automated. Maybe you can just squash-merge?

PR #1904, opened 2026-09-08
Pager: compute replicas only on trace level

wprzytula:
# Compute replicas only if needed for tracing
    
Non-pager execution APIs already compute replicas only if the execution span is enabled. This is reasonable, because the computation is costly and only surfaces the user at TRACE tracing level, which is unlikely in
prod.

The problem was mentioned by Claude Opus when working on https://github.com/scylladb/cpp-rs-driver/pull/499. Also by @Lorak-mmk in the past IIRC.

This PR makes replicas computed **at most once**, if the span is created at the moment when the tracing level is TRACE, so the execution span is enabled. The _at most once_ semantics is achieved by OnceCell, which BTW is a zero-cost abstraction (compared to a plain Option), yet provides convenient memoization semantics here.

## Pre-review checklist

### DIFF
--- a/scylla/src/client/pager.rs
+++ b/scylla/src/client/pager.rs
@@ -2,6 +2,7 @@
 //! They enable consuming result of a paged query as a stream over rows,
 //! which abstracts over page boundaries.
 
+use std::cell::OnceCell;
 use std::future::Future;
 use std::ops::ControlFlow;
 use std::pin::Pin;
@@ -31,7 +32,9 @@ use crate::frame::frame_errors::ResultMetadataAndRowsCountParseError;
 use crate::frame::request::query::{PagingState, PagingStateResponse};
 use crate::frame::response::NonErrorResponseWithDeserializedMetadataV2 as NonErrorResponseWithDeserializedMetadata;
 use crate::frame::response::result;
-use crate::frame::response::result::{DeserializedMetadataAndRawRows, SchemaChange, SetKeyspace};
+use crate::frame::response::result::{
+    DeserializedMetadataAndRawRows, SchemaChange, SetKeyspace, TableSpec,
+};
 use crate::frame::types::{Consistency, SerialConsistency};
 use crate::network::Connection;
 use crate::observability::driver_tracing::RequestSpan;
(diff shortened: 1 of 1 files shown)


wprzytula:
Benchmark results of the relevant scenario:
```
requests::requests::paged_select counts_0:(setup_default(100))
  ======= CALLGRIND ====================================================================
  Baselines:                               |base
  Instructions:                   221979761|221721149            (+0.11664%) [+1.00117x]
  L1 Hits:                        330373299|330023053            (+0.10613%) [+1.00106x]
  LL Hits:                         11184265|11257520             (-0.65072%) [-1.00655x]
  RAM Hits:                             636|638                  (-0.31348%) [-1.00314x]
  Total read+write:               341558200|341281211            (+0.08116%) [+1.00081x]
  Estimated Cycles:               386316884|386332983            (-0.00417%) [-1.00004x]
  ======= DHAT =========================================================================
  Total bytes:                     13478964|13478964             (No change)
  Total blocks:                      151701|151701               (No change)
  At t-gmax bytes:                        0|0                    (No change)
  At t-gmax blocks:                       0|0                    (No change)
  At t-end bytes:                         0|0                    (No change)
  At t-end blocks:                        0|0                    (No change)
  Reads bytes:                     33288100|33227300             (+0.18298%) [+1.00183x]
  Writes bytes:                    66464900|66458500             (+0.00963%) [+1.00010x]
```

scylla/src/client/pager.rs
```diff
@@ -936,14 +936,17 @@ If you are using this API, you are probably doing something wrong."
...
                 partition_key.as_ref().map(|pk| pk.iter()),
                 token,
                 serialized_values_size,
             );
-            if let Some(replicas) = replicas {
+            if !span.span().is_disabled()
+                && let Some(replicas) = replicas
+            {
                 span.record_replicas(replicas.iter().map(|(node, shard)| (node, *shard)));
             }
```
Lorak-mmk: Commit: "pager: pass &ClusterState to create_span()"

Commit message says "Just threading, no behavioural changes." but you are adding a check for `!span.span().is_disabled()`

Lorak-mmk (changes requested):

scylla/src/client/pager.rs
wprzytula: Keen eye — miscommitted change.

Lorak-mmk (approved):

PR #1905, opened 2026-09-08
Driver config reporting

### LINKED ISSUE #1859: Driver config reporting: Initial (downcasting-based) implementation

Lorak-mmk:
This PR introduces support for reporting driver configuration to the server.

Configuration report is sent to the server as JSON, only on the control connection. It is sent in the STARTUP frame, in `DRIVER_CONFIG` key. It can then be read from `system.clients.client_options['DRIVER_CONFIG']`.

Format of the config report is the same for all drivers, which unfortunately means:
1. We can't send everything
2. Sometimes we need to lose some information in order to conform to the schema.

Schema file is added to the codebase, because it is used for the tests.

## AI usage and review

This PR makes a change that requires tons of relatively simple code, that is almost fully isolated from the rest of the driver.
Carefully reviewing all of it would take a lot of time, and I don't think it would be very beneficial.

I did carefully review all the changes outside `driver_config.rs` / `driver_config_tests.rs`.
For the `driver_config.rs`:
 - Last 2 commits introduce tests that prove two things: 1. Reports always conform to the schema. 2. We send all the optional stuff we realistically can
 - I did review the tests above

Given all of the above, I only briefly reviewed the actual implementation in `driver_config.rs`  / `driver_config_tests.rs`, to make sure there are no stupid bugs, panics etc. I encourage reviewers to take the same approach.

Fixes: https://github.com/scylladb/scylla-rust-driver/issues/1859
Fixes: https://scylladb.atlassian.net/browse/DRIVER-378

## Pre-review checklist

### DIFF
--- a/scylla/src/client/driver_config.rs
+++ b/scylla/src/client/driver_config.rs
@@ -1,0 +1,108 @@
+//! Reporting of the driver's effective configuration to the cluster.
+//!
+//! The driver sends its configuration as a JSON document under the
+//! `DRIVER_CONFIG` `STARTUP` option, where ScyllaDB exposes it in
+//! `system.clients.client_options`. That makes the configuration a session
+//! actually runs with inspectable from the server side, without having to
+//! correlate it with client-side logs. The document follows an external schema
+//! shared by all ScyllaDB drivers, so that one
+//! query answers the same question regardless of which driver a client uses.
+//! Only the control connection sends it: the value is identical for every
+//! connection of a session, so sending it on all of them would only bloat their
+//! `STARTUP` frames.
+
+use serde::Serialize;
+
+/// The major version of the reported configuration schema.
+///
+/// Adding keys is backwards compatible and does not bump this; only changing or
+/// removing the meaning of an existing key does.
+const REPORT_VERSION: u32 = 1;
+
+/// Reports larger than this are omitted instead of being sent.
+///
+/// ScyllaDB's protocol-extensions spec requires drivers to validate the size
+/// client side and omit the option above this limit, so that reporting the
+/// configuration never prevents a connection from being established. 32 KiB is
+/// the limit gocql and csharp-driver use, keeping it consistent across drivers.
+///
+/// It also leaves a 2x margin to the driver's own hard limit: `write_string`
+/// prefixes every `STARTUP` value with a 16-bit length via `write_short_length`,
+/// so a value above 64 KiB would fail `STARTUP` serialization and take down the
+/// control-connection handshake, and with it session creation.
+const MAX_REPORT_SIZE: usize = 32 * 1024;
+
+/// The reported configuration document.
+///
+/// The `connection`, `control-plane` and `query` groups are yet to be added.
+/// The schema requires all three, so the report this currently produces does
+/// not validate against it.
+#[derive(Serialize)]
+struct Report {
+    version: u32,
+}
+
+/// Builds the configuration report of a session.
+///
+/// Carries no configuration yet, because `{"version":1}` needs no inputs; the
+/// configuration each group reports is added alongside that group.
+pub(crate) struct DriverConfigReporter {}
+
+impl DriverConfigReporter {
+    pub(crate) fn new() -> Self {
+        Self {}
+    }
+
+    /// Renders the report, or returns `None` when there is nothing safe to send.
+    ///
+    /// Never fails and never panics: the report is a diagnostic aid, and must
+    /// never prevent a connection from being established. Problems are logged
+    /// and the option omitted.
+    ///
+    /// The report is rebuilt on every call rather than cached, because it is not
+    /// constant for the lifetime of a session: `target_is_scylladb` is only
+    /// known after the `OPTIONS`/`SUPPORTED` exchange of the connection being
+    /// established, and configuration read through an `ExecutionProfileHandle`
+    /// can be remapped at runtime.
+    ///
+    /// `target_is_scylladb` tells whether the peer is a ScyllaDB node, as some
+    /// reported values only apply to ScyllaDB. It is not read yet, hence the
+    /// scoped `expect`: a leading underscore would only have to be renamed once
+    /// a group starts using it.
+    pub(crate) fn report(
+        &self,
+        #[expect(unused_variables)] target_is_scylladb: bool,
+    ) -> Option<String> {
+        let report = Report {
+            version: REPORT_VERSION,
+        };
+
+        // `serde_json::to_string` cannot fail for the report's current shape; the
+        // branch keeps reporting best effort once a future one can fail. It will
+        // not catch a non-finite `f64`: `serde_json` emits those as `null`, which
+        // the schema rejects under a numeric key, so such values have to be
+        // filtered out where they are read rather than here.
+        let report = match serde_json::to_string(&report) {
+            Ok(report) => report,
+            Err(err) => {
+                tracing::warn!("Failed to serialize the driver configuration report: {err}");
+                return None;
+            }
+        };
+
+        if report.len() > MAX_REPORT_SIZE {
+            tracing::warn!(
+                "Driver configuration report is too large to be reported: {} bytes, the limit is {} bytes",
+                report.len(),
+                MAX_REPORT_SIZE
+            );
+            return None;
+        }
+
+        Some(report)
+    }
+}
+
+#[cfg(test)]
+#[path = "driver_config_tests.rs"]
+mod tests;
--- a/scylla/src/client/driver_config.rs
+++ b/scylla/src/client/driver_config.rs
@@ -11,6 +11,7 @@
 //! connection of a session, so sending it on all of them would only bloat their
 //! `STARTUP` frames.
 
+use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -30,6 +31,11 @@ use crate::policies::reconnect::{
 };
 use crate::policies::retry::{
     DefaultRetryPolicy, DowngradingConsistencyRetryPolicy, FallthroughRetryPolicy, RetryPolicy,
+};
+#[cfg(feature = "metrics")]
+use crate::policies::speculative_execution::PercentileSpeculativeExecutionPolicy;
+use crate::policies::speculative_execution::{
+    SimpleSpeculativeExecutionPolicy, SpeculativeExecutionPolicy,
 };
 use crate::routing::NodeLocationPreference;
 use crate::statement::PageSize;
@@ -56,9 +62,9 @@ const MAX_REPORT_SIZE: usize = 32 * 1024;
 /// The reported configuration document.
 ///
 /// Carries every group the schema requires, so the produced document validates
-/// against it. Of the optional ones only `query.speculative-execution` is yet
-/// to be added; each type below states why the keys it does not populate are
-/// absent.
+/// against it. The schema's optional keys are populated where this driver has
+/// something to report; each type below states why it does not where it does
+/// not.
 #[derive(Serialize)]
 #[serde(rename_all = "kebab-case")]
 struct Report {
@@ -280,6 +286,10 @@ struct QueryReport {
     defaults: QueryDefaultsReport,
     retry: QueryRetryReport,
     load_balancing: QueryLoadBalancingReport,
+    /// Omitted when speculative execution never fires, which the schema gives
+    /// the group's absence as its meaning.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    speculative_execution: Option<QuerySpeculativeExecutionReport>,
 }
 
 /// The defaults a statement gets when it overrides nothing, read from the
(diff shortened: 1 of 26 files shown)


scylla/src/network/connection.rs
```diff
@@ -2545,11 +2545,15 @@ struct StreamIdSet {
...
+/// requests in flight on it. Every id in `0..MAX_IN_FLIGHT_REQUESTS` is
+/// allocatable; none is reserved.
+const MAX_IN_FLIGHT_REQUESTS: usize = i16::MAX as usize + 1;
+
 impl StreamIdSet {
     fn new() -> Self {
-        const BITMAP_SIZE: usize = (i16::MAX as usize + 1) / 64;
         Self {
-            used_bitmap: vec![0; BITMAP_SIZE].into_boxed_slice(),
+            used_bitmap: vec![0; MAX_IN_FLIGHT_REQUESTS / 64].into_boxed_slice(),
```
wprzytula, line 2588: ⛏️ `64` is a magic constant here. I don't immediately know what it comes from.

scylla/src/client/session.rs
```diff
@@ -36,9 +36,7 @@ use crate::observability::tracing::TracingInfo;
 use crate::policies::address_translator::AddressTranslator;
```
wprzytula, line 37: commit `session: derive the socket options and reconnect policy once`:
🤔 I fails to understand the commit message, mostly because of the _derive_ term. 
> Both were assembled inline where they are consumed: the socket options inside the `ConnectionConfig` literal, the effective reconnect policy as a pair of cfg-gated `PoolConfig` fields. Deriving each once, ahead of the literals, gives every future consumer the same value instead of a parallel derivation of it.

What is _deriving_ in this context? What is a _parallel derivation_ and how this move of some code prevents it?

scylla/src/policies/mod.rs
```diff
@@ -29,3 +29,23 @@ pub(crate) mod reconnect;
...
+        Some(idx) => &full[..idx],
+        None => full,
+    };
+    match head.rfind("::") {
+        // The `name` field is declared non-empty by the schema, so fall back to
+        // the unstripped name rather than returning "".
+        Some(idx) if idx + 2 < full.len() => &full[idx + 2..],
+        _ => full,
+    }
+}
```
wprzytula, line 51: 🔧 I'd like to have unit tests for this function.

scylla/src/client/driver_config.rs
```diff
@@ -11,8 +11,20 @@
 //! connection of a session, so sending it on all of them would only bloat their
```
wprzytula, line 11: commit `Report the connection group`:

> shard-aware.enabled accounts for PoolSize too, since a per-host pool never dials the shard-aware port whatever disallow_shard_aware_port says.

🤔 Is `shard-aware.enabled` about _basic shard awareness_ or _advanced shard awareness_? I'd think it should about _basic shard awareness_. Then it has nothing to do with the shard-aware port and the `disallow_shard_aware_port` option.

scylla/src/client/driver_config.rs
```diff
@@ -34,23 +46,235 @@ const MAX_REPORT_SIZE: usize = 32 * 1024;
...
+#[derive(Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct RequestsReport {
+    in_flight: MaxReport,
+    /// The driver's bound on orphaned requests is not quite what the schema
+    /// asks for: it counts only orphans older than `OLD_AGE_ORPHAN_THRESHOLD`,
+    /// a qualifier the schema cannot express. Reporting the threshold is still
+    /// much closer to the truth than omitting the group, which means "no bound
+    /// at all".
+    orphaned: MaxReport,
```
wprzytula, line 164: ❓ Can you make it clearer here what this discrepancy means in practice? How big can the difference between the reported and the real orphaned request counts be?

scylla/src/client/driver_config.rs
```diff
@@ -34,23 +46,235 @@ const MAX_REPORT_SIZE: usize = 32 * 1024;
...
+#[serde(rename_all = "kebab-case")]
+struct LingerReport {
+    /// Whole seconds, the only granularity the schema offers - and the only one
+    /// `SO_LINGER` offers either. A configured linger below one second reports
+    /// 0, which reads as "reset the connection on close", and that is exact
+    /// rather than approximate: `apply_socket_options` sets the option through
+    /// `socket2`, whose `into_linger` truncates with the very same `as_secs`, so
+    /// the kernel is given the 0 this reports. Do not "fix" this to floor at 1;
+    /// it would describe a lingering close the socket does not perform.
+    interval_s: u64,
```
wprzytula, line 209: 💭 Should I be able to understand this? I cannot...

scylla/src/client/driver_config.rs
wprzytula: OK, I checked against the schema and indeed it's _advanced shard awareness_. Too bad the naming doesn't imply it.

scylla/src/client/driver_config.rs
```diff
@@ -101,6 +326,86 @@ impl DriverConfigReporter {
...
+    fn node_preference_report(&self) -> Option<NodePreferenceReport> {
+        let filter = self.host_filter.as_ref()?.as_any()?;
+        let filter = filter.downcast_ref::<DcHostFilter>()?;
+
+        // `local-dc` is a `nonEmptyString`, and an empty DC name accepts no node
+        // anyway, so there is no preference worth reporting.
+        (!filter.local_dc.is_empty()).then(|| NodePreferenceReport::Dc {
+            local_dc: filter.local_dc.clone(),
+        })
+    }
```
wprzytula, line 841: 🔧 A comment would be useful — unexplained unconditional `?` after a downcast is suspicious.

scylla/src/client/driver_config.rs
```diff
@@ -0,0 +1,1032 @@
...
+            },
+            fallback_to_non_preferred_nodes: fallback_to_non_preferred_nodes(
+                node_preference,
+                default_policy.permit_dc_failover,
+            ),
+            adaptive_ordering: default_policy.latency_awareness.is_some().then_some(
+                AdaptiveOrderingReport {
+                    signals: &["latency"],
+                },
+            ),
```
wprzytula, line 984: Quoting the specification:
> policy.adaptive-ordering.signals | 
> array of enum(**latency, response-rate, in-flight-requests, recovery-state**), non-empty and unique, required within adaptive-ordering — the runtime observations that reorder otherwise eligible candidates. The whole group is absent when adaptive ordering is disabled, and it deliberately names signals rather than an algorithm or its thresholds

❓ What is `recovery-state` and doesn't Rust Driver reorder based on this? After all, `is_alive()` arguably filters out node that haven't recovered after a reboot or network partition.

scylla/src/network/connection.rs
Lorak-mmk: Its not introduced by this PR, and it appears in other places of this impl.

scylla/src/client/driver_config.rs
Lorak-mmk: Well it is a property of the `connection` group. The only shard awareness relevant there is advanced shard awareness (shard aware port).

"Basic shard awareness" (which is imo even worse term than advanced shard awareness) is a property of load balancing, not connection management.

scylla/src/client/driver_config.rs
Lorak-mmk: Hmm... is the doc comment on this method not enough? It does explain why we are only interested in DcHostFilter.

scylla/src/client/driver_config.rs
Lorak-mmk: The config schema defines orphans as "Requests that the client stopped waiting for but whose stream identifiers cannot yet be safely reused.".  

In the driver we do have the same definition for orphaned requests, but we also define old orphans as orphans that are still waiting after some defined time. Driver only imposes a limit on a number of old orphans, not orphans in general. 

The difference? Theoretically, I guess, however many requests you can issue in OLD_AGE_ORPHAN_THRESHOLD (1s) time. So possibly a lot, but it would require extremely pathological scenario. Typically the difference should be small.

scylla/src/client/driver_config.rs
Lorak-mmk: @sylwiaszunejko 
- What is "recovery-state" and "response-rate" here? Other signals are quite clear.
- If our DefaultPolicy always checks if Node is alive (= we have any connection opened to it) and puts unalive nodes at the end of the plan, should it always report adaptive ordering with "recovery-state"?

scylla/src/client/driver_config.rs
Lorak-mmk: Our linger config is `Option<Duration>`. When calling Tokios `socket.set_linger`, `socket2 (which is used internally)  needs to convert this duration to pass it to a syscall:
```rust
    pub fn set_linger(&self, linger: Option<Duration>) -> io::Result<()> {
        let linger = into_linger(linger);
        unsafe { setsockopt(self.as_raw(), sys::SOL_SOCKET, sys::SO_LINGER, linger) }
    }
```

That conversion looks like this:
```rust
const fn into_linger(duration: Option<Duration>) -> sys::linger {
    match duration {
        Some(duration) => sys::linger {
            l_onoff: 1,
            l_linger: duration.as_secs() as _,
        },
        None => sys::linger {
            l_onoff: 0,
            l_linger: 0,
        },
    }
}
```

So, if the duration is less than a second, it will behave exactly as a 0 in the driver. This comment explains this behavior and says this is why we also floor sub-second durations to 0 in the report.

Lorak-mmk:
Addressed 2 comments, responded to all others.

Lorak-mmk:
Rebased on main

scylla/src/client/driver_config.rs
sylwiaszunejko: "recovery-state" and "response-rate" are right now java specific configuration related to slow replica avoidence, I think the only option here applicable to rust driver is latency if latency awareness is enabled.

sylwiaszunejko: Probably the documentation is not the best. Every driver avoids nodes it can't reach: gocql skips `!h.IsUp()` hosts in the plan loop, Java builds plans from live nodes, Python from `host.is_up`. If connectedness counted, all five drivers would report recovery-state unconditionally and the enum member would distinguish nothing. 
`recovery-state` = how recently a node returned to service, used as a graded signal against a node that is up and connected. In Java: `upTimes` stores the timestamp of each node's last UP transition; a replica whose UP is younger than NEWLY_UP_INTERVAL (1 min) gets bubbled to the back of the plan with probability 3/4, and its mere presence also disables the unhealthy-demotion branch for that plan. The node is a perfectly good replica by every other measure — connected, not busy, answering — and is deliberately deprioritized anyway, because it restarted a moment ago with cold caches and a half-filled pool.

scylla/src/client/driver_config.rs
```diff
@@ -0,0 +1,1032 @@
...
+            node_preference: self.node_preference_report(),
+            tls: self.tls_configured.then_some(TlsReport {}),
+        }
+    }
+
+    fn reconnection_policy_report(&self) -> ReconnectionPolicyReport {
+        let Some(policy) = self.reconnect_policy.as_any() else {
+            return ReconnectionPolicyReport::Custom {
+                name: self.reconnect_policy.reported_name(),
+            };
```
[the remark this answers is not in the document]
Lorak-mmk: Our implementation of those APIs returns non empty strings, and is doc(hidden), so an override returning empty strings is totally a users fault.

scylla/src/client/driver_config.rs
wprzytula: It is enough. For some reason, I didn't see it when writing the comment.

scylla/src/client/driver_config.rs
wprzytula: "Basic shard awareness" combines two behaviours:
1. Sending requests to the shard being the replica (load balancing department — as you said).
2. Being _aware_ what shard the connection is open to; and keeping connections opened to all shards. This is definitely a property of connection management.

wprzytula (approved):

PR #1922, opened 2026-09-17
Fetch and expose index metadata

hmarszalek:
The driver reads every `system_schema` table describing a keyspace except `indexes`, so a user who wanted to know which indexes a table has had to query the table themselves.
Add `Table::indexes`, keyed by the index name, which is what `system_schema.indexes` is keyed by within a table.
An unrecognised kind becomes `IndexKind::Other` rather than failing the fetch.
The query runs concurrently with the one for partitioners, which is keyed the same way and reads at the same point of the fetch, so the extra read costs no extra round trip.

## Pre-review checklist


scylla/src/cluster/metadata/mod.rs
```diff
@@ -271,6 +273,10 @@ pub struct Table {
     pub clustering_key: Vec<String>,
     /// Name of the partitioner used by the table.
     pub partitioner: Option<String>,
+    /// Indexes defined on the table, keyed by the index name.
+    ///
+    /// Empty HashMap may as well mean that the client disabled schema fetching in SessionConfig.
+    pub indexes: HashMap<String, Index>,
     /// Column specs for the partition key columns.
     pub(crate) pk_column_specs: Vec<ColumnSpec<'static>>,
```
Lorak-mmk, line 281: Interesting, you made it part of `Table`. 

The argument I see for it is that index is tied to a single table.
The argument I see against is that an index is usually a type of materialized view, which we store in `Keyspace`, not `Table` - so for consistency index should do the same.

What's your reasoning for doing it that way?

hmarszalek: `system_schema.views` is keyed `(keyspace_name, view_name)` and `system_schema.indexes` is keyed `(keyspace_name, table_name, index_name)`.

From what I understand, a view is an object that you can run `SELECT` against and that's why it wraps a whole `Table` in metadata. An index is just a property of one table. So it seems more natural to me to put it as part of the `Table`.

